PR #13, opened 2018-11-17
✨ RSP-631: Datepicker range styles added

GarthDB:


src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+      border-top: 0;
+      border-right: 0;
+      border-bottom: 0;
+    }
+    /* Horisontal line */
+    /* TODO: Add class to HR */
+    hr {
+      background-color: var(--spectrum-textfield-placeholder-text-color);
+      border: 0;
+      margin: 15px 0px;
```
NateBaldwinDesign: Do not use hard pixel values as they will not scale

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+      height: 1px;
+      width: 10px;
```
NateBaldwinDesign: image

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+      border-right: 0;
+      border-bottom: 0;
+    }
+    /* Horisontal line */
+    /* TODO: Add class to HR */
+    hr {
+      background-color: var(--spectrum-textfield-placeholder-text-color);
+      border: 0;
+      margin: 15px 0px;
+      height: 1px;
```
NateBaldwinDesign: use `var(--spectrum-alias-border-size-thin)`

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+      width: 10px;
+      position: absolute;
+      top: 0;
+      left: calc(50% - 27px); /*22px half button width and 5px hald of HR width*/
```
NateBaldwinDesign: also should be calculated using spectrum tokens rather than pixels so that it scales properly

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+    }
+    /* Range */
+    /* TODO: use defined variable for width and padding */
+    &.spectrum-Datepicker--range {
+      /* Inputs */ 
+      .spectrum-InputGroup-field {
+        width: 100px;
+        min-width: 100px;
```
NateBaldwinDesign: These widths don't seem appropriate. We would hope that the minimum width for each input would relate to the date format being used, which may not be able to be expressed here (fyi @GarthDB )

To that I would suggest leaving this out since textfields default min-width should already be set to 72px.

Ultimately we would hope that the spacing between the HR and the inputs always be the mid-way point, even if the field has a fixed width.
image

src/inputgroup/skin.css
GarthDB: It would get converted to rems, but do you mean using the scaling tokens from dna?

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+        &:nth-of-type(1) {
+          padding-right: 12px;
```
NateBaldwinDesign: use `var(--spectrum-global-dimension-size-150)`

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+    &.spectrum-Datepicker--range {
+      /* Inputs */ 
+      .spectrum-InputGroup-field {
+        width: 100px;
+        min-width: 100px;
+        &:nth-of-type(1) {
+          padding-right: 12px; 
+        }
+        &:nth-of-type(2) {
+          padding-left: 12px;
```
NateBaldwinDesign: use `var(--spectrum-global-dimension-size-150)`

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+  &.spectrum-Datepicker {
+    /** Normal **/
+    &:not(.spectrum-InputGroup--quiet) {
+      border: 1px solid var(--spectrum-textfield-border-color);
+    }
+    /** Quiet **/
+    &.spectrum-InputGroup--quiet {
+      border-bottom: 1px solid var(--spectrum-textfield-border-color);
+      hr {
+        left: calc(50% - 21px); /*22px half button width and 5px hald of HR width*/
```
NateBaldwinDesign: should not use pixel values or it won't scale properly. Be sure to test locally by swapping themes and scales in order to see this in action (note the location of the <hr>):
image

src/inputgroup/skin.css
```diff
@@ -86,3 +86,89 @@
...
+*/
+.spectrum-InputGroup {
+  &.spectrum-Datepicker {
+    /** Normal **/
+    &:not(.spectrum-InputGroup--quiet) {
+      border: 1px solid var(--spectrum-textfield-border-color);
+    }
+    /** Quiet **/
+    &.spectrum-InputGroup--quiet {
+      border-bottom: 1px solid var(--spectrum-textfield-border-color);
```
NateBaldwinDesign: Can't this be inherited from quiet textfields? @lazd ?

src/inputgroup/skin.css
NateBaldwinDesign: That’s correct it should be driven by DNA size tokens

src/inputgroup/skin.css
majornista: @NateBaldwinDesign Should this be an actual em dash added with the :before pseudo-selector instead?

src/inputgroup/skin.css
majornista: @NateBaldwinDesign Adjusting the margin of the em dash based on the text width of the first input's value or placeholder requires a javascript method to execute when the value changes. Should we add a utility method to the docs in order to do this?

GarthDB:
Possible duplicate of https://github.com/adobe/spectrum-css/pull/53

GarthDB:
Closing as the work has moved to #53

PR #25, opened 2018-11-21
RSP-674: Add option to disable items in the TreeView component

davidjocic93:


NateBaldwinDesign:
@GarthDB this seems reasonable to include in case a tree node needs to be disabled. I've commented with the variables that should be used

PR #36, opened 2018-12-04
Fix stray & in build output due to assumptions about nesting

### LINKED ISSUE #34: CSS output for typography contains stray &
### Expected Behavior

Shouldn't have a stray `&`, break everything.

### Actual Behavior

Has a stray `&`, breaks everything.

#### Steps to Reproduce

1. Build project
2. Look at `dist/components/typography/index.css`
3. See this:
```
&.spectrum,
&.spectrum-Body,
.spectrum,
.spectrum-Body {
  font-size: 14px;
  font-weight: 400;
  line-height: 1.5;
  font-style: normal;
}
```
4. Cry

#### Spectrum-CSS version

2.6.0

lazd:
## Description

This PR adds a quick postcss plugin that removes selectors that make assumptions about nesting. This enables us to add behaviors that only apply when the interface is scaled. 

This PR also reverts d65d0ab2827285fbb5e931eb4270301f6c8cd786 (which fixes #34, but breaks 94ade838d3e84bb2279f6d71e6e7ecfcfa8443f2).

## Related Issue

#34 

## Motivation and Context

94ade838d3e84bb2279f6d71e6e7ecfcfa8443f2 introduced a fix that increased the specificity of a selector such that large interfaces would get the correct font size, but it broke this for non-scaled interfaces as the assumption of nesting resulted in an invalid selector.

This change removes the invalid selectors, making it possible to safely assume nesting in selectors.

After this change, if you use an `&` in `index.css` as follows:

```css
&.spectrum-Body,
.spectrum-Body {
  font-size: var(--spectrum-body-4-text-size);
}
```

For the standalone medium scale, this compiles to:

```css
.spectrum-Body {
  font-size: 14px;
}
```

And for the standalone large scale, this compiles to:

```css
.spectrum-Body {
  font-size: 17px;
}
```

However, for the diff scale (which is applied after medium), we get:
```css
.spectrum--large.spectrum-Body,
.spectrum--large .spectrum-Body {
  font-size: 17px;
}
```

This means the rule now takes effect for both a`.spectrum-Body` nested within `.spectrum--large` AND when both `.spectrum-Body` and `.spectrum--large` are present on the same element, which fixes the original issue 94ade838d3e84bb2279f6d71e6e7ecfcfa8443f2 was trying to fix without causing the issue described in #34.

Additionally, inside of `index.css` for any component, you can now write:

```css
&.spectrum-Component {
  width: var(--spectrum-global-dimension-size-6000); // must use a variable, otherwise the prop is stripped
}
```

And this will only show up if the component is scaled -- the rule is completely omitted for the standard, medium scale. I don't have a reason why we'd need to do this right now, but it is an interesting side effect of this fix.

Yes, this is disgusting. Yes, there is probably a better way. No, it is not a priority to explore that right now.

## How Has This Been Tested?

I diffed the output of the project against output generated before this commit to ensure the only thing that changed was the bug in question.

## Types of changes

### DIFF
--- a/tasks/build-css.js
+++ b/tasks/build-css.js
@@ -89,6 +89,7 @@ var processors = [
     noValueNotifications: 'error'
   }),
   require('./lib/postcss-custom-properties-passthrough')(),
+  require('./lib/postcss-notnested')(),
   require('postcss-calc'),
   require('postcss-svg'),
   require('postcss-functions')({
@@ -318,7 +319,7 @@ gulp.task('build-css:concat-standalone-md', function() {
     ])
       // .pipe(concat('spectrum-' + colorStop + '-md.css'))
       .pipe(concat('spectrum-' + colorStop + '.css'))
-      // Replace instances of & that refer to the colorstop selector with .secptrum
+      // Replace instances of & that refer to the colorstop selector with .spectrum
       .pipe(replace(/^&/gm, '.spectrum'))
       .pipe(gulp.dest('dist/standalone'));
   }
@@ -333,7 +334,7 @@ gulp.task('build-css:concat-standalone-lg', function() {
       'dist/spectrum-' + colorStop + '.css'
     ])
       .pipe(concat('spectrum-' + colorStop + '-lg.css'))
-      // Replace instances of & that refer to the colorstop selector with .secptrum
+      // Replace instances of & that refer to the colorstop selector with .spectrum
       .pipe(replace(/^&/gm, '.spectrum'))
       .pipe(gulp.dest('dist/standalone'));
   }
--- a/tasks/build-css.js
+++ b/tasks/build-css.js
@@ -7,6 +7,7 @@ var insert = require('gulp-insert');
 var concat = require('gulp-concat');
 var merge = require('merge-stream');
 var plumb = require('./lib/plumb.js');
+var notnested = require('./lib/postcss-notnested');
 
 var colorStops = [
   'darkest',
@@ -89,7 +90,6 @@ var processors = [
     noValueNotifications: 'error'
   }),
   require('./lib/postcss-custom-properties-passthrough')(),
-  require('./lib/postcss-notnested')(),
   require('postcss-calc'),
   require('postcss-svg'),
   require('postcss-functions')({
@@ -122,7 +122,7 @@ gulp.task('build-css:individual-components-md', function() {
     .pipe(plumb())
     .pipe(insert.prepend('@import "../../vars/spectrum-medium.css";'))
     .pipe(insert.prepend('@import "../../vars/spectrum-global.css";\n'))
-    .pipe(postcss(processors))
+    .pipe(postcss(processors.concat([notnested()])))
     // .pipe(rename(function(path) {
     //   path.basename += '-md';
     // }))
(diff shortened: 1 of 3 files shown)


tasks/build-css.js
```diff
@@ -333,7 +334,7 @@ gulp.task('build-css:concat-standalone-lg', function() {
       'dist/spectrum-' + colorStop + '.css'
     ])
       .pipe(concat('spectrum-' + colorStop + '-lg.css'))
-      // Replace instances of & that refer to the colorstop selector with .secptrum
```
lazd: Secptrum is my new band name 🤘

tasks/lib/postcss-notnested.js
```diff
@@ -0,0 +1,25 @@
...
+    root.walkRules((rule, ruleIndex) => {
+      var matchFound = false;
+      if (rule.selectors) {
+        var selectors = rule.selectors.filter(selector => {
+          // Kill the selector with the stray ampersand -- it's not nested!
+          return !re.test(selector)
+        });
+
+        // Only replace the selectors if we changed something (avoids extra work for every selector)
+        if (selectors.length != rule.selectors.length) {
```
lazd: This avoids running this code for every single selector in the project: https://github.com/postcss/postcss/blob/8157278c743564555a22da67da2b5972945462d8/lib/rule.es6#L42-L46

lazd: It also seems like, if length goes to zero, we should remove the rule completely... That way, we can add a single selector that assumes nesting, and the rule is killed if its not nested...

tasks/build-css.js
```diff
@@ -318,8 +319,6 @@ gulp.task('build-css:concat-standalone-md', function() {
     ])
       // .pipe(concat('spectrum-' + colorStop + '-md.css'))
       .pipe(concat('spectrum-' + colorStop + '.css'))
-      // Replace instances of & that refer to the colorstop selector with .secptrum
-      .pipe(replace(/^&/gm, '.spectrum'))
```
lazd: This is no longer needed as the generated file already has `&` replaced by `.spectrum`.

tasks/build-css.js
```diff
@@ -333,8 +332,6 @@ gulp.task('build-css:concat-standalone-lg', function() {
       'dist/spectrum-' + colorStop + '.css'
     ])
       .pipe(concat('spectrum-' + colorStop + '-lg.css'))
-      // Replace instances of & that refer to the colorstop selector with .secptrum
-      .pipe(replace(/^&/gm, '.spectrum'))
```
lazd: This is no longer needed as the generated file already has `&` replaced by `.spectrum`.

tasks/lib/postcss-notnested.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = postcss.plugin('postcss-notnested', function (opts) {
+  opts = opts || {};
+
+  // Match ampersands at the start of a given selector
+  var re = /^&.*/;
+
+  return function (root, result) {
+    root.walkRules((rule, ruleIndex) => {
+      if (rule.selectors) {
+        if (opts.replace) {
```
lazd: If `opts.replace` is passed, stray `&` are replaced with the passed string.

tasks/lib/postcss-notnested.js
```diff
@@ -0,0 +1,46 @@
...
+            else {
+              return selector;
+            }
+          });
+
+          if (replaced) {
+            rule.selectors = selectors;
+          }
+        }
+        else {
```
lazd: Otherwise, all selectors that use `&` are killed, and if no selectors remain, the entire rule is killed.

Issue #44, opened 2018-12-09
Create release scripts for npm and artifactory releases


stevengill:
Need to document and automate the release process for both public and private 

@lazd's current steps are 

```
npm install @spectrum/dna-npm-release@1.x @spectrum/spectrum-icons@2.0.0
git checkout package.json package-lock.json
npm run release # publishes to Artifactory as @adobe/spectrum-css, even though we don't really want to do that, but also creates tags and stuff
# edit package.json to change @adobe to @spectrum
# publish to artifactory as @spectrum/spectrum-css
npm publish --registry=https://artifactory.corp.adobe.com:443/artifactory/api/npm/npm-spectrum-release-local
# now we're ready to publish to public
npm publish
```

stevengill:
Related: https://github.com/adobe/spectrum-css/pull/56

PR #45, opened 2018-12-09
Added new release scripts

stevengill:
## Description
Automated some of the release process. Check out the readme changes for the new release process. 

Created two scripts. 

The first script handles updating the version, tagging, committing, pushing commits and tags to git. We may want to split out the pushing to git part into its own command. 

The second script handles updating the package scope in `package.json` for the artifactory release. It also installs the internal only dependency of spectrum-icons. 

I don't have anything about docs release in this yet.

## Related Issue
https://github.com/adobe/spectrum-css/issues/44

## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,3 +283,43 @@ To note: the JenkinsCI server will almost always have a green check mark in the 
 Check out the [contributing guidelines](.github/CONTRIBUTING.md)!
 
 [topdoc-link]: https://github.com/Topdoc/topdoc/wiki
+
+## Releasing
+
+We are currently releasing this pacakage on `npm`. We are also releasing an internal version of this package for Adobe on an internal instance of Artifactory. 
+
+First, prepare the package for release by bumping the version (minor version bump by default), committing the changes, tagging the repo, and pushing the changes and tags up to GitHub (origin remote). We have a command that does this for you. 
+
+```
+npm run bump (patch/minor/major)
+```
+
+This command takes an optional argument of the type of version increase you want. You can pass either patch, minor or major. If you don't supply an argument, the command will do a minor version bump by default. You could also pass a version you want the package to be. Ex: `npm run bump 5.0.0`.
+
+Now that we have prepared the repo for publishing, let's publish the package to `npm`! Run the following command. 
+
+```
+npm publish
+```
+
+Now let's prepare to release to artifactory. For legacy reasons, we need to update the scope from `@adobe` to `@spectrum` and add the internal only `@spectrum/spectrum-icons` dependency. We have a command that does this for you.
+
+```
+npm run prepare-artifactory
+```
+
+Now let's publish the package to artifactory. Run the following command.
+
+```
+npm publish --registry=https://artifactory.corp.adobe.com:443/artifactory/api/npm/npm-spectrum-release-local
+```
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -283,3 +283,43 @@ To note: the JenkinsCI server will almost always have a green check mark in the
 Check out the [contributing guidelines](.github/CONTRIBUTING.md)!
 
 [topdoc-link]: https://github.com/Topdoc/topdoc/wiki
+
+## Releasing
+
+We are currently releasing this pacakage on `npm`. We are also releasing an internal version of this package for Adobe on an internal instance of Artifactory.
```
filmaj:
```suggestion
We are currently releasing this package on `npm`. We are also releasing an internal version of this package for Adobe on an internal instance of Artifactory. 
```

scripts/prepare-release.js
```diff
@@ -0,0 +1,26 @@
...
+if(!version) {
+	version = 'minor'
+}
+
+// bump version & tag
+console.log(version);
+shell.exec(`npm version ${version} -m 'Release %s' `)
+
+// push up to github, assumes origin is where you want to push
+shell.exec(`git push origin && git push --tags`)
```
filmaj: Instead of assuming `origin` as the remote name, can we parse the correct remote name out based on URL? Something like:

```sh
for remote in $(git remote);
do
    if [[ $(git remote get-url $remote) == *"adobe/spectrum-css"* ]]; then
        echo "the adobe github remote is under the $remote name";
    fi
done
```

... yields the following in my repo w/ github.com/adobe/spectrum-css under 'origin':

> the adobe github remote is under the origin name

README.md
```diff
@@ -283,3 +283,43 @@ To note: the JenkinsCI server will almost always have a green check mark in the
 Check out the [contributing guidelines](.github/CONTRIBUTING.md)!
 
 [topdoc-link]: https://github.com/Topdoc/topdoc/wiki
+
+## Releasing
+
+We are currently releasing this pacakage on `npm`. We are also releasing an internal version of this package for Adobe on an internal instance of Artifactory. 
+
+First, prepare the package for release by bumping the version (minor version bump by default), committing the changes, tagging the repo, and pushing the changes and tags up to GitHub (origin remote). We have a command that does this for you.
```
filmaj: Assuming my suggested change around parsing the remote is incorporated, let's make sure to remove the note around the origin remote in here.

scripts/prepare-release.js
lazd: Oh dang this will push all the tags, so if we have any local/wrong tags, they'll go too. We should only push the tag we created for this release.

stevengill: Fixed to only push the release tag now

stevengill: It will also figure out the correct remote name to push to (finds the one that has `adobe/spectrum-css` in it.

stevengill:
Made some updates to this script @filmaj @lazd

stevengill:
Still don't have anything about docs release in these scripts yet. But will do an additional PR for that stuff once I go through the next release.

stevengill:
Merging in. We can make changes later

PR #49, opened 2018-12-12
Typography Margins Fixed

GarthDB:
added a means for the margins to be returned to the typography components and cleaned up docs examples

### DIFF
--- a/docs/typography/typography-body.yml
+++ b/docs/typography/typography-body.yml
@@ -5,7 +5,7 @@ description: Default body text sizes.
 status: Verified
 ignoreDNA: true
 markup: |
-  <div class="spectrum" lang="en">
+  <div class="spectrum-Typography" lang="en">
     <p class="spectrum-Body1">Body1 Text <em>Body1 Emphasis</em> <strong>Body1 Strong</strong>.</p>
     <p class="spectrum-Body2">Body2 text <em>Body2 Emphasis</em> <strong>Body2 Strong</strong>.</p>
     <p class="spectrum-Body3">Body3 Text <em>Body3 Emphasis</em> <strong>Body3 Strong</strong>.</p>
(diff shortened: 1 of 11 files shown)


GarthDB:
@devongovett let me know if this works for you.

lazd:
This definitely gets the job done, but I can't help but feel we've introduced overly complex API to put margins on headings. It's not obvious, there are two ways to do it (with both mechanisms not entirely true to SUIT), the number of selectors is huge, and the duplication of properties between selectors results in a lot of bloat.

We need to re-approach this. Perhaps a simpler fix is to:

1. Drop `.spectrum-Typography` completely
2. Introduce `u-noMargin { margin: 0; }`
3. Restore margins to `.spectrum-Heading*`

Then, if people want a properly styled heading, they simply use the class:

```html
<h2 class="spectrum-Heading2">Lorem Ipsum Dolor</h2>
```

If they want a heading without margin, they add the utility class:

```html
<h2 class="spectrum-Heading2 u-noMargin">Lorem Ipsum Dolor</h2>
```

GarthDB:
Seems very similar, add a class to add margins or add a class to get rid of margins.

GarthDB:
The majority uses should need the margins unless you’re using the margins in a whole block of typography, which is why we have the wrapper class and no margins by default.

lisekyle:
Another drive-by comment: For those projects that have already implemented the typography, having margins is going to cause **a lot** of unnecessary regressions (this happened to me when I updated spectrum-css and the Spectrum website basically exploded). In my opinion, our components should be as modular as possible, which means they come out of the box with no extra styling. We should give our customers an option to use their own custom margins or the margins we suggest based on what they need.

GarthDB:
> Drive by comment: whenever I start a new project, 99% of the time I forget to add the `spectrum-Typography` class to a parent element. So I am in favour of 🔪 that class 😄

So does that mean you want the margins, or do you have any trouble with them being on by default?

filmaj:
If I may be honest, I don't have an opinion on margins. I'd rather be told what my opinion on margins should be. I was making more of a comment on what trips me up w/ spectrum setup than whether or not stylistic specifics clash with my own opinions.

GarthDB:
@filmaj the way it’s designed in this pr is that mother has margins by default so if you want to use a heading or some body text you can have those styles without it impacting layout with things around it. Where you likely do want margins is when you have headings and body together which is why there’s a wrapper provided to give them all margins.

PR #53, opened 2018-12-12
✨ RSP-631 InputGroup: add Datepicker range variant

majornista:
Add Datepicker range variants to InputGroup to support date and datetime range.

## Description
Add `.spectrum-Datepicker--range` and `.spectrum-Datepicker--datetimeRange` to `.spectrum-InputGroup` style definitions.

Update docs to support adding `.is-focused` style to InputGroup on focus, similar to how we do it for `.spectrum-Slider`. 

## Related Issues
#54 
[✨ RSP-631: Datepicker range styles added](https://github.com/adobe/spectrum-css/pull/13)
Spectrum/spectrum-css/issues/663
JIRA issue RSP-631

## Motivation and Context
Calendar supports range selection but Datepicker component does not display the date range.

## How Has This Been Tested?
Tested in Spectrum docs and as part of a related React-Spectrum PR.

Datepicker Range Input Group
Datepicker Range Input Group (Quiet variant)

## Types of changes

## Checklist:

### DIFF
--- a/docs/inputgroup/datepicker-range-quiet.yml
+++ b/docs/inputgroup/datepicker-range-quiet.yml
@@ -2,9 +2,9 @@ name: Date Picker - Range Quiet
 description: A date picker uses the input group component to display a field with a button next to it
 markup: |
   <div class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range" role="combobox" aria-expanded="false" aria-haspopup="dialog">
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--start" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
     <div class="spectrum-Datepicker--rangeDash"></div>
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="end" value="2018-10-30">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--start" placeholder="mm/dd/yyyy" name="end" value="2018-10-30">
     <button type="button" class="spectrum-FieldButton spectrum-FieldButton--quiet" tabindex="-1" aria-label="Calendar">
       <svg class="spectrum-Icon spectrum-UIIcon-ChevronDownMedium spectrum-Dropdown-icon" focusable="false" aria-hidden="true">
         <use xlink:href="#spectrum-css-icon-ChevronDownMedium" />
@@ -17,9 +17,9 @@ markup: |
   <p />
 
   <div aria-invalid="true" class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range is-invalid" role="combobox" aria-expanded="false" aria-haspopup="dialog">
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" aria-invalid="true" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--start" aria-invalid="true" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
     <div class="spectrum-Datepicker--rangeDash"></div>
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field is-invalid" aria-invalid="true" placeholder="mm/dd/yyyy" name="end" value="2018-10-30">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--end is-invalid" aria-invalid="true" placeholder="mm/dd/yyyy" name="end" value="2018-10-30">
     <button type="button" class="spectrum-FieldButton spectrum-FieldButton--quiet is-invalid" tabindex="-1" aria-expanded="false" aria-haspopup="dialog" aria-label="Calendar">
       <svg class="spectrum-Icon spectrum-UIIcon-ChevronDownMedium spectrum-Dropdown-icon" focusable="false" aria-hidden="true">
         <use xlink:href="#spectrum-css-icon-ChevronDownMedium" />
@@ -32,9 +32,9 @@ markup: |
   <p />
 
   <div aria-disabled="true" class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range is-disabled" role="combobox" aria-expanded="false" aria-haspopup="dialog">
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="start" value="2018-10-01" disabled>
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--start" placeholder="mm/dd/yyyy" name="start" value="2018-10-01" disabled>
     <div class="spectrum-Datepicker--rangeDash"></div>
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="end" value="2018-10-30" disabled>
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--end" placeholder="mm/dd/yyyy" name="end" value="2018-10-30" disabled>
     <button type="button" class="spectrum-FieldButton spectrum-FieldButton--quiet" tabindex="-1" disabled aria-expanded="false" aria-haspopup="dialog" aria-label="Calendar">
       <svg class="spectrum-Icon spectrum-UIIcon-ChevronDownMedium spectrum-Dropdown-icon" focusable="false" aria-hidden="true">
         <use xlink:href="#spectrum-css-icon-ChevronDownMedium" />
@@ -46,9 +46,9 @@ markup: |
   <p/>
 
   <div class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range  spectrum-Datepicker--datetimeRange" role="combobox" aria-expanded="false" aria-haspopup="dialog">
-    <input type="text" class="spectrum-Textfield  spectrum-Textfield--quiet spectrum-InputGroup-field" aria-invalid="false" placeholder="mm/dd/yyyy hh:mm a" name="start" value="">
+    <input type="text" class="spectrum-Textfield  spectrum-Textfield--quiet spectrum-InputGroup-field spectrum-Datepicker-field--start" aria-invalid="false" placeholder="mm/dd/yyyy hh:mm a" name="start" value="">
     <div class="spectrum-Datepicker--rangeDash"></div>
-    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet  spectrum-InputGroup-field" aria-invalid="false" placeholder="mm/dd/yyyy hh:mm a" name="end" value="">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet  spectrum-InputGroup-field spectrum-Datepicker-field--end" aria-invalid="false" placeholder="mm/dd/yyyy hh:mm a" name="end" value="">
     <button type="button" class="spectrum-FieldButton spectrum-FieldButton--quiet" tabindex="-1" aria-expanded="false" aria-haspopup="dialog" aria-label="Calendar">
       <svg class="spectrum-Icon spectrum-UIIcon-ChevronDownMedium spectrum-Dropdown-icon" focusable="false" aria-hidden="true">
         <use xlink:href="#spectrum-css-icon-ChevronDownMedium" />
--- a/docs/inputgroup/datepicker-range-quiet.yml
+++ b/docs/inputgroup/datepicker-range-quiet.yml
@@ -3,7 +3,7 @@ description: A date picker uses the input group component to display a field wit
 markup: |
   <div class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range" role="combobox" aria-expanded="false" aria-haspopup="dialog">
     <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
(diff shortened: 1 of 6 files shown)


src/inputgroup/index.css
```diff
@@ -63,3 +73,74 @@
...
+      }
+    }
+    &:nth-of-type(2) {
+      border-left: 0;
+      border-radius: 0;
+      padding-left: var(--spectrum-padding);
+    }
+  }
+  /* Em dash */
+  hr {
```
lazd: This should get a classname instead of just styling the bare element.

lazd: This should get a classname instead of just styling the bare element.

src/inputgroup/index.css
```diff
@@ -12,12 +12,20 @@
...
+{{ inputgroup/datepicker-range-quiet.yml }}
+*/
 
 :root {
   /* Todo: move to DNA */
   --spectrum-combobox-quiet-fieldbutton-border-radius: 0;
   --spectrum-combobox-field-border-width-right: 0;
   --spectrum-combobox-quiet-fieldbutton-padding-right: 0;
+  --spectrum-datepicker-input-width: calc(var(--spectrum-global-dimension-size-1600) -  2 * var(--spectrum-padding));
+  --spectrum-datepicker-datetime-input-width: calc(var(--spectrum-datepicker-input-width) + var(--spectrum-global-dimension-size-700) - 0.5em);
```
lazd: Perhaps @NateBaldwinDesign can provide guidance on the right width to use here? Also, I assume this will work fine with dates using a period or `-` separator instead of `/` (the date won't overlap the em dash)?

docs/inputgroup/datepicker-range-quiet.yml
```diff
@@ -0,0 +1,47 @@
...
+markup: |
+  <div class="spectrum-InputGroup spectrum-InputGroup--quiet spectrum-Datepicker spectrum-Datepicker--range" role="combobox" aria-expanded="false" aria-haspopup="dialog">
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="start" value="2018-10-01">
+    <hr />
+    <input type="text" class="spectrum-Textfield spectrum-Textfield--quiet spectrum-InputGroup-field" placeholder="mm/dd/yyyy" name="end" value="2018-10-30">
+    <button type="button" class="spectrum-FieldButton spectrum-FieldButton--quiet" tabindex="-1">
+      <svg viewBox="0 0 36 36" focusable="false" aria-hidden="true" role="img" class="spectrum-Icon spectrum-Icon--sizeS"><path d="M33 6h-5V3a1 1 0 0 0-1-1h-2a1 1 0 0 0-1 1v3H10V3a1 1 0 0 0-1-1H7a1 1 0 0 0-1 1v3H1a1 1 0 0 0-1 1v26a1 1 0 0 0 1 1h32a1 1 0 0 0 1-1V7a1 1 0 0 0-1-1zm-1 26H2V8h4v1a1 1 0 0 0 1 1h2a1 1 0 0 0 1-1V8h14v1a1 1 0 0 0 1 1h2a1 1 0 0 0 1-1V8h4z"></path><path d="M6 12h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4zM6 18h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4zM6 24h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4z"></path></svg>
```
lazd: Please use `<use>` for icons instead of pasting the paths.

src/inputgroup/skin.css
```diff
@@ -86,3 +102,77 @@
...
+        box-shadow: 0 0 0 1px var(--spectrum-dropdown-border-color-error);
+      }
+    }
+  }
+
+  &.is-invalid {
+    .spectrum-InputGroup-field {
+      border-color: var(--spectrum-dropdown-border-color-error) !important;
+    }
+    :focus-ring {
```
lazd: I think this needs to be `&:focus-ring`?

src/inputgroup/skin.css
```diff
@@ -86,3 +102,77 @@
...
+    }
+    &[disabled] {
+      ~ hr {
+        color: var(--spectrum-textfield-text-color-disabled);
+      }
+    }
+  }
+
+  /* Focus ring */
+  :focus-ring {
```
lazd: This might also need to be `&:focus-ring`?

docs/inputgroup/datepicker-range.yml
```diff
@@ -0,0 +1,48 @@
...
+    <input type="text" class="spectrum-Textfield spectrum-InputGroup-field" placeholder="mm/dd/yyyy hh:mm a" name="start" value="">
+    <hr>
+    <input type="text" class="spectrum-Textfield spectrum-InputGroup-field" placeholder="mm/dd/yyyy hh:mm a" name="end" value="">
+    <button type="button" class="spectrum-FieldButton" tabindex="-1" aria-expanded="false" aria-haspopup="dialog">
+      <svg viewBox="0 0 36 36" focusable="false" aria-hidden="true" role="img" class="spectrum-Icon spectrum-Icon--sizeS"><path d="M33 6h-5V3a1 1 0 0 0-1-1h-2a1 1 0 0 0-1 1v3H10V3a1 1 0 0 0-1-1H7a1 1 0 0 0-1 1v3H1a1 1 0 0 0-1 1v26a1 1 0 0 0 1 1h32a1 1 0 0 0 1-1V7a1 1 0 0 0-1-1zm-1 26H2V8h4v1a1 1 0 0 0 1 1h2a1 1 0 0 0 1-1V8h14v1a1 1 0 0 0 1 1h2a1 1 0 0 0 1-1V8h4z"></path><path d="M6 12h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4zM6 18h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4zM6 24h4v4H6zm6 0h4v4h-4zm6 0h4v4h-4zm6 0h4v4h-4z"></path></svg>
+    </button>
+    <div class="spectrum-Datepicker-focusRing" role="presentation"></div>
```
lazd: Can we accomplish this with `:before`? I'd rather not introduce another element/classname if we can do it with a pseudo-element.

topdoc/resources/js/docs.js
```diff
@@ -84,16 +84,30 @@ window.addEventListener('click', function(event) {
   }
 });
 
-// Display Slider focus style
-function toggleSliderFocus(event) {
```
lazd: Was this supposed to be nuked? Seems like you meant to add `toggleInputGroupFocus`, not replace this.

src/inputgroup/skin.css
majornista: In this case, we're creating a selector for `.spectrum-Datepicker-focusRing` following and adjacent to any keyboard focused sibling. `.spectrum-Datepicker--range :focus-ring ~ .spectrum-Datepicker-focusRing`. The either of the `input` elements or the `button` can have `:focus-ring`.

docs/inputgroup/datepicker-range.yml
majornista: `:after` would apply to `.spectrum-Datepicker--range` and we can't able to apply `:focus-ring ~` style to a pseudo element on the parent.

src/inputgroup/skin.css
majornista: As above In this case, we're creating a selector for `.spectrum-Datepicker-focusRing` following and adjacent to any keyboard focused sibling. `.spectrum-Datepicker--range :focus-ring ~ .spectrum-Datepicker-focusRing`. Either of the `input` elements or the `button` can have `:focus-ring`.

topdoc/resources/js/docs.js
majornista: `toggleInputGroupFocus` works for Slider as well: https://github.com/adobe/spectrum-css/pull/53/files#diff-85f7fb7fa03b4cdf706f93ebc1b243deR97

src/inputgroup/skin.css
lazd: Ah ha! Clever. It would be nice to comment about that so future do-gooders don't undo your goodness.

docs/inputgroup/datepicker-range.yml
lazd: Got it, makes sense.

topdoc/resources/js/docs.js
lazd: Godliness.

src/inputgroup/index.css
```diff
@@ -63,3 +74,77 @@
...
+    &:nth-of-type(1) {
+      border-right: 0;
+      padding-right: var(--spectrum-padding);
+      &.is-invalid,
+      &:invalid {
+        background-image: none;
+        padding-right: var(--spectrum-padding);
+      }
+    }
+    &:nth-of-type(2) {
```
lazd: @GarthDB how does this make you feel? Do you think we should have selectors for these two bad boys instead of doing `nth-of-type`?

lazd:
@majornista almost good to go! Just those icons remaining, and pending some feedback from @GarthDB on the `nth-of-type` situation.

GarthDB:
@majornista is this a duplicate of https://github.com/adobe/spectrum-css/pull/13 ?

src/inputgroup/index.css
GarthDB: It has full browser support, so I don't really have any issue with it. https://caniuse.com/#feat=css-sel3

GarthDB: The only other reason we might be concerned is performance and/or code readability/maintainability, but I think it's fine on both of those.

majornista:
@GarthDB Yes, this supercedes #13.

majornista:
> @majornista is this a duplicate of #13 ?

Yes, this supersedes #13.

GarthDB:
@majornista perfect, I'll close that one in favor of this.

majornista:
@GarthDB and @lazd Where are we with this one?

src/inputgroup/index.css
```diff
@@ -63,3 +76,84 @@
...
+      border-left: 0;
+      border-radius: 0;
+      padding-left: var(--spectrum-padding);
+    }
+  }
+  /* Em dash */
+  .spectrum-Datepicker--range-dash {
```
lazd: This should be `.spectrum-Datepicker--rangeDash`

src/inputgroup/skin.css
```diff
@@ -86,3 +102,79 @@
...
   }
 }
+
+.spectrum-Datepicker--range {
+  .spectrum-InputGroup-field {
+    &:focus-ring {
+      box-shadow: none !important;
+    }
+    &[disabled] {
+      ~ hr {
```
lazd: Can we give the `<hr>` a classname and target that?

src/inputgroup/index.css
lazd: I think the specificity is pretty high here. I'm not 100% against it, but it feels like we could throw a class on  the inputs and simplify the CSS... The real question is: do we want markup simplicity (the PR, as-is), or CSS simplicity (throw an extra class on the fields)?

src/inputgroup/skin.css
majornista: I there shouldn't be an `<hr>`, just `.spectrum-Datepicker--range-dash`.  I think this is a holdover from an earlier version.

docs/inputgroup/datepicker-range.yml
```diff
@@ -0,0 +1,60 @@
...
+    <input type="text" class="spectrum-Textfield spectrum-InputGroup-field" aria-invalid="true" placeholder="mm/dd/yyyy" name="start" value="">
+    <div class="spectrum-Datepicker--rangeDash"></div>
+    <input type="text" class="spectrum-Textfield spectrum-InputGroup-field is-invalid" aria-invalid="true" placeholder="mm/dd/yyyy" name="end" value="">
+    <button type="button" class="spectrum-FieldButton is-invalid" tabindex="-1" aria-expanded="false" aria-haspopup="dialog" aria-label="Calendar">
+      <svg class="spectrum-Icon spectrum-UIIcon-ChevronDownMedium spectrum-Dropdown-icon" focusable="false" aria-hidden="true">
+        <use xlink:href="#spectrum-css-icon-ChevronDownMedium" />
+      </svg>
+    </button>
+    <!-- To render focus ring around entire input group when one of the inputs has keyboard focus, we need this: -->
+    <div class="spectrum-Datepicker-focusRing" role="presentation"></div>
```
lazd: Mmmm, this can't be done with `:before`/`:after`? I don't see anything using those pseudo-elements right now.

src/inputgroup/index.css
```diff
@@ -63,3 +76,84 @@
...
+      margin: 0 var(--spectrum-datepicker-range-dash-margin-left);
+      overflow: hidden;
+      text-align: center;
+      vertical-align: middle;
+      width: var(--spectrum-global-dimension-static-font-size-100);
+    }
+  }
+  /* Focus ring */
+  &.is-focused {
+    .spectrum-Datepicker-focusRing {
```
lazd: Can we try this as `:before`?

docs/inputgroup/datepicker-range.yml
lazd: @majornista oh, I see -- it's being selected with the sibling selector, so we can't use pseudo-elements. Perhaps, instead, we should be programmatically applying `.is-focused` to the outer `.spectrum-Datepicker` and using `:focus-within` (for futureproofing)

majornista: Only with additional javascript to check for `.focus-ring` className on a descendant element. And, we'd probably need an extra `.focus-ring` state, to distinguish between mouse a keyboard focus. With the added element, the style can be applied with pure CSS.

lazd: Got it, so we would need to know "is a child element keyboard focused?" in JavaScript, and a separate class for keyboard vs mouse focus. @GarthDB thoughts on this? This is the first component that has anything like this.

GarthDB: I prefer pure CSS for accessibility things like keyboard focus. It would be preferable to implement it with the same strategy as all the others, but I'm not seeing an obvious way to do that.
We could add something to the focus-ring polyfill, but that would kind of defeat the point of a standards-based polyfill.
The focus-withing strategy would be nice, it just means we're adding another polyfill requirement to focus. What would you prefer @lazd?

lazd: Though I would prefer the polyfill, I don't want to hold up this PR. I agree with your sentiments and suppose I'm fine with the current approach.

src/inputgroup/index.css
lazd: @majornista We talked a bit with @devongovett and think these selectors should be classes added to the markup.

docs/inputgroup/datepicker-range.yml
majornista: > Got it, so we would need to know "is a child element keyboard focused?" in JavaScript, and a separate class for keyboard vs mouse focus. @GarthDB thoughts on this? This is the first component that has anything like this.

The Rating component has a similar issue with how the focus outline is applied to the wrapper element. Devon was somewhat resistant to the approach of evaluating for descendant input  with `.focus-ring`. The difference for the Datepicker is that text inputs show blue border for mouse focus and border with box-shadow for keyboard focus.

src/inputgroup/index.css
majornista: @lazd Would `spectrum-Datepicker-field--startDatetime` and `spectrum-Datepicker-field--endDatetime` work?

lazd:
Thanks for working through all of the changes, merging!

Issue #58, opened 2018-12-14
Overflow Tabs


NateBaldwinDesign:
### Expected Behavior
When there are too many tabs to fit horizontally across the viewport, the tabs component can be displayed as a quiet dropdown.

(Known to be examples only, not functional)

### Actual Behavior
Old design spec implemented where the remaining tab items are hidden under a dropdown menu in an action button.

(Known to be examples only, not functional)

#### Screenshots (if applicable)
Current CSS implementation:
![image](https://user-images.githubusercontent.com/13972198/50014624-b48fff00-ff81-11e8-801a-599dc526d962.png)

Updated design guideline:
![image](https://user-images.githubusercontent.com/13972198/50014669-ce314680-ff81-11e8-8033-ec2e199728d8.png)

PR #59, opened 2018-12-14
Tabs Overflow Example

NateBaldwinDesign:
## Description

Updated the example of how to handle overflowing tabs based on the updated guidelines. "When there are too many tabs to fit horizontally across the viewport, the tabs component can be displayed as a quiet dropdown."

## Related Issue

https://github.com/adobe/spectrum-css/issues/58

## Motivation and Context

This example is necessary to aid in proper implementations across component libraries

image

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.


lazd:
Well that's a whole heck of a lot simpler. Thanks @NateBaldwinDesign!

Harbs:
The doc was confusing to me.

The screenshot of the spec had no extra space for more tabs, but the doc does.

I'm assuming the tabs are supposed to collapse one by one as there becomes less space for them, so if there's room for three tabs and there's a total of 5, the third tab becomes a dropdown for 3, 4 and 5) as shown, while tabs 1 and 2 are normal tabs. Is this correct?

If yes, the generated docs could be clearer. (If no, they could be clearer too...) ;)

NateBaldwinDesign:
As soon as there is no room to display any single tab item, all tab items are placed within the dropdown displayed in the screenshot above. The single "tab" that is visible changes it's label depending on the selected tab item. For example, looking at the screenshot above, if the user selects "Tab 3" that respective tab content will be displayed and the Tabs component will have the label "Tab 3". This interaction is consistent with our dropdown.

Harbs:
Thanks. I'd suggest making the docs a bit clearer.

Issue #81, opened 2019-01-24
Accordion: Update implementation to follow WAI-ARIA 1.1 design pattern 


majornista:
### Expected Behavior
The WAI-ARIA design pattern for Accordion has changed from WAI-ARIA 1.0 to WAI-ARIA 1.1. 

WAI-ARIA 1.0 implemented Accordion using `role="tablist"`, which, in addition to being somewhat confusing to a keyboard user, was problematic for users of the JAWS screen reader. Basically, using `role="tablist"` on the Accordion switches JAWS into forms mode when an Accordion item header receives focus. In forms mode, content within the expanded panels cannot be accessed using the screen reader. One can exit forms mode to access panel content by pressing the Esc key, but if the Accordion happens to be a descendant of a Dialog, the Esc key will also close the Dialog. There doesn't seem to be a good workaround for this other than switching to the WAI-ARIA 1.1 design pattern that doesn't use `role="tablist"`.

In WAI-ARIA 1.1, each Accordion header is presented as a `button` element wrapped in a header, `h3` element. JAWS won't shift into forms mode when an item header receives focus and the panel content will be easier to access. 

See: https://w3c.github.io/aria-practices/#accordion for more detail.

#### Spectrum-CSS version
2.7.2

#### Sample Code that illustrates the problem
WAI-ARIA 1.0 Implementation:
```html
<div class="spectrum-Accordion" role="tablist" aria-multiselectable="true">
  <div class="spectrum-Accordion-item is-open" role="presentation">
    <div class="spectrum-Accordion-itemHeader" tabindex="0" role="tab" aria-selected="true" aria-expanded="true" aria-controls="spectrum-accordion-item-0-content" id="spectrum-accordion-item-0-header">Header 1</div>
    <svg class="spectrum-Icon spectrum-UIIcon-ChevronRightMedium spectrum-Accordion-itemIndicator" focusable="false" aria-hidden="true">
      <use xlink:href="#spectrum-css-icon-ChevronRightMedium" />
    </svg>
    <div class="spectrum-Accordion-itemContent" role="tabpanel" id="spectrum-accordion-item-0-content" aria-labelledby="spectrum-accordion-item-0-header">Item 1</div>
  </div>
  <div class="spectrum-Accordion-item is-disabled" role="presentation">
[...]

PR #83, opened 2019-01-24
Accordion: Update implementation to follow WAI-ARIA 1.1 design pattern (#81)

majornista:
## Description
Fix for Issue #81, related to RSP-786

The WAI-ARIA design pattern for Accordion has changed from WAI-ARIA 1.0 to WAI-ARIA 1.1.

WAI-ARIA 1.0 implemented Accordion using role="tablist", which, in addition to being somewhat confusing to a keyboard user, was problematic for users of the JAWS screen reader. Basically, using role="tablist" on the Accordion switches JAWS into forms mode when an Accordion item header receives focus. In forms mode, content within the expanded panels cannot be accessed using the screen reader. One can exit forms mode to access panel content by pressing the Esc key, but if the Accordion happens to be a descendant of a Dialog, the Esc key will also close the Dialog. There doesn't seem to be a good workaround for this other than switching to the WAI-ARIA 1.1 design pattern that doesn't use role="tablist".

In WAI-ARIA 1.1, each Accordion header is presented as a button element wrapped in a header, h3 element. JAWS won't shift into forms mode when an item header receives focus and the panel content will be easier to access.

See: https://w3c.github.io/aria-practices/#accordion for more detail.

## Related Issue
#81 

## Motivation and Context
Fix for Issue #81, related to RSP-786

The WAI-ARIA design pattern for Accordion has changed from WAI-ARIA 1.0 to WAI-ARIA 1.1.

WAI-ARIA 1.0 implemented Accordion using role="tablist", which, in addition to being somewhat confusing to a keyboard user, was problematic for users of the JAWS screen reader. Basically, using role="tablist" on the Accordion switches JAWS into forms mode when an Accordion item header receives focus. In forms mode, content within the expanded panels cannot be accessed using the screen reader. One can exit forms mode to access panel content by pressing the Esc key, but if the Accordion happens to be a descendant of a Dialog, the Esc key will also close the Dialog. There doesn't seem to be a good workaround for this other than switching to the WAI-ARIA 1.1 design pattern that doesn't use role="tablist".

In WAI-ARIA 1.1, each Accordion header is presented as a button element wrapped in a header, h3 element. JAWS won't shift into forms mode when an item header receives focus and the panel content will be easier to access.

See: https://w3c.github.io/aria-practices/#accordion for more detail.

## How Has This Been Tested?
Tested against React-Spectrum. Style changes should be backward compatible.

## Types of changes

## Checklist:

### DIFF
--- a/docs/accordion/accordion.yml
+++ b/docs/accordion/accordion.yml
@@ -1,33 +1,41 @@
 ---
 name: Accordion
 markup: |
-  <div class="spectrum-Accordion">
-    <div class="spectrum-Accordion-item is-open">
-      <div class="spectrum-Accordion-itemHeader" tabindex="0">Header 1</div>
-      <svg class="spectrum-Icon spectrum-UIIcon-ChevronRightMedium spectrum-Accordion-itemIndicator" focusable="false" aria-hidden="true">
-        <use xlink:href="#spectrum-css-icon-ChevronRightMedium" />
-      </svg>
-      <div class="spectrum-Accordion-itemContent">Item 1</div>
+  <div class="spectrum-Accordion" role="region">
+    <div class="spectrum-Accordion-item is-open" role="presentation">
+      <h3 class="spectrum-Accordion-itemHeading">
+        <button class="spectrum-Accordion-itemHeader" type="button" id="spectrum-accordion-item-0-header" aria-controls="spectrum-accordion-item-0-content" aria-expanded="true">Header 1</button>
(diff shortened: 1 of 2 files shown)


lazd:
Looks like it is indeed backwards compatible. @GarthDB, what's your take on this? I feel it's OK to introduce new markup, as long as the old markup works and call it a minor version.

majornista:
@GarthDB and @lazd Where are we with this one?

GarthDB:
@majornista could we include a comment in the CSS or in the markup on which markup is preferred and why?

Issue #88, opened 2019-01-29
CircleLoader animation does not work in CEP7


Harbs:
### Expected Behavior
The CircleLoader animation should work in CEP 7 and earlier
### Actual Behavior
There is no animation
### Reproduce Scenario (including but not limited to)
http://opensource.adobe.com/spectrum-css/2.7.2/docs/#circle-loader on Chromium version 41.x
#### Steps to Reproduce
Run http://opensource.adobe.com/spectrum-css/2.7.2/docs/#circle-loader on Chromium version 41.x

#### Browser name/version/os (ie Chrome Version 62.0.3202.94 (Official Build) (64-bit) MacOS)
Chromium 41.0.2272.118
#### Spectrum-CSS version
2.7.2
#### Proposed Fix
Chromium 41.x requires -webkit- prefixes for animations.

Harbs:
FYI, I tried writing a PR to fix this, but it seems like the fix needs to be in the autoprefixer (or browserlist) configuration. I have no experience with autoprefixer and I was not sure how to do that.

Harbs:
FYI 2: I did test manually adding `-webkit-` and it fixed the problem.

lazd:
@Harbs thanks for the report and the insight into the fix! It seems like the fix could be simple enough...

Harbs:
Great. There are other similar issues, but if the autoprefixer is configured to fix this, I'm pretty sure the other issues will be fixed as well.

Once a fix is in place, I'll give another go-through to see if there are still issues...

Harbs:
I found where to change the autoprefixer config and submitter a PR.

GarthDB:
@larz0 could we surface a way for people to make custom builds passing in browser preferences into autoprefixer?

With the extension developers, I suspect they'll have a very specific browser targe that no one else will.

Harbs:
@GarthDB Good idea. Requiring -webkit- prefixes for everyone is not ideal...

Harbs:
What about a "cep7" gulp script?

GarthDB:
That might be a better strategy than #90; I'm worried it might have unintended consequences.

Harbs:
Also: please see my comment here: https://github.com/adobe/spectrum-css/issues/80#issuecomment-458335335

Changing the autoprefixer does not fix that problem, and manually changes get stripped out by the css processing. I'm not sure how to fix that one...

Harbs:
I'll close #90

[...]

PR #91, opened 2019-01-30
Added CEP7 gulp task so Autoprefixer works with older Chromium

Harbs:
Added cep7 gulp task

## Description
Added a gulp task to support Chromium 41.x used in CEP7 and earlier

## Related Issue
#88 

## Motivation and Context
Many -webkit- prefixes are needed for CEP7 that are not generally needed for modern browsers. The additional gulp task allows a special build to be run for CEP panel developers.

## How Has This Been Tested?
The cep7 target was run locally and tested on Chromiuim 41.0.2272.118

## Types of changes

## Checklist:


Harbs:
There might be a simpler way to create the gulp task than what I did. I'm not exactly a gulp expert...

It also might make sense adding to the readme that CEP panel developers should run `gulp cep7` if they are developing panels for CC 2017 and earlier.

lazd:
This is very duplicative and will be hard to maintain if we ever make changes to the build script. Gulp is just JavaScript, so you can export functions, share them between tasks, etc to avoid duplication. 

That said, I would much rather see a way to pass a custom autoprefixer configuration like @GarthDB suggested than another task specific to CEP.

Harbs:
What do you think of my latest commit? Building for cep7 can be done by `gulp --cep7`

lazd:
@Harbs much better! But I think we can do even more... Maybe we can pass an arg to define the location of a package.json that contains [`browserslists`](https://github.com/browserslist/browserslist#queries)... I'm not sure what the right way to do this is... Let's explore it...

lazd:
Hey @harbs, is this something you still need? If so, let's look into the `.browserslistrc` option mentioned above. Thanks!

lazd:
Let's go ahead and close this for now, we can rebase and re-open if this is still an issue.

PR #93, opened 2019-02-01
📝 updating link to spectrum icons in readme to point to non-internal url

filmaj:
I'm not exactly sure if the wording and accuracy here is quite right, so would appreciate a review from a core contributor!

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,7 @@ Finally, you'll need workflow icons for your app. The `icons/spectrum-icons.svg`
 AdobeSpectrum.loadIcons('icons/spectrum-icons.svg');
 ```
 
-***NOTE**: the following link is internal-to-Adobe-only*
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -126,10 +126,10 @@ Note that the Spectrum CSS UI icons must change as well, see below for a full ex
 
 ### Importing Icons
 
-To get icons, you'll need to use the `AS.loadIcons()` function (also available as `AdobeSpectrum.loadIcons()`). This function lives in the `icons/AS.loadIcons.js` file and has the following signature:
+To get icons, you'll need to use the `loadIcons()` function. This function lives in the `icons/AS.loadIcons.js` file and has the following signature:
```
lazd: It's now `icons/loadIcons.js`

lazd: But maybe we should just tell them to use https://www.npmjs.com/package/loadicons ?

filmaj: Hmm.... yes maybe... so we should turf it from the repo then?

lazd: It's included in the `dist/`, but the file is actually pulled from an npm install of `loadicons`, so we're good there. I think this is fine for now!

Issue #41, opened 2018-12-06
Update Rating Component


GarthDB:
Use prototype made by @lazd in #30 

Also update tokens https://github.com/adobe/spectrum-css/issues/30#issuecomment-444572462

GarthDB:
- [ ] DNA token change pull-request
- [ ] Final design review
- [ ] Accessibility review

Issue #30, opened 2018-11-27
Prototype updated Rating component (stars)


GarthDB:
Design is provided by Mathieu Badimon from the Adobe Design Team.

lazd:
Prototype happening here: https://codepen.io/lazd/pen/WYggmO

The opensource.adobe.com domain is misconfigured so https doesn't work, so you have to visit https://opensource.adobe.com/spectrum-css/2.6.0/docs/ first, click Advanced, Proceed.

lazd:
Prototype approved by @majornista as far as accessibility!

@NateBaldwinDesign, what do you think about the tokens in the prototype? Are they organize/named properly? If so, I can send a PR to DNA.

NateBaldwinDesign:
@lazd the token names look good except the following:
`--spectrum-rating-icon-spacing` should be `--spectrum-rating-icon-gap`
`--spectrum-rating-indicator-offset` should probably be `--spectrum-rating-indicator-gap` (although this one would be fine either way)

Side note, `--spectrum-rating-indicator-height` should be global, not scale dependent (it's treated like a border).

GarthDB:
@NateBaldwinDesign, fixing the tokens, what would design want to see/do before we start working on a released version of rating?

PR #94, opened 2019-02-01
Update Rating implementation, Update to DNA 5.9.0

lazd:
## Description

Updates Rating component based on prototype in #30 

Uses DNA 5.9.0 to get the new tokens.

This introduces a new state class on `spectrum-Rating-icon`. In addition to `is-selected`, one must set `is-currentValue` on the star that represents the currently selected value. This is to enable the underline beneath the last selected star when hovering to change the rating. If this class is omitted, the component still renders correctly (sans the underline), so therefore this change is backwards compatible. Implementations will need to be informed of this change, however, so they can have full design compliance.

## Related Issue

#41 

## Motivation and Context

So much prettier.

## How Has This Been Tested?

iScrolled.exe

image
image

## Types of changes


lazd:
@GarthDB if you can give this a quick once-over, it should be an easy merge.

PR #96, opened 2019-02-05
✨RSP-629 - Fill Slider - with Offset

sakshigupta:


GarthDB:
@NateBaldwinDesign do we want to include this? Was there any design review?
image

GarthDB:
@sakshigupta what's the use case for it?

NateBaldwinDesign:
Yes, the offset typically occurs at the 50% mark, where the user can make adjustments to increase/decrease properties (such as in photography editing) starting from an origin point. 

Likewise, there are scenarios where users may apply a preset and then make adjustments to the preset configurations. In that use-case, a product may want to have the fill originate from the position on the slider defined by the preset as a way of showing the user how much they've deviated from the preset they originally selected.

GarthDB:
@devongovett are you ok with the js used here to implement the offset?

GarthDB:
@NateBaldwinDesign thanks!

devongovett:
Doesn't seem to me like we need the JS in spectrum-css for it. Could just hard code the offsets in the doc examples.

GarthDB:
@devongovett do you think this dom setup would work with how you would want to do the js in react?

devongovett:
made a simplified version in https://github.com/adobe/spectrum-css/pull/116

sakshigupta:
Closing it, as its no longer required

Issue #105, opened 2019-02-27
Card: display focus ring when a parent row element has focus


majornista:
### Expected Behavior
In a grid view, where the `.spectrum-Card` is has `role="gridcell"` within a parent element with `role="row"`, the`.spectrum-Card` should render its focus style when the parent row receives focus. 

### Actual Behavior
`.spectrum-Card[role="gridcell"]` does not account for use case where a parent element with `[role="row"]` has keyboard focus.

`.spectrum-Card` only accounts for `:focus` style on the card itself.

#### Spectrum-CSS version
v2.7.4

PR #106, opened 2019-02-27
Issue-105 Card: display focus ring when a parent row element has focus

majornista:
`.spectrum-Card` only accounts for `:focus` style on the card itself.

## Description
Currently, `.spectrum-Card` only accounts for `:focus` style on the card itself.

This PR adds `:focus-ring` styles so that the `.spectrum-Card` will render focused style when its parent has focus.

## Related Issue

#105 

## Motivation and Context
In a grid view, where the `.spectrum-Card` is has `role="gridcell"` within a parent element with `role="row"`, the`.spectrum-Card` should render its focus style when the parent row receives focus.

## How Has This Been Tested?
Tested against React-Spectrum PR #495

## Types of changes

## Checklist:

### DIFF
--- a/docs/card/card-quiet-small.yml
+++ b/docs/card/card-quiet-small.yml
@@ -4,7 +4,7 @@ description: A spectrum Quiet Card for an image
 status: Beta
 markup: |
   <div style="width: 112px; height: 136px;">
-    <div class="spectrum-Card spectrum-Card--quiet spectrum-Card--small">
+    <div class="spectrum-Card spectrum-Card--quiet spectrum-Card--small" tabindex="0" role="figure">
       <div class="spectrum-Card-preview">
         <div class="spectrum-Asset">
           <img class="spectrum-Asset-image" src="img/example-ava.jpg" style="max-width: 75%; max-height: 75%; object-fit: contain;">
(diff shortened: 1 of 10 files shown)


docs/card/card-quiet-small.yml
```diff
@@ -42,3 +42,46 @@ markup: |
       </div>
     </div>
   </div>
+  
+  <h2>Focused Card item within a grid</h3>
+  <div role="grid">
+    <div style="width: 112px; height: 136px; outline: 0;" role="row" tabindex="0" class="focus-ring">
+      <div class="spectrum-Card spectrum-Card--quiet spectrum-Card--small is-focused" role="gridcell">
```
lazd: Ok so this is interesting -- you're simulating `:focus-ring` in this example, but also applying `.is-focused` to the Card under it... Doesn't it seem like `.is-focused` should be the only thing required to make it look focused?

lazd: Ok so this is interesting -- you're simulating `:focus-ring` in this example, but also applying `.is-focused` to the Card under it... Doesn't it seem like `.is-focused` should be the only thing required to make it look focused?

src/card/index.css
```diff
@@ -189,6 +190,7 @@
     border-radius: var(--spectrum-card-quiet-border-radius);
     position: relative;
     transition: background-color var(--spectrum-global-animation-duration-100);
+    overflow: visible;
```
lazd: Specifically needed here to override something?

src/card/skin.css
```diff
@@ -95,3 +95,25 @@
     color: var(--spectrum-card-quiet-subtitle-text-color);
   }
 }
+
+:focus-ring {
```
lazd: Does this need to be under `:focus-ring` at all? Shouldn't we just leave it as `.spectrum-Card.is-focused` programmatically when the grid cell is focused and be done?

src/card/index.css
majornista: `.spectrum-Card-preview` was clipping the focus styling.

docs/card/card-quiet-small.yml
majornista: I changed this so that the Card itself receives focus, which is how it was originally implemented in React-Spectrum.

src/card/skin.css
majornista: It should be `:focus-ring` otherwise, the focused item in the grid will render with blue focus ring even on touch devices or when a user deselects a Card using the mouse.

lazd:
Looks good to me, thanks @majornista!

PR #108, opened 2019-02-28
Add a build using unique CSS variables

devongovett:
We are considering how to improve the theming support in react-spectrum for v3. Currently by default, the css we build includes all themes and scales, which is massive (885 KB). It is possible to only include the themes you use by setting environment variables, but this is hard to configure depending on your build tool (e.g. create-react-app doesn't allow it at all). It is also very manual - you need to know all of the themes you need to use in your app, including for components you might include from libraries in node_modules.

In order to improve this, we are considering just using CSS variables natively in the browser. They are supported by all browsers except IE11, and it's possible to polyfill there for those applications that still need to.

However, even the variable files exported by DNA are quite large. They currently include 2573 variables per theme, and 4630 variables per scale. Most of these variables are just aliasing other spectrum global variables though, so we can reduce this considerably.

This PR adds a new build that supports using CSS variables natively. It includes 2 parts:

1. A new variable build per theme/scale, which includes only unique variables by value. The variables above reduce to just 72 variables per theme, and 79 variables per scale. The variables are attached to the `.spectrum` element, which makes it possible to nest themes since variables will overwrite each other and be inherited by children.
2. A new CSS file per component which includes the CSS for both index and skin, but leaves some variables uncompiled. The original variable names are replaced with the mapped unique variables produced in the first step. If a variable is static across themes, the static value is just inlined into the CSS.

This should make the included CSS MUCH smaller than compiling everything to static values per theme/scale. We will just include all themes by default, since they are just a small list of variables, and only need to swap classnames on the `.spectrum` element to switch themes and scales as before.


devongovett:
Thinking about this further in the context of allowing custom theming, I think potentially we can take advantage of variable fallback. I think there are two types of themes we want to support:

1. Global palette changes across all components, e.g. changing all blues to red. This involves changing the values of spectrum global variables, which cascade across all components. This would be supported just fine by this PR.
2. Component specific variable changes, e.g. changing only cta buttons to red. This involves changing the values of specific component-level variables, either to map to a different global, or to a specific value. This is not supported by the PR as it is today since the individual component variables are eliminated.

Ideally we want to allow individual overrides of variables, while still producing a minimal build including only the 72 unique variables per theme by default. This could be accomplished by using variable fallback, leaving the original variable name in addition to the mapped unique variable.

```css
.spectrum-Button {
  background: var(--spectrum-button-background, var(--spectrum-global-color-grey-100))
}
```

The default theme would only include definitions for the global variables (e.g. `--spectrum-global-color-grey-100`) so we'd still only have 72 variables, but if a custom theme provided a value for `--spectrum-button-background` it would be used instead of the global color. This would give us a minimal build by default while still allowing custom themes to override individual variables where needed.

benjamind:
I like this @devongovett - this is exactly the sort of approach we are taking with the web components implementations. They have undefined component specific variables that fallback to the defaults from spectrum.  As you say, its pretty easy to reason about, keeps the scope of variables most themes needs small. Feels like the best we can do until webcomponents get `::part` and `::theme`.

lazd:
@devongovett killer idea, this will make it very simple to apply a custom theme without having to target CSS selectors themselves.

Issue #109, opened 2019-03-01
Rating: should not appear interactive on hover in readOnly mode


majornista:
### Expected Behavior
In some use cases, a Rating component needs to be readOnly, and simply show the current rating. When readOnly is true, the Rating should not respond to hover events.

### Actual Behavior
`.spectrum-Rating.is-readonly` style does not have `cursor: default; pointer-events: none`, so hovering over the control renders as if the component were still interactive.

### Reproduce Scenario (including but not limited to)
React Spectrum

See https://codepen.io/mijordan/pen/KEVEKO for fixed example.

#### Spectrum-CSS version
2.8.0

#### Sample Code that illustrates the problem
```css
.spectrum-Rating {
  &.is-disabled {
    cursor: default;
    pointer-events: none;
  }
}
```
Should be:
```css
.spectrum-Rating {
  &.is-disabled,
  &.is-readonly {
    cursor: default;
    pointer-events: none;
  }
}
```

PR #110, opened 2019-03-01
Issue-109 Rating: add readOnly state

majornista:
## Description
In some use cases, a Rating component needs to be `readOnly`, and simply show the current rating. When `readOnly` is true, the Rating should not respond to hover events.

### Sample Code that illustrates the problem
```css
.spectrum-Rating {
  &.is-disabled {
    cursor: default;
    pointer-events: none;
  }
}
```
Should be:
```css
.spectrum-Rating {
  &.is-disabled,
  &.is-readonly {
    cursor: default;
    pointer-events: none;
  }
}
```

## Related Issue
#109 

## Motivation and Context
In some use cases, a Rating component needs to be `readOnly`, and simply show the current rating. When `readOnly` is true, the Rating should not respond to hover events.

## How Has This Been Tested?
Tested in React Spectrum. And Spectrum-CSS docs.

Also, see working prototype: https://codepen.io/mijordan/pen/KEVEKO

## Types of changes

- [NA] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/docs/rating/rating.yml
+++ b/docs/rating/rating.yml
@@ -110,7 +110,7 @@ components:
     status: Verified
     description: The readOnly state of the Rating element.
     markup: |
-      <div class="spectrum-Rating is-readonly">
+      <div class="spectrum-Rating is-readOnly">
         <input class="spectrum-Rating-input" type="range" min="0" max="5" value="0" readOnly aria-label="Rating">
 
         <span class="spectrum-Rating-icon is-selected">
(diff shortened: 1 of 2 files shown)


docs/rating/rating.yml
```diff
@@ -161,7 +214,7 @@ components:
     name: Quiet Selected
     status: Verified
     markup: |
-      <div class="spectrum-Rating">
+      <div class="spectrum-Rating spectrum-Rating--quiet">
```
lazd: Good catch here :)

src/rating/index.css
```diff
@@ -31,7 +31,8 @@
 
   cursor: pointer;
 
-  &.is-disabled {
+  &.is-disabled,
+  &.is-readonly {
```
lazd: I think this is gonna have to be `is-readOnly` as it's two words, and this is more in line with `is-currentValue` and `is-outsideMonth` (used elsewhere).

PR #118, opened 2019-03-12
Action Bar and Table with Action Bar

jianliao:
## Description
A floating action bar appearing in selection mode. It could be used with spectrum div table on web desktop or card grid on mobile device.

## Related Issue
https://jira.corp.adobe.com/browse/ADUI-556

## Motivation and Context
Internal precursor link:
* http://spectrumprecursors.corp.adobe.com/submissions/action-bar

Context:
* User can use action bar within table for bulk action

## How Has This Been Tested?
The table integration part of this component rely on `position:sticky` and css component itself has no special dependency.
I've verified it with Safari, Firefox and Chrome on MacOS. Please see here -https://caniuse.com/#feat=css-sticky for browser supporting status.

### Known Issues
* While reviewing my impl with Lindsay Browne, she pointed out that the action button group needs to moved right for 9px and checkbox needs to move left for 1px. But I am not sure if I should enforce it in this component's css style. Maybe it is user's responsible to align it correctly.
* While integrating with Table - Div, I need to hard code style on table cell element to make checkbox alignment correctly.
* No color and animation defined for this component. All are reusing `spectrum-Popover`.
* Two missing css variables (maybe should be defined in DNA)
   * --spectrum-actionbar-min-width: 280px;
   * --spectrum-actionbar-max-width: 1280px;

image
image

## Types of changes

## Checklist:


src/actionbar/index.css
```diff
@@ -0,0 +1,46 @@
...
+  --spectrum-actionbar-offset-y: var(--spectrum-global-dimension-size-200);
+
+  --spectrum-actionbar-min-width: 280px;
+  --spectrum-actionbar-max-width: 1280px;
+}
+
+/* topdoc
+{{ actionbar/actionbar.yml }}
+*/
+.spectrum-Actionbar {
```
lazd: I _think_ this should be `ActionBar` since Bar is its own word...

src/actionbar/index.css
```diff
@@ -0,0 +1,46 @@
...
+  padding: 0 var(--spectrum-actionbar-padding-x);
+  min-width: var(--spectrum-actionbar-min-width);
+  max-width: var(--spectrum-actionbar-max-width);
+  width: 100%;
+}
+
+/* topdoc
+{{ actionbar/table-actionbar.yml }}
+*/
+.spectrum-Actionbar-tableContainer {
```
lazd: Hmmm, is there any way to do this without a parent element? Could we put `position: sticky` on a modifier class that we add, something like`spectrum-ActionBar spectrum-ActionBar--sticky` that makes it sticky and sets the transform?

jianliao: The container element is for centering the action bar and enforcing the left and right margin. I will give it  an another try. 

I totally agree with you on `position: sticky`, it should not be hard coded. `position: fixed` will be another possible value for a super long table which longer than the current viewport. Any thoughts that we can make it more flexible?

lazd: @jianliao ok, let's see if we can pull it off without the container. Thanks!

jianliao: @lazd I tried, but seems like container element is a must for supporting both `fixed` and `sticky` strategies. When applied `position: fixed`, action bar content lost the connection with its parent element e.g. table body. A container element is for simulating its parent to set a width on and make centered easy.

I generalized it by renaming `.spectrum-Actionbar-tableContainer` to `.spectrum-Actionbar-container`. Here is a [live demo](http://angular-spectrum.corp.adobe.com/component-viewer/table-basic) for `position: fixed`.

jianliao:
@lazd I created a pen to demo this new css component. https://codepen.io/jianliao/pen/ROmMdj

lazd:
I'll be out until Tuesday next week, will review this again then. Thanks @jianliao !

src/actionbar/index.css
```diff
@@ -0,0 +1,53 @@
...
+    &.is-open {
+       transform: translateY(calc(-1*(var(--spectrum-actionbar-height) + var(--spectrum-actionbar-offset-y))));
+    }
+  }
+
+  &.spectrum-ActionBar-sticky {
+    position: sticky;
+  }
+
+  &.spectrum-ActionBar-fixed {
```
lazd: Assuming we go with the suggestion I made above, since these are [modifiers](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#ComponentName--modifierName) (not children), they should be double dashes. And, since they're not states (no `.is-` prefix), they don't need to be nested and defined with `&.`:

Move them to the top level and rename them as:

```
.spectrum-ActionBar--sticky {  ... }
.spectrum-ActionBar--fixed {  ... }
```

src/actionbar/index.css
```diff
@@ -0,0 +1,53 @@
...
+  padding: 0 var(--spectrum-actionbar-padding-x);
+  min-width: var(--spectrum-actionbar-min-width);
+  max-width: var(--spectrum-actionbar-max-width);
+  width: 100%;
+}
+
+/* topdoc
+{{ actionbar/table-actionbar.yml }}
+*/
+.spectrum-ActionBar-container {
```
lazd: Since this is an outer component and not a child of ActionBar, what if we made this be the actual `.spectrum-ActionBar`, and changed the existing `.spectrum-ActionBar` to be `.spectrum-ActionBar-popover` (since it's really meant to modify a popover).

Then we can make this container basically do nothing unless you apply `--sticky` or `--fixed` to it, if that's indeed what we want to happen. 

That would change the markup to:

```
<div class="spectrum-ActionBar spectrum-ActionBar--sticky">
  <div class="spectrum-Popover spectrum-ActionBar-popover is-open ">
    <!-- ... -- >
  </div>
</div>
```

Thoughts?

src/actionbar/index.css
```diff
@@ -0,0 +1,53 @@
+@import '../commons/index.css';
+
+:root {
+  --spectrum-actionbar-height: var(--spectrum-global-dimension-size-600);
+  --spectrum-actionbar-padding-x: var(--spectrum-global-dimension-size-200);
+  --spectrum-actionbar-margin-x: var(--spectrum-global-dimension-size-200);
+  --spectrum-actionbar-offset-y: var(--spectrum-global-dimension-size-200);
+
+  --spectrum-actionbar-min-width: 280px;
+  --spectrum-actionbar-max-width: 1280px;
```
lazd: This seems a little arbitrary. Where does this value come from?

src/actionbar/index.css
```diff
@@ -0,0 +1,53 @@
...
+  min-width: var(--spectrum-actionbar-min-width);
+  max-width: var(--spectrum-actionbar-max-width);
+  width: 100%;
+}
+
+/* topdoc
+{{ actionbar/table-actionbar.yml }}
+*/
+.spectrum-ActionBar-container {
+  margin: 0 var(--spectrum-actionbar-margin-x);
```
lazd: An example would probably make this clear, but I'm not sure what this is intended to do, and I wonder if it should have been padding?

src/actionbar/index.css
jianliao: It is from Lindsay's XD file. https://precursors.s3-sj1.corp.adobe.com/files/action-bar.xd
I am thinking moving it to dna, but don't how.

src/actionbar/index.css
jianliao: This margin setting comes from here
image

src/actionbar/index.css
jianliao: min-width and max-width are from here: 
image

lazd: Ok excellent, as long as it has a source. I will help you get a PR into DNA after this is in and we can hardcode for now.

src/actionbar/index.css
lazd: Ok, looks like it needs be padding on `.spectrum-ActionBar`

jianliao: I tried it [here](https://codepen.io/jianliao/pen/ROmMdj), does not seem right.

lazd: Hmm ok, my concern is that margin will have no effect. But if it's working correctly with margin, then I'm fine with it.

lazd:
Some additional exploration of techniques has been done here: https://codepen.io/lazd/pen/PvZvRb?editors=0100

@jianliao and I are working with Lindsay to figure out expected behavior when scrolled to the bottom of a table with a sticky ActionBar.

lazd:
@jianliao can you review the two latest commits and let me know if this is good to merge? I incorporated the changes we worked on together as well as the feedback from the Design Workshop this morning.

jianliao:
> @jianliao can you review the two latest commits and let me know if this is good to merge? I incorporated the changes we worked on together as well as the feedback from the Design Workshop this morning.

All look good to me. Thanks.

Issue #124, opened 2019-03-18
Checkbox shouldn't have margin-right


buddhari:
### Expected Behavior
Checkboxes shouldn't have a built-in margin-right. 

### Actual Behavior
There's a 16px margin-right for checkboxes right now, which is not as designed.
![image (17)](https://user-images.githubusercontent.com/45472337/54570174-d80c8080-499a-11e9-90e7-5556b80ed48f.png)

#### Spectrum-CSS version
2.8.0

synced to jira: https://jira.corp.adobe.com/browse/SDS-7323

lazd:
So this is definitely wrong, but there are a couple reasons for it the fix will need to take into account:

1. Because the box of the checkbox must be aligned with the leftmost side its parent, the hitarea for the component falls outside of the bounds of root element. The right margin makes sure there's space for the hit area on the right side so it doesn't overlap with other controls
2. Because the hit area of the checkbox falls out, the right margin makes sure it doesn't overlap with a checkbox that follows it (whose hit area also falls out).

That said, we could fix this by reducing that margin by half and adding a little left margin to a checkbox when it follows another checkbox.

misterbrownlee:
I brought this up in #qotd ... will follow up with any progress.

lazd:
Per discussion on slack, we should be getting margin from a group, not from components. Right now, we have [Field Group](https://opensource.adobe.com/spectrum-css/fieldgroup.html), but it does not apply margin. We should require the use of a Field Group (or add a new component, Checkbox Group) that applies the expected margin.

PR #125, opened 2019-03-19
WIP: Remove checkbox right margin

lazd:
# WIP, DO NOT MERGE

## Description

This PR removes the right margin that all Checkboxes have to they look nice when placed next to each other, fixing the usage of an individual checkbox as defined in #124. However, this has a side effect -- checkboxes that wrap get pushed to the left since they technically follow another checkbox, but don't appear on the same line...

This is the example from the docs, but with the `<br>` removed between the long checkbox and the 3 above it:

image

This may not be a good solution to the issue, as the new bug it introduces could show up elsewhere. More digging is required before merging this.

## Related Issue

#124 

## Types of changes

### DIFF
--- a/src/checkbox/index.css
+++ b/src/checkbox/index.css
@@ -9,9 +9,19 @@
   height: var(--spectrum-checkbox-height);
   max-width: 100%;
 
-  margin-right: calc(var(--spectrum-checkbox-cursor-hit-x) * 2);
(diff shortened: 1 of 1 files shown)


lazd:
This is the wrong approach to this problem. We should be using Field Group to apply margins.

PR #126, opened 2019-03-25
 🐛 disable hover highlight for table header

junqdu:
## Description
When `class="spectrum-Table-row"` is added to the `tr` of `thead`, header shows highlight on hovering.

## Related Issue
https://jira.corp.adobe.com/projects/CORAL/issues/CORAL-156

## Motivation and Context
https://jira.corp.adobe.com/projects/CORAL/issues/CORAL-156, coral-spectrum has a temporary fix for it, but we would like the fix happening in spectrum-css.

## How Has This Been Tested?

Manually test, hosted on my machine, added `class="spectrum-Table-row"` to the `tr` of `thead`, and applying hover state to the `tr`, header row is not highlighted.

As you can see after adding `class="spectrum-Table-row"` to the `tr` of `thead`, and applying hover state to the `tr`, header row is not highlighted.
Screen Shot 2019-03-22 at 4 45 36 PM

## Types of changes

## Checklist:

### DIFF
--- a/src/table/skin.css
+++ b/src/table/skin.css
@@ -75,7 +75,42 @@ tbody.spectrum-Table-body {
   border: none;
 
   .spectrum-Table-row {
+    background-color: var(--spectrum-table-row-background-color);
+    border-bottom: 1px solid var(--spectrum-table-cell-border-color);
     border-top: none;
+
+    &:hover {
+      background-color: var(--spectrum-table-row-background-color-hover);
+    }
+
+    &:focus-ring,
+    &.is-focused {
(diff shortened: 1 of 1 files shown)


lazd:
@junqdu, the docs do not suggest that you apply `.spectrum-Table-row` to the `<tr>` inside of `<thead>`, it's only supposed to be used for `<tr>`s inside of `<tbody>`. The table seems to work fine if used as the docs suggest.

Is there a reason why you need to add the class to the `<tr>` inside of `<thead>`?

junqdu:
@lazd fair enough, I don't have a strong opinion on this change.

Issue #149, opened 2019-05-08
index-vars.css build output inconsistent between builds


lazd:
### Expected Behavior

Build output is consistent

### Actual Behavior

Build output sometimes has vars, sometimes doesn't. This changes between builds.

#### Steps to Reproduce

1. Run this 
```
rm -rf dist1 dist; gulp build-css:individual-components-vars; cp -r dist dist1; gulp build-css:individual-components-vars; diff -rq dist dist1
```
2. Observe occasional errors such as:
```
postcss-import: spectrum-css/src/accordion/index-vars.css:55:1: @import must precede all other statements (besides @charset)
```
3. Observe various differences that change between runs:
```
Files dist/components/calendar/index-vars.css and dist1/components/calendar/index-vars.css differ
```
4. Goto 1

#### Wat

What's happening here is that the skin and index files are sometimes out of order, because the glob passed to concat is not ordered.

#### Spectrum-CSS version

2.10.0

Issue #134, opened 2019-04-27
Page component's index-vars.css includes & instead of selector


lazd:
### Expected Behavior

Selector should be `.spectrum`

### Actual Behavior

Is `&`

#### Steps to Reproduce

1. Visit https://unpkg.com/@adobe/spectrum-css@2.10.0/dist/components/page/index-vars.css
2. Observe /title

#### Spectrum-CSS version

2.10.0

PR #150, opened 2019-05-08
Make sure glob passed to concat for vars is ordered

lazd:
## Description

Basically, the glob being passed to concat wasn't ordered, so we would get inconsistent output between builds. This change fixes that by explicitly ordered the glob, and ignores non-directories in figuring out what components to process.

## Related Issue

#149 

## Motivation and Context

The build was breaking randomly.

## How Has This Been Tested?

I ran the following command over and over:

```
rm -rf dist1 dist; gulp build-css:individual-components-vars; cp -r dist dist1; gulp build-css:individual-components-vars; diff -rq dist dist1
```

No output = success

## Types of changes

### DIFF
--- a/tasks/build-css.js
+++ b/tasks/build-css.js
@@ -131,6 +131,7 @@ function getProcessors(keepVars = false) {
 }
 
 var processors = getProcessors();
+var varsProcessors = getProcessors(true);
 
 /**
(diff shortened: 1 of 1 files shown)


lazd:
Gonna go ahead and merge this since it's blocking #134

Issue #156, opened 2019-05-10
Prototype Floating Panel


lazd:
Work is happening here: https://github.com/adobe/spectrum-css/tree/floatingpanel

Deployed demos:
https://git.corp.adobe.com/pages/lawdavis/spectrum-panels-study/docs/panel.html
https://git.corp.adobe.com/pages/lawdavis/spectrum-panels-study/docs/panelFloating.html

lazd:
Prototypes are considered complete. We'll make these a real thing after Design has had time to poke at them and deliver revisions.

PR #200, opened 2019-06-06
Form: labelsAbove

lazd:
## Description

This PR adds support for a structured Form layout with labels above the form fields.

## Related Issue

#199 

## Motivation and Context

This was needed for the Panels prototype #156 

## How Has This Been Tested?

iScrolled

image

## Types of changes

### DIFF
--- a/docs/form/index.yml
+++ b/docs/form/index.yml
@@ -1,92 +1,279 @@
 ---
 name: Form
-description: >-
-  Spectrum Form with Table layout
-markup: |
-  <form class="spectrum-Form">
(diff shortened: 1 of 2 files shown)


NateBaldwinDesign:
This looks good to me. Products have the option to increase those margins if they wish (between each form item), so long as that’s possible we should be good. But this does match the default I believe. Only questionable part is the distance between the label and radio/checkbox options... but if they’re spaced the same between those items as they are for text fields (which I assume they are), then it’s better to be consistent IMO

Issue #157, opened 2019-05-10
Prototype App Frame


lazd:
An early exploration of the App Frame is here: https://git.corp.adobe.com/pages/lawdavis/dsa/

Demo of recent progress: https://git.corp.adobe.com/pages/lawdavis/spectrum-panels-study/docs/panel.html

lazd:
Prototypes are considered complete. We'll make these a real thing after Design has had time to poke at them and deliver revisions.

PR #201, opened 2019-06-06
Vertical Rule

lazd:
## Description
This PR adds support for vertical Rules. The examples are a bit silly, but they're there.

## Related Issue

#198 

## Motivation and Context

This was needed for the App Frame prototype #157 

## How Has This Been Tested?

iScrolled

image

## Types of changes

### DIFF
--- a/src/rule/index.css
+++ b/src/rule/index.css
@@ -16,6 +16,8 @@ governing permissions and limitations under the License.
 {{ rule/rule.yml }}
 */
 .spectrum-Rule {
+  width: 100%;
+
   /* Show the overflow for hr in Edge and IE. */
   overflow: visible;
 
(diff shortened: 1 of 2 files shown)


src/rule/index.css
```diff
@@ -16,6 +16,8 @@ governing permissions and limitations under the License.
 {{ rule/rule.yml }}
 */
 .spectrum-Rule {
+  width: 100%;
```
lazd: This is helpful in flex layouts.

src/rule/index.css
```diff
@@ -41,3 +43,19 @@ governing permissions and limitations under the License.
 
   border-radius: var(--spectrum-rule-small-border-radius);
 }
+
+.spectrum-Rule--vertical {
+  height: 100%;
```
lazd: This means you need height defined on the parent container of the Rule.

lazd:
Got confirmation from the team that the small/medium examples are good. The Large rule really shouldn't be used, so the example has been removed

PR #217, opened 2019-07-10
Fix #215 [Accessibility] StepList should include interactive example with links

### LINKED ISSUE #215: [Accessibility] StepList should include interactive example with links
### Expected Behavior
StepList should provide an example that follows WAI tutorial recommendation for [Using a Step by Step indicator](https://www.w3.org/WAI/tutorials/forms/multi-page/#using-step-by-step-indicator).

### Actual Behavior
StepList only shows static examples. 
Does not have example where current step receives keyboard focus `:focus-within` or `.is-focused` on a descendant link.
Tooltip will not display on focus.

#### Browser name/version/os (ie Chrome Version 62.0.3202.94 (Official Build) (64-bit) MacOS)
Google Chrome Version 75.0.3770.100 (Official Build) (64-bit)

#### Spectrum-CSS version
v 2.13.0

#### Sample Code that illustrates the problem
```css
  /* Interactivity Styles */
  &.spectrum-Steplist--interactive {
    .spectrum-Steplist-item {
      /* Step Focused */
      &:focus {
        .spectrum-Steplist-marker {
          background-color: var(--spectrum-steplist-current-marker-color-key-focus)
        }
      }
    }
  }
```
Should be:
```css
  /* Interactivity Styles */
  &.spectrum-Steplist--interactive {
    .spectrum-Steplist-item {
      /* Step Focused */
      &:focus,
      &:focus-within,
      &.is-focused {
        .spectrum-Steplist-marker {
          background-color: var(--spectrum-steplist-current-marker-color-key-focus)
        }
      }
    }
  }
```

majornista:
Fixes:

- StepList only shows static examples.
- Does not have example where current step receives keyboard focus :focus-within or .is-focused on a descendant link.
- Tooltip will not display on focus.

## Description

Fixes:

- StepList only shows static examples.
- Does not have example where current step receives keyboard focus :focus-within or .is-focused on a descendant link.
- Tooltip will not display on focus.

## Related Issue
#215 

## Motivation and Context
React-Spectrum defaults to interactive variant which should follow design pattern exhibited in the WAI Multipage form tutorial under [Using Step by Step Indicator](https://www.w3.org/WAI/tutorials/forms/multi-page/#using-step-by-step-indicator)

StepList should render focused style when a link descendant of a `.spectrum-Steplist-item` has focus.

## How Has This Been Tested?
Tested using documentation examples and in an updated React-Spectrum implementation.

## Types of changes

## Checklist:


majornista:
@lazd bump!

lazd:
@majornista thanks for your patience! We just got done refactoring the entire project. Can you rebase this? Some files have moved from `src/` to `components/`

components/steplist/index.css
```diff
@@ -126,7 +126,9 @@ governing permissions and limitations under the License.
   /* Selected state */
   &.is-selected {
     /* Focus */
-    &:focus {
+    &:focus,
+    &.is-focused,
+    *:focus {
```
lazd: This is a little curious; you're setting `outline: none` on any child element of a selected step that has focus (i.e. this selector resolves to `.spectrum-Steplist-item.is-selected *:focus`. Was that intended?

components/tooltip/index.css
```diff
@@ -203,23 +203,38 @@ governing permissions and limitations under the License.
     transform: translate(-50%, var(--spectrum-tooltip-tip-margin));
   }
 
-  &:hover .spectrum-Tooltip {
+  &:hover .spectrum-Tooltip,
+  &:focus .spectrum-Tooltip,
+  &.is-focused .spectrum-Tooltip,
+  *:focus .spectrum-Tooltip {
```
lazd: This one here makes sense (show the tooltip if a child element of something that should show a tooltip has focus), but it will not active if a child element has `.is-focused` or `:hover`. was that intended?

components/steplist/index.css
majornista: This is to account for the interactive use case where a link is the child of the `.spectrum-Steplist-item`.

majornista: https://github.com/adobe/spectrum-css/pull/217/files#diff-90810980a9333ecd68c5d9e338f0e818R40

components/tooltip/index.css
majornista: Again, this is for the interactive use case where the element receiving focus would be the descendant link.

PR #273, opened 2019-09-07
Remove package-lock.json of all the lerna managed package

jianliao:
## Description
As we use [hoist](https://github.com/lerna/lerna/blob/master/doc/hoist.md) when bootstrap the lerna, only the top-level `package-lock.json` matter. We don't need to keep package-lock file per package.

## Related Issue
n/a

## Motivation and Context
1. package-lock per package is hard to maintain
2. keep it simple 

## Types of changes

## Checklist:


jianliao:
> Excellent. I assume these are not re-created during normal events like `npm install`, `npm run bootstrap`, etc, right?

Yes, package-lock would not be re-created.

PR #276, opened 2019-09-12
Move docs to metadata

GarthDB:
removed old data files

## Description
Moved the metadata content.

## Related Issue
Some jira issues

## Motivation and Context
Just trying to standardize the metadata

## How Has This Been Tested?
not yet.

## Types of changes

- [x] Docs change

## Checklist:


lazd:
@GarthDB looks like meter still doesn't have an ID, hmm...

lazd:
@GarthDB looks like `status:` needs to be moved into examples too. I'll handle the remaining changes here.

GarthDB:
@lazd I'll go through and move those.

GarthDB:
@lazd It looks like barloader has a `metadata.yml` file and a `metadata` dir - I'll merge those.

lazd:
@GarthDB uh oh, some data was lost. Take a look at Breadcrumb: https://github.com/adobe/spectrum-css/pull/276/files#diff-a839ea5216faadd7c0d011cc76cd1be3

It previously had `id: breadcrumb` for the examples so they got the right DNA status, but your script swapped in `breadcrumb-nested` from the object key there, so the DNA status is now wrong. Can you redo the conversion and make sure those IDs override the key names?

GarthDB:
@lazd ah, yes. What should we call that thing? It's like a slug or a name? because we need the DNA id as well.

lazd:
@GarthDB so basically, if there is an `id` field for the example, it overrides the object key. There's no need for a separate field.

lazd:
This looks great. I just tested it out locally and everything is looking right; DNA status, descriptions, all that. Thank you!

Issue #272, opened 2019-09-06
Alert icon color is wrong


lazd:
### Expected Behavior

It's slightly darker than the border
![image](https://user-images.githubusercontent.com/201344/64464404-ad36f580-d0bc-11e9-9528-6ea818ea75ae.png)

### Actual Behavior
It's the same color as the border
![image](https://user-images.githubusercontent.com/201344/64464409-b0ca7c80-d0bc-11e9-91b8-1635ecf332fc.png)

#### Spectrum-CSS version

2.13.0

PR #284, opened 2019-09-20
Generate linked CSS variables for design tokens

lazd:
## Description

This PR generates separate CSS files for each component and global using the new linked DNA output.

The experimental JS API for `@spectrum-css/vars` has been removed for now so it can be added back later after formal review by the team in a minor version, instead of being a major version bump when its changed. The code that generates the JS API remains, though is commented out.

In addition, several components that use tokens incorrectly (or didn't use tokens and should have) were updated. In the process, #272 is fixed (SDS-2285).

Finally, documentation has been updated to reference the `spectrum-*-unique.css` files included in bundles properly, closing #299.

**Note:** ~~This PR should not be squashed, otherwise semantic changelogs will not be generated correctly.~~ Since the first release for components will happen after this, it's ok to squash.

 - **How this was tested:** Each component was looked at to ensure they still worked, and build output tasks report that all used variables still exist. Visual regression testing reports failures for changed values (investigation needed, see below)
 - **Browser(s) and OS(s)** this was tested with: Chrome 75 on macOS, Chrome 77 on Linux

## To-do list

- [x] Review each test failure in visual regression testing locally
- [x] Document use of `spectrum-*-unique.css` variable files in bundle output


jianliao:
My local test result:
image

Reviewed with @lazd, all expected. Need to update reference bitmap package.

components/vars/tasks/updateDNA.js
```diff
@@ -0,0 +1,565 @@
...
+    value = value.replace(/colorStopData\..*?\.colorSemantics\./, 'semantic.');
+    // Strip the scale information
+    value = value.replace(/scaleData\..*?\.dimensionTokens\./, 'global-dimension.');
+    value = value.replace(/scaleData\..*?\.dimensionAliases\./, 'alias.');
+    // Sub in proper names for globals
+    value = value.replace(/colorGlobals\./, 'global-color.');
+    value = value.replace(/dimensionGlobals\./, 'global-dimension.static-');
+    value = value.replace(/fontGlobals\./, 'global-font.');
+    value = value.replace(/animationGlobals\./, 'global-animation.');
+    value = value.replace(/staticAliases\./, 'alias.');
```
bernhard-adobe: just an idea but should we maybe try to catch some errors here if any of these replaces go wrong?

components/vars/tasks/updateDNA.js
```diff
@@ -0,0 +1,565 @@
...
+        }
+      }
+
+      for (let elementName of elements) {
+        generateCSSFile([
+          elementVariables[elementName]
+        ], elementName, 'css/components');
+        cssFilesGenerated[elementName] = true;
+
+        // generateJSFile([
```
bernhard-adobe: maybe remove or add as an optional parameter if the old files should for any reason be compiled.

tools/component-builder/css/index.js
```diff
@@ -18,10 +18,14 @@ const processors = require('./processors').processors;
...
 
+// Read in all variables used
+// Read in all vars from recent DNA
+// Include definitions if they refer to a variable, static if not
+
 function buildIndexVars() {
   return gulp.src([
     'index.css',
-    'skin.css'
+    'skin.css',
```
bernhard-adobe: remove comma?

tools/component-builder/css/lib/varUtils.js
```diff
@@ -0,0 +1,180 @@
...
+  let varNames = Object.keys(vars);
+  if (varNames.length) {
+    return `
+${classNames.map((className) => `${className}`).join(',\n')} {
+${varNames.map((varName) => `  ${varName}: ${vars[varName]};`).join('\n')}
+}
+`;
+  }
+
+  return '';
```
bernhard-adobe: not sure if that's important but returning an empty string could lead to issues? 
Do you do a length check later on this or just export the empty string?

components/vars/tasks/updateDNA.js
lazd: They'll either match or they won't; these operate on a single variable name.

We catch errors during build -- if this somehow references a non-existent variable, the build errors.

components/vars/tasks/updateDNA.js
lazd: I mentioned in the PR description I've commented them out for now since reviewing a JS API is outside of the scope of this PR. I'd rather not put them behind a flag or anything, just leave the code intact so we can add it back later. I don't think it's a big problem.

tools/component-builder/css/lib/varUtils.js
lazd: This function needs to return a string value, the output is written to a file. If there are no defined variables, I still want a string, but an empty one.

bernhard-adobe:
Syntax wise this looks good @lazd .

PR #290, opened 2019-09-22
test(backstopjs): support configurable bitmaps reference path

jianliao:
## Description

1. Enhance backstopjs testing script to support configurable bitmaps reference path. It can make update bitmaps reference easier.
2. Add `npm run watch`.
3. Update travis ci script to use `watch` instead of `dev` to avoid unnecessary rebuild.

Test with latest mac and travis-ci passed

## To-do list

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,10 @@ node_js:
 services:
   - docker
 before_install:
+  # Enforce upgrading to the latest yarn to avoid random version of yarn assigned by travis virtual environment.
   - npm install -g yarn
 install:
   - yarn
-before_script:
-  - yarn global add gulp-cli
 script:
   # This script should be the first that runs to reduce the risk of
   # executing a script from a compromised NPM package.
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ script:
   # This script should be the first that runs to reduce the risk of
   # executing a script from a compromised NPM package.
   - audit-ci --moderate
-  - gulp build
-  - gulp dev &
+  - yarn run build
+  - yarn run watch &
```
lazd: Hmm, why not just use `gulp` directly?

jianliao: Just in case there is no global gulp instance setup or global gulp version does not meet our need. Actually, we should encourage people using `yarn run` or `npm run` as it will guarantee picking up the right one(./node_modules/.bin/gulp) that we specify in package.json file.

lazd: Ah, interesting. I didn't realize it could execute tasks without `gulp-cli` if you're using `yarn run`. If you don't need a global `gulp-cli` at all, that's a good deal.

Note that there is no global `gulp`, only global `gulp-cli`, which knows how to run the local `gulp`.

jianliao: Good point. I'll give it a try.

jianliao: It works. I removed installing global gulp-cli

PR #301, opened 2019-10-01
docs: fix scale switcher dropdown state when resizing window

lazd:
## Description

This implements feedback from design re: automatic scale switching when resizing the window.

 - **How this was tested:** resize window, look at scale switcher, smile

## To-do list

### DIFF
--- a/site/resources/js/site.js
+++ b/site/resources/js/site.js
@@ -188,6 +188,10 @@ window.addEventListener('DOMContentLoaded', function() {
     else {
       switcher.scale = 'medium';
     }
+    let scaleDropdown = document.querySelector('#switcher-scale');
+    if (scaleDropdown) {
+      setDropdownValue(scaleDropdown, switcher.scale);
+    }
   }
   scaleMQL.addListener(handleScaleMQLChange);
 
--- a/site/resources/js/site.js
+++ b/site/resources/js/site.js
@@ -188,7 +188,6 @@ window.addEventListener('DOMContentLoaded', function() {
     else {
       switcher.scale = 'medium';
     }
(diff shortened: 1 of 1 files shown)


site/resources/js/site.js
```diff
@@ -188,6 +188,10 @@ window.addEventListener('DOMContentLoaded', function() {
     else {
       switcher.scale = 'medium';
     }
+    let scaleDropdown = document.querySelector('#switcher-scale');
```
bernhard-adobe: since this is part of the event listener matching a media query for max-width 768px won't this be fired up every time the window sizes in that viewport range? Does this mean the query selector will fire up multiple times?
Maybe this should be moved outside the window.matchMedia listener.  

```
 var scaleMQL = window.matchMedia('(max-width: 768px)');
  function handleScaleMQLChange() {
    if (scaleMQL.matches) {
      switcher.scale = 'large';
    }
    else {
      switcher.scale = 'medium';
    }
    let scaleDropdown = document.querySelector('#switcher-scale');
    if (scaleDropdown) {
      setDropdownValue(scaleDropdown, switcher.scale);
    }
  }
  scaleMQL.addListener(handleScaleMQLChange);
```

lazd: Oh it totally could be moved out. I'll do that.

lazd: Wait I lied, it has to run inside of the listener; some pages have it, some don't, and we get a new dropdown every time we move to a new page (asynchronously).

lazd: I'll cache it in FastLoad's listener!

PR #307, opened 2019-10-03
Added flex-wrap: wrap to FieldGroup

wenzdey:
Added radio warp for when there are too many radios or radio labels exceed container width.

## Description
https://github.com/adobe/spectrum-css/issues/268

BEFORE:
image

AFTER:
image

## To-do list

### DIFF
--- a/components/fieldgroup/index.css
+++ b/components/fieldgroup/index.css
@@ -15,6 +15,7 @@ governing permissions and limitations under the License.
 .spectrum-FieldGroup {
   display: flex;
   vertical-align: top;
+  flex-wrap: wrap;
 }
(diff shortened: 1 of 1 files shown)


lazd:
@wenzdey, it would be nice to add an example that illustrates the fix that that this change made. Can you add an example, or if one already exists, provide a before/after screenshot? Thanks!

wenzdey:
> 
> 
> @wenzdey, it would be nice to add an example that illustrates the fix that that this change made. Can you add an example, or if one already exists, provide a before/after screenshot? Thanks!

Sure thing. Added screenshots

Issue #269, opened 2019-09-05
Calendar heading should be flex positioned, not absolute


lazd:
Right now, the buttons and the heading are absolute positioned. They should be flex to make RTL work nice.

### Spectrum-CSS version

2.14.0-alpha.15

PR #315, opened 2019-10-07
fix: fix Calendar title wrapping/alignment, related to #269

lazd:
## Description

While testing for the 2.14 release, I discovered some extra padding on the calendar heading, and when removed, I found that calendar headings no longer truncate correctly or align vertically in the large scale. Also, the horizontal alignment of the text was wrong due to an Action Button margin applied to adjacent siblings. This fixes all that. 

This relates to work done in #269 

 - **How this was tested:** Load up docs, give a calendar a large heading like `August of the good year 2017`, it should truncate, switch to large scale, it should line up vertically
 - **Browser(s) and OS(s) this was tested with:** Chrome 77 on macOS

## Screenshots

image
image

image

## To-do list

### DIFF
--- a/components/calendar/index.css
+++ b/components/calendar/index.css
@@ -17,9 +17,11 @@ governing permissions and limitations under the License.
   --spectrum-calendar-border-width-reset: 0;
   --spectrum-calendar-margin-y: 24px;
   --spectrum-calendar-margin-x: 32px;
+  --spectrum-calendar-width: calc((var(--spectrum-calendar-day-width) + var(--spectrum-calendar-day-padding) * 2) * 7);
 }
 
 .spectrum-Calendar {
+  width: var(--spectrum-calendar-width);
   display: inline-block;
 }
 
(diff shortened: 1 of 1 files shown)


lazd:
@Martskin looks like we didn't quite get it in the first PR. Does this look good to you?

Martskin:
Ahh I see @lazd .

Looks good to me. Although the 3px margin on either side of the next/prev month buttons was there to align the caret icon with the column:

With margins:
Screen Shot 2019-10-07 at 1 03 55 PM

Now without margins:
Screen Shot 2019-10-07 at 1 04 14 PM

lazd:
Excellent catch, @Martskin. I added the margin back, but added it on both sides of the button so its language direction agnostic. Thanks!

components/calendar/index.css
```diff
@@ -40,23 +42,26 @@ governing permissions and limitations under the License.
...
+
   text-align: center;
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
 }
 
+.spectrum-Calendar-prevMonth,
+.spectrum-Calendar-nextMonth {
+  margin: 0 var(--spectrum-calendar-button-gap) !important; /* override adjacent sibling selector from ActionButton */
```
bernhard-adobe: There was a task for Spectrum CSS to remove the !important bangs. Is there maybe a different solution possible to se the correct margin for prev-Month, next-Month?

lazd: Yes, the only solution is to increase selector specificity. I went ahead and did that.

Issue #202, opened 2019-06-06
Menu icons need their own classes


lazd:
Right now, menu icons are just targeting `.spectrum-Icon`, when they should target `.spectrum-Menu-itemIcon`.

PR #331, opened 2019-10-14
Add menu icon class

Martskin:
Adds `spectrum-Menu-itemIcon` classname.

## Description
Adds menu specific icon classname for issue #202 and changes all menu instance with icons.

Screen Shot 2019-10-14 at 1 18 05 PM

### DIFF
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -99,7 +99,7 @@ governing permissions and limitations under the License.
     }
   }
 
-  .spectrum-Icon {
+  .spectrum-Menu-itemIcon {
     /* Don't get smaller, you're an icon! */
     flex-shrink: 0;
     align-self: flex-start;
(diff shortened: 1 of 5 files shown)


components/menu/index.css
```diff
@@ -99,7 +99,7 @@ governing permissions and limitations under the License.
     }
   }
 
-  .spectrum-Icon {
+  .spectrum-Menu-itemIcon {
     /* Don't get smaller, you're an icon! */
     flex-shrink: 0;
     align-self: flex-start;
```
lazd: There's another instance below that needs this change. I would say we should also preserve support for the existing markup so we don't have to make a breaking change.

Martskin: Ahh! Good catch. Fixed!

lazd:
This looks good. I'd say it's a `feat` since we're adding a new class in a backwards compatible way.

Martskin:
> This looks good. I'd say it's a `feat` since we're adding a new class in a backwards compatible way.

Thanks @lazd ! Do I need to add a new commit with `feat` to this PR?

PR #345, opened 2019-10-21
fix: preprocess percentage values and convert them to floats, fixes #340

### LINKED ISSUE #340: [Avatar] Opacity is processed at NaN
## Description
The opacity code available at https://github.com/adobe/spectrum-css/blob/master/components/avatar/skin.css#L14-L18 renders to `opacity: NaN;` in the published package.

## Steps to reproduce

1. `yarn add @spectrum-css/avatar`
2. `cd node_modules/@spectrum-css/avatar/dist`
3. `vim index-vars.css`

## Environment
 - **Spectrum CSS version:** 
 - **Browser(s) and OS(s):**

lazd:
## Description

Instead of attempting to convert percentage values to floats at the CSS implementation level, we instead convert them at the variable import level. This means you can directly use opacity variables and they'll already be real CSS values.

 - **How this was tested:**
```
cd components/vars
gulp update
git diff
```
 - **Browser(s) and OS(s) this was tested with:** n/a

## To-do list

### DIFF
--- a/components/avatar/skin.css
+++ b/components/avatar/skin.css
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-Avatar {
-  opacity: percent(var(--spectrum-avatar-small-opacity));
+  opacity: var(--spectrum-avatar-small-opacity);
 
   &.is-disabled {
-    opacity: percent(var(--spectrum-avatar-small-opacity-disabled));
+    opacity: var(--spectrum-avatar-small-opacity-disabled);
   }
 }
--- a/components/tags/skin.css
+++ b/components/tags/skin.css
@@ -75,7 +75,7 @@ governing permissions and limitations under the License.
 
     .spectrum-Avatar {
       /* Duplicated so state is on the tag */
-      opacity: percent(var(--spectrum-avatar-small-opacity-disabled));
+      opacity: var(--spectrum-avatar-small-opacity-disabled);
     }
   }
 }
--- a/components/vars/css/components/spectrum-avatar.css
+++ b/components/vars/css/components/spectrum-avatar.css
@@ -1,7 +1,7 @@
 :root {
-  --spectrum-avatar-small-opacity: 100%;
-  --spectrum-avatar-small-opacity-disabled: 30%;
-  --spectrum-avatar-small-border-radius: 50%;
+  --spectrum-avatar-small-opacity: 1;
+  --spectrum-avatar-small-opacity-disabled: 0.3;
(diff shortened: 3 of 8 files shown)


components/avatar/skin.css
```diff
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-Avatar {
-  opacity: percent(var(--spectrum-avatar-small-opacity));
+  opacity: var(--spectrum-avatar-small-opacity);
```
Westbrook: How does this exemplify in the output of `index-vars.css`? I've see some packages export with a fixed fallback `var(--spectrum-avatar-small-opacity, 0.3)` (seen in `@spectrum-css/actionbar`) and some with "system fallbacks" `var(--spectrum-avatar-small-opacity, --spectrum-global-disabled-opacity)` (seen in `@spectrum-css/avatar` ) , and am not sure what _should_ be expected in this area. Are there docs I could read on this subject?

lazd: So we look at every instance of a variable, and we include a fallback -- either the variable referenced by the instance, or, if the instance does not reference another variable (i.e. an alias or global), the actual value.

This is (sort of) documented here: https://github.com/adobe/spectrum-css#variable-fallbacks

This allows us to reduce bundle sizes by not including every component's variable file while still allowing users to customize those variables (sort of documented at the end of this section https://github.com/adobe/spectrum-css#using-components).

Now, in the case of these percentage values, what we're going to see now is: 

```
.spectrum-Avatar {
  opacity: var(--spectrum-avatar-small-opacity, 1);
}
```

That is, if you override the opacity of an avatar with the following:

```
.spectrum-Avatar {
  --spectrum-avatar-small-opacity: 0.8;
}
```

The override will take effect. If that variable is not defined, you'll get the default of `1`.

Westbrook: And the disabled version will output?
```
opacity: var(--spectrum-avatar-small-opacity, 0.3);
```

If so, great! Thanks for clarifying, I'll check out that link to see what else I can learn about the processing here.

lazd: Yup, sure will! Thanks!

lazd: @Westbrook ah, it shows up as:

```css
.spectrum-Avatar.is-disabled {
   opacity: var(--spectrum-avatar-small-opacity-disabled, 0.3);
}
```

lazd:
Note that the single failing visual regression test is because Avatar disabled now actually shows as disabled. This is ready to merge.

PR #350, opened 2019-10-21
docs: fix detail table styling

lazd:
## Description

This fixes the detail table font size on large scale as well as line height for the first entry.

 - **How this was tested:** Load, use large, examine detail table
 - **Browser(s) and OS(s) this was tested with:** Chrome macOS

## Screenshots

image

## To-do list

### DIFF
--- a/components/css/componentPage.scss
+++ b/components/css/componentPage.scss
@@ -15,6 +15,18 @@
   margin-bottom: 73px;
 }
 
+:global(.spectrum--large),
+:global(.spectrum--medium) {
+  :local(.compStatus) {
+    min-height: 0;
+    padding: 0;
+
+    &:before {
+      margin-left: 0;
+    }
+  }
+}
+
 :local(div.subHeadStatusLight) {
   display: inline-flex;
 }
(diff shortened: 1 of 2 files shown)


components/css/componentPage.scss
```diff
@@ -15,6 +15,15 @@
   margin-bottom: 73px;
 }
 
+:local(.compStatus) {
+  min-height: 0 !important;
```
GarthDB: Can we do this without `!important`?

lazd: If we redefine it for `.spectrum--large .compStatus` (which overrides it otherwise), but TBH, this is a hack, and `!important` makes that very clear :)

lazd: @GarthDB explored fixing this using CSS variable overrides, but we can't do it as the component references global variables and does not differentiate between left and right padding on the pseudo element. It's either get more specific or be `!important`, and TBH i think it's fine to be `!important` as this is a hack for this specific use case.

lazd: @GarthDB updated to explicitly define `.spectrum--large` etc and removed the `!important`s.

PR #361, opened 2019-10-23
docs: fix dialog presentation on docs page

lazd:
## Description

Fix Dialog page on next.js documentation

 - **How this was tested:** open Dialog and check it out
 - **Browser(s) and OS(s) this was tested with:** Chrome macOS

## Screenshots

image

## To-do list

### DIFF
--- a/components/css/componentPage.scss
+++ b/components/css/componentPage.scss
@@ -28,6 +28,7 @@
 }
 
 :local(.example) {
+  position: relative;
   background-color: var(--spectrum-global-color-gray-100);
   border-radius: var(--spectrum-global-dimension-size-50) var(--spectrum-global-dimension-size-50) 0 0;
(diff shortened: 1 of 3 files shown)


components/css/componentPage.scss
```diff
@@ -69,3 +70,15 @@
   text-align: left;
   display: inline-block;
 }
+
+:global(.spectrum-CSSExample-example--overlay) {
+  background: rgba(0,0,0,0.4);
```
bernhard-adobe: no Spectrum var here for black?

GarthDB: @bernhard-adobe there should be, but not with the alpha I don't think there is one with the correct the alpha.

bernhard-adobe: ah ok, thanks for the clarification.

PR #363, opened 2019-10-26
SDS 3087 - Adding Apache 2.0 headers to new NextJS.

bernhard-adobe:
[SDS-3087](https://jira.corp.adobe.com/browse/SDS-3087)

## Description
Adding Apache 2.0 headers to new NextJS.
I split this in three commits in case we don't want to add a specific set of headers. 

N/A

## Screenshots
N/A

## To-do list


GarthDB:
We don't need to include the header on `package-lock.json`; it is a generated file

PR #373, opened 2019-10-30
docs: add Rating enhancement to show focus, closes #249

### LINKED ISSUE #249: Rating is missing keyboard focus
@NateBaldwinDesign is there a design spec for the keyboard focus for Rating?

lazd:
## Description

This enables focus and clicking on Ratings to show the focus and selected states.

 - **How this was tested:** Tab to focus on Rating, observe focus.

## Screenshots

image

## To-do list

### DIFF
--- a/components/rating/metadata/rating.yml
+++ b/components/rating/metadata/rating.yml
@@ -109,7 +109,7 @@ examples:
     description: A non-interactive rating.
     markup: |
       <div class="spectrum-Rating is-readOnly">
-        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="0" readOnly aria-label="Rating">
+        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="3" readonly aria-label="Rating">
 
         <span class="spectrum-Rating-icon is-selected">
           <svg class="spectrum-Icon spectrum-UIIcon-Star spectrum-Rating-starActive" focusable="false" aria-hidden="true">
@@ -262,7 +262,7 @@ examples:
     name: Quiet (read-only)
     markup: |
       <div class="spectrum-Rating spectrum-Rating--quiet is-readOnly">
-        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="0" readOnly aria-label="Rating">
+        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="3" readonly aria-label="Rating">
 
         <span class="spectrum-Rating-icon is-selected">
           <svg class="spectrum-Icon spectrum-UIIcon-Star spectrum-Rating-starActive" focusable="false" aria-hidden="true">
--- a/site/resources/js/enhancement.js
+++ b/site/resources/js/enhancement.js
@@ -10,6 +10,62 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+// Rating
+(function() {
+  function setFocus(rating, focused) {
+    rating.classList[focused ? 'add' : 'remove']('is-focused');
+  }
+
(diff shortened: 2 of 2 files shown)


components/rating/metadata/rating.yml
```diff
@@ -262,7 +262,7 @@ examples:
     name: Quiet (read-only)
     markup: |
       <div class="spectrum-Rating spectrum-Rating--quiet is-readOnly">
-        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="0" readOnly aria-label="Rating">
+        <input class="spectrum-Rating-input" type="range" min="0" max="5" value="0" readOnly disabled aria-label="Rating">
```
Martskin: I see why this has to be disabled but will this make the whole component hidden from screen readers in the Read-only variant?

lazd: Yeah, I wasn't sure about this... The `readonly` attribute does nothing at all, so I think I'll have to remove `disabled` and reset the value in JS... But not really important since this is just for demo purposes...

site/resources/js/enhancement.js
```diff
@@ -10,6 +10,62 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+// Rating
+(function() {
+  function setFocus(rating, focused) {
+    rating.classList[focused ? 'add' : 'remove']('is-focused');
+  }
+
+  function setValue(rating, value) {
+    var input = rating.querySelector('.spectrum-Rating-input');
+    input.value = value;
+    Array.prototype.slice.call(rating.querySelectorAll('.spectrum-Rating-icon')).forEach(function(el, index) {
```
bernhard-adobe: Does this have to be added to the Array prototype for the slice method?

site/resources/js/enhancement.js
```diff
@@ -10,6 +10,62 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+        setValue(rating, value);
+      }
+    }
+  });
+
+  document.addEventListener('click', function(event) {
+    var icon = event.target.closest('.spectrum-Rating-icon');
+    if (icon) {
+      var rating = event.target.closest('.spectrum-Rating');
+      var value = Array.prototype.slice.call(rating.querySelectorAll('.spectrum-Rating-icon')).indexOf(icon) + 1;
```
bernhard-adobe: ditto, same comment as above

site/resources/js/enhancement.js
lazd: Here we're invoking the `slice` method of `Array` with the `this` value set to the `NodeList` returned by `querySelectorAll`. This is a common mechanism to convert a `NodeList` to an array.

We could invoke `forEach` directly here instead of converting to an Array first. I'll go ahead and make that change.

bernhard-adobe: you could use the spread operator instead:
``[...rating.querySelectorAll(".spectrum-Rating-icon")]`
https://stackoverflow.com/a/44439293

lazd: @bernhard-adobe we can't use new features here, this is ES5 land (no babel involved).

lazd:
@bernhard-adobe given the ES5 limitations, is this PR ready to merge?

components/rating/metadata/rating.yml
Martskin: > I updated the enhancement to reset the value on change for readonly ratings and removed `disabled` from the readonly examples. I think this should do the trick.

Looks good to me!

site/resources/js/enhancement.js
bernhard-adobe: oh, in this case carry on.

PR #415, opened 2019-11-22
Fix button height in flex container

devongovett:
When a button is inside a flex container and some of the items are taller than it, currently the button will expand to match the height of the other items. This causes it to look strange and not rounded correctly. This is due to a `min-height` but no `height` setting. Setting a height of 0% fixes the problem, and still allows it to grow when the button is multi-line. CSS is weird man. See https://stackoverflow.com/questions/27575779/prevent-a-flex-items-height-from-expanding-to-match-other-flex-items.

Before:

image

After:

image

### DIFF
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -99,7 +99,7 @@ governing permissions and limitations under the License.
   border-radius: var(--spectrum-button-primary-border-radius);
 
   min-height: var(--spectrum-button-primary-height);
-  height: auto;
+  height: 0%;
   min-width: var(--spectrum-button-primary-min-width);
 
   padding: var(--spectrum-button-padding-y) calc(var(--spectrum-button-primary-padding-x) - var(--spectrum-button-primary-border-size));
--- a/site/wrapping.pug
+++ b/site/wrapping.pug
@@ -174,6 +174,22 @@ html(lang='en-US').spectrum.spectrum--light.spectrum--medium
                   </div>
 
             article.spectrum-CSSExample
+              h5.spectrum-CSSExample-heading.spectrum-Heading4 Button (within flexbox)
+              section.spectrum-CSSExample-container
+                .spectrum-CSSExample-example
+
+
+                  <div style="margin: 0 0 10px 0; display: flex; flex-direction: row;">
+                    <div style="width: 32px; height: 96px; background-color: pink; margin: 0 10px;"></div>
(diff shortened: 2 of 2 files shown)


lazd:
So I tested this in IE 11 and... it actually works... I don't see a problem with this approach given that fact, I assumed it would just implode.

More testing is required, I'm working to get the old wrapping/truncation studies restored so we can make sure this didn't break them. Once those are back in, we can properly test this.

lazd:
Tested the following HTML:

```html
<div style="margin: 0 0 10px 0; display: flex; flex-direction: row;">
  <div style="width: 32px; height: 96px; background-color: pink; margin: 0 10px;"></div>
  <button class="spectrum-Button spectrum-Button--cta" style="width: 186px">
    <span class="spectrum-Button-label">Informationen über Creative Cloud</span>
  </button>
  <button class="spectrum-Button spectrum-Button--primary">
    <span class="spectrum-Button-label">Informationen</span>
  </button>
</div>
```

Chrome 78.0.3904.108:
image

Safari 13.0.3:
image

Firefox 70.0.1: 
image

IE 11:
image

Safari 11.1.2:
image

With this, I think we can approve this pull request. @GarthDB @jianliao @bernhard-adobe, let me know if you want to weigh in, otherwise I'll go ahead and merge this EOD.

lazd:
Uh oh, this breaks when you set height on the flex container:


image

```html
<div style="margin: 0 0 10px 0;display: flex;flex-direction: row;height: 120px;">
    <div style="width: 32px; height: 96px; background-color: pink; margin: 0 10px;"></div>
    <button class="spectrum-Button spectrum-Button--cta" style="width: 186px">
        <span class="spectrum-Button-label">Informationen über Creative Cloud</span>
    </button>
    <button class="spectrum-Button spectrum-Button--primary">
        <span class="spectrum-Button-label">Informationen</span>
    </button>
</div>
```

It seems the `align-self: start` solution might be better. @devongovett thoughts? It should be easily overridden.

snowystinger:
Looks like align-self is a better solution and is noted in the stackoverflow answer
https://stackoverflow.com/questions/27575779/prevent-a-flex-items-height-from-expanding-to-match-other-flex-items/40156422#40156422

devongovett:
Closing. You should use `align-items: flex-start` on the container.

Issue #451, opened 2019-12-13
Long words in Dropdown menu do not wrap


matthewdeutsch:
## Description
Long words in Dropdown menu do not wrap

## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/components/dropdown/
2. Edit a menu option with a long word (30 characters), as shown below

![image](https://user-images.githubusercontent.com/700930/70761749-2a4cb900-1d03-11ea-81a5-37043294dbfe.png)

## Expected behavior

Word wraps, as shown below. Can be fixed with `word-break: break-all;`

![image](https://user-images.githubusercontent.com/700930/70761766-389ad500-1d03-11ea-9061-9605d6809f37.png)

## Environment
 - **Spectrum CSS version:** latest
 - **Browser(s) and OS(s):** all

## Additional context
- filed by Asset Link team https://jira.corp.adobe.com/browse/CCE-4119600

PR #452, opened 2019-12-13
fix: wrapping on long words indropdowns (SDS-4649) #451

bernhard-adobe:
adding `word-break: break-all;` to dropdown label.

## Description
In order to break "unbreakable words" like German connected nouns or urls / paths we need to allow break-all for words.

 - **How this was tested:**  go to `http://localhost:3000/docs/dropdown.html` and inspect the dropdown label
 - **Browser(s) and OS(s) this was tested with:**  Google Chrome Canary Version 81.0.3993.0 (Official Build) canary (64-bit) on macOS Mojave 0.14.5 (18F203)

## Screenshots
Before:

Screen Shot 2019-12-13 at 23 15 45

After:
Screen Shot 2019-12-13 at 23 19 48

## To-do list

### DIFF
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -115,6 +115,7 @@ governing permissions and limitations under the License.
 .spectrum-Menu-itemLabel {
   flex: 1 1 auto;
   line-height: var(--spectrum-selectlist-option-label-line-height);
+  word-break: break-all;
 }
 
 .spectrum-Menu-itemLabel--wrapping {


dannify:
I don't think this is the solution we should employ. We still want it to keep whole words that fit on the line and only break if there is overflow. 
So probably `overflow-wrap: break-word` instead of using break-word?
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
It should be tested with a string that has long words and other words with spaces in it.
`The fox jumpedsdfjhsfiuhsgfilubsliubwrfliuwbfliubfiusbvoiwsdjvhbskfhubvsdkufvlsudhfvsduhv over the lazy dog.`

dannify:
@lazd @bernhard-adobe ^

Issue #508, opened 2020-01-29
[Accessibility] StatusLight indicator dot disappears in Windows High Contrast Mode


majornista:
## Description
StatusLight indicator dot disappears in Windows High Contrast Mode.

## Steps to reproduce
1. Enable Windows High Contrast Mode:
  1. Select the Start button , and then select Settings > Ease of Access > High contrast.
  2. To turn on high contrast mode, select the toggle button under Turn on high contrast. Windows may 
  3. display a “Please wait” screen for a few seconds after which the colors on the screen change.
  4. Select the theme that works best for you from the Choose a theme dropdown menu.
  5. To turn off high contrast mode, select the Turn on high contrast toggle button again. Windows may again display a “Please wait” screen for a few seconds after which the colors on the screen change back to default.
3. Using Edge Chromium, Go to https://opensource.adobe.com/spectrum-css/components/statuslight/#Colors
4. Observe StatusLight indicator dots do not display in high contrast mode.

## Expected behavior
StatusLight indicator dots should display in the correct color in Windows High Contrast Mode

## Screenshots
Current behavior:
![StatusLight without support for Windows High Contrast Mode](https://user-images.githubusercontent.com/154077/73407438-1a932f00-42c7-11ea-85b3-0660f67c59e5.png)

Expected behavior:
![StatusLight with support for Windows High Contrast Mode](https://user-images.githubusercontent.com/154077/73407882-70b4a200-42c8-11ea-81c1-dcd2b158ffad.png)

## Environment
 - **Spectrum CSS version:** @spectrum-css/statuslight@2.0.4
 - **Browser(s) and OS(s):** Windows 10 Edge 81.0.408.0 

## Additional context
https://jira.corp.adobe.com/browse/RSP-1401

PR #509, opened 2020-01-30
fix: [Accessibility] StatusLight dot disappears in High Contrast Mode

majornista:
## Description
#508 

## Steps to reproduce
1. Enable Windows High Contrast Mode:
  1. Select the Start button , and then select Settings > Ease of Access > High contrast.
  2. To turn on high contrast mode, select the toggle button under Turn on high contrast. Windows may 
  3. display a “Please wait” screen for a few seconds after which the colors on the screen change.
  4. Select the theme that works best for you from the Choose a theme dropdown menu.
  5. To turn off high contrast mode, select the Turn on high contrast toggle button again. Windows may again display a “Please wait” screen for a few seconds after which the colors on the screen change back to default.
3. Using Edge Chromium, Go to https://opensource.adobe.com/spectrum-css/components/statuslight/#Colors
4. Observe StatusLight indicator dots do not display in high contrast mode.

## Expected behavior
StatusLight indicator dots should display in the correct color in Windows High Contrast Mode

 - **How this was tested:** Using steps in issue #508  -->
 - **Browser(s) and OS(s) this was tested with:** Windows 10 Edge 81.0.408.0 

## Screenshots
Current behavior:
StatusLight without support for Windows High Contrast Mode

Fixed behavior:
StatusLight with support for Windows High Contrast Mode

## To-do list

### DIFF
--- a/components/statuslight/index.css
+++ b/components/statuslight/index.css
@@ -49,6 +49,10 @@ governing permissions and limitations under the License.
 
     /* margin on the dog is the opposite of the padding in order to align the dot as if the padding hack was not present */
     margin: var(--spectrum-statuslight-padding-bottom) var(--spectrum-statuslight-text-gap) var(--spectrum-statuslight-padding-top) var(--spectrum-statuslight-text-gap);
+
+    /* support high contrast/forced-color-adjust mode for dot */
+    -ms-high-contrast-adjust: none;
+    forced-color-adjust: none;
   }
 }
 


components/statuslight/index.css
```diff
@@ -49,6 +49,10 @@ governing permissions and limitations under the License.
 
     /* margin on the dog is the opposite of the padding in order to align the dot as if the padding hack was not present */
     margin: var(--spectrum-statuslight-padding-bottom) var(--spectrum-statuslight-text-gap) var(--spectrum-statuslight-padding-top) var(--spectrum-statuslight-text-gap);
+
+    /* support high contrast/forced-color-adjust mode for dot */
+    -ms-high-contrast-adjust: none;
```
majornista: This proprietary property forces the `.spectrum-StatusLight::before` dot to render the `background-color` in high contrast mode and will work in IE or MS Edge. Note that the dot may not render in Spectrum-CSS documentation using MS Edge 15, because of poor support for CSS variables in pseudo elements.

components/statuslight/index.css
```diff
@@ -49,6 +49,10 @@ governing permissions and limitations under the License.
...
+    forced-color-adjust: none;
```
majornista: This [CSS Color Module Level 4 property](https://www.w3.org/TR/css-color-adjust-1/#forced-color-adjust-prop) forces the `.spectrum-StatusLight::before` dot to render the `background-color` in high contrast mode and will work in Chromium Edge.

Issue #507, opened 2020-01-29
Spectrum DNA 5.22.0 1px offset issues


bernhard-adobe:
## Description

After updating to Spectrum DNA 5.22.0 a lot of components are off by 1px. 

Example VR test is: https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/186/artifact/spectrum-css/backstop_data/html_report/index.html

This came up with the visual testing on #408

## Steps to reproduce

Update DNA to 5.23.0. 
run:

```
yarn install 
yarn add -W -O @spectrum/spectrum-dna

cd components/vars
npm run update
```

a test branch is `SDS-2513-dna-test` has the package.json updates already. 

## Expected behavior
no visual changes

## Screenshots

one of the 77 examples: 

![](https://user-images.githubusercontent.com/52184321/73322915-95dadf00-41fa-11ea-95d8-17d26e35b5f9.png)

## Environment
Every browser independent of Spectrum CSS version

bernhard-adobe:
VR test for DNA  `5.22.0` reduces the issues down from 70 to 40: https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/190/artifact/spectrum-css/backstop_data/html_report/index.html

bernhard-adobe:
This is a bug that got introduced in DNA 5.22.0 by removing the following tokens:

```  --spectrum-alias-heading-display1-text-size: var(--spectrum-global-dimension-font-size-1100);
  --spectrum-alias-heading-xxxl-text-size: var(--spectrum-global-dimension-font-size-1100);
  --spectrum-alias-heading-display1-margin-top: var(--spectrum-global-dimension-font-size-1000);
  --spectrum-alias-heading-xxxl-margin-top: var(--spectrum-global-dimension-font-size-1000);
  --spectrum-alias-heading-display2-text-size: var(--spectrum-global-dimension-font-size-1000);
  --spectrum-alias-heading-xxl-text-size: var(--spectrum-global-dimension-font-size-1000);
  --spectrum-alias-heading-display2-margin-top: var(--spectrum-global-dimension-font-size-900);
  --spectrum-alias-heading-xxl-margin-top: var(--spectrum-global-dimension-font-size-900);
  --spectrum-alias-heading1-text-size: var(--spectrum-global-dimension-font-size-900);
  --spectrum-alias-heading-xl-text-size: var(--spectrum-global-dimension-font-size-900);
[...]

PR #408, opened 2019-11-15
feat: adding t-shirt sized typography (SDS-2513) #210

bernhard-adobe:
Covers:
[SDS-2513](https://jira.corp.adobe.com/browse/SDS-2513) Adding t-shirt sized typography. Issue: #210

[SDS-3259](https://jira.corp.adobe.com/browse/SDS-3259 ) Update Spectrum CSS Typography documentation: Spectrum CSS NextJS & legacy documentation

[SDS-4405](https://jira.corp.adobe.com/browse/SDS-4405) Verify Spectrum CSS Typography Updates

[SDS-4403](https://jira.corp.adobe.com/browse/SDS-4403) Split Typography documentation for Spectrum CSS

[SDS-4460](https://jira.corp.adobe.com/browse/SDS-4460) Update Typekit / Adobe Fonts (Thanks @misterbrownlee )

[SDS-4453](https://jira.corp.adobe.com/browse/SDS-4453) adobe-clean-ux support can be removed #416

## Description

0. Updating Spectrum DNA to get t-shirt sized tokens.
1. Modifying processors.js to allow t-shirt sizes. Some minor restructuring of shared methods.
2. Adding t-shirt sized: sizes, margins and color
3. Enabling the debug statement the processors.js for colors and sizes.
4. Splitting Typography pages 
5. Moving old Typography into Depreciated

For validation the following tokens can be searched:

Paperdoc summary of all changes: https://paper.dropbox.com/doc/Typography-Update--AopzJsKrFFzH_O8gexjcxdEdAg-bluAkl9HGke8WKkJvow2N
(updated t-shirt sized): https://git.corp.adobe.com/gist/nbaldwin/3364cf137d8bd50097c3bedbde9218e5

(original t-shirt sized): https://git.corp.adobe.com/Spectrum/spectrum-dna/pull/310

## How and where has this been tested?
 - **How this was tested:** 
 - **Browser(s) and OS(s) this was tested with:** 

Ran Backstop.js tests a couple of times: Latest test is  https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/157/artifact/spectrum-css/backstop_data/html_report/index.html

Tested on MacOS 10.14.5 (18F203)
on Chrome Version 80.0.3968.0 (Official Build) canary (64-bit)

1. Run: 
```
yarn install 
gulp dev
```

2. Visit locally:

**New pages:**

[Body](http://localhost:3000/docs/typography-body.html)
[...]

Issue #185, opened 2019-05-29
Button text vertical alignment is incorrect in Chrome


lazd:
### Expected Behavior

Looks nice

### Actual Behavior

Looks like this
![image](https://user-images.githubusercontent.com/201344/58207570-cb0c5600-7c97-11e9-9d2f-45dedfd9b18c.png)

#### Browser name/version/os (ie Chrome Version 62.0.3202.94 (Official Build) (64-bit) MacOS)

Chrome 74.0.3729.169 on macOS
Safari 12
Safari 11

#### Spectrum-CSS version

master

#### Notes

[This is the hack that was making text vertically align in Chrome](https://github.com/adobe/spectrum-css/pull/176#discussion_r286679290), and it was removed to allow wrapping buttons... We'll have to either find another hack or live with it. 

The design team has decided that we'll be switching to 14px, which also fixes this issue. If we make that change, we can close this issue.

PR #511, opened 2020-01-31
Upping DNA to 5.23.0 WIP

bernhard-adobe:
Do not merge, WIP while we are going through the DNA updates.
This PR will help us to share the output of the DNA updates.

https://git.corp.adobe.com/Spectrum/spectrum-dna/issues/365
https://github.com/adobe/spectrum-css/issues/507
https://git.corp.adobe.com/Spectrum/spectrum-dna/releases/tag/v5.22.0

run 
```
  yarn install 
  yarn add -W -O @spectrum/spectrum-dna@5.23.0
  cd components/vars
  npm run update
```

## To-do list


components/vars/css/scales/spectrum-large.css
```diff
@@ -97,43 +79,43 @@
   --spectrum-breadcrumb-button-touch-hit-y: var(--spectrum-global-dimension-static-size-50);
   --spectrum-breadcrumb-multiline-item-touch-hit-y: var(--spectrum-global-dimension-static-size-50);
   --spectrum-breadcrumb-multiline-button-touch-hit-y: var(--spectrum-global-dimension-static-size-50);
-  --spectrum-button-cta-text-padding-bottom: [object Object];
+  --spectrum-button-cta-text-padding-bottom: var(--spectrum-global-dimension-size-100);
```
bernhard-adobe: this is a good fix from 5.22.0 and does not create the 1px offset bug.

components/vars/css/scales/spectrum-medium.css
```diff
@@ -57,26 +57,8 @@
   --spectrum-global-dimension-font-size-900: 36px;
   --spectrum-global-dimension-font-size-1000: 40px;
   --spectrum-global-dimension-font-size-1100: 45px;
-  --spectrum-alias-heading-display1-text-size: var(--spectrum-global-dimension-font-size-1100);
```
bernhard-adobe: removing these changes a lot of components by 1px offset

package.json
```diff
@@ -57,7 +57,7 @@
   },
   "optionalDependencies": {
     "@a4u/a4u-collection-essential-ui-icons-release": "^4.1.0",
-    "@spectrum/spectrum-dna": "^5.21",
+    "@spectrum/spectrum-dna": "5.23.0",
```
bernhard-adobe: This is temporary for now.

components/vars/css/globals/spectrum-dimensionAliases.css
```diff
@@ -18,22 +18,42 @@
   --spectrum-alias-single-line-height: var(--spectrum-global-dimension-size-400);
   --spectrum-alias-single-line-width: var(--spectrum-global-dimension-size-2400);
   --spectrum-alias-workflow-icon-size: var(--spectrum-global-dimension-size-225);
-  --spectrum-alias-heading-han-display1-text-size: var(--spectrum-global-dimension-font-size-1000);
-  --spectrum-alias-heading-han-xxxl-text-size: var(--spectrum-global-dimension-font-size-1000);
-  --spectrum-alias-heading-han-display1-margin-top: var(--spectrum-global-dimension-font-size-900);
-  --spectrum-alias-heading-han-xxxl-margin-top: var(--spectrum-global-dimension-font-size-900);
+  --spectrum-alias-heading-display1-text-size: var(--spectrum-global-dimension-font-size-1300);
```
lazd: @bernhard-adobe `--spectrum-alias-heading-display1-text-size` is present here, it seems it was moved from `spectrum-medium.css` to `spectrum-dimensionAliases.css`. The variable is present on the dev site and works, but its value is now `--spectrum-global-dimension-font-size-1300`, whereas previously it was `--spectrum-global-dimension-font-size-1100`. for medium and `--spectrum-global-dimension-font-size-1000` for large.

@NateBaldwinDesign, was this intended?

NateBaldwinDesign: Yes this was an intentional increase in size for the larger type styles

jianliao: Thanks @lazd @NateBaldwinDesign. I manually reverted the change in `spectrum-dimensionAliases.css` and restored it with the old value in `spectrum-medium.css`. Here is the VR result: https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/197/artifact/spectrum-css/backstop_data/html_report/index.html

It seems like all the 1px offset issue gone. So the question now would be what should we do ?

lazd: Ok, so now that the sizes are the same between medium and large, the alias is no longer present in `spectrum-medium.css` and `spectrum-large.css`, and instead is moved up into `spectrum-dimensionAliases.css`. These variables are definitely not missing, though the change in their value could be related to the 1px shift we're seeing. @bernhard-adobe hope this helps.

lazd:
I inspected BarLoader, wondering why the label and the text were misaligned, and found every bit of CSS and rendered sizes to be identical. However, the size and line-height of the header text above the examples changed...

When I delete the header from the page and manually compare, the alignment issue goes away. I think we're dealing with rendering issues in the browser as a result of where the component appears in the page and what has pushed it down a half pixel or otherwise above it.

Looking at the rendered boxes for the header in question (`spectrum-CSSComponent-title spectrum-Article spectrum-Heading1--display`) I see that the previous header had a rendered height of `62.0938px;`, whereas the new header has a rendered height of `84.4688px;`. Perhaps that `0.4688px` extra vs the `0.0938px` on the old one is enough to push some things down by a half-pixel, causing the off-by-one errors.

Sure enough, adding `.spectrum-CSSComponent-title { margin-bottom: -0.5px; }` fixes the issue, confirming there is a sub-pixel problem happening.

Now, why is that happening? The new styles have whole pixel values for the height of the header? Looking at the header, it contains an inline-block anchor. Because its effectively inline, it's rendering some space above the anchor, subject to the line height. Adding `.spectrum-BigSubtleLink { display: block;`
}``

Old: 
image


New: 
image

jianliao:
Here is the latest VR result: https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/200/artifact/spectrum-css/backstop_data/html_report/index.html

It seems like all the 1px up change gone. What's left are the 1px down change.

bernhard-adobe:
looks like some buttons are still impacted. This helped in the typography update: https://github.com/adobe/spectrum-css/pull/408#issuecomment-579561986

>I am able to reduce the the 1px offset issue at least for CTA buttons by altering this hack from #185 by changing `components/buttons.css https://github.com/adobe/spectrum-css/blob/master/components/button/index.css#L107:
>```
>  /padding-bottom: calc(var(--spectrum-button-padding-y) + 0.5px);
> padding-top: calc(var(--spectrum-button-padding-y) - 0.5px);
> ````
> to
>````
>  padding-bottom: calc(var(--spectrum-button-padding-y) -0.5px);
>  padding-top: calc(var(--spectrum-button-padding-y) - 1.5px);
>````
>URL: http://localhost:3000/docs/button-cta.html

lazd:
Looking at Button, I see that the text shifts based on what's above it. That said, the previous reference was closer to the expected design. @bernhard-adobe's hack fix might help.

Looking at Card and Dropdown, it's clear the previous position was due to the composition bug moving things around by a half pixel.

Looking at Stepper, Datepicker, and Quiet Dropdown, I'm unable to reproduce the issue.

I can't reproduce the issue on Small Card either.

I will play with the Button fix and push a commit shortly.

lazd:
We're much closer after the latest test: https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/202/artifact/spectrum-css/backstop_data/html_report/index.html

Almost all of the Button regressions are fixed.

The Toast failures are because Toast was wrong before (confirmed by pasting the reference into XD).

Now we have to deal with SplitButton, Breadcrumbs, and Accordion. I will continue with that shortly.

lazd:
Ok, I went through every remaining failure and compared them to the design files. It turns out we have lots of issues introduced by the composition error, making our reference files mismatch the designs. That, and we have a few bugs. See below:

✅ AssetList: The block/inline-block composition error was pushing the labels around, the new position is correct (verified against Miller Columns XD file)

⚠️ Accordion: The font is wrong here, and it does seem like the reference is slightly closer. That said, there's tons wrong here and I say we avoid trying to handle it as part of this PR. I've filed https://github.com/adobe/spectrum-css/issues/517 to track this.

✅ Breadcrumbs (Title): Though I'm not sure why it shifted, I didn't invest additional time trying to figure it out; this variant is deprecated, so it doesn't matter if it matches the previous design.

✅ Card: The shift is due to a heading above.

✅ Dropdown: The shift is due to a heading above.

✅ Dropdown (quiet): The shift is due to a heading above. The second example was wrong previously (doesn't even match the first). Now they both match.

✅ Datepicker: Reference was wrong. Current matches the design from Contributions.

✅ Datepicker (quiet): Reference was wrong. Current matches the design from Contributions.

⚠️ Pagination Button: The reference is correct -- the test is 1px too high. However, it's looking wrong even in 2.16.0, so there's a bug here. Filed https://github.com/adobe/spectrum-css/issues/518 to track that.

✅ Pagination Listing: The existing reference was wrong (due to #516)

✅ SplitButton: The existing reference assets were wrong. Current matches the design from Contributions.

✅ Stepper (quiet): The existing reference assets were wrong.  Current matches the design from Contributions.

✅ Textfield (quiet): The existing reference assets were wrong.  Current matches the design from Contributions.

✅ Toast: The existing reference assets were wrong.  Current matches the design from Contributions.

With this, I think we might be ok here, but I want another set of eyes on every single failure in the VR test to double check my work. @GarthDB @bernhard-adobe @jianliao someone please double check, and a little cross-browser testing wouldn't hurt either.

bernhard-adobe:
VR test is https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/206/artifact/spectrum-css/backstop_data/html_report/index.html

bernhard-adobe:
Thanks @lazd for all your work on this ticket. I went through the VR test 207 and found the following changes: Let me know if those are still good to go: (I added screenshots below the item.)



> ✅ Card: The shift is due to a heading above.

There is another card called "card light" that seems to also have text shifting. Is that ok with you? 

Screen Shot 2020-02-05 at 13 26 42



>  ✅ Dropdown: The shift is due to a heading above.

It looks to me that inside the dropdown the light gray horizontal rule in the bottom is also shifting. Is that ok with you? 
Screen Shot 2020-02-05 at 13 32 47

bernhard-adobe:
I added also semver back. I took that perviously out for testing while we were still chasing the origin of the bug.

lazd:
@bernhard-adobe I will go ahead and merge it, a second set of eyes on my analysis was all that was needed.

GarthDB:
I was looking over it yesterday but was pulled into a few meetings. I came back to check again this morning and it was merged.

This is good work @bernhard-adobe; thanks for doing it!

bernhard-adobe:
All credits go to @lazd . :)

Issue #512, opened 2020-02-03
Global sizes in CSS dimension globals are out of order


alangwilson:
## Steps to reproduce

1. Go to /components/vars/css/globals/spectrum-dimensionGlobals.css
2. 3 should come before 4.

lazd:
This is interesting, I'm not sure why generated output wouldn't be in order here... Maybe we can just sort the keys before dumping them out...

misterbrownlee:
This is fixed now

PR #513, opened 2020-02-03
Fixed the order of a few of the globals sizes

alangwilson:
## Description
https://github.com/adobe/spectrum-css/issues/512

No tests or screenshot.

## To-do list

- [x ] If my change impacts other components, I have tested to make sure they don't break.
- [x ] If my change impacts documentation, I have updated the documentation accordingly.
- [ x] I have read the [CONTRIBUTING document](/.github/CONTRIBUTING.md).

- [x ] This pull request is ready to merge.


lazd:
Hey @alangwilson, thanks for the PR! These variable files are generated from DNA, and order generally doesn't matter for vars that don't reference one another, so this shouldn't be causing any problems. I'm not sure why the order is wrong, but this fix for it would be overwritten next time we update DNA, so I'll go ahead and close this PR.

Issue #450, opened 2019-12-12
Table should support a non-interactive mode


lazd:
## Description

There should be an option to disable the hover/active states of table rows.

## Why do you need this feature or component?

There are cases where a table will contain non-interactive data that does not require a hover state.

## Additional context

This was requested by a React Spectrum consumer.

mitchell-foote-adobe:
Here is the context and use case. 

We have a table that holds data, but doesn't have any interactive action to it. When a user hovers or clicks, it shows a clicking cursor, and they will expect interactive behavior. We don't provide such interactive behavior. Customers would be confused at the lack of this behavior. 

![Screen Shot 2019-12-12 at 1 18 02 PM](https://user-images.githubusercontent.com/50968569/70745789-da112f00-1ce1-11ea-91f3-505342337730.png)

We need the ability to turn off the interactivity to improve the experience.

lazd:
I wonder if this should be the default, and you have to add `--interactive` to get the row highlight and cursor... Perhaps you can add it on a per-row basis or on the table itself... Thoughts?

bernhard-adobe:
my my POV that decision could be made by asking
What is the more common use case?
What might be more common in the future?

GarthDB:
More than a `noninterative` mode we should probably make a new component that has similar styling as table.

misterbrownlee:
This is an ongoing conversation with React, but we don't need to track the work here.  Outcomes from that conversation will turn into tasks in Jira. Closing this.

PR #526, opened 2020-02-07
Typography migration guide

lazd:
## Description

This pull request does the following:

* Add migration guide for new Typography API
* Switch from `markdown` (deprecated) to `markdown-it`
* Implement Spectrum classes for markdown'd links, tables, and code samples
* Use new Typography components on site template for header/body text
* Use new Typography Code component for markup examples
* Fix color styles that leaked into `site/index.css` (should be in `site/skin.css`)
* Fix header for Code on Typography landing page
* Support `sections` property in metadata to define sections
* Update documentation for component metadata to document new properties and correct errors

 - **How this was tested:** Load up https://git.corp.adobe.com/pages/lawdavis/spectrum-css-typography/docs/typography.html#migratingfromdeprecatedtypography, use eyes

### Typography migration guide

image

### Pretty new links and code

image

### Pretty sections in component pages
image

## To-do list

### DIFF
--- a/components/site/index.css
+++ b/components/site/index.css
@@ -442,8 +442,6 @@ governing permissions and limitations under the License.
 
   border-radius: 0 0 var(--spectrum-cssexample-border-radius) var(--spectrum-cssexample-border-radius);
 
-  font-size: var(--spectrum-global-dimension-font-size-25);
-
   overflow: hidden;
 }
 
--- a/components/site/index.css
+++ b/components/site/index.css
@@ -362,16 +362,6 @@ governing permissions and limitations under the License.
   margin-bottom: var(--spectrum-global-dimension-size-300);
 }
 
-.spectrum-CSSComponent-resource--adobe {
-  color: rgb(255, 2, 1) !important;
-  background-color: var(--spectrum-global-color-gray-100) !important;
-}
-
-.spectrum-CSSComponent-resource--github {
-  background-color: rgba(0, 0, 0, 0.1) !important;
-  color: black;
-}
-
 .spectrum--dark,
 .spectrum--darkest {
   .spectrum-CSSComponent-resource--github {
--- a/components/typography/metadata/typography.yml
+++ b/components/typography/metadata/typography.yml
@@ -43,3 +43,69 @@ examples:
       [View the Code typography component](typography-code.html)
     markup: |
       <code class="spectrum-Code spectrum-Code--S">alert("Hello world");</code>
+  - id: migrating-typography
+    status: Verified
+    name: Migrating from deprecated Typography
+    description: |
+      See the table below to reference what will need to change when migrating to the new Typography API. Cells with a dash indicate the typography style did not exist in the previous API.
+
+      Note that all instances of typography now require the component class in addition to the modifier class. For example, to get a large Heading, you will need the `.spectrum-Heading--L` modifier class as well as the `.spectrum-Heading` component class.
+
+      #### Heading
+
+      | Deprecated Classname                                    | New Classname                                                      |
+      | ------------------------------------------------------- | ------------------------------------------------------------------ |
+      | `~.spectrum-Heading1--display`                           | `.spectrum-Heading.spectrum-Heading--XXXL`                         |
+      | `~.spectrum-Heading1--display.spectrum-Heading1--quiet`  | `.spectrum-Heading.spectrum-Heading--XXXL.spectrum-Heading--light` |
+      | `~.spectrum-Heading1--display.spectrum-Heading1--strong` | `.spectrum-Heading.spectrum-Heading--XXXL.spectrum-Heading--heavy` |
+      | `~.spectrum-Heading2--display`                           | `.spectrum-Heading.spectrum-Heading-XXL`                           |
+      | `~.spectrum-Heading2--display.spectrum-Heading2--quiet`  | `.spectrum-Heading.spectrum-Heading--XXL.spectrum-Heading--light`  |
+      | `~.spectrum-Heading2--display.spectrum-Heading2--strong` | `.spectrum-Heading.spectrum-Heading--XXL.spectrum-Heading--heavy`  |
+      | `~.spectrum-Heading1`                                    | `.spectrum-Heading.spectrum-Heading-XL`                            |
+      | `~.spectrum-Heading1--quiet`                             | `.spectrum-Heading.spectrum-Heading--XL.spectrum-Heading--light`   |
+      | `~.spectrum-Heading1--strong`                            | `.spectrum-Heading.spectrum-Heading--XL.spectrum-Heading--heavy`   |
+      | `~.spectrum-Heading2`                                    | `.spectrum-Heading.spectrum-Heading-L`                             |
+      | `~.spectrum-Heading1--quiet`                             | `.spectrum-Heading.spectrum-Heading--L.spectrum-Heading--light`    |
+      | `~.spectrum-Heading2--strong`                            | `.spectrum-Heading.spectrum-Heading--L.spectrum-Heading--heavy`    |
+      | `~.spectrum-Heading3`                                    | `.spectrum-Heading.spectrum-Heading--M`                            |
+      | `~.spectrum-Heading4`                                    | `.spectrum-Heading.spectrum-Heading--S`                            |
+      | `~.spectrum-Heading5`                                    | `.spectrum-Heading.spectrum-Heading--XS`                           |
+      | `~.spectrum-Heading6`                                    | `.spectrum-Heading.spectrum-Heading--XXS`                          |
+
+
+      #### Body
+
+      | Deprecated Classname | New Classname                        |
+      | -------------------- | ------------------------------------ |
+      | -                    | `.spectrum-Body.spectrum-Body--XXXL` |
+      | -                    | `.spectrum-Body.spectrum-Body--XXL`  |
+      | `~.spectrum-Body1`    | `.spectrum-Body.spectrum-Body--XL`   |
+      | `~.spectrum-Body2`    | `.spectrum-Body.spectrum-Body--L`    |
+      | `~.spectrum-Body3`    | `.spectrum-Body.spectrum-Body--M`    |
+      | `~.spectrum-Body4`    | `.spectrum-Body.spectrum-Body--S`    |
+      | `~.spectrum-Body5`    | `.spectrum-Body.spectrum-Body--XS`   |
+
+
+      #### Detail
+
+      | Deprecated Classname   | New Classname                                 |
+      | ---------------------- | --------------------------------------------- |
+      | -                      | `.spectrum-Detail--XL`                        |
+      | -                      | `.spectrum-Detail--XL.spectrum-Detail--light` |
+      | -                      | `.spectrum-Detail--L`                         |
+      | -                      | `.spectrum-Detail--L.spectrum-Detail--light`  |
+      | -                      | `.spectrum-Detail--M`                         |
+      | -                      | `.spectrum-Detail--M.spectrum-Detail--light`  |
+      | `~.spectrum-Subheading` | `.spectrum-Detail--S`                         |
+      | `~.spectrum-Detail`     | `.spectrum-Detail--S.spectrum-Detail--light`  |
+
+
+      #### Code
+
+      | Deprecated Classname | New Classname       |
+      | -------------------- | ------------------- |
+      | `~.spectrum-Code-1`   | `.spectrum-Code--XL`|
+      | `~.spectrum-Code-2`   | `.spectrum-Code--L` |
+      | `~.spectrum-Code-3`   | `.spectrum-Code--M` |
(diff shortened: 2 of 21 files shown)


components/typography/metadata/typography.yml
```diff
@@ -35,11 +35,77 @@ examples:
...
+    description: |
+      See the table below to reference what will need to change when migrating to the new Typography API. Cells with a dash indicate the typography style did not exist in the previous API.
+
+      Note that all instances of typography now require the component class in addition to the modifier class. For example, to get a large Heading, you will need the `.spectrum-Heading--L` modifier class as well as the `.spectrum-Heading` component class.
+
+      #### Heading
+
+      | Deprecated Classname                                    | New Classname                                                      |
+      | ------------------------------------------------------- | ------------------------------------------------------------------ |
+      | `~.spectrum-Heading1--display`                           | `.spectrum-Heading.spectrum-Heading--XXXL`                         |
```
lazd: Note the `~` at the start of the inline code block -- I extended `markdown-it` to add a class to turn this code red per @NateBaldwinDesign

jianliao:
https://spectrum-visual-regression.ci.corp.adobe.com/job/spectrum-css-vr-test-asset/221/

components/typography/metadata/typography.yml
bernhard-adobe: I personally would have solved this by id'ing the markdown and add the styling via CSS. Your approach requires to loop over the content of the markdown.
I heard this is a way that works for some teams: https://lifelongprogrammer.blogspot.com/2019/01/how-to-style-markdown-with-css.html

site/util.js
```diff
@@ -12,7 +12,77 @@ governing permissions and limitations under the License.
...
+  'thead_open': 'spectrum-Table-head',
+  'tr_open': 'spectrum-Table-row',
+  'tbody_open': 'spectrum-Table-body',
+  'th_open': 'spectrum-Table-headCell',
+  'td_open': 'spectrum-Table-cell',
+  'code_inline': 'spectrum-Code spectrum-Code--S',
+  'code_block': 'spectrum-Code spectrum-Code--S'
+};
+
+for (let [rule, className] of Object.entries(ruleClassnames)) {
```
bernhard-adobe: yeah this approach requires a lot of processing in util.js.

bernhard-adobe:
there is a hover effect and a cursor change to pointer on those example rows?
I clicked on it but didn't notice an action - are you planning on adding an action this? 
Screencast:

components/site/index.css
```diff
@@ -362,16 +362,6 @@ governing permissions and limitations under the License.
   margin-bottom: var(--spectrum-global-dimension-size-300);
 }
 
-.spectrum-CSSComponent-resource--adobe {
```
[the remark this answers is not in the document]
lazd: All colors need to be defined in `skin.css`.

components/site/index.css
```diff
@@ -452,8 +442,6 @@ governing permissions and limitations under the License.
 
   border-radius: 0 0 var(--spectrum-cssexample-border-radius) var(--spectrum-cssexample-border-radius);
 
-  font-size: var(--spectrum-global-dimension-font-size-25);
```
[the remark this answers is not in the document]
lazd: It got deleted; it was replaced by applying `.spectrum-Code--S` on the outer element (yay for new typography styles!)

jianliao:
Not sure why the action-bar case failed.
image

site/util.js
lazd: This is applying Spectrum classnames to links, tables, and code elements. It was actually easier than I thought it would be!

components/typography/metadata/typography.yml
lazd: Unfortunately, that syntax doesn't seem to be supported: 
image

It does seem we could use [custom containers](https://github.com/markdown-it/markdown-it-container) and apply `::: spectrum-CSSComponent-migrationTable` around the table, then style `.spectrum-Table-body .spectrum-Table-cell:first-child` to accomplish this, but that's quite a bit less direct than the (admittedly weird) approach taken. Thoughts?

lazd:
@bernhard-adobe so unfortunately, Table be like that. We have an issue filed to support a non-interactive mode to remove the hover state and cursor: https://github.com/adobe/spectrum-css/issues/450

components/typography/metadata/typography.yml
bernhard-adobe: Adding a class name on custom containers is what I would do and not id-ing each element
would this work around the content placed in a the table?

```
::: warning
*here be dragons*
:::
```
to hopefully:
```
<div class="warning">
<em>here be dragons</em>
</div>
```
and then we could target the 'warning' class by renaming it and styling it in CSS?

Another alternative could be to use the sibling selector to style the content followed after a selector.  https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator

This personally feels to me a bit more direct than looping over a utility function over the content. But that's just my own personal feeling and I am open for other suggestions.

site/util.js
bernhard-adobe: This is my personal opinion but I always thought having the ids and classnames on the element is more implicit. but that's my personal opinion.

components/typography/metadata/typography.yml
lazd: Yeah we could wrap every single instance of the code block with the `:::warning` bit, that would do it. I'll make that change.

lazd: @bernhard-adobe uhg, that doesn't work. Because of the way tables are formatted, we can't put a block element inside of it. I can fall back to the outer class and use the `:first-child` bit, but to be honest, that's not really gaining us anything. 

The tokenization function runs already, we're just replacing the rule with our own, so we're not adding any runtime complexity or slowing things down (docs generation takes 5 seconds).

Would you block this pull request pending a different solution, or are you ok with the existing solution?

site/util.js
lazd: I think you misunderstand; this is effectively turning markdown into Spectrum CSS-ified code. We don't want to go and add classes to every instance of `[link](to something else)` or `# Heading` -- instead, we use the renderer architecture of `mardown-it` exactly as it was designed to be used to apply our own custom classes to the rendered HTML markup. We're not writing elements, we're writing markdown, and this is the accepted way to add classes

bernhard-adobe: I think my proposed solution is on how to apply those custom classes directly in markdown without the use of a custom processor.

components/typography/metadata/typography.yml
bernhard-adobe: Thank you @lazd  for your customization.
It is still unclear to me why you would prefer writing those rules in JavaScript as an utility function if you could use a wrapper / container ? https://github.com/markdown-it/markdown-it-container This feels more like a straight forward solution to me that might help in the future with area in the documentation that could benefit from customization.

lazd: So currently, we use markdown in our docs. When we write the following:

```
      Heading is used to create various levels of typographic hierarchies.
      [View the Heading typography component](typography-heading.html)
```

It automatically turns into this:

```
<p>Heading is used to create various levels of typographic hierarchies.</p>
<p><a href="typography-heading.html">View the Heading typography component</a></p>
```

That looks like the following:
image

You can see it looks wrong because we're missing the `spectrum-Link` class, right?

So, I've added rules to the markdown parser such that it will output the following HTML:

```
<p>Heading is used to create various levels of typographic hierarchies.</p>
<p><a href="typography-heading.html" class="spectrum-Link">View the Heading typography component</a></p>
```

That looks like the following:

image

Now you can see, it's a real Spectrum link! 

Of course, we could write HTML inside of our markdown and get the same result:

```
      Heading is used to create various levels of typographic hierarchies.
      <a href="typography-heading.html">View the Heading typography component</a>
```

But why would we want to do that? Instead, our markdown is automatically rendered with the correct Spectrum CSS classes.

I did the same for `### Headings`, ``code``, and tables. This is awesome, right?

In the case of the red code blocks, however, I extended the existing code block to add another class if it finds a `~` as the first char. Other approaches with the wrapper were not feasible, and wrapping the tables and using overly specific CSS to target it doesn't make sense.

It's magical and super awesome, and all of the potential solutions either don't work or are tedious and silly. We're using `markdown-it` precisely how it was designed, and there are tons of issues filed against their repository asking how to do exactly what we're doing, documentation that shows how to do exactly what we're doing etc.

site/util.js
lazd: So currently, we use markdown in our docs. When we write the following:

```
      Heading is used to create various levels of typographic hierarchies.
      [View the Heading typography component](typography-heading.html)
```

It automatically turns into this:

```
<p>Heading is used to create various levels of typographic hierarchies.</p>
<p><a href="typography-heading.html">View the Heading typography component</a></p>
```

That looks like the following:
image

You can see it looks wrong because we're missing the `spectrum-Link` class, right?

So, I've added rules to the markdown parser such that it will output the following HTML:

```
<p>Heading is used to create various levels of typographic hierarchies.</p>
<p><a href="typography-heading.html" class="spectrum-Link">View the Heading typography component</a></p>
```

That looks like the following:

image

Now you can see, it's a real Spectrum link! 

Of course, we could write HTML inside of our markdown and get the same result:

```
      Heading is used to create various levels of typographic hierarchies.
      <a href="typography-heading.html">View the Heading typography component</a>
```

But why would we want to do that? Instead, our markdown is automatically rendered with the correct Spectrum CSS classes.

I did the same for `### Headings`, ``code``, and tables. This is awesome, right?

In the case of the red code blocks, however, I extended the existing code block to add another class if it finds a `~` as the first char. Other approaches with the wrapper were not feasible, and wrapping the tables and using overly specific CSS to target it doesn't make sense.

It's magical and super awesome, and all of the potential solutions either don't work or are tedious and silly. We're using `markdown-it` precisely how it was designed, and there are tons of issues filed against their repository asking how to do exactly what we're doing, documentation that shows how to do exactly what we're doing etc. Again, these aren't utility functions, we're extending the markdown parser's built-in rules and adding our classes to them. There is no performance penalty, and instead we just get beautifully formatted, Spectrum HTML without having to lift a finger. There are zero drawbacks.

Your proposed solution is to simply write HTML within the `yml` files instead of markdown. If that's the case, we might as well completely rip out the markdown parser, edit every instance of a link or backtick code block in every component, and change it all to HTML. We don't want to do that, markdown is here to make writing docs easier, and we should use it.

bernhard-adobe: Thanks for the in-depth explanation @lazd . I think I fully understand your approach but I really appreciate your explanation.
In our direct message on Slack I shared with you the example how Boostrap is for example trying to solve this with classnames as well. 
Your approach of applying automatically classnames on HTML elements  might be confusing in the future for someone who is writing our documentation as they might experience classnames being added to elements that they did not expect to be added. This might not be ideal as it might require a previous understanding that post-processing on markdown will be applied. 
For you this might feel like a good solution for simply elements like a header or link. For me it feels like this will get a bit more complex for table and items nested in each other and it may lead to unexpected results and side-effects where wrong styling might need to be applied. 
For example the table classes you are applying are turning the documentation table into an interactive table with hover styles and cursor changes. We would have to apply in utils.js a rule that applies either way an "interactive" class or removes one in the future depending how this might be fixed.

lazd: All tables are interactive, so even if we wrote all of the HTML hand, it would still have hover styles and cursor changes. See #450, where I have proposed that non-interactive tables are default, and you must add `--interactive` to make them interactive. Once that's implemented, if we ever wanted interactive tables in our examples (which we likely will not), we could simply write the HTML, as you have proposed.

As far as Bootstrap's use of HTML to render tables, we absolutely could do that ourselves, but why? I would expect tables in our markdown to come out looking Spectrum. In addition, Bootstrap doesn't need to add classes to their headings, because Bootstrap indiscriminately  styles naked `<h2>` elements out of the box, and we don't. We specifically require that people opt into our styles by adding classnames, and this change simply makes follow our own recommendation and apply classes to heading, link, and table elements.

The only reason I did not do this in the first place when I added `markdown` to our build process was because it was not possible. Now that we switched to `markdown-it`, this is the obvious, simple approach to generating documentation that matches Spectrum.

That said, I hear you that people might be surprised by Spectrum CSS classes being added automatically. In response to that, I've added a note to the documentation that defines how the metadata is structured. Does that satisfy your concerns?

bernhard-adobe: Thank you @lazd  for the clarification again. It sounds like you are proceeding then with adding those classnames to elements automatically and you are suggesting that future users should read more documentation to understand how to document the documentation. 
This still feels a bit complicated to me but that might be my personal feeling about documentation.

lazd: @bernhard-adobe yes, I believe it's the simplest and most elegant solution. Unless you choose to block the PR, I say we go forward with the Spectrum CSSified markdown and the documentation of how it works.

bernhard-adobe: Thanks @lazd , it looks like you are set in your opinion here. I don't think this should be blocking the PR as I feel this discussion might not move forward at this point.

PR #603, opened 2020-03-03
feat: halo focus ring

GarthDB:
## Description
Added support for the halo focus ring on:

- Button
- Logic Button
- Checkbox
- Radio
- ToggleSwitch
- Slider

### Steps
1. Run `gulp dev`
2. Visit http://localhost:3000/docs/halo-focus.html
3. Tab around like a maniac

### Tested in:
* ✅ Chrome
* ✅ Safari
* ✅ Firefox
* ❌ IE 11
    - Checked Checkboxes have incorrect halo size (it's inside of the box)
    - Checked Radios have incorrect halo size (doesn't have gap)

## To-do list

- [x] Make animation of ring look nicer
- [x] Create stand-in DNA vars for some components (right now, reusing the Button's vars for Slider)
- [x] Fix IE 11 bugs
- [x] Final design signoff


components/button/skin.css
```diff
@@ -227,10 +234,20 @@ governing permissions and limitations under the License.
...
+    box-shadow: none;
+    &::after {
+      content: '';
+      width: 100%;
+      height: 100%;
+      border: 2px solid var(--spectrum-button-over-background-border-color-key-focus);
+      padding: 4px;
+      position: absolute;
+      top: -6px;
+      border-radius: 22px;
```
lazd: Put all dimensional values in `index.css`, and create variables for each one, preferably `calc`'d based on existing vars (even if DNA doesn't exist yet) so we have some semantic meaning.

components/button/skin.css
```diff
@@ -227,10 +234,20 @@ governing permissions and limitations under the License.
...
-    box-shadow: 0 0 0 var(--spectrum-button-primary-border-size-increase-key-focus) var(--spectrum-button-quiet-over-background-border-color-key-focus);
+    background-color: var(--spectrum-button-quiet-over-background-background-color-hover);
+    border-color: var(--spectrum-button-quiet-over-background-border-color-hover);
+    color: var(--spectrum-button-quiet-over-background-text-color-hover);
+    box-shadow: none;
+    &::after {
+      content: '';
+      width: 100%;
+      height: 100%;
+      border: 2px solid var(--spectrum-button-over-background-border-color-key-focus);
```
lazd: Just define the color here, everything else goes in `index.css`.

components/button/skin.css
```diff
@@ -227,10 +234,20 @@ governing permissions and limitations under the License.
...
 
   &:focus-ring {
-    background-color: var(--spectrum-button-quiet-over-background-background-color-key-focus);
-    border-color: var(--spectrum-button-quiet-over-background-border-color-key-focus);
-    color: inherit;
-    box-shadow: 0 0 0 var(--spectrum-button-primary-border-size-increase-key-focus) var(--spectrum-button-quiet-over-background-border-color-key-focus);
+    background-color: var(--spectrum-button-quiet-over-background-background-color-hover);
+    border-color: var(--spectrum-button-quiet-over-background-border-color-hover);
+    color: var(--spectrum-button-quiet-over-background-text-color-hover);
+    box-shadow: none;
```
lazd: `index.css`

components/button/skin.css
```diff
@@ -10,14 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-:root {
-  --spectrum-button-primary-border-size-increase-key-focus: 1px;
-  --spectrum-actionbutton-quiet-border-size-key-focus: 1px;
-}
 
 .spectrum-Button {
   &:focus-ring {
-    box-shadow: 0 0 0 var(--spectrum-button-primary-border-size-increase-key-focus) var(--spectrum-button-primary-border-color-key-focus);
+    box-shadow: 0 0 0 2px var(--spectrum-button-primary-focus-ring-gap-color-key-focus), 0 0 0 4px var(--spectrum-button-primary-focus-ring-color-key-focus);
```
lazd: Define vars with semantic names for those `2px` and `4px` values.

components/button/skin.css
```diff
@@ -185,6 +181,7 @@ governing permissions and limitations under the License.
   background-color: var(--spectrum-button-over-background-background-color);
   border-color: var(--spectrum-button-over-background-border-color);
   color: var(--spectrum-button-over-background-text-color);
+  position: relative;
```
lazd: `index.css`

components/checkbox/skin.css
```diff
@@ -10,106 +10,96 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 }
 
 /* Focus */
 .spectrum-Checkbox-input {
   &:focus-ring + .spectrum-Checkbox-box {
-    /* Since the specificity of .focus-ring is less than :hover, we need important */
-    border-color: var(--spectrum-checkbox-emphasized-box-border-color-key-focus) !important;
-
-    box-shadow: 0 0 0 1px var(--spectrum-checkbox-emphasized-box-border-color-key-focus) !important;
+    box-shadow:  0 0 0 2px var(--spectrum-checkbox-focus-ring-gap-color-key-focus), 0 0 0 4px var(--spectrum-checkbox-focus-ring-color-key-focus);
```
lazd: Ideally, you would create a global DNA var (by adding a [static alias](https://git.corp.adobe.com/Spectrum/spectrum-dna/blob/master/data/aliases/StaticAliases.mjs)) for the focus ring width and offset, and use that (if it doesn't already exist)...

If you don't do that, note that you can use any global DNA vars (non-scale dependent) inside of `skin.css`:

* `--spectrum-alias-border-size-thick` or `--spectrum-global-dimension-static-size-25` for 2px
* `--spectrum-alias-border-size-thicker` or `--spectrum-global-dimension-static-size-50` for 4px

As a last resort, you can put them in `commons/index.css` or create a `commons/skin.css` and import it inside of `commons/index.css`, but it would be best to create a DNA var. Can you create static aliases instead?

components/button/index.css
```diff
@@ -125,6 +125,43 @@ governing permissions and limitations under the License.
   }
 }
 
+.spectrum-Button--overBackground,
+.spectrum-Button--overBackground.spectrum-Button--quiet {
```
lazd: Is this extra selector required because `.spectrum-Button--quiet` overrides some of this?

components/button/index.css
```diff
@@ -236,6 +287,25 @@ a.spectrum-ActionButton {
...
+
+  &:hover {
+    box-shadow: none;
+  }
+  &:active {
+    box-shadow: none;
+  }
+  &.is-selected {
+    &:focus-ring,
+    &.is-focused {
```
lazd: Does this also need `:focus`? For anything that can physically be focused (i.e. `tabindex` or already focusable), if we ever have `.is-focused`, we need `:focus` on the side, IIRC.

components/button/skin.css
```diff
@@ -11,18 +11,14 @@ governing permissions and limitations under the License.
 */
 
 :root {
-  --spectrum-button-primary-border-size-increase-key-focus: 1px;
-  --spectrum-actionbutton-quiet-border-size-key-focus: 1px;
+  --spectrum-button-primary-focus-ring-gap-key-focus: var(--spectrum-alias-focus-ring-gap);
+  --spectrum-button-primary-focus-ring-size-key-focus: calc(var(--spectrum-alias-focus-ring-gap) + var(--spectrum-alias-focus-ring-size));
```
lazd: +1

lazd:
From today's discussion re: Clear Button in toast, the outline should probably be in the same place as the over background button next to it.

jianliao:
Hi, I found a few build errors in this PR. 
`Error running 'checkbox:build': CssSyntaxError in plugin "gulp-postcss"`

components/button/index.css
lazd: Er why was this there at all? It makes it so you can't focus on selected field buttons. Thoughts?

lazd: TBH I don't think any of those `box-shadow` resets were required... Removing 'em!

GarthDB: Most of that was to keep was existing and move it to the correct css file. I'm fine deleting it if box shadows aren't involved anymore and there are no visible problems.

components/button/index.css
lazd: Removed in latest commits.

components/button/index.css
lazd: Removed! Things look right still.

jianliao:
Pull the trigger too early. Is this an issue?
image

lazd:
@jianliao fixed the Dial issue, good catch!

Now will just fixup the DNA vars and then we can merge.

components/button/index.css
```diff
@@ -25,6 +25,9 @@ governing permissions and limitations under the License.
 }
 
 %spectrum-BaseButton {
+  /* Contain halo */
+  position: relative;
```
lazd: I hope this doesn't break anything for anyone.

AShroti: @lazd yes, unfortunately, it causes major regression CUI-7431

PR #650, opened 2020-03-19
feat: link component change for rsp@v3

jianliao:
## Description
#553 

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/link/index.css
+++ b/components/link/index.css
@@ -26,16 +26,6 @@ governing permissions and limitations under the License.
     text-decoration: underline;
     text-decoration-style: double;
   }
-
-  &.is-disabled {
-    cursor: default;
-    pointer-events: none;
-
-    &:hover,
-    &:focus {
-      text-decoration: none;
-    }
-  }
 }
 
 .spectrum-Link--quiet {
--- a/components/link/index.css
+++ b/components/link/index.css
@@ -17,9 +17,10 @@ governing permissions and limitations under the License.
   background-color: transparent;
   /* Remove gaps in links underline in iOS 8+ and Safari 8+. */
   -webkit-text-decoration-skip: objects;
-  text-decoration: none;
+  text-decoration: underline;
   transition: color var(--spectrum-global-animation-duration-100) ease-in-out;
   outline: none;
+  cursor: pointer;
 
   &:hover {
     text-decoration: underline;
@@ -45,11 +46,14 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Link--subtle,
-/** @deprecated */.spectrum-Link--quiet {
+.spectrum-Link--secondary {
   text-decoration: underline;
 }
 
 .spectrum-Link--overBackground {
   text-decoration: underline;
+}
+
+.spectrum-Link--quiet {
+  text-decoration: none;
 }
--- a/components/link/index.css
+++ b/components/link/index.css
@@ -22,14 +22,6 @@ governing permissions and limitations under the License.
   outline: none;
   cursor: pointer;
 
-  &:hover {
-    text-decoration: underline;
-  }
-
-  &:focus-ring {
-    text-decoration: underline;
-  }
-
   &:focus-ring {
     text-decoration: underline;
     text-decoration-style: double;
@@ -46,14 +38,10 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Link--secondary {
-  text-decoration: underline;
-}
-
-.spectrum-Link--overBackground {
-  text-decoration: underline;
-}
-
 .spectrum-Link--quiet {
   text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
 }
--- a/components/link/metadata/link.yml
+++ b/components/link/metadata/link.yml
@@ -1,6 +1,11 @@
 id: link-primary
 name: Link
 SpectrumSiteSlug: https://spectrum.adobe.com/page/link/
+sections:
+  - name: Migration Guide
+    description: |
+      ### Subtle variant
+      Subtle variant was removed. Please use Secondary.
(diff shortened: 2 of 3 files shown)


components/link/index.css
```diff
@@ -17,9 +17,10 @@ governing permissions and limitations under the License.
   background-color: transparent;
   /* Remove gaps in links underline in iOS 8+ and Safari 8+. */
   -webkit-text-decoration-skip: objects;
-  text-decoration: none;
+  text-decoration: underline;
```
lazd, line 20: Not sure about this. I don't believe our links have an underline by default.

components/link/index.css
```diff
@@ -17,9 +17,10 @@ governing permissions and limitations under the License.
...
   transition: color var(--spectrum-global-animation-duration-100) ease-in-out;
   outline: none;
+  cursor: pointer;
```
lazd, line 23: Hmm, I don't think this is necessary for `<a>`, but perhaps they're using the styles elsewhere, so it's fine.

components/link/index.css
```diff
@@ -45,11 +46,14 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Link--subtle,
-/** @deprecated */.spectrum-Link--quiet {
+.spectrum-Link--secondary {
   text-decoration: underline;
```
lazd, line 50: I think this is duplicative? We already have `text-decoration: underline` above

components/link/index.css
```diff
@@ -45,11 +46,14 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Link--subtle,
-/** @deprecated */.spectrum-Link--quiet {
+.spectrum-Link--secondary {
```
lazd, line 49: This needs to have `color: inherit` or something -- secondary links are supposed to look like the paragraph text. Not sure what to do on hover? https://spectrum.adobe.com/page/link/#Secondary-variant

components/link/metadata/link.yml
```diff
@@ -1,34 +1,50 @@
+id: link-primary
 name: Link
 SpectrumSiteSlug: https://spectrum.adobe.com/page/link/
+sections:
+  - name: Migration Guide
+    description: |
+      ### Subtle variant
+      Subtle variant was removed. Please use Secondary.
```
lazd, line 8: I believe `quiet` is the substitution for `subtle`, but I could be wrong.

components/link/metadata/link.yml
```diff
@@ -1,34 +1,50 @@
...
       <div style="background-color: rgb(15, 121, 125); padding: 15px 20px; display: inline-block;">
         <p class="spectrum-Body" style="color: rgb(255, 255, 255);">
           Hello, this is my <a href="#" class="spectrum-Link spectrum-Link--overBackground"> spectrum-Link--overBackground</a>. This is just filler text, but if you keep reading maybe something good will happen.
         </p>
       </div>
+  - id: link-quiet-primary
+    name: Link - Quiet (Primary)
+    status: Verified
+    markup: |
+      <p> This uses <a href="#" class="spectrum-Link spectrum-Link--quiet">spectrum-Link--quiet</a>.</p>
```
lazd, line 30: We don't have `primary` modifier, it seems links are primary by default. Is that what we want?

components/link/skin.css
```diff
@@ -30,8 +30,7 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Link--quiet,
-/** @deprecated */.spectrum-Link--subtle {
+.spectrum-Link--secondary{
   color: inherit;
```
lazd, line 34: That looks right. Not sure what to do on hover though, is it correct as-is? 
https://spectrum.adobe.com/page/link/#Secondary-variant

components/link/index.css
jianliao: I've cleaned up the duplicated `text-decoration` property.

components/link/metadata/link.yml
jianliao: Fixed.

components/link/metadata/link.yml
jianliao: @GarthDB, could you please confirm this? `primary` modifier got removed prior to this change.

components/link/skin.css
jianliao: hmm... This change was pulled in from RSP.

According to the description of Secondary variant on the spectrum-documentation site, the color should be [the same gray color as the paragraph text](https://spectrum.corp.adobe.com/page/link/#Secondary-variant).

Is it this reason that it used `inherit`? But I did find DNA tokens for  Secondary variant.

` --spectrum-link-secondary-text-color: var(--spectrum-alias-text-color);
   --spectrum-link-secondary-text-color-hover: var(--spectrum-alias-text-color-hover);
  --spectrum-link-secondary-text-color-key-focus: var(--spectrum-alias-text-color-hover);
`

lazd:
The hover behavior seems to be missing, not sure what it's supposed to do? https://opensource.adobe.com/spectrum-css-vr-results/30854f8-result-2020-04-22_19-57-27/dist/docs/link.html

components/link/skin.css
NateBaldwinDesign: General rule of thumb we should try to abide by: Find DNA tokens and use them whenever available. For now, inheriting is displaying the correct colors, but if we change this in the future for *secondary links only* (no other type), then the change will not be picked up via DNA upgrade as expected.

lazd:
@jianliao @GarthDB pushed a commit that drops the `.is-disabled` class and example -- this has been removed from the design system and now is a fine time to remove it.

lazd:
AFAIK, @NateBaldwinDesign said there are no hover styles for Primary and Secondary links, so I believe this PR is good to go!

PR #664, opened 2020-04-05
update path to icons dist in readme quickstart icon usage

filmaj:
the path to the icons directory in a typical node install was one character off 😄


jianliao:
Sorry, I approved too quickly. It turns out package [@spectrum-css/icons](https://www.npmjs.com/package/@spectrum-css/icons) is a testing package. The one in used should be [@spectrum-css/icon](https://www.npmjs.com/package/@spectrum-css/icon). cc: @lazd

PR #710, opened 2020-04-29
feat: RSP V3 Dialog

GarthDB:
## Description
Syncing up with React Spectrum V3 work on Dialog.

It has some breaking stuff by using grid and such, but it also breaks our examples.

 - **How this was tested:** 

## Screenshots

image

## To-do list


components/dialog/index.css
```diff
@@ -34,337 +34,321 @@ governing permissions and limitations under the License.
...
 }
 
 .spectrum-Dialog--small {
-  inline-size: var(--spectrum-dialog-small-width);
+  width: var(--spectrum-dialog-small-width);
 }
 
 .spectrum-Dialog--medium {
-  inline-size: var(--spectrum-dialog-medium-width);
+  width: var(--spectrum-dialog-medium-width);
```
bernhard-adobe, line 70: are we switching back to width instead of inline-size? I thought this was for RTL support

components/dialog/index.css
```diff
@@ -34,337 +34,321 @@ governing permissions and limitations under the License.
...
-.spectrum-Dialog-header {
-  display: flex;
-  /* Without this, buttons will be stretched */
-  align-items: center;
+.spectrum-Dialog .spectrum-Dialog-grid {
+  display: grid;
+  grid-template-columns: var(--spectrum-dialog-padding) auto 1fr auto minmax(0, auto) var(--spectrum-dialog-padding);
+  grid-template-rows: auto var(--spectrum-dialog-padding) auto auto 1fr auto var(--spectrum-dialog-padding);
+  grid-template-areas:
+    "hero hero    hero    hero        hero        hero"
```
bernhard-adobe, line 94: nice. I used http://asciiflow.com on the past but this feels kind of self-explaining.

Screen Shot 2020-05-29 at 17 29 34

components/dialog/index.css
```diff
@@ -34,337 +34,321 @@ governing permissions and limitations under the License.
...
   font-weight: var(--spectrum-dialog-content-text-font-weight);
   line-height: var(--spectrum-dialog-content-text-line-height);
-}
 
-@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
-  .spectrum-Dialog-content {
-    /* Temporary IE 11 fix */
-    max-block-size: 70vh;
-  }
+  /* this is kinda dumb, but needed for the keyboard focus rings so they don't get clipped. is there a better way to treat this */
```
bernhard-adobe, line 172: not sure if there is a better way. space needs to be made to ensure the focus ring is visible. 
only other option I would see requires scripting and placing buttons via `position:absolute` on top of the dialog. not the best solution IMO.

components/dialog/index.css
```diff
@@ -34,337 +34,321 @@ governing permissions and limitations under the License.
...
+  width: 100%;
+  height: 100%;
 }
+.spectrum-Dialog--fullscreenTakeover {
+  width: 100%;
+  height: 100%;
 
+  border-radius: 0;
+}
+/** @unofficial */
```
bernhard-adobe, line 254: why is this @unofficial?

components/popover/index.css
```diff
@@ -31,8 +37,8 @@ governing permissions and limitations under the License.
   flex-direction: column;
   box-sizing: border-box;
 
-  min-inline-size: var(--spectrum-popover-min-width);
-  min-block-size: var(--spectrum-popover-min-height);
+  min-width: var(--spectrum-popover-min-width);
```
bernhard-adobe, line 40: same comment as before: won't this impact RTL?

components/dialog/index.css
GarthDB: This comes from @snowystinger and I think it means that the fullscreen and fullscreen takeover are in contributions.

components/dialog/index.css
lazd: Negative margin is fine here, it's hacky, but that's the entire web.

lazd: @garthdb Would prefer that these globals are assigned to a diff variable though.

GarthDB: Good call.

snowystinger:
Why do so many of the dialogs show up with `deprecated`?

PR #734, opened 2020-05-19
fix: correct Treeview icon/text alignment

lazd:
## Description

* fix: correct Treeview icon/text alignment
* BREAKING CHANGE: spectrum-Treeview-icon now required on all non-indicator icons

 - **How this was tested:** 

## To-do list

- [x] Resolve design issues with hover state


components/treeview/metadata/treeview.yml
```diff
@@ -77,117 +77,383 @@ examples:
...
         <li class="spectrum-TreeView-item is-open">
           <a class="spectrum-TreeView-itemLink" href="#">
-            <svg class="spectrum-Icon spectrum-UIIcon-ChevronRightMedium spectrum-TreeView-indicator" focusable="false" aria-hidden="true">
+            <svg class="spectrum-Icon spectrum-UIIcon-ChevronRightMedium spectrum-TreeView-itemIndicator" focusable="false" aria-hidden="true">
               <use xlink:href="#spectrum-css-icon-ChevronRightMedium" />
             </svg>
-            <svg class="spectrum-Icon spectrum-Icon--sizeS" focusable="false" aria-hidden="true" aria-label="Folder">
-              <use xlink:href="#spectrum-icon-18-Folder"></use>
+            <svg class="spectrum-TreeView-itemIcon spectrum-Icon spectrum-Icon--sizeS" focusable="false" aria-hidden="true" aria-label="Folder">
+              <use xlink:href="#spectrum-icon-18-Folder"><span class="spectrum-TreeView-itemLabel"></span></use>
```
jianliao, line 138: I think this `<span />` should not inside `<use />` and it is redundant. I will remove it along with the change to add thumbnail component VRT scenario.

lazd: Ah, looks like I screwed up the regex when I added the label. Thanks!

lazd:
Thanks for the fix, @jianliao. I made the `<use>` not self-close as that's how we do it elsewhere, and I think this is ready for merge.

PR #889, opened 2020-09-03
Fixes Toast separator for color stop builds

### LINKED ISSUE #884: Toast missing separator in color stop build
## Description
Toast is supposed to have a separator

The following CSS is output for the colorstop css:

````
.spectrum--light [dir="ltr"] .spectrum-Toast-buttons {
  border-left-color: rgba(255, 255, 255, 0.2);
}
.spectrum--light [dir="rtl"] .spectrum-Toast-buttons {
  border-right-color: rgba(255, 255, 255, 0.2);
}
````
I don't understand all the nitty gritty of CSS, but for some reason, browsers don't seem to see this CSS. I thought  that it might be the dir='ltr' between the selectors, but moving it to the beginning does not help. Somehow, the combination of descendants and dir pseudo-selectors don't seem to work.

Considering that the colors are the same for all color stops, it seems to me that the fix is simply to have a single set of declarations for all color stops:
````
[dir="ltr"] .spectrum-Toast-buttons {
  border-left-color: rgba(255, 255, 255, 0.2);
}
[dir="rtl"] .spectrum-Toast-buttons {
  border-right-color: rgba(255, 255, 255, 0.2);
}
````
The above CSS works as expected.

### PR DESCRIPTION
**Pull request by Harbs**
Fixes Toast separator for color stop builds

## Description
Details in #884

 - **How this was tested:** CSS copied to my project which uses the color stops and inspected visually.
 - **Browser(s) and OS(s) this was tested with:** Firefox and Chrome

## To-do list

### DIFF
--- a/components/toast/index.css
+++ b/components/toast/index.css
@@ -59,6 +59,7 @@ governing permissions and limitations under the License.
   display: flex;
   flex: 0 0 auto;
   align-items: flex-start;
+  border-inline-start-color: rgba(255, 255, 255, 0.2);
 
   .spectrum-Button,
   .spectrum-ClearButton {
--- a/components/searchwithin/index.css
+++ b/components/searchwithin/index.css
@@ -39,7 +39,6 @@ governing permissions and limitations under the License.
 
 .spectrum-SearchWithin-input {
   flex: 1;
-  min-width: inherit;
   margin-inline-start: calc(-1 * var(--spectrum-textfield-border-size)); /* hides left border */
   border-start-start-radius: var(--spectrum-searchwithin-border-radius);
   border-end-start-radius: var(--spectrum-searchwithin-border-radius);
@@ -53,8 +52,4 @@ governing permissions and limitations under the License.
   position: absolute;
   inset-block-start: 0;
   inset-inline-end: 0;
-}
-
-.spectrum-SearchWithin > .spectrum-Textfield {
-  flex-grow: 1;
 }
--- a/components/searchwithin/index.css
+++ b/components/searchwithin/index.css
@@ -39,6 +39,7 @@ governing permissions and limitations under the License.
 
 .spectrum-SearchWithin-input {
   flex: 1;
+  min-width: inherit;
   margin-inline-start: calc(-1 * var(--spectrum-textfield-border-size)); /* hides left border */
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Harbs** [comment]
(I mistakenly pushed the next fix to the same branch.)

File: components/toast/index.css [hunk 1]
```diff
@@ -59,6 +59,7 @@ governing permissions and limitations under the License.
   display: flex;
   flex: 0 0 auto;
   align-items: flex-start;
+  border-inline-start-color: rgba(255, 255, 255, 0.2);
```
GarthDB, line 62: We put colors in the skin.css. Can you have border-color still in skin.css and then set the border-width for inline-start in the index.css file?

GarthDB: Also, make your prs against the `spectrum-tokens-17` branch while we're working on it make it easier to merge.

GarthDB (changes requested):

components/toast/index.css
Harbs: This commit moved the color out of the skin file. The reason I did so was to fix the bug. When compiled for color stops when the css is in `skin.css`, the css compiles like so:
````
.spectrum--light [dir="ltr"] .spectrum-Toast-buttons {
  border-left-color: rgba(255, 255, 255, 0.2);
}
````
That does not work because it makes assumptions on the hierarchy where you need to have nesting like this: `.spectrum--light` > `[dir="ltr"]` > `.spectrum-Toast-buttons`

Changing the order of processing can turn that into:
````
[dir="ltr"] .spectrum--light .spectrum-Toast-buttons {
  border-left-color: rgba(255, 255, 255, 0.2);
}
````
But that assumes a hierarchy of `[dir="ltr"]` > `.spectrum--light` > `.spectrum-Toast-buttons`

My current implementation applies `.spectrum--light` and `[dir="ltr"]` both to the same lavel i.e. body so neither of those assumed hierarchies are true. I can obviously change my implementation, but I guess that begs the question of whether you want that restriction.

Keeping the color (which is the same for all color stops) in the `skin.css` file also results in the css being duplicated four times for color stops.

GarthDB:
Closing for now, @Harbs we can look at it again if you find a workable solution.

PR #902, opened 2020-09-10
Spectrum tokens 17 card

### LINKED ISSUE #811: Check card against DL and VRT
Check [card](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/card)
- [ ] matches DL
- [ ] check VRT passes

jianliao:
## Description
Fixed #811 
1. Text line-height change caused by typography - expected
2. `More` icon size change - fixed
3. Card title font-color and font-weight change to match the DL

 - **How this was tested:** 

## Screenshots

Line-height change caused by the text in footer and card title font color/weight change.
image

DL:
image

## To-do list

### DIFF
--- a/components/card/package.json
+++ b/components/card/package.json
@@ -21,6 +21,7 @@
     "@spectrum-css/checkbox": "^3.0.0-beta.0",
     "@spectrum-css/icon": "^2.0.0",
     "@spectrum-css/quickaction": "^3.0.0-beta.0",
+    "@spectrum-css/typography": "^2.1.3",
     "@spectrum-css/vars": "^2.0.0"
   },
   "devDependencies": {
@@ -30,6 +31,7 @@
     "@spectrum-css/component-builder": "^1.2.0",
     "@spectrum-css/icon": "^2.1.1",
     "@spectrum-css/quickaction": "^3.0.0-beta.3",
+    "@spectrum-css/typography": "^2.1.3",
     "@spectrum-css/vars": "^2.3.0",
     "gulp": "^4.0.0"
   },
--- a/components/card/skin.css
+++ b/components/card/skin.css
@@ -30,10 +30,6 @@ governing permissions and limitations under the License.
     background-color: var(--spectrum-alias-highlight-selected);
   }
 
-  .spectrum-Card-title {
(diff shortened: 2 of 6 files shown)


GarthDB:
I chatted with @NateBaldwinDesign why to find out why we don't have more typography tokens for `spectrum-Card-title` and `spectrum-Card-subtitle` and it is so the products can have more control over which size title and subtitle to use with the `spectrum-Heading` and `spectrum-Detail` components.

components/card/package.json
```diff
@@ -30,6 +31,7 @@
     "@spectrum-css/component-builder": "^1.2.0",
     "@spectrum-css/icon": "^2.1.1",
     "@spectrum-css/quickaction": "^3.0.0-beta.3",
+    "@spectrum-css/typography": "^2.1.3",
```
lazd, line 34: Since this is in deps, we don't need it in devDeps too
```suggestion
```

components/card/skin.css
```diff
@@ -31,7 +31,7 @@ governing permissions and limitations under the License.
   }
 
   .spectrum-Card-title {
-    color: var(--spectrum-alias-text-color);
+    color: var(--spectrum-global-color-gray-900);
```
lazd, line 34: Any reason for this? I'd rather see an alias than a global, and it seems that maybe this was changed? I see

```
  --spectrum-alias-text-color: var(--spectrum-global-color-gray-800);
  --spectrum-alias-text-color-hover: var(--spectrum-global-color-gray-900);
```

lazd: Perhaps this should be `gray-800`?

GarthDB: Sorry, I should have removed that as well, one sec, and I'll update title and subtitle to just use the Heading and Detail components and delete those.

jianliao: @GarthDB I just removed it.

GarthDB: @jianliao beat me to it.

components/card/package.json
GarthDB: That's `peerDependencies`, not `dependencies`.

PR #903, opened 2020-09-10
Spectrum tokens 17 - Pagination

### LINKED ISSUE #859: Check pagination against DL and VRT
Check [pagination](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/pagination)
- [ ] matches DL
- [ ] check VRT passes

jianliao:
## Description

Fixed #859 
Change pagination counter text color to match with DL.

This issue is blocked by the solution of #913 

 - **How this was tested:** 

## Screenshots
DL:
image

## To-do list

### DIFF
--- a/components/pagination/skin.css
+++ b/components/pagination/skin.css
@@ -26,5 +26,5 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Pagination-counter {
-  color: var(--spectrum-global-color-gray-700);
+  color: var(--spectrum-alias-label-text-color);
 }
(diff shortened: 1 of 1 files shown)


lazd:
Hmm, looks like this is something that should be addressed with a solution to #913... However, this solution isn't the worst in the meantime, but I think I'd rather see it solved properly and left the wrong color in the meantime.

GarthDB:
I'm closing this because it actually appears fixed in a future release of DNA for some of the components and we can do it at the same time. I'll open an issue specifically for pagination.

PR #908, opened 2020-09-15
Spectrum tokens 17 Checkbox

### LINKED ISSUE #816: Check checkbox against DL and VRT
Check [checkbox](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/checkbox)
- [x] matches DL
- [x] check VRT passes
- [x] check and update migration guides [example](https://github.com/adobe/spectrum-css/blob/f58d8e071e5cce8db653e61c5a09b7690c236f63/components/pagination/metadata/pagination-explicit.yml#L3)

lazd:
## Description

Closes #816

Checkbox Emphasized vs Quiet was not implemented correctly. That has been fixed.

Checkbox has two global variables:

```
[08:33:16] @spectrum-css/checkbox directly uses global variable --spectrum-global-dimension-size-65
[08:33:16] @spectrum-css/checkbox directly uses global variable --spectrum-global-animation-duration-100
```

These are both used for properties that are currently unavailable in DNA.

A note was written on quiet vs emphasized.

 - **How this was tested:** 

## Screenshots

image
image

## To-do list

### DIFF
--- a/components/checkbox/metadata/checkbox.yml
+++ b/components/checkbox/metadata/checkbox.yml
@@ -3,9 +3,11 @@ status: Verified
 description: "Checkboxes allow users to select multiple items from a list of individual items, or mark one individual item as selected."
 SpectrumSiteSlug: https://spectrum.adobe.com/page/checkbox/
 sections:
-  - name: A note on emphasized and quiet
+  - name: Migration Guide
     description: |
       Spectrum has chosen the variant previously known as `quiet` to be the default and has added an `emphasized` variant with the same styles as the previous default.
+      If you were using the `quiet` variant, the `.spectrum-Checkbox--quiet` class is no longer required and can be removed.
+      If you need a checkbox to look like it did before this change, add `.spectrum-Checkbox--emphasized`.
 examples:
   - id: checkbox-default
     name: Standard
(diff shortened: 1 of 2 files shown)


jianliao (approved):

components/checkbox/metadata/checkbox.yml
```diff
@@ -2,6 +2,10 @@ name: Checkbox
 status: Verified
 description: "Checkboxes allow users to select multiple items from a list of individual items, or mark one individual item as selected."
 SpectrumSiteSlug: https://spectrum.adobe.com/page/checkbox/
+sections:
+  - name: A note on emphasized and quiet
```
GarthDB, line 6: Should this be a migration guide?

GarthDB (changes requested):

components/checkbox/metadata/checkbox.yml
lazd: Hmm, I don't know. I have just been writing "A note on emphasized and quiet, " but technically, they should _not_ change their checkboxes to emphasized where they were previously default, and their quiet will already look like the default ones looked before, so... Maybe it needs wording changes?

lazd: @GarthDB I went ahead and made the section a `Migration Guide` and added instructions. Makes sense!

Issue #829, opened 2020-08-27
Check progressbar against DL and VRT


GarthDB:
Check [progressbar](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/progressbar)
- [x] matches DL
- [ ] check VRT passes
- [x] check and update migration guides [example](https://github.com/adobe/spectrum-css/blob/f58d8e071e5cce8db653e61c5a09b7690c236f63/components/pagination/metadata/pagination-explicit.yml#L3)

GarthDB:
VRT not returning image for progress bar

bernhard-adobe:
I did not notice any differences between the latest release and the version in branch `spectrum-tokens-17`. The renaming of _Bar Loader_ into _Progress Bar_ is intended in `DNA-beta-6-17
`
![Screen Shot 2020-09-15 at 22 03 10](https://user-images.githubusercontent.com/52184321/93284820-7b617000-f7a1-11ea-9e27-9a3f59cbe4c9.png)
![Screen Shot 2020-09-15 at 22 04 07](https://user-images.githubusercontent.com/52184321/93284823-7dc3ca00-f7a1-11ea-8a07-cc1119fcf898.png)
![Screen Shot 2020-09-15 at 22 04 21](https://user-images.githubusercontent.com/52184321/93284827-7ef4f700-f7a1-11ea-8ee9-f160cafa5181.png)

Issue #850, opened 2020-08-27
Check slider against DL and VRT


GarthDB:
Check [slider](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/slider)
- [ ] matches DL
- [ ] check VRT passes
- [ ] check and update migration guides [example](https://github.com/adobe/spectrum-css/blob/f58d8e071e5cce8db653e61c5a09b7690c236f63/components/pagination/metadata/pagination-explicit.yml#L3)

GarthDB:
Some label sizing changes and this variant is broken in VRT
![image](https://user-images.githubusercontent.com/125516/92627507-b25af180-f288-11ea-9545-04a32c9bcc01.png)

bernhard-adobe:
also slider-variant color is missing from code. 
![Screen Shot 2020-09-16 at 17 34 47](https://user-images.githubusercontent.com/52184321/93394899-ef078980-f842-11ea-9e39-df5b0a00cd6d.png)

PR #923, opened 2020-09-17
fix: ramp slider (#829)

bernhard-adobe:
Adding back ramp-slider and removing color-slider.

## Description
Addresses issue #829 

- added back ramp-slider
- removed color-slider from slider page that is now it's own component under [/docs/colorslider.html](/docs/colorslider.html)

URL: http://localhost:3000/docs/slider.html
Chrome Version 85.0.4183.102 (Official Build) (64-bit) on MacOS

## Screenshots

Screen Shot 2020-09-22 at 13 16 22

## To-do list

### DIFF
--- a/components/slider/metadata/slider.yml
+++ b/components/slider/metadata/slider.yml
@@ -1,5 +1,11 @@
 name: Slider
 SpectrumSiteSlug: https://spectrum.adobe.com/page/slider/
+sections:
+  - name: Migration Guide
+    description: |
+      ### Color Slider moved  out of Sliders into it's own component
+      Color Slider moved to [Color slider](colorslider.html).
+      Replace class names starting with `.spectrum-Slider` with `.spectrum-ColorSlider`.
 examples:
   - id: slider
     name: Standard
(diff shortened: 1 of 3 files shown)


lazd (changes requested):
Commit message needs to include:

```
BREAKING CHANGE: Slider color variant removed, used Color Slider component
```

Also need migration guide.

bernhard-adobe:
Good points @lazd .
I added a migration guide. Unfortunately your suggested commit message is not working because of the commit rules with uppercases:
```
$ g commit -m "BREAKING CHANGE: Slider color variant removed, used Color Slider component"
husky > commit-msg (node v12.16.2)
⧗   input: BREAKING CHANGE: Slider color variant removed, used Color Slider component
✖   header must not be longer than 72 characters, current length is 74 [header-max-length]
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 3 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

```

I used instead `g commit -m "fix: breaking change: slider color variant moved to color slider. (#850)"`
I also added the breaking-change label.

GarthDB:
The disabled color seems to have changed for ramp slider. Is that a result of the tokens changing?

components/slider/metadata/slider.yml
```diff
@@ -1,5 +1,11 @@
 name: Slider
 SpectrumSiteSlug: https://spectrum.adobe.com/page/slider/
+sections:
+  - name: Migration Guide
+    description: |
+      ### Color Slider moved  out of Sliders into it's own component
```
lazd, line 6:
```suggestion
      ### Color slider is now a separate component
```

components/slider/metadata/slider.yml
```diff
@@ -1,5 +1,11 @@
...
+      Color Slider moved to [Color slider](colorslider.html).
+      Replace class names starting with `.spectrum-Slider` to `.spectrum-ColorSlider`.
```
lazd, line 8:
```suggestion
      Replace class names starting with `.spectrum-Slider` with `.spectrum-ColorSlider`.
```

lazd (approved):

PR #961, opened 2020-09-22
Spectrum tokens 17 - Textfield Textarea

### LINKED ISSUE #860: Check textfield against DL and VRT
Check [textfield](https://github.com/adobe/spectrum-css/tree/spectrum-tokens-17/components/textfield)
- [ ] matches DL
- [ ] check VRT passes

GarthDB:
## Description

- fix: wip
- fix: adjusted textfield icon placement with updated tokens. closes #860

 - **Browser(s) and OS(s) this was tested with:** Tested in latest Chrome

## To-do list


GarthDB:
Looks like I have an issue with the icon placement on the quiet variants. I'll fix that.

components/textfield/index.css
```diff
@@ -231,35 +231,28 @@ governing permissions and limitations under the License.
...
-        calc(var(--spectrum-icon-alert-medium-width) / 2)
+      calc(var(--spectrum-textfield-height) / 2) -
+        calc(var(--spectrum-textfield-validation-icon-height-error) / 2)
     );
 
-    padding-inline-end: calc(
-      var(--spectrum-textfield-padding-x) / 2 -
-        var(--spectrum-global-dimension-size-10)
-    );
+    inset-inline-end: var(--spectrum-textfield-icon-margin-left);
```
lazd, line 241: Why change from padding to inset?

components/textfield/metadata/textfield.yml
```diff
@@ -208,4 +208,4 @@ examples:
           <use xlink:href="#spectrum-css-icon-AlertMedium" />
         </svg>
         <input type="text" placeholder="Enter your name" name="field" value="Invalid input" class="spectrum-Textfield-input" pattern="[\d]+" required>
-      </div>
\ No newline at end of file
+      </div>
```
lazd, line 211: Yay for newlines at EOF ♥️

lazd (changes requested):
Not sure about the `padding` -> `inset` change, was that a mistake?

components/textfield/index.css
GarthDB: So it used padding and positioning to get it offset but the valid icon doesn't. This gets the icon in the same place, but just using tokens and not using global dimensions.

jianliao:
@GarthDB VRT result is [here](https://opensource.adobe.com/spectrum-css-vr-results/vrt-9f69a87-2020-09-22_17-50-20/html_report/)

But I don't know why spectrum-github-bot could not post the comment here.

GarthDB:
So the VRT shows a change in the quiet textfield/textarea, but the VRT is not in sync with the DL UI kit. From the UI kit:
image

While this is the VRT reference:
image

It shouldn't have the side padding. So this PR is ready to go.

bernhard-adobe:
This looks good to me, we had to do the same update in the SVG-Generator. I think it also fixes a bug where the green checkmark validation icon was gray and now is green. Maybe worth double checking.

bernhard-adobe (approved):

PR #1030, opened 2020-09-29
Action Button t-shirt sizing, separate component

lazd:
### Action Button changes
This PR adds t-shirt sizing to Action Button, which means that `.spectrum-ActionButton--sizeM` is now required to make ActionButtons size correctly. This is a breaking change.

This PR also splits out Action Button from `button` into its own component, `actionbutton`.

Additionally, documentation was reformatted and a migration guide was included.

The big mess of padding problems was also fixed, so things should match designs now.

### Legacy build system disabled
This PR also disables the legacy build system. ActionButton with t-shirt sizing does not build with legacy.

### Unsupported icon sizes removed

Finally, this PR removes legacy icon sizes that are unsupported in Spectrum.

 - **How this was tested:** scrolly scroll, VRT

## Screenshots

image

## To-do list


tools/bundle-builder/subrunner/index.js
```diff
@@ -91,6 +91,9 @@ function runTaskOnPackages(task, packages) {
 
       if (packageDir) {
         runComponentTask(packageDir, task, function(err) {
+          if (err) {
+            process.exit();
```
lazd, line 95: When running tasks on all components within the monorepo, exit if a failure occurs. `runComponentTask` and gulp itself will log the error already, so we don't need to print it again here.

tools/component-builder/css/lib/varUtils.js
```diff
@@ -38,6 +38,23 @@ function getVarsFromCSS(css) {
   return variableList;
 }
 
+function getVarsDefinedInCSS(css) {
```
lazd, line 41: This gets the list of all variables defined within the component's CSS; i.e. the `--spectrum-actionbutton-height` which is based on `--spectrum-actionbutton-m-height` for the medium size, etc. This makes it so we don't throw when the variable has been defined inside of the component, instead we warn.

tools/component-builder/css/vars.js
```diff
@@ -44,26 +44,39 @@ function bakeVars() {
...
-        logger.warn(`${pkg.name} directly uses global variable ${varName}`);
+        logger.warn(`⚠️  ${pkg.name} directly uses global variable ${varName}`);
       }
       else if (!allVars[varName]) {
-        logger.warn(`${pkg.name} uses undefined variable ${varName}`);
+        if (componentVars.indexOf(varName) === -1) {
+          errors.push(`${pkg.name} uses undefined variable ${varName}`);
+        }
+        else {
+          logger.warn(`🔶 ${pkg.name} uses locally defined variable ${varName}`);
```
lazd, line 68: This is new: previously, locally defined variables were issued the same warning (`uses undefined variable`). Now they get their own warning, and it's not a failure.

tools/component-builder/css/vars.js
```diff
@@ -44,26 +44,39 @@ function bakeVars() {
...
     let usedVars = {};
     variableList.forEach(varName => {
       if (varName.indexOf('spectrum-global') !== -1) {
-        logger.warn(`${pkg.name} directly uses global variable ${varName}`);
+        logger.warn(`⚠️  ${pkg.name} directly uses global variable ${varName}`);
       }
       else if (!allVars[varName]) {
-        logger.warn(`${pkg.name} uses undefined variable ${varName}`);
+        if (componentVars.indexOf(varName) === -1) {
+          errors.push(`${pkg.name} uses undefined variable ${varName}`);
```
lazd, line 65: Undefined variables result in a failure.

tools/component-builder/index.js
```diff
@@ -23,7 +23,7 @@ function clean() {
 const build = gulp.series(
   clean,
   gulp.parallel(
-    css.buildCSS,
+    css.buildVars,
```
lazd, line 26: The old, legacy build tasks are still in the code, but instead of calling them, we run the modern vars build tasks.

tools/bundle-builder/index.js
```diff
@@ -215,4 +215,4 @@ exports.dev = devTask;
 exports.clean = clean;
 exports.build = build;
 exports.watch = dev.watch;
-exports.default = build;
+exports.default = buildLite;
```
lazd, line 218: The default action changed from a full build to a "lite" build (CSS vars only).

components/icon/workflow-icons.css
```diff
@@ -15,28 +15,6 @@ governing permissions and limitations under the License.
...
-  }
-}
-
-.spectrum-Icon--sizeXS {
-  &,
-  img,
-  svg {
-    block-size: var(--spectrum-icon-size-xs);
-    inline-size: var(--spectrum-icon-size-xs);
-  }
```
lazd, line 39: We don't have to do this in this PR, but I figured why not drop the unsupported icon sizes? @jianliao thoughts?

components/actionbutton/index.css
```diff
@@ -0,0 +1,165 @@
...
+  --spectrum-actionbutton-text-padding-x: var(--spectrum-actionbutton-xl-text-padding-x);
+  --spectrum-actionbutton-text-size: var(--spectrum-actionbutton-xl-text-size);
+}
+
+a.spectrum-ActionButton {
+  @inherit: %spectrum-AnchorButton;
+}
+
+.spectrum-ActionButton {
+  @inherit: %spectrum-BaseButton;
```
lazd, line 101: Now that we've broken this code out into BaseButton, we will have duplicate code, i.e. it will be defined for `.spectrum-ActionButton` as well as `.spectrum-Button`. This will increase output code size when both `@spectrum-css/button` and `@spectrum-css/actionbutton` are included in the same project.

components/actionbutton/index.css
```diff
@@ -0,0 +1,214 @@
...
+@import "../commons/basebutton.css";
+
+/*
+small: 5x5px hold icon, 3px from the edge
+medium: 5x5px hold icon, 4px from the edge
+large: 6x6px hold icon, 5px from the edge
+extra-large: 7x7pm hold icon, 6px from the edge
+*/
+
+.spectrum-ActionButton--sizeM {
```
lazd, line 23: This simple change would avoid breaking changes by making medium variables default (and thus not requiring `spectrum-ActionButton--sizeM` to render a medium Action Button):

```suggestion
.spectrum-ActionButton,
.spectrum-ActionButton--sizeM {
```

jianliao:
[Cycle button](https://opensource.adobe.com/spectrum-css-vr-results/vrt-cfe2f6a-2020-10-05_19-04-58/dist/docs/cyclebutton.html) example looks wrong.

components/actionbutton/README.md
```diff
@@ -0,0 +1,6 @@
+# @spectrum-css/actionbuttonbutton
```
jianliao, line 1: Should be `@spectrum-css/actionbutton`

components/actionbutton/skin.css
```diff
@@ -0,0 +1,340 @@
...
+    color: var(--spectrum-actionbutton-m-hold-icon-color);
+  }
+
+  &:hover {
+    background-color: var(--spectrum-actionbutton-m-background-color-hover);
+    border-color: var(--spectrum-actionbutton-m-border-color-hover);
+    color: var(--spectrum-actionbutton-m-text-color-hover);
+
+    .spectrum-Icon {
+      color: var(--spectrum-actionbutton-m-icon-color-hover);
```
jianliao, line 37: Just curious what is the meaning of `-m-` here? Is it `medium`, why even color has a medium size?

components/icon/workflow-icons.css
jianliao: Ah, I dropped them in my original UI icon t-shirt sizes PR. But I saw @GarthDB added them back, so I added them back, too.

components/actionbutton/skin.css
lazd: Yeah, it's medium. All variables are re-defined for each t-shirt size, and `--spectrum-actionbutton-icon-color-hover` is no longer defined. We have to use `--spectrum-actionbutton-m-icon-color-hover`.

lazd:
> Here are some examples that looks wrong.
> 
> 1. [Cycle button](https://opensource.adobe.com/spectrum-css-vr-results/vrt-cfe2f6a-2020-10-05_19-04-58/dist/docs/cyclebutton.html)
> 2. [Action button in pagination explicit](https://opensource.adobe.com/spectrum-css-vr-results/vrt-cfe2f6a-2020-10-05_19-04-58/dist/docs/pagination-explicit.html)
> 3. [Action button in pagination listing](https://opensource.adobe.com/spectrum-css-vr-results/vrt-cfe2f6a-2020-10-05_19-04-58/dist/docs/pagination-listing.html)

These should be fixed with the latest commit

components/icon/workflow-icons.css
lazd: Hmm, @GarthDB should we kill them or leave them?

lazd:
@jianliao think we're good to merge here?

jianliao (approved):
LGTM

components/actionbutton/index.css
GarthDB: Not ideal, but gzipped shouldn't be much. Are you thinking we run a deduper?

lazd: @GarthDB, nah, can't really de-dupe this. I don't think it's a big problem, RSP was asking for this anyway.

lazd:
@jianliao turns out those Pagination examples needed a little more love. Fixed now!

components/icon/workflow-icons.css
GarthDB: Have these tokens not been killed in DNA yet? @jianliao do you know what version of DNA removes them?

jianliao: They are not included in 6.0.0-beta.20. But seems like DNA never provides them, we calculated them out based on medium and small.

lazd: Right, so they've never been in DNA, but we've always provided those non-standard icon sizes as a holdover from Coral. We should kill them, but lemme know if you think a different PR is a better place for it @GarthDB.

PR #1034, opened 2020-10-05
ColorLoupe: add checkerboard background

mischnic:
## Description
Only the colorhandle had a checkerboard background (visible when alpha != 1), but not colorloupe.

 - **How this was tested:** I've updated the examples on the colorhandle and colorloupe docs
 - **Browser(s) and OS(s) this was tested with:** Chrome 85.0.4183.121, macOS 10.15

## Screenshots

Before (colorloupe displaying red with alpha=0.5):
Bildschirmfoto 2020-10-05 um 12 13 33

After (same color):
Bildschirmfoto 2020-10-05 um 12 13 57

## To-do list

### DIFF
--- a/components/colorhandle/metadata/colorhandle.yml
+++ b/components/colorhandle/metadata/colorhandle.yml
@@ -28,6 +28,12 @@ examples:
         <div class="spectrum-ColorHandle-color" style="background-color: rgba(255, 0, 0, 0.5)"></div>
         <svg class="spectrum-ColorLoupe is-open">
           <g transform="translate(1 1)">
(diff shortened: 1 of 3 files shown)


Westbrook:
It looks like `colorhandle` is managing this in CSS only: https://github.com/adobe/spectrum-css/blob/d24c17e06b7b23893bf9467d161894275da6dc06/components/colorhandle/index.css#L36-L43

Is similar possible here?

lazd:
@Westbrook yeah that's exactly what we want to do here, despite the fact that previous prototypes used the SVG approach, the CSS approach is lighter.

GarthDB:
This is fixed with @lazd's latest work on Color Loupe.

PR #1044, opened 2020-10-07
T-shirt sizing tooling

lazd:
## Description

This PR adds tools for t-shirt sizing that makes it simple to bolt it on to any existing component.

Action Button has been retrofitted with the new tooling to highlight the differences

## How does it work?

1. Write out each t-shirt size, call `@remapvars {}` with `find` and `replace`, then simply use the size-agnostic variables. Most components already use size-agnostic variable names, so its possible no additional code will have to be modified:

```css
.spectrum-ActionButton--sizeS {
  @remapvars {
    find: --spectrum-actionbutton-s-;
    replace: --spectrum-actionbutton-;
  }
}

.spectrum-ActionButton--sizeM {
  @remapvars {
    find: --spectrum-actionbutton-m-;
    replace: --spectrum-actionbutton-;
  }
}

/* ... */

.spectrum-ActionButton {
  block-size: var(--spectrum-actionbutton-height);
  /* ... */
}
```

2. During processing, the `postcss-remapvars` plugin takes all every possible variable for each t-shirt size and remaps them to a variable that does not have a size prefix:

```css
.spectrum-ActionButton--sizeS {
  --spectrum-actionbutton-height: var(--spectrum-actionbutton-s-height);
  /* ... */
}

.spectrum-ActionButton--sizeM {
  --spectrum-actionbutton-height: var(--spectrum-actionbutton-m-height);
  /* ... */
}

/* ... */
```

3. Next, the `postcss-dropunusedvars` plugin checks the file to see which variables are used, then drops everything we didn't use so we're not publishing thousands of variables we don't use. 

4. Finally, the `postcss-dropdupedvars` will remove variables we have overridden so we don't publish unnecessarily re-defined variables.

5. The end result is a set of classes that modify only the necessary variables to change the t-shirt sizes of the component.

## To-do list

- [x] Write documentation for postcss plugins
- [x] Discuss whether plugins should be in a separate repo

### DIFF
--- a/components/actionbutton/index.css
+++ b/components/actionbutton/index.css
@@ -12,8 +12,14 @@ governing permissions and limitations under the License.
 
 @import "../commons/index.css";
 @import "../commons/basebutton.css";
+@import "../vars/css/components/spectrum-actionbutton.css";
 
 .spectrum-ActionButton--sizeM {
+  @remapvars {
+    find: -spectrum-actionbutton-m-;
+    replace: -spectrum-actionbutton-;
+  }
+
   /* Set up default variables, size medium */
   --spectrum-actionbutton-padding-left: calc(var(--spectrum-actionbutton-m-padding-left) - var(--spectrum-actionbutton-m-border-size));
   --spectrum-actionbutton-padding-right: calc(var(--spectrum-actionbutton-m-padding-right) - var(--spectrum-actionbutton-m-border-size));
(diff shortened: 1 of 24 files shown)


lazd:
@misterbrownlee @GarthDB @jianliao @bernhard-adobe I think these PostCSS plugins should live in their own, separate repos. For the purpose of reviewing this pull request, I'd like to keep them in the monorepo. However, after this PR is approved, I would like to move them out to separate repositories. Does that sound like a plan?

misterbrownlee:
I'm good with moving the plug-in out to their own open source release.

jianliao (approved):

Issue #1051, opened 2020-10-13
Calendar: is-focused style for today's date is insufficient to communicate the focused state


majornista:
## Description
See https://opensource.adobe.com/spectrum-css/calendar.html#focused

The grey background color for the .is-focused style on today's date is insufficient to communicate that the date has focus. We should use either the blue focus ring as we do for other focused styles, or to continue communicating that the date is today, some sort of double border.

## Steps to reproduce

1. https://opensource.adobe.com/spectrum-css/calendar.html#focused

2. The grey background color for the .is-focused style on today's date is insufficient to communicate that the date has focus. We should use either the blue focus ring as we do for other focused styles, or to continue communicating that the date is today, some sort of double border.

## Expected behavior
Keyboard focus should be clearly indicated when focus is on today's date.

## Environment
 - **Spectrum CSS version:** 
 - **Browser(s) and OS(s):**

PR #1052, opened 2020-10-14
fix(calendar): .is-today.is-focused should have blue focus ring #1051

majornista:
## Description

See https://opensource.adobe.com/spectrum-css/calendar.html#focused

The grey background color for the .is-focused style on today's date is insufficient to communicate that the date has focus. We should use either the blue focus ring as we do for other focused styles, or to continue communicating that the date is today, some sort of double border.

#1051

 - **How this was tested:** Verified in documentation example.
 - **Browser(s) and OS(s) this was tested with:** Chrome Version 85.0.4183.121 (Official Build) (64-bit) on macOS 

### Before (with black focus ring, making it impossible to perceive Today as the focused date)
Calendar - Before, with black focus ring on Today

### After (with blue focus ring, making it possible to perceive Today as the focused date, but Today's is only distinguished by the bold text when the date has focus)
Calendar - After, with blue focus ring on Today

## To-do list

### DIFF
--- a/components/calendar/skin.css
+++ b/components/calendar/skin.css
@@ -116,6 +116,10 @@ governing permissions and limitations under the License.
         background: var(--spectrum-calendar-day-background-color-selected-hover);
       }
     }
+
+    &:before {
+      border-color: var(--spectrum-calendar-day-border-color-key-focus);
+    }
   }
 
   &.is-disabled {


majornista:
@GarthDB or @lazd  Can I get a review?

lazd:
@majornista I've roped in Dr. Downtown Lindsay Browne @labrowne to verify the design, but I think you may have nailed it based on this screenshot @NateBaldwinDesign posted in Slack:

image

lazd:
Thanks @majornista, this is a fine fix!

PR #1056, opened 2020-10-15
fix: #1045 chevron icon in picker got squeezed

### LINKED ISSUE #1045: ChevronDown UI Icon in Picker component is vertically off.
image

jianliao:
## Description
Since the new chevron ui icon increased its size to 10x10. It got squeezed in some scenarios.

Fixed #1045 

 - **How this was tested:** 

## Screenshots

Before:
image

After:
localhost_3000_docs_picker html (1)

## To-do list

### DIFF
--- a/components/picker/index.css
+++ b/components/picker/index.css
@@ -108,6 +108,7 @@ governing permissions and limitations under the License.
   position: relative;
   vertical-align: top;
   transition: color var(--spectrum-global-animation-duration-100) ease-out;
+  flex-shrink: 0;
 
   /* Fix Safari 10 bug where align-items is ignored inside of buttons */
   margin-block-start: calc(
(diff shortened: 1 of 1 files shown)


lazd:
Interesting, so it has 2px of extra whitespace or something? 

What about the fact that the class and the ID don't match?


```
          <svg class="spectrum-Icon spectrum-UIIcon-ChevronDown100 spectrum-Picker-icon" focusable="false" aria-hidden="true">
            <use xlink:href="#spectrum-css-icon-Chevron100" />
          </svg>
```

We have `ChevronDown100` and `Chevron100`, what's up with that?

jianliao:
If I understand correctly, we have icon css class `spectrum-UIIcon-ChevronDown100`. But we don't have an SVG icon id named `spectrum-UIIcon-ChevronDown100`.

jianliao:
The chevron icon size changed from 10x6 to 10x10.

lazd:
@jianliao then let's rename the class to match, classes and icon names should be identical, I don't know if they've ever been different before.

PR #1063, opened 2020-10-23
feat: upgrade to @adobe/spectrum-tokens@6.0.0-beta.28

lazd:
## Description

✨  feat: update to Spectrum Tokens 6.0.0-beta.28, fix broken components

✨  feat: implement t-shirt sizing for Link, closes #1066

✨  feat: implement t-shirt sizing for Divider, closes #972

✨  feat: implement t-shirt sizing for Status Light, closes #686

✨  feat: implement t-shirt sizing for Field Label, closes #943 

✨  feat: implement t-shirt sizing for ProgressBar, update notice classes, closes #955 

## How to work with this, it's broken?

Use `FORCE=true gulp dev` to develop with broken components.

## To-do list

- [x] Fix all components that fail to build
  - [x] Button
  - [x] Textfield
  - [x] Tags
  - [x] Menu
  - [x] Picker
  - [x] Popover
  - [x] QuickAction
  - [x] Checkbox
  - [x] Radio
  - [x] Switch
  - [x] SearchWithin
  - [x] Search
  - [x] ProgressBar
  - [x] AssetList
  - [x] InputGroup
  - [x] Stepper
  - [x] FieldLabel
  - [x] Tabs
  - [x] Sidenav
  - [x] Dial
  - [x] Slider
- [x] Check VRT result and fix anything that's broken
  - [x] Checkbox line-height is too high, it's been changed to component line height now
  - [x] Table padding - it was wrong before! we're good now
  - [x] ActionGroup quiet - action button quiet was never supposed to have border when selected
  - [x] Action button text size - test is correct
  - [x] CTA/Primary/Over Background font weight - should be bold, ~overridden temporarily waiting for DNA fix~, fixed in DNA
  - [x] Quiet Textfield left/right padding - should be zero, ~overridden temporarily waiting for DNA fix~, fixed in DNA
  - [x] Menu left padding - too big due to CSS bug
  - [x] Height of small Button - too big due to CSS bug, needs to be 24px
  - [x] Dialog line height - Asked @NateBaldwinDesign, it is intended to use component line heights, so test is correct and reference is wrong
  - [x] ButtonGroup - seems like left/right padding is different on buttons, changed in DNA
  - [x] Textfield (Quiet) - vertical alignment of text is too high?
  - [x] Textarea min-height and top padding - height is smaller, and missing padding vars in DNA were taken from Textfield, no longer matches specs
  - [x] Split Button - Width and icon position changed?
  - [x] Slider - track color changed?
[...]

Issue #753, opened 2020-06-13
icon should allow resizing with a css variable


adixon-adobe:
## Description
By default icon should have it's width/height using an overridable css variable. I think ideally the size would default to the equivalent of the small size now. So 18x18 in medium scale and 22x22 in large.

## Why do you need this feature or component?
Using a variable will allow overriding for t-shirt size components that use icons. E.g. a "large" action button can adjust the expected size of icons in the control using the variable, rather than relying on clients to provide the correct sized icon.

## Additional context
This is related to #726 and https://github.com/adobe/spectrum-web-components/issues/669

pfulton:
This should be addressed in our core tokens migration.

PR #1070, opened 2020-11-09
feat: further simplify icon markup in t-shirt size component

jianliao:
### The targets that I want to achieve include:
1. #753 - icon should allow resizing with a css variable.
2. Further simplify icon markup in T-shirt size component requested by this [comment](https://github.com/adobe/spectrum-css/pull/1063#issuecomment-723197689)

### Implementation detail:

1. I followed the [article](https://css-tricks.com/using-custom-property-stacks-to-tame-the-cascade/) mentioned by @Westbrook to build a css variable cascading(fallback) chain. It looks like:

```
block-size: var(--spectrum-icon-user-defined-height, var(--spectrum-icon-tshirt-size-height, var(--spectrum-icon-default-height, var(--spectrum-alias-workflow-icon-size-m)))) 

inline-size: var(--spectrum-icon-user-defined-width, var(--spectrum-icon-tshirt-size-width, var(--spectrum-icon-default-width, var(--spectrum-alias-workflow-icon-size-m))))
```

- `--spectrum-icon-user-defined-height` is for component team(spectrum-react, spectrum-webcomponents) to customize the height.
- `--spectrum-icon-tshirt-size-height` will be provided by each t-shirt size component in its size class like `.spectrum-ActionButton--sizeL`. See [here](https://github.com/adobe/spectrum-css/pull/1070/files#r520738656).
- `--spectrum-icon-default-height`. It is the default value when no user-defined and t-shirt size variable provided. For workflow icon height, it is default as DNA variable `--spectrum-alias-workflow-icon-size-m`.

2. To remove the class like `spectrum-Icon--sizeS` and `spectrum-UIIcon-CornerTriangle100` from workflow icon and ui icon svg element, we need to provide the t-shirt size variable in its wrapper element CSS class like `.spectrum-ActionButton--sizeM`. See [here](https://github.com/adobe/spectrum-css/pull/1070/files#r520738656). The html markup for an ActionButton would look like:

```
<button class="spectrum-ActionButton spectrum-ActionButton--sizeM">
  <svg class="spectrum-UIIcon spectrum-ActionButton-hold" focusable="false" aria-hidden="true">
    <use xlink:href="#spectrum-css-icon-CornerTriangle100" />
  </svg>
  <svg class="spectrum-Icon" focusable="false" aria-hidden="true" aria-label="Edit">
    <use xlink:href="#spectrum-icon-18-Edit" />
  </svg>
</button>
``` 

### Open issues
1. To distinguish workflow icon and ui icon, I need to use `spectrum-UIIcon` for ui icon. This is not a new class, but I could not find anywhere we use it. It would be a breaking change.

2. I defined a series of CSS variables that are not from DNA. Not sure where to put them, I just insert them into `components/vars/css/components/spectrum-icon.css`. It is not ideal as this file will be overridden each time we upgrade the DNA.

3. In each component css, we have to define t-shirt size variable unused anywhere in that file. I have to comment out `postcss-dropunusedvars` for this purpose. Need to enhance this postcss plugin to provide a white list maybe?

4. All the new variables name are TBD. For example:
```
  --spectrum-icon-user-defined-height: var(--spectrum-icon-tshirt-size-height);
  --spectrum-icon-user-defined-width: var(--spectrum-icon-tshirt-size-width);
  --spectrum-icon-tshirt-size-height: var(--spectrum-icon-default-height);
  --spectrum-icon-tshirt-size-width: var(--spectrum-icon-default-width);
  --spectrum-icon-default-height: var(--spectrum-alias-workflow-icon-size-m);
  --spectrum-icon-default-width: var(--spectrum-alias-workflow-icon-size-m);

  --spectrum-ui-icon-user-defined-height: var(--spectrum-ui-icon-tshirt-size-height);
  --spectrum-ui-icon-user-defined-width: var(--spectrum-ui-icon-tshirt-size-width);
  --spectrum-ui-icon-tshirt-size-height: var(--spectrum-ui-icon-default-height);
  --spectrum-ui-icon-tshirt-size-width: var(--spectrum-ui-icon-default-width);
  --spectrum-ui-icon-default-height: var(--spectrum-alias-workflow-icon-size-m);
  --spectrum-ui-icon-default-width: var(--spectrum-alias-workflow-icon-size-m);
```

 - **How this was tested:** 

## To-do list


components/actionbutton/index.css
```diff
@@ -45,6 +55,11 @@ governing permissions and limitations under the License.
   /* todo: remove hardcoded values when we update DNA */
   --spectrum-actionbutton-hold-icon-padding-bottom: var(--spectrum-global-dimension-size-50);
   --spectrum-actionbutton-hold-icon-padding-right: var(--spectrum-global-dimension-size-50);
+
+  --spectrum-icon-tshirt-size-height: var(--spectrum-alias-workflow-icon-size-l);
+  --spectrum-icon-tshirt-size-width: var(--spectrum-alias-workflow-icon-size-l);
+  --spectrum-ui-icon-tshirt-size-height: var(--spectrum-alias-ui-icon-cornertriangle-size-200);
+  --spectrum-ui-icon-tshirt-size-width: var(--spectrum-alias-ui-icon-cornertriangle-size-200);
```
jianliao, line 62: Defined the actual t-shirt size here.

components/vars/css/components/spectrum-icon.css
```diff
@@ -42,4 +42,18 @@
...
+  --spectrum-icon-tshirt-size-width: var(--spectrum-icon-default-width);
+  --spectrum-icon-default-height: var(--spectrum-alias-workflow-icon-size-m);
+  --spectrum-icon-default-width: var(--spectrum-alias-workflow-icon-size-m);
+
+  --spectrum-ui-icon-user-defined-height: var(--spectrum-ui-icon-tshirt-size-height);
+  --spectrum-ui-icon-user-defined-width: var(--spectrum-ui-icon-tshirt-size-width);
+  --spectrum-ui-icon-tshirt-size-height: var(--spectrum-ui-icon-default-height);
+  --spectrum-ui-icon-tshirt-size-width: var(--spectrum-ui-icon-default-width);
+  --spectrum-ui-icon-default-height: var(--spectrum-alias-workflow-icon-size-m);
+  --spectrum-ui-icon-default-width: var(--spectrum-alias-workflow-icon-size-m);
```
lazd, line 58: Let's move these into the `components/icon/index.css` file so they don't get blown away when we update DNA.

jianliao: I tried moving them to either `components/icon/index.css` or `components/icon/icons.css`. But none of them work, these variable never show up in `dist/index-vars.css`.

lazd: Hmm interesting. We should figure another way of shoe-horning these in, then, as this makes DNA updates painful. Any ideas?

jianliao: @lazd,
Looks like these spectrum-css private custom properties could be placed in `components/vars` and should be part of `spectrum-global.css`. Can I create folder named -`components/vars/private` and change the gulp script to concat these properties to `spectrum-global.css`?

lazd: That sounds like a great idea, though I'm not sure about `private`... Maybe `local` or `custom` or something?

jianliao:
@lazd, could you please review it again? So here is change:

1. Based on the what I learned from CSS variable spec, the dynamic cascade variables chain could not be defined on `.spectrum` scope caused it is like root of all elements. Once a variable got evaluated, it could not be changed even in an inner/specific scope. We could only leave static global variables there.

2. I moved the cascade variable chain to `components/vars/dynamic/*.css`, variables defined there are all only could be figured out at runtime.
```
  --spectrum-icon-user-defined-height: var(--spectrum-icon-tshirt-size-height);
  --spectrum-icon-user-defined-width: var(--spectrum-icon-tshirt-size-width);
  --spectrum-icon-tshirt-size-height: var(--spectrum-alias-workflow-icon-size-m);
  --spectrum-icon-tshirt-size-width: var(--spectrum-alias-workflow-icon-size-m);
```
--spectrum-icon-user-defined-xxxx is for component team to defined. --spectrum-icon-tshirt-size-xxxx will be defined by component's t-shirt size specific class like `.spectrum-ActionButton--sizeM`.

3. Build script changed for cheating the `bakeVar` compile function to believe that above tokens in #2 are existing.

Westbrook:
Will these icon specific size classes be required after this change? https://github.com/adobe/spectrum-css/blob/tshirt-size-workflow-icon-continue/components/icon/workflow-icons.css#L21-L65

jianliao:
No more  specific size classes required.

Westbrook:
I think I misstated my question...

Good to hear that is wouldn't be required!

Would Spectrum CSS still maintain those classes for optional use, going forward?

lazd:
@Westbrook yes absolutely; we do want all of the standard icon sizes to be usable. This PR simply changes their sizes contextually where we know what size the icons need to be.

We still need to really chew on this and understand if we're being consistent if we go this route; take a look at ProgressBar, it has t-shirt sizing and uses FieldLabel, which also has t-shirt sizing. You must specify the size of the FieldLabel used within the ProgressBar manually. It would be overreach for ProgressBar to change FieldLabel's variables for font size, weight, margin, etc contextually.

Given the ProgressBar and FieldLabel issue, does it make sense for Button to reach into Icon and do the same thing? Yes, it's only dimensions, but where do we draw the line?

Westbrook:
I'd less call this "reaching into" a component than defining what _scope_ a component is delivered in, which would lead me to think that a ProgressBar effecting a FieldLabel internal to it would be little different than a Button effecting the Icon within it.

In either case, would delivering the internal component at a different size than the external element be considered "Spectrum correct"? If not, I'd hope that the cascade managed this without need of specific classes, this goes the extra mile of reducing the likelihood of breaking changes like we've seen in the move from `sizeS` to `sizeM` as far as Icons in Buttons, recently. In the end, I'm predominately looking for ways to better prevent these sorts of changes making their way down stream to my consumers. Then, the use of classes to make a purposeful departure from the design spec would still be possible as a specific opt-in, and would make sense why the classes I referenced above would continue to exist.

lazd:
> I'd less call this "reaching into" a component than defining what _scope_ a component is delivered in, which would lead me to think that a ProgressBar effecting a FieldLabel internal to it would be little different than a Button effecting the Icon within it.

I can definitely see that, but where do we draw the line? Only icons can be contextually sized? I think that's fair, honestly, but need to hear from the rest of the team.

> In either case, would delivering the internal component at a different size than the external element be considered "Spectrum correct"? 


Almost certainly not, which is definitely a good argument for this approach.

> If not, I'd hope that the cascade managed this without need of specific classes, this goes the extra mile of reducing the likelihood of breaking changes like we've seen in the move from `sizeS` to `sizeM` as far as Icons in Buttons, recently.

My point exactly on the other PR comment; token-driven icon sizing means that consumers aren't on the hook for updating markup when a design decision is made. That said, the `sizeS` to `sizeM` change was a LOOONNGGG time coming, we had legacy code from 3 years ago in there for icon sizing, and we finally unified it. I absolutely don't see that happening again any time soon, and if it did, it would be an update to the size of a medium icon, not a change in the scale (i.e. small is renamed to medium).

> In the end, I'm predominately looking for ways to better prevent these sorts of changes making their way down stream to my consumers. Then, the use of classes to make a purposeful departure from the design spec would still be possible as a specific opt-in, and would make sense why the classes I referenced above would continue to exist.

Agreed. I think we're on the right track, but let's be sure to think about the system as a whole when we make changes that introduce a completely new API, as this one does (variable driven icon sizing is API!).

lazd:
@jianliao please rebase 🙏

components/icon/icons.css
```diff
@@ -28,3 +28,21 @@ governing permissions and limitations under the License.
   /* Don't catch clicks or hover, otherwise they may not escape the SVG */
   pointer-events: none;
 }
+
+.spectrum-Icon {
+  &,
+  img,
```
Westbrook, line 34: FYI: Nesting the size of `img` and `svg` in the same rules as `.spectrum-Icon` will likely cause issue in code like: https://github.com/adobe/spectrum-css/blob/731aeab0a10355dc274efd19d64a58fe71b2ef87/components/tooltip/index.css#L137-L149

Maybe the goal is to ALWAYS use CSS Custom Properties in this sort of context? In that case we'd need to have any such instance updated as part of this PR or prior to an actual release of these changes.

An alternative would be to not directly size the `img` and `svg` content and set them as:
```
    block-size: 100%;
    inline-size: 100%;
```

Thoughts?

castastrophe:
@pfulton As we're reworking the tooling elsewhere and seem to have a new approach to t-shirt size variables, should this PR be closed? Is there anything else in here we want to maintain?

PR #1097, opened 2021-01-08
Vrt exclude animation scenario

jianliao:
 - **How this was tested:** 

## To-do list


lazd:
Wow so the same error happens even without changes! Exciting. Thanks for looking into this!

PR #1110, opened 2021-01-26
fix: link shouldn't remap color variables, fixes #1108

### LINKED ISSUE #1108: [link] Pattern color themeing scoped to element.
## Description
As per https://unpkg.com/browse/@spectrum-css/link@3.1.0-beta.0/dist/index-vars.css the "t-shirt" sizing approach to "remap" all `-m-` vars to the `.spectrum-Link` selector causes color themeing to be scoped directly on the selector along with the t-shirt sizing scoping. This situation means that color customization can no longer be passed to these elements from outside of shadow root. Can we move the color theming off of the selector root so that this (and likely a number of other elements taking on t-shirt sizing) can be customized appropriately?

## Steps to reproduce
Compare https://webcomponents.dev/edit/cIYQ1Cgeo77nfSPgWlvu?file=src%2Findex.stories.js where the lint test has been customized as `red` via the deprecated CSS Custom Property token `--spectrum-link-text-color` to https://webcomponents.dev/edit/YW7jcmS06Fi1BLOqLuaE?file=src%2Findex.stories.js where the current CSS Custom Property Token `--spectrum-link-primary-text-color` is defined directly on the element so that it cannot be given a new value from the outside.

## Expected behavior
A CSS Custom Property based themeing system should allow for shadow root piercing customization of its values.

## Screenshots
Yay!
image

No...
image

CC: @simshanith

lazd:
## Description

fixes #1108

 - **How this was tested:** scrollyscroll.exe
 - **Browser(s) and OS(s) this was tested with:** a computer

### DIFF
--- a/components/link/index.css
+++ b/components/link/index.css
@@ -12,13 +12,6 @@ governing permissions and limitations under the License.
 
 @import "../vars/css/components/spectrum-link.css";
 
-.spectrum-Link {
-  @remapvars {
-    find: /--spectrum-link-(.*?)-m-/;
-    replace: --spectrum-link-$1-;
-  }
-}
-
 .spectrum-Link--sizeS {
   @remapvars {
     find: /--spectrum-link-(.*?)-s-/;
--- a/components/link/skin.css
+++ b/components/link/skin.css
@@ -11,18 +11,18 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-Link {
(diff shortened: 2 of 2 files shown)


lazd:
This PR fixes the issue and don't cause VRT to fail, it should be an easy merge. @GarthDB @jianliao

Westbrook (approved):
L
G
T
M

PR #1132, opened 2021-02-10
feat: align tags with more recent DNA tokens

GarthDB:
Added T-shirt sizes to tag and realigned it to the most recent design and DNA values.

Not ready to merge until there is a design review first.

## To-do list


Westbrook:
With the move away form `tagS` to `taggroup` is there a logic future that `tabS` will also make the move to `tabgroup` at some point? That would absolve the last plural pattern name.

PR #1150, opened 2021-03-03
feat: drop even more unused vars

lazd:
## Description

This PR makes `postcss-dropunusedvars` even more aggressive, removing variables if everything that references them has been removed.

 - **How this was tested:** tests!
 - **Browser(s) and OS(s) this was tested with:** n/a/


lazd:
This should be good to go, it builds just fine with no missing variables -- it broke at first before I pushed https://github.com/adobe/spectrum-css/pull/1150/commits/a04051fcd3b7d515dc5295358ca0d9a2b10ca820

Ready for review!

PR #1182, opened 2021-05-07
fix(textfield): prevent pointer-events on icons over the form element

### LINKED ISSUE #1181: [Textfield] Icons should be "pointer-events: none"
## Description
When using a Textfield, I'd like users who accidentally click on the Icon element that is available when valid, invalid, a Search Field, etc to be able to focus the field the same way they would when they clicked a small number of pixels up/down/left/right of where they actually clicked. Can we add `pointer-events: none;` to those icons?

## Why do you need this feature or component?
Clarify usage of the field.

## Additional context
Currently if you visit https://opensource.adobe.com/spectrum-css/textfield.html#valid and click exactly on the Checkmark you get the "active" styling while the pointer is down, but nothing will have focus when you complete the click.

Westbrook:
## Description
fixes #1181

Addes `pointer-events: none` to non-interactive icons used over the form element in a Textfield.

## To-do list

### DIFF
--- a/components/textfield/skin.css
+++ b/components/textfield/skin.css
@@ -74,6 +74,11 @@ governing permissions and limitations under the License.
   color: var(--spectrum-textfield-m-icon-color);
 }
 
+.spectrum-Textfield-icon,
+.spectrum-Textfield-validationIcon {
+  pointer-events: none;
+}
+
 .spectrum-Textfield-input {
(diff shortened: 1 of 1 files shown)


GarthDB:
Already incorporated with https://github.com/adobe/spectrum-css/commit/f90026541b00a6ca3d2b30228bb475f03bb91587

PR #1193, opened 2021-05-19
fix: adjusted button label padding

GarthDB:
## Description
Using top and bottom padding around `label` in `button` instead of the button itself.

## To-do list


lazd:
Yay for removing some hacks!

PR #1208, opened 2021-06-11
fix: use opacity: 0 and appearance: none of form elements

### LINKED ISSUE #1195: [Checkbox][Swtich][Radio] Native form elements show in Firefox for Arch Linux
## Description
Not sure where this falls in our support matrix, but it seems that the native form element underlying the Checkbox, Radio, and Switch patters are poking out on Firefox for Gnome 40.1.0 (on Arch Linux), see: https://github.com/adobe/spectrum-web-components/issues/1460

It seems that were we to use `opacity: 0` as opposed to `opacity: 0.0001` or `apperance: none` that this would not occur. Is there a specific reasoning behind our choice of this approach? Would it be acceptable to submit a PR with one of these changes underlying?

## Expected behavior
Native form elements shouldn't pop out like this.

## Screenshots
Skärmbild från 2021-05-18 14-46-50

Skärmbild från 2021-05-18 14-47-12

## Environment
 - **Browser(s) and OS(s):** Firefox for Gnome 40.1.0 (on Arch Linux)

Westbrook:
## Description
Use `opacity: 0;` and `appearance: none` to ensure that form elements don't force their visibility in contexts like Arch Linux.\

fixes #1195

## To-do list


components/colorarea/index.css
```diff
@@ -1,5 +1,6 @@
 %spectrum-ColorControl-hiddenField {
-  opacity: 0.0001;
+  opacity: 0;
+  apperance: none;
```
GarthDB, line 3: Isn't `appearance:none` unnecessary as part of a hidden component?

components/colorarea/index.css
```diff
@@ -1,5 +1,6 @@
 %spectrum-ColorControl-hiddenField {
-  opacity: 0.0001;
+  opacity: 0;
```
GarthDB, line 2: I'm not sure why we had `0.0001`. It shouldn't be needed for accessibility, but it's possible that was the intention. Let me git blame and see who to check with on this.

PR #1259, opened 2021-08-31
fix: mini card broken in docs site

### LINKED ISSUE #1240: Mini card broken in docs site
## Description
I updated card in `DNA-7.0-beta` and it broke the docs site on the branch.

image

pfulton:
This fixes an issue where the Card component's width was not properly expanding to accommodate its content.

Fixes #1240

 - Tested locally in the Card Component pages themselves, but also on other non-Card specific pages.
 - Tested in the latest versions of Firefox, Chrome, Safari, and Edge on MacOS.

#### Before:
Screen Shot 2021-08-31 at 4 30 48 PM

#### After:
Screen Shot 2021-08-31 at 4 31 10 PM

## To-do list

- [X] If my change impacts other components, I have tested to make sure they don't break.


pfulton:
I'd love a second set of eyes on this before it gets merged. I visually checked the instances of the Card component that I could find, but I'm not as familiar with the codebase as other folks are.

Westbrook:
@GarthDB was this originally applied to appease some of the VRT failures we were looking at in SWC? If so, is the "more correct" path here to follow the comments in the code and set `width` at the usage point rather than in the actual component?

GarthDB:
@Westbrook yes, but this card isn't super official. Still the width probably would be determined by the container rather than having a width set on the card itself. @pfulton Can you remove the width from the component and apply it in just the documentation website?

Westbrook:
I'm OK with changing the use pattern if we need. If we expect users of the pattern to set a width (or have it applied by content) we can manage that on our end just the same. I had brought this up previously because it was a change from the pre-DNA@7 versions.

In the case that we aren't setting a width, would `inline-size: auto` be needed at all?

GarthDB:
@Westbrook we removed the `auto` but also fixed gallery card which had a fixed width for some reason.

PR #1267, opened 2021-09-16
fix: use correct custom properties for table t-shirt sizing

pfulton:
## Description
This fixes a slew of incorrect custom properties which changed due to the addition of t-shirt sizes.

Additionally, this adds:
* New examples for `compact` and `spacious`
* A new example for `small` 

- Manually visually compared against the existing production site.
- Visual regression tests are a bit unhelpful because the HTML examples did not have the correct IDs when the baseline snapshots were generated, and the addition of new examples throws off all of the snapshots by one test.
- This change also impacts the ActionBar's example, so I've manually checked my changes there, as well. 

## To-do list

- [X] If my change impacts other components, I have tested to make sure they don't break.


GarthDB:
@pfulton
you are the best

PR #1268, opened 2021-09-20
feat: show vars switch when running in dev environments

### LINKED ISSUE #1263: Add dev env for express switch
null

synced to jira: https://jira.corp.adobe.com/browse/SDS-10340

pfulton:
This changes the `dev` task by setting a `development` NODE_ENV variable at run time.

The `siteComponent` Pug template has been updated to conditionally display the vars version switch, depending on what value is read from the NODE_ENV.

--

Currently, we'd only like to show the "Vars Version" switch when running the site locally. We can do this by restricting that drop down to only run when a specified environment variable is set. We're not currently asking folks to use `.env` files for environment variable management, so I've gone the route of setting the `NODE_ENV` to `development` when the `yarn start` task is invoked.

Once merged, these changes will close #1263 

 - **How this was tested:** 
   - Tested locally by changing the environment variables to see how the Pug conditional renders

## To-do list

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "gulp -LL",
-    "dev": "gulp dev",
+    "dev": "NODE_ENV=development gulp dev",
     "clean": "gulp clean",
     "start": "npm run dev",
     "ci-all": "npm run build",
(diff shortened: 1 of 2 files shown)


GarthDB:
success

PR #1271, opened 2021-09-22
fix: remove duplicate class names in table examples

pfulton:
Removes duplicate classes from the Table examples

### DIFF
--- a/components/table/metadata/table.yml
+++ b/components/table/metadata/table.yml
@@ -53,7 +53,7 @@ examples:
   - id: table-medium
     name: Size Medium
     markup: |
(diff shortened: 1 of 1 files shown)


lazd (approved):

lazd:
@pfulton @garthdb should this be sent against #dna-7-merge?

pfulton:
> @pfulton @GarthDB should this be sent against #dna-7-merge?

Good call out. Looks like it should be. I'll make that change, but correct me if I'm wrong, @GarthDB

GarthDB:
I think `docs` is a better fit for a commit message so I amended it.

PR #1280, opened 2021-10-01
docs(tabs): resolve flaky visual regression tests

pfulton:
## Description

This updates the Tabs component's examples page to add unique IDs for the examples. The Backstop visual regression tests were not accurate because we previously didn't provide unique IDs to each of our examples.

Refs: #1275 

N/A

## Screenshots
N/A

## To-do list

- [X] If my change impacts documentation, I have updated the documentation accordingly.


pfulton:
@GarthDB for some reason, the Picker vars got updated after I pulled and ran the updates. It feels like we might not want that to ride along with this PR, so I kept those changes to their own commit. I can drop that commit if that's appropriate for a separate PR or if it's an error.

PR #1307, opened 2021-11-05
fix: remove logical property usage from media queries

### LINKED ISSUE #1306: Logical Property in Media Query
Should there be a postCSS plugin or similar that washes the following out in the build?

https://github.com/adobe/spectrum-css/blob/f36154fa03c3d64a5fcf9d70399e6b1b2e4f01d3/components/dialog/index.css#L314

I'm not seeing it working in any version of Chrome right now: https://codepen.io/Westbrook/pen/jOLZZEM so it seems weird that this would be shipping in the default packages.

Westbrook:
## Description
fixes #1306

I couldn't find any information about logical properties in media queries being a "thing" or tools to process them, so I've removed the two instances that I could find.

 - **How this was tested:** Tried to run locally, but the build scripts failed. Will review that separately.
 - **Browser(s) and OS(s) this was tested with:** See above.

## To-do list

### DIFF
--- a/components/dialog/index.css
+++ b/components/dialog/index.css
@@ -311,7 +311,7 @@ governing permissions and limitations under the License.
   }
 }
 
-@media screen and (max-inline-size: 700px) {
+@media screen and (max-width: 700px) {
   .spectrum-Dialog .spectrum-Dialog-grid {
     grid-template-columns:
       var(--spectrum-dialog-confirm-padding) auto 1fr auto minmax(0, auto)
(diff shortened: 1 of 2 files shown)


lazd (approved):
LGTM, this never should have been there and it feels like `postcss-logical` should have done something about it.

lazd:
image
Ran it locally and it seemed to look fine. I was able to build the branch no problem, not sure why you had build issues.

Westbrook:
My build is running again! Looks like I had a bunch of old packages directories that have been deprecated clogging up the works. Excited to get this in.

PR #1319, opened 2021-11-10
fix: vars switcher issues

### LINKED ISSUE #1308: Vars switcher does not reflect selected value when navigating
## Steps to reproduce

1. Go to https://git.corp.adobe.com/pages/lawdavis/spectrum-css-ccx-verify/docs/search.html
2. Select Express vars
3. Go to another page
4. See that Express is no longer selected

## Expected behavior

Express should be selected. Large/Dark correctly stay selected.

## Screenshots

image

### LINKED ISSUE #1309: Scale/Theme/Vars switchers are missing sizing class, appear funny
## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/actionbutton.html
2. Observe the pickers look funny in the upper right corner
3. Inspect the source, see they are missing `spectrum-Picker--sizeM`

## Expected behavior
They are lovely and have the right classes

## Screenshots

image

lazd:
## Description

Fixes #1309, they're pretty now
image

Fixes #1308, selection persists when navigating (fastload only, of course)

### DIFF
--- a/site/resources/js/site.js
+++ b/site/resources/js/site.js
@@ -193,6 +193,9 @@ window.addEventListener('DOMContentLoaded', function() {
         case 'direction':
           setPickerValue(directionPicker, event.value);
           break;
+        case 'vars':
+          setPickerValue(varsPicker, event.value);
+          break;
       }
     }
   });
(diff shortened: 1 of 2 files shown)


pfulton:
This looks great, @lazd - thank you for taking care of this. Just one thought: should we also toggle a class on the `html` element when the switcher changes [like we do for `scale`](https://github.com/adobe/spectrum-css/blob/d4637f1aeee8185124477871d88b7730f879a82f/site/resources/js/SpectrumSwitcher.js#L162) etc.?

GarthDB:
We could add a class, but we can't use it in our component selectors.

pfulton:
> We could add a class, but we can't use it in our component selectors.

Noted! I saw that we were toggling the other classes yesterday when I was digging into a different issue and noticed that I completely missed this when I implemented the original switcher and wondered if we should be consistent with the other switchers.

I'm fine with or without it.

lazd:
I don't think we need to toggle a class on the `<html>` element -- this is just a swapping of core tokens. We don't require a classname to say `.express` and we should not have any custom CSS that relies on it.

pfulton (approved):
I'm good with these changes @lazd. Thanks!

PR #1327, opened 2021-11-16
feat: ActionButton staticColor

### LINKED ISSUE #877: Actionbutton emphasized down state color incorrect
## Description

The `--emphasized` variant of ActionButton is not properly displaying colors for the de-selected `down`/`:active` state. 

## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/actionbutton.html#emphasized
2. Click on any of the first (deselected) emphasized buttons 
3. Observe the background and border color does not get darker when clicking the button

## Expected behavior

Background and border colors should change and match appearance of the non-emphasized action buttons. It appears that this may be somehow overridden by the `.spectrum-Actionbutton--emphasized:hover` styles

## Screenshots

Currently shown, when pressing down on deselected emphasized Action Button:
Screen Shot 2020-08-28 at 12 31 12 PM

Expected appearance:
Screen Shot 2020-08-28 at 12 31 57 PM

## Environment
 - **Spectrum CSS version:**  @spectrum-css/button@3.0.0-beta.3
 - **Browser(s) and OS(s):** Chrome Version 84.0.4147.135 (Official Build) (64-bit) on MacOS

## Additional context

synced to jira: https://jira.corp.adobe.com/browse/SDS-7318

### LINKED ISSUE #1177: Keyboard focus states wrong for some ActionButton variants
### Quiet Selected
image
Should be thick

### Emphasized Quiet
image
Should have focus-ring

## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/actionbutton.html
2. Tab to the mentioned variants

## Environment
 - **Spectrum CSS version:** @spectrum-css/actionbutton@1.0.3
 - **Browser(s) and OS(s):** Chrome on macOS

### LINKED ISSUE #1224: ActionButton Static Color API Change

### LINKED ISSUE #1237: icon-only Action Button/Action Menu small size not to spec
## Description

The small size in SpectrumCSS is 30x24: https://spectrum.adobe.com/page/action-button/ (it's the only size/variant that's not square, so that's a clue)

## Expected behavior

The small size is 24x24 per the XD designs here https://spectrum.adobe.com/page/action-button/ (sizing examples near the bottom)

## Workaround

`--spectrum-actionbutton-s-min-width: 24px`

lazd:
## Description

This PR adds static colors for ActionButton with `.spectrum-ActionButton--staticWhite` and  `.spectrum-ActionButton--staticBlack`, closes #1224.

image

image

It also fixes the focus-ring for Emphasized Selected ActionButtons, fixes #1177

image

It also fixes the down state of the emphasized ActionButton, closes #877 

image

Finally, it fixes the incorrect sizing of the small icon-only ActionButton, closes #1237

image

## To-do list

- [x] ~CSS generation discussion~ Generated CSS removed and refactored manually
- [x] Design verification (deployed [here](https://git.corp.adobe.com/pages/lawdavis/spectrum-css-ccx-verify/docs/actionbutton.html))
  - [x] Update border for Spectrum non-quiet per feedback
  - [x] Get design re-verified for border change

### DIFF
--- a/components/actionbutton/index.css
+++ b/components/actionbutton/index.css
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 
 @import "../commons/basebutton.css";
 @import "../vars/css/components/spectrum-actionbutton.css";
+@import "./actionbutton-generated.css";
 
 .spectrum-ActionButton--sizeS {
   @remapvars {
@@ -22,6 +23,9 @@ governing permissions and limitations under the License.
   /* todo: remove hardcoded values when we update DNA */
   --spectrum-actionbutton-hold-icon-padding-bottom: var(--spectrum-global-dimension-size-25);
   --spectrum-actionbutton-hold-icon-padding-right: var(--spectrum-global-dimension-size-25);
+
+  /* hack to fix the incorrect min-width */
+  --spectrum-actionbutton-textonly-min-width: var(--spectrum-global-dimension-size-300);
 }
 
 .spectrum-ActionButton--sizeM {
@@ -88,6 +92,9 @@ a.spectrum-ActionButton {
   font-weight: var(--spectrum-actionbutton-textonly-text-font-weight);
   line-height: var(--spectrum-actionbutton-textonly-text-line-height);
 
+  /* Let static variants inherit their color */
+  color: inherit;
+
   .spectrum-Icon {
     @inherit: %spectrum-ButtonIcon;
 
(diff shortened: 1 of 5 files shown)


components/actionbutton/metadata/actionbutton.yml
```diff
@@ -652,3 +652,500 @@ examples:
           </button>
         </div>
       </div>
+
+  - id: actionbutton
```
pfulton, line 656: Could you make each of the `id`s in this file unique?

Ideally, I think we'd have some kind of linter running on these to prevent duplicative `id`s, but that's for future work.

components/actionbutton/index.css
```diff
@@ -22,6 +23,9 @@ governing permissions and limitations under the License.
   /* todo: remove hardcoded values when we update DNA */
```
pfulton, line 23: Has this been addressed in this work?

pfulton (changes requested):
Just a few things here, @lazd

pfulton:
Regarding the focus ring & express: does this look correct for the standard examples (first screenshot)? It looks like it's there, but it wasn't immediately apparent to me. I haven't looked at this as closely as you have, so if this is the way it's meant to be, then go forth!

It is a bit more apparent on the emphasized & static examples (second screenshot)

Screen Shot 2021-12-08 at 10 52 51 AM

Screen Shot 2021-12-08 at 10 53 57 AM

components/actionbutton/index.css
lazd: Nah, looks like they're still wrong in DNA or need to be overridden due to the border model.

lazd:
> Regarding the focus ring & express: does this look correct for the standard examples (first screenshot)? It looks like it's there, but it wasn't immediately apparent to me. I haven't looked at this as closely as you have, so if this is the way it's meant to be, then go forth!

Oh no, that's totally wrong, good catch! Investigating...

lazd:
@pfulton oof, the focus-ring bits are non-trivial. Basically, the Express Selected varieties need the same treatment as the Emphasized Selected varieties on both Spectrum and Express.

I'll need to add a ton more core-level aliases to set the gap, refactor ActionButton to always use `:before` for its focus-ring, and then we should be good. Plenty more work to do here.

components/actionbutton/metadata/actionbutton.yml
pfulton: I thought it had implications for VRT! Maybe I've been misunderstanding its importance :-)

lazd:
@pfulton fixed! Should be looking right in both Spectrum and Express now. Good catch!

components/actionbutton/metadata/actionbutton.yml
lazd: Hmmm I don't think it does, but I may be wrong... I know VRT isn't smart enough to realize when the examples get out of order, so it's clearly not using an id, but you know what? *IT SHOULD!*

You're a genius, @pfulton

GarthDB:
What about the focus ring being overlapped by other components? Do we want to change the docs to space the buttons out more? 
image

PR #1329, opened 2021-11-16
build: run buildMedium when built fron the top level

lazd:
## Description

This makes it so `yarn build` will catch undefined variable usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "release:site": "gulp prepareSite && gh-pages -d dist-site/ -f -e .",
     "release:bundles": "gulp releaseBundles",
     "version": "gulp version",
+    "preversion": "yarn build",
     "backstop:approve": "backstop approve --config=backstop_data/backstop_test.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
     "release:site": "gulp prepareSite && gh-pages -d dist-site/ -f -e .",
     "release:bundles": "gulp releaseBundles",
     "version": "gulp version",
+    "preversion": "yarn build",
```
lazd, line 15: @GarthDB I think this works, but I'm too afraid to test it. Basically, this should run the build before versioning, so if it's broken, then it will stop the versioning process from happening, which leaves you with a clean working directory. Thoughts? Can we test it somehow without like, versioning and pushing tags and stuff?

lazd: Ok I just ran `yarn lerna version` and can confirm that the build runs BEFORE versioning, but that doesn't stop it from creating commits and tags, uhg!

lazd: IT WORKS

PR #1331, opened 2021-11-17
Thumbnail object-fit

### LINKED ISSUE #1134: Are the mitigating factors behind the "fill" version of the thumbnail pattern
## Description
Currently the "fill" delivery of the following DOM structure:
```
<div class="spectrum-Thumbnail">
  <div class="spectrum-Thumbnail-background" style="background-image: url(img/example-card-portrait.jpg)" title="Eiffel Tower at night"></div>
</div>
```
As opposed to the default version featuring an `` tag:
```
<div class="spectrum-Thumbnail is-focused">
  <img class="spectrum-Thumbnail-image" src="img/example-card-portrait.jpg" alt="Eiffel Tower at night">
</div>
```

Was there something in particular about `background-size: cover` and its requirement for this alternate DOM that lead to its use over something like `object-fit: cover` on the image?
```
<div class="spectrum-Thumbnail is-focused">
  <img class="spectrum-Thumbnail-image" src="img/example-card-portrait.jpg" alt="Eiffel Tower at night" style="object-fit: cover;">
</div>
```
In this way a `spectrum-Thumbnail--cover` or `spectrum-Thumbnail-image--cover` toggle would be all that's needed to support this delivery, as opposed to change to the entire DOM.

Thoughts?

GarthDB:
## Description

Since we dropped support for IE, we can eliminate the hack. This is
not a breaking change because background still has the same properties,
but if image is dropped it, it will fill by default without the
background pseudo element hack.

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/thumbnail/index.css
+++ b/components/thumbnail/index.css
@@ -81,9 +81,11 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Thumbnail-image {
-  max-height: 100%;
-  max-width: 100%;
+  height: 100%;
+  width: 100%;
   z-index: 1;
+  object-fit: cover;
+  object-position: center;
 }
 
 .spectrum-Thumbnail-background {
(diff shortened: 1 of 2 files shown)


pfulton (approved):
Looks great, @GarthDB. Thanks for catching those `id`s in the Yaml file!

PR #1333, opened 2021-11-18
Express Dialog

GarthDB:
## Description
Changes to dialog to use aliases better so it supports express vars.

Spectrum dialog:
image

Express dialog:
image

 - **How this was tested:** 

## To-do list


GarthDB:
Looks like the title font weight isn't switching. I'll dig into it.

GarthDB:
Waiting on close button to finish this up.

GarthDB:
The close button X icon is smaller, but not actually a part of the design system ui kit, so we don't have a definition of the size. I went with the default Close Button size.

components/dialog/metadata/dialog.yml
```diff
@@ -77,9 +77,9 @@ examples:
                 Quisque egestas diam in arcu cursus euismod quis viverra. Posuere ac ut consequat semper viverra nam libero justo laoreet. Enim ut tellus elementum sagittis vitae et leo duis ut. Neque laoreet suspendisse interdum consectetur libero id
                 faucibus nisl. Diam volutpat commodo sed egestas egestas. Dolor magna eget est lorem ipsum dolor. Vitae suscipit tellus mauris a diam maecenas sed. Turpis in eu mi bibendum neque egestas congue. Rhoncus est pellentesque elit ullamcorper
                 dignissim cras lobortis.</section>
-              <button aria-label="Dismiss" class="spectrum-ActionButton spectrum-ActionButton--quiet spectrum-ActionButton--sizeM spectrum-Dialog-closeButton" type="button" onclick="closeDialog(this.closest('.spectrum-Modal-wrapper'))">
-                <svg class="spectrum-Icon spectrum-UIIcon-Cross500" focusable="false" aria-hidden="true">
-                  <use xlink:href="#spectrum-css-icon-Cross500"></use>
+              <button aria-label="Dismiss" class="spectrum-CloseButton spectrum-CloseButton--sizeM spectrum-Dialog-closeButton" type="button" onclick="closeDialog(this.closest('.spectrum-Modal-wrapper'))">
```
lazd, line 80: 👏

lazd: We sure `sizeM` is right here?  it's quite a bit smaller than the existing X...

lazd: Ah, `sizeM`'s icon size is supposed `cross100` (quiet a bit smaller than the existing X), but I see you've used has `cross300` here to match the old design. I suppose that's ok, as long as design approved of the way the hover state looks.

pfulton (approved):
LGTM pending design approval. Thanks for this!

components/dialog/metadata/dialog.yml
lazd: Looks like it's the Icon Size Large version, I think we're good! https://opensource.adobe.com/spectrum-css/closebutton.html#iconsize-large

PR #1338, opened 2021-12-13
chore: update @adobe/spectrum-tokens and @adobe/spectrum-express-tokens

GarthDB:
BREAKING CHANGE: changes to button token api for static and style
variants. Removal of over background.

## Description

This breaks the build, so I moved it to a pr.


lazd:
Looks like the build is fixed, I manually looked at a few components and they look fine.

PR #1341, opened 2021-12-16
fix(thumbnail): prefer separate class for image covering

### LINKED ISSUE #1134: Are the mitigating factors behind the "fill" version of the thumbnail pattern

pfulton:
## Description

This adds an additional class (`.spectrum-Thumbnail-image--cover`) for supporting image "covering" via `object-fit` and `object-position` and restores the previous behavior wherein landscape or portrait images filled horizontally or vertically, but left space between their container edges.

Closes #1134 

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/thumbnail/index.css
+++ b/components/thumbnail/index.css
@@ -81,9 +81,14 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Thumbnail-image {
+  max-height: 100%;
+  max-width: 100%;
+  z-index: 1;
+}
+
+.spectrum-Thumbnail-image--cover {
   height: 100%;
   width: 100%;
-  z-index: 1;
   object-fit: cover;
   object-position: center;
 }
--- a/components/thumbnail/index.css
+++ b/components/thumbnail/index.css
@@ -86,11 +86,13 @@ governing permissions and limitations under the License.
   z-index: 1;
 }
 
-.spectrum-Thumbnail-image--cover {
-  height: 100%;
-  width: 100%;
-  object-fit: cover;
-  object-position: center;
+.spectrum-Thumbnail--cover {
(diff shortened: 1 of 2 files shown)


lazd (changes requested):
So to follow [SUIT](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#ComponentName--modifierName) the way we have, we shouldn't put modifier classes on descendent classnames. The CSS API should be expressed entirely with classes on the outermost element.

With that, this class should be `spectrum-Thumbnail--cover` on the outer element, and the selector for your new class should be `.spectrum-Thumbnail--cover .spectrum-Thumbnail-image`.

Otherwise, it looks great!

lazd (approved):

PR #1345, opened 2022-01-10
feat: add InfieldButton component

lazd:
## Description

Adds the (currently unused) InfieldButton component.

## Screenshots

image

## To-do list

- [x] Basic implementation
- [ ] Audit classnames, make sure we're following convention
- [ ] Figure out why padding on the fill needs to be forced to 0 to render right

### DIFF
--- a/components/infieldbutton/gulpfile.js
+++ b/components/infieldbutton/gulpfile.js
@@ -1,0 +1,1 @@
+module.exports = require('@spectrum-css/component-builder');
--- a/components/infieldbutton/index.css
+++ b/components/infieldbutton/index.css
@@ -1,0 +1,35 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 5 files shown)


pfulton (changes requested):
I think the code looks good here, @lazd! Nice work.

Just double-checking on a few focus-ring related things:

* For the default theme, is this the correct focus-ring for the `invalid` state?
Screen Shot 2022-01-28 at 3 19 51 PM

* For express, it looks like there's no focus-ring on any of the buttons
* For express, the `invalid` state looks like the default state. Is that correct?
Screen Shot 2022-01-28 at 3 20 02 PM

lazd:
@pfulton I think we're good to merge! See below.

> For the default theme, is this the correct focus-ring for the invalid state?

So I took a look at this again, checked out the data in DSS, and I believe the invalid + keyboard focus is correct. It's using the same alias as Textfield and behaves the same way, so it should look correct as is.

> For express, it looks like there's no focus-ring on any of the buttons

Yup! This is because focus is never given to the button itself, instead the outer field shows focus.

> For express, the invalid state looks like the default state. Is that correct?

Yup! For the same reason as above -- the invalid state is reflected on the outer field.

pfulton (approved):
@lazd awesome! I think we're good to go here. Thanks for investigating the things that I had inquired about.

PR #1363, opened 2022-01-26
Express tag

GarthDB:
## Description
Add express support to Tag

## To-do list


components/tag/index.css
```diff
@@ -18,6 +18,7 @@ governing permissions and limitations under the License.
     filter: color;
     replace: --spectrum-tag$1;
   }
+  --spectrum-tag-texticon-text-font-size: var(--spectrum-global-dimension-font-size-75);
```
lazd, line 21: Seems like this should be driven by tokens, is this missing from DSS?

pfulton, line 21: I'm confused about this definition (and the ones on line `30` and `39`). Do we need to define these, or should we refer to what's coming across in the vars file? It looks like `--spectrum-tag-s-texticon-text-font-size` is defined in there.

components/tag/index.css
```diff
@@ -70,26 +68,29 @@ governing permissions and limitations under the License.
...
     /* Add padding for "avatar" variant */
-    margin-inline-start: calc(var(--spectrum-tag-texticon-avatar-padding-x) - var(--spectrum-tag-texticon-padding-left));
+    margin-inline-start: calc(var(--spectrum-tag-texticon-padding-left) - var(--spectrum-tag-textonly-padding-left));
 
-    ~ .spectrum-Tag-label {
-      margin-inline-end: calc(var(--spectrum-tag-texticon-avatar-padding-x) - var(--spectrum-tag-texticon-padding-right));
-    }
+    /* ~ .spectrum-Tag-label {
+      margin-inline-end: calc(var(--spectrum-tag-texticon-icon-gap) - var(--spectrum-tag-texticon-padding-right));
+    } */
```
pfulton, line 78: Is it safe to delete this commented out code now?

pfulton (changes requested):
This is awesome, @GarthDB. Thanks for getting this in.

I had a few questions here.

Additionally, just double-checking the focus-ring on the regular theme—do these look correct?

Screen Shot 2022-01-26 at 1 53 24 PM

---

Screen Shot 2022-01-26 at 1 53 37 PM

components/tag/index.css
GarthDB: it is incorrect in DSS. I'll circle back and fix it afterwards; I'm mostly expediting this because it's blocking releases of other things like tabs.

pfulton (approved):
Lookin' good!

ship it

PR #1368, opened 2022-01-26
Actiongroup focus ring

GarthDB:
## Description
SWC requested that we update this property by changing a vars definition.

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/actiongroup/index.css
+++ b/components/actiongroup/index.css
@@ -70,7 +70,6 @@ governing permissions and limitations under the License.
       }
     }
   }
-
   &:not(.spectrum-ActionGroup--quiet) {
     flex-wrap: nowrap;
 
@@ -79,12 +78,20 @@ governing permissions and limitations under the License.
       border-radius: 0;
       z-index: 0;
 
+      &:after {
+        --spectrum-actionbutton-focus-ring-border-radius-adjusted: 0;
+        test: var(--spectrum-actionbutton-focus-ring-border-radius-adjusted);
+      }
+
       &:first-child {
         border-start-start-radius: var(--spectrum-actionbutton-m-texticon-border-radius);
         border-end-start-radius: var(--spectrum-actionbutton-m-texticon-border-radius);
         margin-inline-end: calc(
           -1 * var(--spectrum-actionbutton-m-texticon-border-size) / 2
         );
+        &:after {
+          --spectrum-actionbutton-focus-ring-border-radius-adjusted: calc(var(--spectrum-actionbutton-m-quiet-textonly-border-radius) + var(--spectrum-actionbutton-m-textonly-focus-ring-gap)) 0 0 calc(var(--spectrum-actionbutton-m-quiet-textonly-border-radius) + var(--spectrum-actionbutton-m-textonly-focus-ring-gap));
+        }
       }
 
       &:last-child {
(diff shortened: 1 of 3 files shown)


components/actiongroup/index.css
```diff
@@ -79,12 +78,20 @@ governing permissions and limitations under the License.
       border-radius: 0;
       z-index: 0;
 
+      &:after {
+        --spectrum-actionbutton-focus-ring-border-radius-adjusted: 0;
+        test: var(--spectrum-actionbutton-focus-ring-border-radius-adjusted);
```
lazd, line 83: Looks like some leftover props slipped in.

components/actiongroup/index.css
```diff
@@ -79,12 +78,20 @@ governing permissions and limitations under the License.
       border-radius: 0;
       z-index: 0;
 
+      &:after {
+        --spectrum-actionbutton-focus-ring-border-radius-adjusted: 0;
```
lazd, line 82: Honestly I'd set this on the actual element, not the pseudo-element. It seems strange to define it initially on `.spectrum-ActionButton`, but then set it in the `:after` later, even if that is where it's being used.

components/actiongroup/index.css
```diff
@@ -79,12 +78,20 @@ governing permissions and limitations under the License.
...
+      }
+
       &:first-child {
         border-start-start-radius: var(--spectrum-actionbutton-m-texticon-border-radius);
         border-end-start-radius: var(--spectrum-actionbutton-m-texticon-border-radius);
         margin-inline-end: calc(
           -1 * var(--spectrum-actionbutton-m-texticon-border-size) / 2
         );
+        &:after {
+          --spectrum-actionbutton-focus-ring-border-radius-adjusted: calc(var(--spectrum-actionbutton-m-quiet-textonly-border-radius) + var(--spectrum-actionbutton-m-textonly-focus-ring-gap)) 0 0 calc(var(--spectrum-actionbutton-m-quiet-textonly-border-radius) + var(--spectrum-actionbutton-m-textonly-focus-ring-gap));
```
lazd, line 93: Does design want `0` or something tiny like 2px that would show slightly on retina displays?

0:
image


1px:
image

components/button/skin.css
```diff
@@ -10,32 +10,6 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  --spectrum-button-m-accent-outline-texticon-icon-color-key-focus: var(--spectrum-semantic-cta-background-color-key-focus);
-  --spectrum-button-m-accent-outline-texticon-icon-color-key-focus: var(--spectrum-semantic-cta-background-color-key-focus);
-
-  --spectrum-button-m-accent-outline-texticon-text-color: var(--spectrum-semantic-cta-background-color-default);
-  --spectrum-button-m-accent-outline-texticon-text-color-hover: var(--spectrum-semantic-cta-background-color-hover);
-  --spectrum-button-m-accent-outline-texticon-text-color-down: var(--spectrum-semantic-cta-background-color-down);
-  --spectrum-button-m-accent-outline-texticon-text-color-key-focus: var(--spectrum-semantic-cta-background-color-key-focus);
-  --spectrum-button-m-accent-outline-texticon-text-color-key-focus: var(--spectrum-semantic-cta-background-color-key-focus);
-}
-
```
lazd, line 38: Looks like these changes came along for the ride, maybe merge `main` since they're already there, but it probably doesn't matter.

lazd (changes requested):
Let's double check the radius (`0` or `1px`) and move that var out of the pseudo-element to the element itself, then this is good!

components/actiongroup/index.css
GarthDB: Oh, I wanted to ask you about this, I forgot I left it in, but this was my hack to get the unused vars to not clear out my new definition.

components/actiongroup/index.css
GarthDB: It is for specificity to override the original definition.

components/actiongroup/index.css
```diff
@@ -17,6 +17,11 @@ governing permissions and limitations under the License.
   );
 }
 
+/* hack: define var so it's not reported as undefined */
```
lazd, line 20:
```suggestion
/* hack: use var so it's not reported as unused */
```

lazd (approved):

PR #1369, opened 2022-01-28
fix(breadcrumb): use flex-start for justify content

alexander-riss:
ENG: this was probably a typo in the first place - autoprefixer flagged this when using spectrum-css, so fixing it here to get rid of the linter warning

## To-do list

### DIFF
--- a/components/breadcrumb/index.css
+++ b/components/breadcrumb/index.css
@@ -36,7 +36,7 @@ governing permissions and limitations under the License.
 
 .spectrum-Breadcrumbs {
   display: flex;
-  justify-content: start;
+  justify-content: flex-start;
   list-style-type: none;
   flex-wrap: nowrap;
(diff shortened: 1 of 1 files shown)


lazd:
Note that `start` is a valid value for `justify-content`, it just doesn't respect `flex-direction: row-reverse`, whereas `flex-start` does. I tested the change locally and found it doesn't break RTL, so I think we're good to merge this.

That said, can you please paste the linter error that autoprefixer was showing you? Thanks!

alexander-riss:
The warning was:
start value has mixed support, consider using flex-start instead

alexander-riss:
any other objections? this warning is effectively blocking us from migrating to create-react-app 5.0 - as autoprefixer will log this warning during build time and in this way break any CI build

lazd:
LGTM!

PR #1387, opened 2022-02-16
Update closebutton generated CSS

lazd:
## Description

Quick regen got the colors where they should be


GarthDB:
I'd prefer to move further from generated code than regenerate it. I believe it is better to use the component tokens as well.

PR #1402, opened 2022-03-09
Express Picker

GarthDB:
 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/picker/index.css
+++ b/components/picker/index.css
@@ -47,7 +47,7 @@ governing permissions and limitations under the License.
     cursor: default;
   }
 
-  . {
+  .spectrum-Picker-icon {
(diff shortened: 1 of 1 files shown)


components/picker/index.css
```diff
@@ -44,7 +47,7 @@ governing permissions and limitations under the License.
     cursor: default;
   }
 
-  .spectrum-Picker-icon {
+  . {
```
pfulton, line 50: I think you may be missing the selector here.

pfulton (changes requested):
I think this looks good, @GarthDB - just one thing that I noted here.

pfulton (approved):

PR #1407, opened 2022-03-17
feat: implement AssetCard

lazd:
## Description
This PR implements Asset Card for both Spectrum and Express.

## To-do list

- [x] Initial hack at markup/css
- [x] Spectrum implementation
- [x] Express implementation
- [x] Figure out if we're doing clip handles and in/out handles (we're not)
- [x] Address todos
- [x] DSS work for transparent aliases, other updates
- [x] DSS work for focus ring border radii
- [x] Border colors for down
- [x] Border colors for hover when selected - no feedback from design, so no changes
- [x] Verify typography styles we're using are correct
- [x] Decide whether we're using QuickActions or hacking up a new little gray box to hold the checkbox
- [x] Double check animation timing and whatnot
- [x] Design verification
  - [x] Fixup CCX issues (list forthcoming)
- [x] Update spectrum-tokens after PR merged/released


components/assetcard/metadata/assetcard.yml
```diff
@@ -0,0 +1,211 @@
...
+          <img class="spectrum-AssetCard-asset" src="img/example-card-portrait.jpg" />
+          <div class="spectrum-AssetCard-selectionOverlay"></div>
+        </div>
+        <div class="spectrum-AssetCard-header">
+          <div class="spectrum-AssetCard-title">Card Title</div>
+          <div class="spectrum-AssetCard-headerContent"></div>
+        </div>
+        <div class="spectrum-AssetCard-content">
+          Image
+        </div>
```
pfulton, line 139: Does the `Highlight Selection` variation not include the `spectrum-AssetCard-selectionIndicator`? Just double-checking.

pfulton (commented):
This looks *awesome*. Nice work, @lazd!

Just one question here, but otherwise, I think we're good to go.

components/assetcard/metadata/assetcard.yml
pfulton: Sounds good! Thanks.

PR #1413, opened 2022-03-28
feat: create -rgb suffixed variables for all rgb() values

lazd:
## Description

This PR adds `-rgb` suffixed variables with the comma separated list of individual color values. This enables usage such as:

```css
--spectrum-assetcard-overlay-background-color: rgba(var(--spectrum-global-color-blue-400-rgb), 0.20);
```


GarthDB:
I think this is a working solution, but should we consider an alpha postcss function?

GarthDB (approved):

lazd:
@GarthDB a PostCSS solution would not work with themes where colors change (i.e. current Express and Spectrum) -- at build time, we would not know the value of the color since it could be Express or Spectrum. It would be fine once we unify colors.

I think this is the best workable solution for now, and it's really nice to use native CSS functionality!

PR #1416, opened 2022-03-30
feat(textfield): add read-only state

pfulton:
In order to support future use cases, this adds and styles a `readonly` state to the textfield component.

It ultimately addresses the request in [CSS-30 (add readonly state to Datepicker)](https://jira.corp.adobe.com/secure/RapidBoard.jspa?rapidView=33123&projectKey=CSS&view=detail&selectedIssue=CSS-30) and adds an example to the Datepicker's documentation to show the proper markup to achieve a `readonly` Datepicker.

Additionally, this fixes two browser console errors where we were attempting to manipulate classes for elements that did not always exist.

URL for validation:
https://git.corp.adobe.com/pages/pfulton/spectrum-css-ccx-verify/docs/textfield.html

 - **How this was tested:** 

## To-do list

### DIFF
--- a/site/resources/js/enhancement.js
+++ b/site/resources/js/enhancement.js
@@ -133,7 +133,7 @@ governing permissions and limitations under the License.
     var pickerButton = inputgroup.querySelector('.spectrum-PickerButton');
     if (focused) {
       inputgroup.classList.add(focusClass);
-      pickerButton.classList.add(focusClass);
+      if (pickerButton) pickerButton.classList.add(focusClass);
       if (event.target.tagName !== 'INPUT') {
         input.focus();
       }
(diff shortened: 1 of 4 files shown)


pfulton:
This has been verified/approved by design (Jacob)

site/resources/js/enhancement.js
```diff
@@ -469,6 +469,7 @@ function toggleInputGroupFocus(event) {
   var classList = event.target.classList;
   var closestSelector;
   // target within InputGroup
+  if (!classList) return;
```
lazd, line 472: Strange, why would this ever happen?

lazd:
LGTM, note that this has left padding when read-only, which I think is correct, but does not match the designs we looked at the other day in Engineering Workshop (there was no left padding, I remember complaining about it).

site/resources/js/enhancement.js
pfulton: FWIW, it's happening in production on any page.

pfulton:
> LGTM, note that this has left padding when read-only, which I think is correct, but does not match the designs we looked at the other day in Engineering Workshop (there was no left padding, I remember complaining about it).

Great call-out. I'll double-check with Jacob once more to be sure.

PR #1422, opened 2022-04-08
fix: stick to a single selector for consistency in tag clearbutton

GarthDB:
## Description

A simple fix.

## To-do list


pfulton (approved):
This looks good, @GarthDB. I notice that the docs were already using `.spectrum-Tag-clearButton`, which is great because less work.

I'm wondering about the severity of the version bump for this, though. Were teams already consuming this? If so, I think this would be a breaking change?

GarthDB:
@pfulton I'm leaving it as a patch because the dom doesn't change - just our use of class selectors. It's a bug and the patch shouldn't actually impact anyone.

PR #1426, opened 2022-04-14
feat: add --spectrum-custom-static-content-color for setting color

### LINKED ISSUE #1425: [Button][ActionButton]: Static Color should use a custom property API, not color: inherit to set text color
## Description
Currently, Static Color buttons rely on `color` being set on the outer element to set the text color of the button. This is because this mechanism was created before custom properties were mainstream.

This has two issues:

1. Setting color on an outer element such that its inherited and used within the button is counter-intuitive, and despite it being a fundamental aspect of how CSS behaves, most folks are unaware of this application. Despite documentation, this has confused users before.
2.  The usage of color precludes the newly added design requirement that the default color be `black` or `white` -- it's not possible to say "inherit unless someone says not to" because of the nature of inheritance. 

To address this, we should use a CSS custom property API to define the text color for Static Color buttons, something like `--spectrum-custom-background-color` (as the text is supposed to match the background color) or `--spectrum-custom-static-text-color` (as its actually affecting the text itself).

Using custom property fallbacks, we could easily use a `white`/`black` fallback color with something like:

```
css
.spectrum-Button--staticBlack .spectrum-Button-label {
  color: var(--spectrum-custom-static-text-color, white);
}
```

## Why do you need this feature or component?

Intuitive API, simpler docs, and correct implementation of design requirements.

## Additional context

Customer confused by existing mechanism: https://adobedesign.slack.com/archives/C5N154FEY/p1649805163901619
Slack discussion with design: https://adobedesign.slack.com/archives/G019JTYMT6H/p1649861392969419

lazd:
## Description

This replaces `color` as a mechanism for customizing the content color and adds a default color for static white and static  black Buttons and ActionButtons

BREAKING CHANGE: this replaces the usage of color

Closes #1425 

## To-do list

- [ ] Validate this API with Spectrum Web Components
- [ ] Consider the `-custom` suffix with engineering
- [ ] Nail the naming of the custom property with content strategy

### DIFF
--- a/components/actionbutton/index.css
+++ b/components/actionbutton/index.css
@@ -14,6 +14,11 @@ governing permissions and limitations under the License.
 @import "../vars/css/components/spectrum-actionbutton.css";
 @import "./actionbutton-generated.css";
 
+/* hack: don't complain about this being undefined */
+.🤫 {
+  --spectrum-custom-static-content-color: none;
+}
+
 .spectrum-ActionButton--sizeS {
   @remapvars {
     find: --spectrum-actionbutton-s-;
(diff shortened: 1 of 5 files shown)


pfulton:
This has been addressed with the move to the new tokens system.

PR #1429, opened 2022-04-25
feat: add core tokens

lazd:
## Description

This PR adds core tokens and a new package, `components/coretokens` that builds the CSS custom properties.

~~This is a BREAKING CHANGE since the lightest theme is removed.~~ Just kidding, we added it back.

It's IS a BREAKING CHANGE only for components that have been migrated to core tokens because `.spectrum--express` is now required for the Express theme.

## To-do list

- [x] Get core tokens building
- [x] Get core tokens included in docs site
- [x] Support core tokens overrides and additions using `custom.css`
- [x] Un-deprecate Lightest removal (just duplicate light theme)
- [x] Add docs for switching to Express
- [x] Validate approach with team
  - [x] repository location
  - [x] system-level overrides overrides using `custom.css`
  - [x] approach of applying tokens using classnames
  - [x] component-level system-specific overrides overrides using classnames within component folders
- [x] Validate implementation with SWC
~- [ ] Explore slimmer build system~

## Next steps

1. Merge feat: implement ActionButton core tokens #1430 into `lazd/CSS-75-core-tokens-pre`
2. Release a beta of `@spectrum-css/tokens` and `@spectrum-css/actionbutton` from the `lazd/CSS-75-core-tokens-pre` branch
3. Socialize the beta with SWC, get approval
4. Merge `lazd/CSS-75-core-tokens-pre` into `main` and graduate ActionButton out of beta (breaking change: requires `spectrum-ActionButton--icon` and needs `spectrum--express` for Express mode)


components/actionbutton/index.css
```diff
@@ -11,183 +11,426 @@ governing permissions and limitations under the License.
...
+  --spectrum-actionbutton-font-size: var(--spectrum-font-size-300);
+  --spectrum-actionbutton-text-to-visual: var(--spectrum-text-to-visual-300);
+  --spectrum-actionbutton-edge-to-hold-icon: var(--spectrum-action-button-edge-to-hold-icon-extra-large);
+  --spectrum-actionbutton-edge-to-visual: calc(var(--spectrum-edge-to-visual-300) - var(--spectrum-actionbutton-border-width));
+  --spectrum-actionbutton-edge-to-text: calc(var(--spectrum-edge-to-text-300) - var(--spectrum-actionbutton-border-width));
+  --spectrum-actionbutton-edge-to-visual-only: calc(var(--spectrum-edge-to-visual-only-300) - var(--spectrum-actionbutton-border-width));
+}
+
+.spectrum {
+  .spectrum-ActionButton {
```
Westbrook, line 79: @lazd I'm pretty worries about this decision. It'll take me some time to see if there's something that SWC can do to work around this, but as I see it, we're making a "system" level decision at a "component" level scope, which will present an issue in our shadow DOM encapsulated context.

Knee jerk possibilities/wishes:
- if only https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context was something all browser were going to ship, we'd be good 💢 
- container queries will rock this at some point in the future. `@container (--spectrum-theme: express)` 🤯. Beyond not wanting to require a polyfill here, I've also not seen a good one for either shadow DOM or CSS Custom Properties.
- there's some work on "toggling" values with CSS Custom Properties which is interesting, but would likely be a PITA at Spectrum scale: https://lea.verou.me/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/ 🪄 
- we may be able to reconstitute the idea of Component Level Global Vars through tooling on our end 🚧 
- even if we had `::theme` which allow us to convert all selectors to `::part` we'd still have to hoist values up to the root to distribute them 🤕 
- we do a lot of dumb work to cover over the `[dir=ltr|rtl]` support, but there's at least a clear future (even if not soon) as to when it could be removed, doing that here might be possible, but it would also involve an extra something on _EVERY_ element we create to bind the theme to it 😭

Sounds like I'll have an official beta to test some of these thoughts out on here soon, but overall all of these require some level of changes on one end of the process or the other to get this to work...

lazd: Good insight, @Westbrook. Perhaps we could write CSS closer to the standards, using `@container` etc, then do some PostCSS magic to get our existing build output. Then, SWC can use our source CSS instead of the built output and process it in a way that works best for SWC. Thoughts?

Westbrook: That is quite an interesting proposal. I feel like that'd be a doozie of a postCSS plugin, but it may be the right call here. 

What do you think you would compile that usage down to in the Spectrum CSS output? It is great that right now our systems work really well together. While you sometimes need a little more on top of SWC to get full CSS, you don't need all of CSS again, and I'd want to make sure we were able to continue that relationship.

I wonder if the lower CSS Custom Property usage in general of this change would put us in a place where we could ship ALL the themes ALL the time and manage `--spectrum-theme`, `--spectrum-color-theme`, and `--spectrum-scale` at the root? 😱

lazd: Yes exactly, we can ship all themes, scales, and colors at the same time with the new tokens, then just manage the top level class to let custom properties be set that drive those changes.

I think it would compile down to something close to what we have right now, the real question is what it would look like for SWC in the absence of `@container` support in the browser...

lazd: A couple potential solutions:

## Move Express overrides into Tokens

One solution that would work immediately would be to move the `.spectrum--express .spectrum-ActionButton` block into `components/tokens/express.css`, and change all of `.spectrum .spectrum-ActionButton` to be simply `.spectrum-ActionButton`. You would then have to import `@spectrum-css/tokens/dist/express.css` if you wanted Express support, and would still have to apply `.spectrum--express` selector higher up for these custom properties to take effect.

Since the specificity of `.spectrum--express .spectrum-ActionButton` is greater than `.spectrum-ActionButton`, the definitions in `express.css` will override and things will work.

The problem with this is that we'll be separating the component-level tokens out from the `components/actionbutton` folder _only for Express_.  This unfortunately means that all Express overrides must all be included at the top level.

## Move component-level tokens into separate files

Exactly the same as above, but we could put the Express overrides in a separate file `components/actionbutton/express.css`, then your build process would have to find if there's an `express.css` in each of the component folders and append that to your top-level imports. This unfortunately means that all Express overrides are manually processed by your scripts and must all be included at the top level.

## Aliases

Another option is to define an alias for each thing that changes between Express and Spectrum, and store those aliases inside of `components/tokens/aliases.css`. Instead of defining things that change between Express and Spectrum inside of `.spectrum--express` inside of `components/actionbutton/`, we'd have that information captured higher up.

We would have to work with design to properly name these aliases, and it would end up looking exactly like what we already do in DSS.

## `system` API option

Default to Spectrum, and allow the `system` API option on ActionButton to control which tokens are used. Rewrite `.spectrum--express .spectrum-ActionButton` selectors to `sp-actionbutton[express]` and be done.

Thoughts?

Westbrook: Quick feedback:
1. **Move Express overrides into Tokens**: Relying on a call above the component is *not acceptable* for us as the relation of `.spectrum--express .spectrum-ActionButton` when not in the same shadow root is not something that can be created in the Spectrum Web Component context. *CONJECTURE* It's _possible_ that were we to move to using all [`:where()` psudo selectors](https://caniuse.com/mdn-css_selectors_where) (see their decent support matrix) at the component level then we could create that specificity relationship without needing the extra classes. 
2. **Move component-level tokens into separate files**: I'm not immediately sure how this approach avoids a class structure like the above. There is theoretically a way that we could go about passing these overrides into the component. At first blush, the techniques to achieve this seem to put unnecessary stress on the element consumer at dev/build time or on the application consumer at run time in a way that we'd likely need to call *not acceptable*, unless we could find absolutely no other way.
3. **Aliases**: This works _really_ well in the current architecture. The explosion of CSS Custom Properties was an issue as part of the DSS approach, but we've found the largess in this area to [not be required](https://github.com/adobe/spectrum-web-components/pull/2174). What would it look like to educate a system earlier in the pipeline not to create those Custom Properties at all? I can see that being more complex at scale, but it's also much more powerful. It clearly outlines at the system level how we got from Spectrum to Express to Next to Infinity and beyond. One of the first conversations I had with @GarthDB about this new system got into customizations, and while we don't spend much time thinking about the system as customizable outside "Spectrum" it's the ability to customize in this way that frees to the design team to be as expressive and creative as they need to be with the system over time.
4. **system API option**: needing to branch _every single_ element at the complexity we're talking about is quite prohibitive. This really runs close to how we might mange the "Move component-level tokens into separate files" option at the SWC level and should really be held as a last resort.

---

As for my options above, the most promising felt like Container Queries or CSS Custom Property toggling, but they both suffer from an abject inability to scale nicely to support Next and Infinity and beyond, etc. I ran into some issues with the conversion of Action Button in the context of the current beta release not being completely converted but I was working on a [manual conversion](https://codepen.io/Westbrook/pen/YzeVxKL) to the toggling approach. It relies on this idea:

```css
:root {
  --ON: ;
  --OFF: inherit;
  --SPECTRUM-THEME: var(--OFF);
  --EXPRESS-THEME: var(--OFF);
}

.spectrum {
  --SPECTRUM-THEME: var(--ON);
}

.spectrum--express {
  --EXPRESS-THEME: var(--ON);
}

.spectrum-ActionButton {
    --_express-actionbutton-background-color-default: var(--EXPRESS-THEME) var(--spectrum-gray-200);
    --_spectrum-actionbutton-background-color-default: var(--SPECTRUM-THEME) var(--spectrum-gray-75);

    --spectrum-actionbutton-background-color-default: var(
      --_express-actionbutton-background-color-default,
      var(--_spectrum-actionbutton-background-color-default)
    );
}
```
It's not the prettiest to write, but might be a viable conversion of a Container query approach (which would be the forward facing solution to a problem like this, if we ignored the scaling issues) until there's acceptable support for `@container style(--theme = express)`, which may be coming sooner than we think...

lazd: 1. **Move Express overrides into Tokens:** totally agree, this solution has a series of issues that make it a non-starter
2. **Move component-level tokens into separate files:** This solution is no different than 1, it's just putting the code in a different place
3. **Aliases:** This, to me, is the ideal solution, but it means lots of thinking at the system level. Alias-driven themes moves the actual differences up to the core level, and as you said works really well in the current approach (despite the extra variables, which we don't need)
4. **system API option:** I agree, this could be very annoying for consumers
5. **toggling:** Oof, the ergonomics of this are really rough if done manually, but if we can automate this with a PostCSS plugin it could work. I also wonder about the performance implication of this, given that each of the new custom properties we define now needs 2 additional custom properties, maybe we can avoid adding additional variables and just inline it all?

I'm going to get a task on my sprint to poke at this problem and see if I can make it less nutty.

Westbrook: Very much agree that toggling would have to be mechanical. If it were possible to do as a compilation target for Container Queries, that would be my favorite option, as we'd also be able to remove it at some point later.

As you get into all this, let me know if you want to throw some things at the wall together to see what sticks.

lazd: Sounds great, yes I definitely want to pair on this and throw some ideas at the wall. I love the notion of having Container Queries compile down to this with a PostCSS plugin, and I'm going to try to get that on my sprint.

PR #1440, opened 2022-05-23
feat: add PostCSS plugin to transform container queries to support SWC use case

lazd:
## Description

This PR adds `postcss-containerizer` to support switching Express/Spectrum using the [var toggle hack](https://lea.verou.me/2020/10/the-var-space-hack-to-toggle-multiple-values-with-one-custom-property/). This plugin changes:

```css
@container(--system: spectrum) {
  .component {
    --background-color: blue;
    --border-color: darkblue;
  }
  .component .icon {
    --color: gray;
  }
}

@container(--system: express) {
  .component {
    --background-color: purple;
    --border-color: darkpurple;

  }
  .component .icon {
    --color: white;
  }
}
```

Into:

```css
.component {
    --system-spectrum-background-color: var(--system-spectrum) blue;
    --system-express-background-color: var(--system-express) purple;
    --background-color: var(--system-express-background-color, var(--system-spectrum-background-color));
    --system-spectrum-border-color: var(--system-spectrum) darkblue;
    --system-express-border-color: var(--system-express) darkpurple;
    --border-color: var(--system-express-border-color, var(--system-spectrum-border-color))
}
.component .icon {
    --system-spectrum-color: var(--system-spectrum) gray;
    --system-express-color: var(--system-express) white;
    --color: var(--system-express-color, var(--system-spectrum-color))
}
```

As the article explains, given the following CSS is present (exists inside of `components/tokens/custom.css`:

```css
:root {
  --ON: ;
  --OFF: inherit;
  --system-spectrum: var(--OFF);
  --system-express: var(--OFF);
}

.spectrum {
  --system-spectrum: var(--ON);
  --system: spectrum;
}

.spectrum--express {
  --system-express: var(--ON);
  --system: express;
}
```

This effectively allows the value of the custom properties to be toggled by the presence of the `--system-spectrum` and `--system-express` properties (!). It doesn't _actually_ use the `--system` custom property _yet_, but once Container Queries land, we can remove `postcss-containerizer` from the pipeline and things will *just work*!

## New builder

This PR adds a new builder called `@spectrum-css/component-builder-simple` that's way faster and doesn't have dependencies or processes that use Spectrum DNA. This builder is separate from `@spectrum-css/component-builder`, so it does not affect packages that have not been migrated to core tokens.

Changes include:

* DNA tokens not allowed
* Undefined custom properties cannot be used (must be defined in Core tokens or in the component itself or the CSS it imports)
* All custom properties defined within a component must be used or an error is thrown
* Tons less PostCSS plugins
* Includes `postcss-containerizer`
* Logical properties are no longer transformed, uses native support for `dir()` and `padding-inline` etc

## Is this invalid CSS?

According to the latest W3 draft, [it's explicitly called out as being valid](https://www.w3.org/TR/css-variables-1/#guaranteed-invalid):

> actually writing an empty value into a custom property, like `--foo: ;`, is a valid (empty) value

This [css-wg issue](https://github.com/w3c/csswg-drafts/issues/3169) surrounding using a whitespace only variable has some conflicting information. 

More [discussion indicating this is valid](https://github.com/w3c/csswg-drafts/issues/774).

### Pros
1. Uses clean `@container` query syntax
2. No additional post-processing required by SWC -- code in `dist/index.css` works with web components and shadow DOM
3. No workarounds required for overrides such as `--custom-actionbutton-static-content-color`, they just work
4. Custom properties for each component are only included if the component is used

### Cons
1. Uses the var toggle hack, and the definition of `--ON` does not validate
2. Hard to debug as custom properties with names that never appear in source are involved
3. Syntax for var toggle hack is confusing and looks foreign if you're unfamiliar

## To-do list

- [x] More tests
- [x] Support more than 2 design systems
- [x] Even more tests
- [x] README for new builder
- [ ] Docs on using `@container`
- [x] Implement new, tiny build system using this
- [x] Use new build system and `postcss-containerizer` in ActionButton to prove concept
- [x] Validate approach with team
- [x] Validate result with SWC


lazd:
Closing this PR in favor of #1441

PR #1441, opened 2022-05-25
feat: add PostCSS plugin to transform container queries to flat variables

lazd:
## Description

This PR implements the second discussed option: turn container queries into flat variables using a new plugin `postcss-splitinator`.

```css
@container(--system: spectrum) {
  .component {
    --background-color: blue;
  }
  .component.is-selected {
    --background-color: darkblue;
  }
  .component .icon {
    --color: gray;
  }
}

@container(--system: express) {
  .component {
    --background-color: purple;
  }
  .component.is-selected {
    --background-color: darkpurple;
  }
  .component .icon {
    --color: white;
  }
}

```

Into this:

```css
.spectrum {
  --system-component-background-color: blue;
  --system-component-selected-background-color: darkblue;
}

.spectrum--express {
  --system-component-background-color: blue;
  --system-component-selected-background-color: darkblue;
}

.component {
  --background-color: var(--system-component-background-color);
}

.component.is-selected {
  --background-color: var(--system-component-selected-background-color);
}
```

## Caveats

~Note that this breaks the usage of `--custom-actionbutton-static-content-color` since the custom property that references it is defined at the `.spectrum`/`.spectrum--express` level instead of the `.spectrum-ActionButton` level, so we'll need to either have a workaround for that in code or modify the plugin to be aware of fallbacks.~ This was fixed, but the fix makes assumptions and is likely brittle.

### Pros
1. No hacks involved
2. Uses clean `@container` query syntax
3. Will have less custom properties than today as SWC is only shipping custom properties for components they implement

### Cons
1. Additional processing required for SWC: must extract `.spectrum` and `.spectrum--express` classes from the resulting `dist/index.css` file and up-level them to globals
2. Mechanism to address `--custom-actionbutton-static-content-color` is likely brittle, doesn't handle all possible cases
4. Hard to debug as custom properties with names that never appear in source are involved
5. Does not support `:where()` selectors as-is (makes tons of assumptions about how selectors are named in order to generate flat custom property names)
6. Simplest approach requires including all variables for all components

## To-do list
- [x] Address issue with `--custom-actionbutton-static-content-color` not working
- [x] Review with SWC
- [x] Review with team and decide
- [x] Break out `.spectrum {}` and `.spectrum--express` blocks as separate files, combine and de-dupe Spectrum and Express custom propeties
- [ ] Someone should add tests for `postcss-splitinator`'s options


components/tokens/custom.css
```diff
@@ -11,9 +11,24 @@ governing permissions and limitations under the License.
...
 .spectrum {
   --spectrum-animation-duration-100: 130ms;
+
+  --spectrum-font-family-base: adobe-clean, 'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Ubuntu, 'Trebuchet MS', 'Lucida Grande', sans-serif;
+  --spectrum-font-family-serif: adobe-clean-serif, 'Source Serif Pro', Georgia, serif;
+  --spectrum-font-family-code: 'Source Code Pro', Monaco, monospace;
+
+  --spectrum-line-height-large: 1.7;
+  --spectrum-line-height-medium: 1.5;
+  --spectrum-line-height-small: 1.3;
```
lazd, line 31: This typographical information is missing from core tokens, I've added it here otherwise the new builder complains because we were getting these tokens from DNA.

tools/component-builder-simple/css/plugins/postcss-transform-logical.js
```diff
@@ -0,0 +1,86 @@
...
+
+              // Remove original declaration
+              decl.remove();
+            }
+          }
+        }
+      });
+    });
+  }
+});
```
lazd, line 86: This not net-new code: these plugin files are all duplicated from the original builder in case we need to make changes specific to core tokens. I anticipate the original builder being deleted once we're done with core tokens migration, so I think this duplication is OK to swallow for now.

tools/component-builder-simple/index.js
```diff
@@ -0,0 +1,41 @@
...
+      }))
+      .pipe(gulp.dest('dist/'))
+  }
+);
+
+exports.default = build;
+exports.build = build;
+exports.buildLite = build;
+exports.buildMedium = build;
+exports.buildHeavy = build;
```
lazd, line 39: We may way to make `buildLite` avoid checking CSS for undefined properties so `gulp dev` works even when things are broken. The advantage is that you can debug live, the disadvantage is you don't see that bug until you try to build for release (full build).


Right now, `gulp dev` will break if there are any unused or undefined variables.

tools/postcss-splitinator/README.md
```diff
@@ -0,0 +1,80 @@
...
+### `options.noSelectors = false`
+
+Whether to avoid including selectors that use the flat variables in the output.
+
+## Usage
+
+This plugin turns this:
+
+```css
+@container(--system: standard) {
```
pfulton, line 34:
```suggestion
@container(--system: spectrum) {
```

tools/postcss-splitinator/README.md
```diff
@@ -0,0 +1,80 @@
...
+Into this:
+
+```css
+.spectrum {
+  --system-component-background-color: blue;
+  --system-component-selected-background-color: darkblue;
+}
+
+.spectrum--express {
+  --system-component-background-color: blue;
```
pfulton, line 69:
```suggestion
  --system-component-background-color: purple;
```

tools/postcss-splitinator/README.md
```diff
@@ -0,0 +1,80 @@
...
+  --system-component-selected-background-color: darkblue;
```
pfulton, line 70:
```suggestion
  --system-component-selected-background-color: darkpurple;
```

tools/postcss-splitinator/index.js
```diff
@@ -0,0 +1,116 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+const postcss = require('postcss');
+
+function getName(selector, prop) {
+  selector = selector.replace(/^:where\((.*?)\)$/, '$1');
+
+  // This regex is designed to pull spectrum-ActionButton out of a selector
```
pfulton, line 18: And for others, too, right?

pfulton (changes requested):
Just a few things here, @lazd!

tools/postcss-splitinator/index.js
lazd: Yep!

PR #1442, opened 2022-05-25
fix: resolve peerDependencies issues

pfulton:
## Description
This updates a number of components' `package.json` files, where the versions of dependencies listed as `peerDependencies` did not match the versions listed as `devDependencies`.

Another team tried to install `@spectrum-css/card` via `npm install` and encountered an "unable to resolve dependency tree" error. This error is due to how npm handles peer dependencies in versions newer than `6.x.x`. Running `npm install xxxx --legacy-peer-deps` will get around this issue, but shouldn't be relied upon in long-term scenarios. There's a good Stack Overflow post [here](https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh) with more information.

Specifically for Spectrum CSS, it appears that we have incorrect peer dependencies defined in a few of our packages which is causing this issue. This PR should fix that issue.

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/accordion/package.json
+++ b/components/accordion/package.json
@@ -16,7 +16,7 @@
     "build": "gulp"
   },
   "peerDependencies": {
-    "@spectrum-css/icon": "^3.0.11-beta.0",
+    "@spectrum-css/icon": "^3.0.21",
     "@spectrum-css/vars": "^8.0.0"
   },
   "devDependencies": {
--- a/components/actionbar/package.json
+++ b/components/actionbar/package.json
@@ -18,7 +18,7 @@
   "peerDependencies": {
     "@spectrum-css/actionbutton": "^1.1.13",
     "@spectrum-css/actiongroup": "^1.0.25",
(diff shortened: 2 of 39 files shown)


pfulton:
@lazd @GarthDB am I missing anything here? Is there any danger in bumping to these newer versions?

lazd:
I think this is good, but isn't `gulp checkPeerDependencies` supposed to do this automatically? Looks like there's a bug in there! Perhaps we should investigate why this script (which runs before every release) is missing the peer deps in this PR? https://github.com/adobe/spectrum-css/blob/main/gulpfile.js#L32-L64

pfulton:
Great thought. I'll dig into it. It would be nice to have this covered automatically and I'm suspicious as to why it didn't catch these.

pfulton:
Closed in favor of #1444

PR #1444, opened 2022-05-26
fix: use stricter comparision for peerDependency version semver checks

pfulton:
## Description
This changes the way that we check to see if a package's `peerDependencies` match (or would be 'satisfied') by the version required in the package's `devDependencies`. Our previous comparison was not properly handling prerelease version satisfaction (or, at least, it wasn't handling it in a way that we needed it to). This change updates the `satisfies` args so that they compare exact version numbers (instead of the "^" range).

More details on semver & prerelease tag satisfaction [here](https://github.com/npm/node-semver#prerelease-tags).

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/accordion/package.json
+++ b/components/accordion/package.json
@@ -16,7 +16,7 @@
     "build": "gulp"
   },
   "peerDependencies": {
-    "@spectrum-css/icon": "^3.0.11-beta.0",
+    "@spectrum-css/icon": "^3.0.21",
     "@spectrum-css/vars": "^8.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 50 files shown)


pfulton:
We can close https://github.com/adobe/spectrum-css/pull/1442 if this one is the way forward.

pfulton:
By the way, this looks like it will _always_ make the `peerDependencies` match the `devDependencies`, which I think is fine and maybe what we want anyway? We definitely need to be sure that previously prerelease `peerDependencies` are updated to match non-prerelease `devDependencies`, and this accomplishes that.

PR #1478, opened 2022-07-21
build: support live-reload for token custom value changes (CSS-225)

lazd:
## Description

Make `watch` work with Tokens

### DIFF
--- a/components/tokens/gulpfile.js
+++ b/components/tokens/gulpfile.js
@@ -31,11 +31,20 @@ function concatIndex() {
     .pipe(gulp.dest('dist/'));
 }
 
-function buildCustoms() {
-  buildCustomSpectrum = () => gulp.src('custom-spectrum/*.css').pipe(gulp.dest('dist/css/spectrum/'));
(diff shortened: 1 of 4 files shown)


pfulton (commented):
✅ the watcher is now picking up changes made to these files. Awesome!

I noticed that if I try to use a newly defined token in the custom dirs, components (I tried with ActionButton) aren't aware of them without stopping & restarting the server. Is there anything we could do to make them aware w/o having to stop and restart?

One more thing: would we be able to add the `<component>/themes/*` files to the watcher as well?

bernhard-adobe:
@lazd yeah, I think we need also the files in the `themes` folder to be watched.
Screen Shot 2022-07-21 at 14 53 34

bernhard-adobe:
Added the watching of the new themes folder for the Core-Tokens migration

bernhard-adobe:
@lazd follow up if you could answer it:
Do you have suggestions how to place the CSS files in the folder `components/commons` under watch as well?

PR #1492, opened 2022-08-05
feat(actiongroup)!: migrate to core tokens

pfulton:
## Description
Migrates ActionGroup to core tokens. Also consumes latest version of ActionButton

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/actiongroup/index.css
+++ b/components/actiongroup/index.css
@@ -44,7 +44,7 @@ governing permissions and limitations under the License.
 .spectrum-ActionGroup {
   display: flex;
   flex-wrap: wrap;
-  gap: var(--spectrum-actiongroup-horizontal-spacing-regular);
+  gap: var(--mod-actiongroup-horizontal-spacing-regular, var(--spectrum-actiongroup-horizontal-spacing-regular));
 
   .spectrum-ActionGroup-item {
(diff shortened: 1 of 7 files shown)


components/actiongroup/index.css
```diff
@@ -114,51 +108,40 @@ governing permissions and limitations under the License.
         z-index: 3;
```
lunarfusion, line 108: I think that &:focus-ring is supposed to be &:focus-visible.

It looks like the intent of the focus behavior on this component is to remove border radius from the corner of the focus ring where compact items connect.

This isn’t working because the focus ring is applied to action button as a box-shadow on the :after element. The focus ring on action button also lacks the focus ring gap, making the focus state almost invisible on emphasized buttons. 

I think the focus ring on action button should be applied as a border instead of a box shadow. Then the border radii of the compact action group items can changed per item.

bernhard-adobe (approved):

components/actiongroup/index.css
pfulton: Thanks @lunarfusion — I think this is a larger question around how we're handling focus rings, but the implementation here is correct.

pfulton:
Released beta:
`3.0.0-beta.0`

PR #1499, opened 2022-08-17
fix(breadcrumb): Update peer deps for breadcrumb

GarthDB:
## Description

Breadcrumb had the incorrect version for actionbutton it peer deps so it was throwing an error/warning.

 - Built locally

## To-do list


pfulton:
So, is this happening because we have a batch of components that have been migrated to the new tokens system, and some that haven't, and some of the ones that **have** been migrated are dependencies of those that have not yet been migrated?

castastrophe (approved):
This compiles on my machine with no errors & the docs site looks good: https://pr-1499--spectrum-css.netlify.app/docs/breadcrumb.html. I did push up the yarn.lock update which is missing from here.

castastrophe:
@pfulton @GarthDB On closer inspection, I believe the error has more to do with how peer dependencies are being defined across the repo. Typically they are more permissive and leverage the `>=` syntax to allow some flexibility in definition. This is a benefit because it will lead to less version management struggles downstream. I think we should tackle this on every component. What do you think?

castastrophe:
I think I'll close this PR out for now in favor of a wholistic solution. I created [CSS-381](https://jira.corp.adobe.com/browse/CSS-381) to track that work.

PR #1508, opened 2022-08-26
feat(menu)!: migrate to core tokens & many new features

pfulton:
 - **How this was tested:** 

## To-do list


bernhard-adobe (approved):
oh wow there is a lot going on here, especially around CSS grid stuff. 
The code looks ok, some minor comments about missing core-tokens as always.

I need to see the VRT for this to get a better judgment.

bernhard-adobe:
@pfulton maybe we could add more screenshots to our PRs. I am running VRT here: https://spectrum-visual-regression.ci.corp.adobe.com/view/Spectrum%20CSS/job/css-vrt-test/43/

components/menu/index.css
```diff
@@ -10,170 +10,661 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  .spectrum-Menu--sizeL {
+    --spectrum-menu-item-selectable-edge-to-text-not-selected: 47px;
+  }
+  .spectrum-Menu--sizeXL {
+    --spectrum-menu-item-selectable-edge-to-text-not-selected: 54px;
+  }
+}
+
+.spectrum--light {
+  --spectrum-menu-item-background-color-default: rgba(0, 0, 0, 0); /* --spectrum-gray-900 */
```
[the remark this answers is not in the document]
pfulton: Great call-out! I actually dug into this a bit [in Slack](https://adobedesign.slack.com/archives/G0143NJRLKY/p1662563496552649), but, the short story is: we need token definitions at the `100` level for `transparent-white` and `transparent-black` which would provide a `0%` in the alpha.

Right now, I'd have to try to combine two tokens together, BUT, if I do so, the resulting rule isn't valid CSS. For instance:

```
--spectrum-menu-item-background-color: --spectrum-gray-900; /* gray-900 === rgb(0,0,0) */
--spectrum-menu-item-background-opacity-hover: 10%;
...

background-color: rgba(var(--spectrum-menu-item-background-color) var(--spectrum-menu-item-opacity-hover));
```

The above ends up resolving to `rgba(rgb(0, 0, 0) 10%)` unfortunately. Sooooo, until we get that token, we shall hard-code.

yosevu:
I only see this padding issue in the deployed PR so I'm not sure if it is env specific. It doesn't appear to be relevant to this work, but I'm sharing the screenshot just in case

Screen Shot 2022-09-22 at 2 10 53 PM

yosevu:
This might be an issue on main since it doesn't appear to be relevant to this work. Capturing the screenshot until I find the source of the issue:

Screen Shot 2022-09-22 at 2 10 53 PM

yosevu:
Is it worth considering other spacing options to justify content when there are only two items so that there isn't such a large gap?

Screen Shot 2022-09-22 at 2 39 33 PM

yosevu:
I noticed there isn't a border around the menu or submenu and I don't see tokens in the designs, but it seems like they would give a more realistic context to the variant examples. Is this out of scope?

Screen Shot 2022-09-22 at 2 55 02 PM
Screen Shot 2022-09-22 at 2 54 27 PM

yosevu:
I think a lot of my questions so far are scope related, so I will list them here:

- Are switch and multiple selection options in scope for this work?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+        --highcontrast-menu-item-background-color-hover: Highlight;
+        --highcontrast-menu-item-background-color-down: Highlight;
+        --highcontrast-menu-item-background-color-key-focus: Highlight;
+
+        --highcontrast-menu-item-value-color-hover: HighlightText;
+        --highcontrast-menu-item-value-color-down: HighlightText;
+        --highcontrast-menu-item-value-color-focus: HighlightText;
+      }
+      
+      &:focus-ring,
```
yosevu, line 272: Yes, there is a bug causing the `spectrum-Menu-itemLabel` text color to resolve as black when the `focus-ring` class is applied.

`focus-ring` can also be replaced with `:focus-visible` [source](https://lists.w3.org/Archives/Public/www-style/2017Dec/0018.html)

Screen Shot 2022-09-22 at 3 23 59 PM

Screen Shot 2022-09-22 at 3 28 41 PM

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  --spectrum-menu-drillin-icon-color-down: var(--spectrum-neutral-subdued-content-color-down);
+  --spectrum-menu-drillin-icon-color-focus: var(--spectrum-neutral-subdued-content-color-key-focus);
+
+  --spectrum-menu-item-value-color-default: var(--spectrum-neutral-subdued-content-color-default);
+  --spectrum-menu-item-value-color-hover: var(--spectrum-neutral-subdued-content-color-hover);
+  --spectrum-menu-item-value-color-down: var(--spectrum-neutral-subdued-content-color-down);
+  --spectrum-menu-item-value-color-focus: var(--spectrum-neutral-subdued-content-color-key-focus);
+
+  /* "one" icon: chevron + additional icon (we don't count the chevron as an icon because it HAS to be there for a collapisble) */
+  --spectrum-menu-item-collapisble-has-icon-submenu-item-padding-x-start:
```
yosevu, line 209: These comments and property names are very helpful for understanding the calc! 👏  Taking notes ✍️

components/menu/index.css
```diff
@@ -182,35 +688,120 @@ governing permissions and limitations under the License.
...
+    .spectrum-Menu-chevron {
+      transform: rotate(90deg);
+    }
+  }
+
+  &.is-open:hover,
+  &.is-open:active,
+  &.is-open:focus,
+  &.is-open.is-focused {
+    background-color: var(--highcontrast-menu-item-background-color-default, var(--mod-menu-item-background-color-default, var(--spectrum-menu-item-background-color-default)));
```
yosevu, line 714: This background color changes in WHCM. What about using a `Background` or transparent for the background-color and `ButtonText` for the text color?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 }
 
 .spectrum-Menu {
-  @remapvars {
-    find: /--spectrum-listitem-m(.*)/;
-    filter: color;
-    replace: --spectrum-listitem$1;
+  --spectrum-menu-item-label-line-height: var(--spectrum-line-height-100);
+  --spectrum-menu-item-label-cjk-line-height: var(--spectrum-cjk-line-height-100);
+  --spectrum-menu-item-label-to-description-spacing: 1px;
```
yosevu, line 160: Should this either have a token or go in the Dropbox Paper doc?

components/menu/index.css
```diff
@@ -10,170 +10,661 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
   margin: 0;
+  text-decoration: none;
+}
 
-  border-inline-start: var(--spectrum-listitem-texticon-focus-indicator-size) solid
-    transparent;
+.spectrum-Menu-item {
+  display: grid;
+  grid-template-areas:
+    "selectedArea chevronAreaCollapsible checkmarkArea iconArea labelArea        valueArea actionsArea chevronAreaDrillIn"
```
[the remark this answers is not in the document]
yosevu: 💯 A great use case for CSS Grid!

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+
+.spectrum-Menu-checkmark {
+  grid-area: checkmarkArea;
+}
+
+/* menus with "selectable" menuitems */
+.spectrum-Menu.is-selectable {
+  /* provide inline-start spacing for potential checkmarks */
+  .spectrum-Menu-item {
+    padding-inline-start: var(--spectrum-menu-item-selectable-edge-to-text-not-selected);
```
yosevu, line 562: Does this need a `--mod` override token?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  }
+
+  .spectrum-Menu-checkmark {
+    display: none;
+  }
+
+  /* remove inline-start spacing once an item is selected and a checkmark appears */
+  .spectrum-Menu-item.is-selected {
+    padding-inline-start: var(--spectrum-menu-item-label-inline-edge-to-content);
```
yosevu, line 571: Does this need a `--mod` override token?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+.spectrum-Menu-chevron {
+  grid-area: chevronArea;
+  align-self: center;
+  block-size: var(--spectrum-menu-item-checkmark-height);
+  height: var(--spectrum-menu-item-checkmark-height);
+  inline-size: var(--spectrum-menu-item-checkmark-width);
+  width: var(--spectrum-menu-item-checkmark-width);
+}
+
+/* TODO: double check this. Is this necessary? An error? */
```
yosevu, line 606: What may not be necessary?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  .spectrum-Menu-chevron {
+    fill: var(--highcontrast-menu-collapsible-icon-color, var(--mod-menu-collapsible-icon-color, var(--spectrum-menu-collapsible-icon-color)));
+  }
+
+  .spectrum-Menu-checkmark {
+    fill: var(--highcontrast-menu-checkmark-icon-color-focus, var(--mod-menu-checkmark-icon-color-focus, var(--spectrum-menu-checkmark-icon-color-focus)));
+  }
+}
+
+.spectrum-Menu-item:focus-ring,
```
lunarfusion, line 420: :focus-ring  is depreciated, it's not giving the 10% black background color like it should:
menu item with focus-ring
_[fig 1]_

we can use `.spectrum-Menu-item:focus-visible` which gives me this (10% black background)
In Chrome:
menu item with focus-visible
_[fig 2]_

In Firefox:
Screen Shot 2022-09-23 at 4 04 49 PM
_[fig 3]_

The background is correct with :focus-visible, but the result does not match the design as shown at https://spectrum.adobe.com/page/menu/
menu design
_[fig 4]_

The incorrect blue outline we are seeing in fig2 and fig3 is the result of the outline being removed from the :focus state on line 389 but not from the :focus-visible state.

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  text-decoration: none;
+.spectrum-Menu-item .spectrum-Menu-item .spectrum-Menu-itemLabel {
+  grid-area: submenuItemLabelArea;
+}
+
+.spectrum-Menu-item .spectrum-Menu-item .spectrum-Menu-itemValue {
+  grid-area: submenuItemValueArea;
+}
+
+.spectrum-Menu-item:focus {
```
lunarfusion, line 388: Lines 388-418 appear to be redundant with lines 420-452. Suggest adding .spectrum-Men-item:focus to line 420.

But I'd check with engineering to make sure that controlling :focus in the CSS won't interfere with their work. @GarthDB can you chime in on this? Should we keep :focus on menu items or remove it?

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  .spectrum-Menu-checkmark {
+    fill: var(--highcontrast-menu-checkmark-icon-color-focus, var(--mod-menu-checkmark-icon-color-focus, var(--spectrum-menu-checkmark-icon-color-focus)));
+  }
+}
+
+.spectrum-Menu-item:focus-ring,
+.spectrum-Menu-item.is-focused {
+  background-color: var(--highcontrast-menu-item-background-color-key-focus, var(--mod-menu-item-background-color-key-focus, var(--spectrum-menu-item-background-color-key-focus)));
+  border-color: var(--highcontrast-menu-item-focus-indicator-color, var(--mod-menu-item-focus-indicator-color, var(--spectrum-menu-item-focus-indicator-color)));
+
```
lunarfusion, line 424: Needs
`outline: none;`

Combined with adding :focus-visible as noted above, this gives me (same results in Chrome and Firefox):
Screen Shot 2022-09-23 at 4 23 19 PM

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  }
+
+  .spectrum-Menu-checkmark {
+    fill: var(--highcontrast-menu-checkmark-icon-color-focus, var(--mod-menu-checkmark-icon-color-focus, var(--spectrum-menu-checkmark-icon-color-focus)));
+  }
+}
+
+[dir="rtl"] .spectrum-Menu-item:focus-ring,
+[dir="rtl"] .spectrum-Menu-item.is-focused {
+  box-shadow: inset calc(var(--mod-menu-item-focus-indicator-width, var(--spectrum-menu-item-focus-indicator-width)) * -1) 0 0 var(--highcontrast-menu-item-focus-indicator-color, var(--mod-menu-item-focus-indicator-color, var(--spectrum-menu-item-focus-indicator-color)));
```
lunarfusion, line 456: Appears to be a redundancy of line 349. Seems to be causing the blue start border to be too thick.
Screen Shot 2022-09-23 at 4 35 08 PM

lunarfusion (changes requested):
Noticed some issues with focus and chevron icons.

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  --spectrum-menu-item-collapisble-has-icon-submenu-item-padding-x-start:
+    calc((
+      var(--spectrum-menu-item-label-inline-edge-to-content)
+        + var(--spectrum-menu-item-checkmark-width)
+        + var(--spectrum-menu-item-text-to-control)
+        + var(--spectrum-menu-item-icon-width)
+        + var(--spectrum-menu-item-label-text-to-visual)
+    ));
+
+  /* "no" icon: just the chevron (we're not counting it because it HAS to be there for a collapisble) */
```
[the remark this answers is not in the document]
pfulton: I must have copy/pasted the misspelling! 'Doh!

components/menu/index.css
pfulton: I think I left this in from an earlier commit when I was taking notes for myself. Removed!

components/menu/index.css
pfulton: Yep, looks like I missed the `--mod`s. I'll add them now.

components/menu/index.css
pfulton: Oh, you're right! Thank you.

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  /* TODO: eventual token name: */
+  /* --spectrum-menu-item-selectable-edge-to-text-not-selected: var(--spectrum-menu-item-edge-to-text-not-selected-extra-large); */
+  /* TODO: remove below once above is defined */
+  --spectrum-menu-item-selectable-edge-to-text-not-selected: 45px;
+
+  --spectrum-menu-item-checkmark-height: 14px;
+  --spectrum-menu-item-checkmark-width: 14px;
+}
+
+/* TODO: temporary & can be removed once the tokens are defined in the core tokens package */
```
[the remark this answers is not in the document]
pfulton: The `TODO` refers to the `--large` stuff as well as the stuff for the `--light`, etc. If you didn't already know, and you need to hard-code things while awaiting the real tokens, these are the class names you can use/this is how you'd do it.

components/menu/index.css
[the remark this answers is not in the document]
bernhard-adobe: Thanks for calling out the issue Patrick. Then hard-coding them for now to avoid those combination issues might be the right choice. 
I can mark this as resolved for now.

components/menu/index.css
pfulton: Woah, wild. I did not see this happening when I worked on it. I've made some changes to this focus state, so hopefully this is fixed now.

components/menu/index.css
pfulton: Yeah those were almost completely redundant. I consolidated.

components/menu/index.css
```diff
@@ -10,170 +10,676 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
     overflow: visible;
-
     inline-size: auto;
-    margin-block: calc(var(--spectrum-listitem-texticon-divider-padding) / 2);
-    margin-inline: var(--spectrum-listitem-texticon-padding-y);
+    margin-block: var(--mod-menu-item-bottom-edge-to-text, var(--spectrum-menu-item-bottom-edge-to-text));
   }
 }
 
+.spectrum-Menu-itemIcon {
```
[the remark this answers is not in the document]
pfulton: Thank you for pointing this out. I needed to account for the scale size changes between `.spectrum--medium` and `.spectrum--large` for these.

components/menu/index.css
pfulton: Yes, thank you for this! Done.

components/menu/index.css
pfulton: Thanks! I fixed these up.

pfulton:
I've fixed up all of the call-outs here, and I've also done some fixing up elsewhere in the docs site where these changes were affecting things. I think we're ready for a beta release.

pfulton:
Released: `@spectrum-css/menu@5.0.0-beta.0`

hunterloftis:
Here are those examples I promised of CSS that's challenging to transmogrify into a web component. I'm pulling these from [this integration branch](https://github.com/adobe/spectrum-web-components/compare/hloftis/update-cssmenu-beta) that I'm currently working on.

## Theming rules where tokens are specified in the component's CSS rather than the theme's

https://github.com/adobe/spectrum-css/blob/f9f5e0b1ef1ad7c90895eb936f83b5931382b43e/components/menu/index.css#L162-L175

These rules end up disappearing with our current import tools. I think we could work around this by looking for specific selectors (like `.spectrum--light`) in the component CSS, and then applying them instead to `sp-theme`.

## State that depends on an ancestor

https://github.com/adobe/spectrum-css/blob/f9f5e0b1ef1ad7c90895eb936f83b5931382b43e/components/menu/index.css#L555-L574

This gets compiled to `.spectrum-Menu.is-selectable .spectrum-Menu-item.is-selected .spectrum-Menu-checkmark`. In "vanilla" HTML & CSS, that relationship is maintained, but in a web component that breaks over shadow boundaries. Since the component is encapsulated, it can't tell if the checkmark `<sp-icon>` is contained within an `<sp-menu-item>` that is contained within a selectable `<sp-menu>`.

Our workaround for that is to [recreate the rule at the level where the state is available](https://github.com/adobe/spectrum-web-components/compare/hloftis/update-cssmenu-beta#diff-5ed00f38f2e20e3cf95d2b9840c848ab2da84a83546bf94daccfdbdb126acaafR57-R67), as a CSS var, which gets consumed inside the component.

## Selectors that depend on DOM relationships

https://github.com/adobe/spectrum-css/blob/f9f5e0b1ef1ad7c90895eb936f83b5931382b43e/components/menu/index.css#L656-L663

We try to match the structure, but it sometimes has to be slightly different in a web component. These are the easiest to work around though; I can either update our DOM to match, or I can update rules to be looser [like this](https://github.com/adobe/spectrum-web-components/compare/hloftis/update-cssmenu-beta#diff-70cded925531a1266c56f9c8af62eaa2fb073ed68136bebd78e478c4d94d9bc3R100).

---

I don't know that we _need_ to do any of these differently - we can work around all of them to one degree or another. But if there are equally-valid approaches at the CSS level that would translate more smoothly to a web component context, that would be awesome!

lazd:
Stoked to see this one land, we're doing awful things over in Project X land to customize menu and need the `--mod-*` bits here to do it right.

pfulton:
@lazd Very 🔜 !!

pfulton:
Released updated beta: `@spectrum-css/menu@5.0.0-beta.1`

components/menu/index.css
```diff
@@ -10,170 +10,623 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 
-  border-inline-start: var(--spectrum-listitem-texticon-focus-indicator-size) solid
-    transparent;
+.spectrum-Menu-item,
+.spectrum-Menu li:not(.spectrum-Menu-item, .spectrum-Menu-divider) {
+  display: grid;
+  grid-template-areas:
+    ".            chevronAreaCollapsible .             iconArea sectionHeadingArea  .         .           .                 "
+    "selectedArea chevronAreaCollapsible checkmarkArea iconArea labelArea        valueArea actionsArea chevronAreaDrillIn"
+    ".            .                      .             .        descriptionArea  .         .           .                 "
```
Westbrook, line 334:
```suggestion
    ".            .                      .             iconArea       descriptionArea  .         .           .                 "
```
This would allow for layouts like (ignoring the misplaced check mark):
image

pfulton:
New beta version is: `@spectrum-css/menu@5.0.0-beta.2`

components/menu/index.css
```diff
@@ -10,170 +10,637 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  transform: scale(1);
+  align-self: center;
+  display: var(--mod-menu-checkmark-display, var(--spectrum-menu-checkmark-display));
+  fill: var(--highcontrast-menu-checkmark-icon-color-default, var(--mod-menu-checkmark-icon-color-default, var(--spectrum-menu-checkmark-icon-color-default)));
   opacity: 1;
 }
 
-.spectrum-Menu-item {
+.spectrum-Menu-item,
+.spectrum-Menu li:not(.spectrum-Menu-item, .spectrum-Menu-divider) {
```
najikahalsema, line 338: Should `:not(.spectrum-Menu-item)` really be getting the same styles as `.spectrum-Menu-item` in this case?

components/menu/index.css
```diff
@@ -10,170 +10,637 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  padding-inline: var(--mod-menu-item-label-inline-edge-to-content, var(--spectrum-menu-item-label-inline-edge-to-content));
+  
   margin: 0;
+  text-decoration: none;
+}
 
-  border-inline-start: var(--spectrum-listitem-texticon-focus-indicator-size) solid
-    transparent;
+.spectrum-Menu-item,
+.spectrum-Menu li:not(.spectrum-Menu-item, .spectrum-Menu-divider) {
```
najikahalsema, line 359: same as previous comment. Should `:not(.spectrum-Menu-item)` have the same rules as `.spectrum-Menu-item`?

components/menu/index.css
```diff
@@ -10,170 +10,637 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+.spectrum-Menu-item--collapsible .spectrum-Menu-sectionHeading {
+  /*
+    start at 5th column, to account for icons & match sectionHeadingArea
+    ". chevronAreaCollapsible . iconArea sectionHeadingArea  .  .  ."
+    stretch to last column
+  */
+  grid-column: 5 / -1;
+}
+
+.spectrum-Menu li:not(.spectrum-Menu-item, .spectrum-Menu-divider) > .spectrum-Menu-sectionHeading {
```
najikahalsema, line 572: Are there times when `.spectrum-Menu-sectionHeading` shouldn't have this padding?

components/menu/index.css
najikahalsema: The rules here for `padding-block-start` and `end` make sense, but I'm not sure about the other rules. It's adding the inline-padding twice if you have a menu heading.

najikahalsema (changes requested):

jawinn:
This work is continued in the newer PR #1942

PR #1513, opened 2022-09-12
fix(tabs): selection indicator scroll overflow border

yosevu:
## Description

Resolves an issue with the selection indicator border when there is scrollable overflow.

Updated with the recommended solution using the background linear gradient rather than background-color.

Set `.spectrum-Tabs` width to less than the widths of its `.spectrum-Tab-item` children combined e.g. `200px`. Set `.spectrum-Tabs` to `overflow: auto`.  See examples of the bug and fix below.

#### Bug
https://user-images.githubusercontent.com/16216104/194005007-e61b01ee-3f0d-4563-b2c4-a127217171a0.mp4

#### Fixed

https://user-images.githubusercontent.com/16216104/194003987-77d9c0f5-9aad-49f2-b57a-229ef2226e51.mp4

## To-do list

### DIFF
--- a/components/tabs/skin.css
+++ b/components/tabs/skin.css
@@ -11,11 +11,27 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-Tabs {
-  border-block-end-color: var(--spectrum-tabs-textonly-divider-background-color);
+  --spectrum-tabs-list-background-direction: top;
+
+  background: linear-gradient(
+    to var(--spectrum-tabs-list-background-direction),
+    var(--spectrum-tabs-textonly-divider-background-color) 0 var(--spectrum-tabs-quiet-textonly-divider-size),
+    transparent var(--spectrum-tabs-quiet-textonly-divider-size)
+  );
 }
 
 .spectrum-Tabs--vertical {
-  border-inline-start-color: var(--spectrum-tabs-vertical-textonly-divider-background-color);
+  --spectrum-tabs-list-background-direction: right;
+}
+
+.spectrum-Tabs--quiet,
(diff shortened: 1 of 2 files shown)


components/tabs/skin.css
```diff
@@ -19,12 +19,26 @@ governing permissions and limitations under the License.
...
 .spectrum-Tabs-selectionIndicator {
-  background-color: var(--spectrum-tabs-textonly-tabitem-selection-indicator-background-color-selected);
+  --spectrum-tabs-list-background-direction: right;
+
+  background: linear-gradient(
+    to var(--spectrum-tabs-list-background-direction),
+    var(--spectrum-tabs-textonly-tabitem-selection-indicator-background-color-selected) 0,
+    var(--spectrum-tabs-textonly-tabitem-selection-indicator-background-color-selected)
+    var(--spectrum-tabs-quiet-textonly-divider-size),
+    transparent var(--spectrum-tabs-quiet-textonly-divider-size)
```
yosevu, line 29: What token should I use in place of `--spectrum-tabs-quiet-textonly-divider-size`?

The widths are currently set as an inline styles, so I'm not sure the purpose of this token, which is only 2px.

Screen Shot 2022-09-12 at 9 02 08 AM

Screen Shot 2022-09-12 at 9 01 52 AM

yosevu: Thanks Patrick, I misunderstood how the width is was supposed to work and you helped clarify that.

I updated the PR and it looks correct now. Can I reproduce the issue in the the PR link to make sure this works as expected?

pfulton (approved):

pfulton (approved):
Awesome. Looks great to me. Thanks for taking care of this.

PR #1514, opened 2022-09-12
refactor(buttongroup)!: migrate to core tokens

yosevu:
## Description

Migrate Button Group to use core tokens.

 - Chrome
 - Safari
 - WHCM emulation

## Screenshots
Screen Shot 2022-09-12 at 8 54 14 AM

## To-do list

### DIFF
--- a/components/buttongroup/index.css
+++ b/components/buttongroup/index.css
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adobe. All rights reserved.
+Copyright 2022 Adobe. All rights reserved.
 This file is licensed to you under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License. You may obtain a copy
 of the License at http://www.apache.org/licenses/LICENSE-2.0
@@ -11,7 +11,28 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-ButtonGroup {
-  --spectrum-buttongroup-button-gap-reset: 0;
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
+}
+
+.spectrum-ButtonGroup--sizeS {
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-200);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-200);
+}
+
+.spectrum-ButtonGroup--sizeM {
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
+}
+
+.spectrum-ButtonGroup--sizeL {
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
+}
+
+.spectrum-ButtonGroup--sizeXL {
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
 }
 
 .spectrum-ButtonGroup {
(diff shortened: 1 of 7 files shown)


components/buttongroup/index.css
```diff
@@ -11,7 +11,28 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-ButtonGroup {
-  --spectrum-buttongroup-button-gap-reset: 0;
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
```
lunarfusion, line 14: Should we update the YAML to show the t-shirt sizes and help confirm correct spacing for them?

lunarfusion (commented):
This looks great! Just one question about YAML examples.

components/buttongroup/index.css
yosevu: I added variants to show the different between M, L, XL and S. Is this what you had in mind?

lunarfusion: Yes! But the buttons within the group need to be changed to --sizeS also.

Good call to show m/l/xl together since their spacing is the same.

yosevu: Ah, thank you. Updated.

components/buttongroup/metadata/buttongroup.yml
```diff
@@ -30,3 +51,20 @@ examples:
           <span class="spectrum-Button-label">Rate now</span>
         </button>
       </div>
+
+  - id: buttongroup-vertical-small
+    name: Vertical - Small
+    description: |
+      Spacing for the Small t-shirt size.
+    markup: |
+      <div class="spectrum-ButtonGroup--sizeS spectrum-ButtonGroup--vertical">
```
lunarfusion, line 60: line 60 should be 
`<div class="spectrum-ButtonGroup spectrum-ButtonGroup--sizeS spectrum-ButtonGroup--vertical">`

(missing spectrum-ButtonGroup base class)

components/buttongroup/metadata/buttongroup.yml
```diff
@@ -16,8 +18,27 @@ examples:
         </button>
       </div>
 
-  - id: buttongroup
+  - id: buttongroup-horizontal-small
+    name: Horizontal - Small
+    description: |
+      Spacing for the Small t-shirt size.
+    markup: |
+      <div class="spectrum-ButtonGroup--sizeS">
```
lunarfusion, line 26: line 26 should be 
`<div class="spectrum-ButtonGroup spectrum-ButtonGroup--sizeS">`

(missing spectrum-ButtonGroup base class)

components/buttongroup/index.css
```diff
@@ -11,7 +11,28 @@ governing permissions and limitations under the License.
...
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
+}
+
+.spectrum-ButtonGroup--sizeXL {
+  --spectrum-buttongroup-spacing-horizontal: var(--spectrum-spacing-300);
+  --spectrum-buttongroup-spacing-vertical: var(--spectrum-spacing-300);
 }
 
 .spectrum-ButtonGroup {
```
lunarfusion, line 38: Line 40 - I think we might need **flex-wrap: wrap** here because worse case scenerio could have horizontal overflow:
Screen Shot 2022-09-27 at 4 33 46 PM

lunarfusion (changes requested):
Sorry I did not catch this before.

components/buttongroup/metadata/buttongroup.yml
yosevu: 🙈 Oops, thank you!

components/buttongroup/index.css
yosevu: It looks like I will end up with something similar to the Vertical variant if I use `flex-wrap` and update the `inline` and `block` margins.

What do you think about using vertical variant styles as the default and applying the horizontal styles at this breakpoint instead?

Screen Shot 2022-09-27 at 4 05 43 PM

yosevu: Discussed with Monet and I think I was overcomplicating it thinking about layout changes.

It looks like I just need to use `gap` with `flex-wrap`:

Screen Shot 2022-09-28 at 9 52 43 AM

lunarfusion (approved):
Looks good!

PR #1535, opened 2022-10-24
refactor(tray): use spectrum tokens

yosevu:
## Description

Refactor Tray to use spectrum tokens.

Remove overlay related styles as these will be likely be included as a part of Modal. See [this conversation](https://adobedesign.slack.com/archives/CS22GNRU2/p1667504713526569) for context.

- This work removes the 375px breakpoint, which was previously used to apply a max-width and border radius to the Tray.
- Instead this uses [an orientation media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) to apply these styles viewport is in landscape orientation or when the width is greater than the height.
- This can be tested by resizing the screen until the viewport width is greater than the height.

See https://spectrum.adobe.com/page/tray for information about maximum height safe zone and orientation requirements.

### Open and close tray by toggling the `is-open` class.

Screenshot 2022-10-31 at 8 16 32 PM

 - **How this was tested:** 

### Landscape
Screenshot 2022-11-07 at 7 11 31 PM
### Portrait
Screenshot 2022-11-07 at 7 11 18 PM

## To-do list


components/commons/overlay-coretokens.css
```diff
@@ -0,0 +1,54 @@
+/*
```
yosevu, line 1: I copied the overlay changes from #1509. Can we create a separate PR to add them since more than one component depends on overaly?

components/tray/metadata/tray.yml
```diff
@@ -1,13 +1,19 @@
 name: Tray
-description: Tray Dialogs are typically used to portray information on mobile device or smaller screens.
+description: |
```
yosevu, line 2: Should I include information about portrait, landscape orientation and the safe zone here?

https://spectrum.adobe.com/page/tray/

components/tray/package.json
```diff
@@ -19,20 +19,20 @@
     "@spectrum-css/button": "^6.0.0",
     "@spectrum-css/icon": "^3.0.21",
     "@spectrum-css/modal": "^3.0.20",
-    "@spectrum-css/vars": "^8.0.0"
+    "@spectrum-css/tokens": "^4.0.0"
   },
   "devDependencies": {
     "@spectrum-css/button": "^6.0.13",
-    "@spectrum-css/component-builder": "^3.2.0",
+    "@spectrum-css/component-builder-simple": "^1.0.0",
```
pfulton, line 26:
```suggestion
    "@spectrum-css/component-builder-simple": "^2.0.0",
```

components/tray/package.json
```diff
@@ -19,20 +19,20 @@
...
     "@spectrum-css/dialog": "^6.0.15",
     "@spectrum-css/divider": "^1.0.27",
```
pfulton, line 28:
```suggestion
    "@spectrum-css/divider": "^2.0.1",
```

components/tray/index.css
```diff
@@ -10,70 +10,96 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  inset-inline-start: 0;
+  /* Fixed to the bottom of the viewport */
+  position: fixed;
+  bottom: 0;
+
+  display: flex;
+  justify-content: center;
+  inline-size: 100%;
+
+  --spectrum-tray-overlay-background-color: var(--spectrum-overlay-color);
```
pfulton, line 103: Can you move this custom property definition from here to being defined under `.spectrum-Tray`? For a consumer to customize this in SWC, it needs to be defined at the root.

pfulton (changes requested):
Just a few things here, @yosevu - this is looking really great.

components/commons/overlay-coretokens.css
bernhard-adobe: Should Overlay than be it's own component like Underlay?: https://github.com/adobe/spectrum-css/blob/main/components/underlay/index.css

components/tokens/custom-spectrum/custom-vars.css
```diff
@@ -18,6 +18,7 @@ governing permissions and limitations under the License.
 .spectrum {
   --spectrum-animation-duration-100: 130ms;
   --spectrum-animation-duration-200: 160ms;
+  --spectrum-animation-duration-500: 250ms;
```
bernhard-adobe, line 21: We will have to do some refactoring once we have those tokens in the package. 
We used those animation tokens very often.

components/tray/index.css
```diff
@@ -10,70 +10,96 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 .spectrum-Tray {
-  /* Bug: this must be 0ms, not 0 */
-  --spectrum-dialog-confirm-exit-animation-delay: 0ms;
-  --spectrum-tray-margin-top: 64px;
+  /* Placeholder tokens */
+  --spectrum-tray-exit-animation-delay: 0ms;
+  --spectrum-tray-entry-animation-delay: 160ms;
+
+  /* Maximum width for landscape */
+  --spectrum-tray-max-inline-size: 375px;
```
bernhard-adobe, line 21: Is there a way to soft-code this value?

components/tray/metadata/tray.yml
bernhard-adobe: I could see examples of this being helpful. Also for VRT.  I think Web Components could also benefit from displaying these cases https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/tray--default

yarn.lock
```diff
@@ -1468,6 +1468,33 @@
   resolved "https://registry.yarnpkg.com/@spectrum-css/closebutton/-/closebutton-1.2.13.tgz#a02a33e211349138bcde0988cd4317749a7618e4"
   integrity sha512-V/P/XVWPh/TR2Qo66vn6o+vi5/utDbrMIdNV+zuF1dbl1wxH5QMOdCQrQOtzHrzcwulJRu3+Sk759VRPojLYCA==
 
+"@spectrum-css/component-builder-simple@^1.0.0":
```
bernhard-adobe, line 1471: I think this will be removed when you install the component builder in v2.

bernhard-adobe (changes requested):

components/commons/overlay-coretokens.css
yosevu: Hmm, I'm not sure, but perhaps this could be done separately to not impact Popover in #1509.

bernhard-adobe: This would be the easiest way to share a component / same code across multiple components if I understand your comment correct. 

> I copied the overlay changes from https://github.com/adobe/spectrum-css/pull/1509. Can we create a separate PR to add them since more than one component depends on overaly @pfulton?

components/tray/metadata/tray.yml
yosevu: I thought about a way to add examples for portrait and landscape, but there didn't seem to be a simple way since these depend on the [orientation media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation) which is determined by the relationship between the viewport height and width.

I meant to include this information in the description.

components/tray/index.css
yosevu: This won't work if I move it to .spectrum-Tray since it isn't a modifier class and needs to be in container that wraps .spectrum-Tray. Is there another way to make this work? I'm not familiar with how SWC consumes components.

components/tray/index.css
yosevu: Not that I know of, but it might be worth adding to the tokens doc. This was the previous static value that was used.

components/commons/overlay-coretokens.css
yosevu: I removed overlay related styles since overlay will likely updated to be a part of Modal.

components/tray/index.css
yosevu: I removed overlay styles as they will likely be included with Modal.

pfulton (approved):
I think this looks good @yosevu!

pfulton:
Rebased against the latest code in `main` and published the beta: `@spectrum-css/tray@2.0.0-beta.0`

PR #1565, opened 2022-12-15
refactor(avatar!): use spectrum tokens

yosevu:
## Description

Refactors avatar to use spectrum tokens.

Tested in Chrome
Tested in WHCM

## To-do list


components/avatar/index.css
```diff
@@ -10,107 +10,128 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    content: "";
     position: absolute;
-    inset-block-start: 0;
-    inset-block-end: 0;
-    inset-inline-start: 0;
-    inset-inline-end: 0;
-    box-sizing: border-box;
-    border-width: var(--spectrum-avatar-border-size);
+    top: calc(
+      var(--mod-avatar-focus-indicator-gap, var(--spectrum-avatar-focus-indicator-gap)) *
```
bernhard-adobe, line 106: nit. I would place the `-2` for both cases on a single line.

bernhard-adobe (approved):
Approving here with a minor question for PJ and Jess in terms of content strategy:

https://adobedesign.slack.com/archives/G019JTYMT6H/p1671723449855909

Screen Shot 2022-12-22 at 16 38 13

pfulton:
> Approving here with a minor question for PJ and Jess in terms of content strategy

I don't have a problem with this, as we seem to use his photo in other examples in other components. I can double-check with design folks for the ultimate answer, though.

I pushed up a few more commits here to correct a few things:
* [Here](https://github.com/adobe/spectrum-css/pull/1565/commits/aa213a82144aeb1c5de99176b2f402be0f509148), I think Yosevu might have just accidentally kept the anchor links in the "no link" examples.
* [Here](https://github.com/adobe/spectrum-css/pull/1565/commits/9cb9ee06c8ea5cc77764e66f62691704c7e568a3), I bumped the version number for the tokens package, and also included the default selector stuff we've been including in the `express.css` and `spectrum.css` files.
* Here, I restored these modified and removed `vars` and `expressvars` files. I don't think these should have been modified or deleted. I'm wondering if Yosevu was having some trouble with style conflicts while he was working on this, but when I restored/reset these files and ran things again, everything looked ok?

components/avatar/index.css
```diff
@@ -10,107 +10,126 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    pointer-events: none;
+    content: "";
     position: absolute;
-    inset-block-start: 0;
-    inset-block-end: 0;
-    inset-inline-start: 0;
-    inset-inline-end: 0;
-    box-sizing: border-box;
-    border-width: var(--spectrum-avatar-border-size);
+    top: calc(
```
lunarfusion, line 105: This should be

```
    top: calc((var(--mod-avatar-focus-indicator-gap, var(--spectrum-avatar-focus-indicator-gap)) 
              + var(--mod-avatar-focus-indicator-thickness, var(--spectrum-avatar-focus-indicator-thickness))) * -1);

```

So that if either the border width or the focus indicator gap ever change, the indicator will still center with the component.

Also, I think it _might_ be better to use margin-block-start instead of "top".

components/avatar/index.css
```diff
@@ -10,107 +10,126 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+      var(--mod-avatar-focus-indicator-gap, var(--spectrum-avatar-focus-indicator-gap)) * -2);
+    left: calc(
```
lunarfusion, line 107: This should be

```
    left: calc((var(--mod-avatar-focus-indicator-gap, var(--spectrum-avatar-focus-indicator-gap)) 
              + var(--mod-avatar-focus-indicator-thickness, var(--spectrum-avatar-focus-indicator-thickness))) * -1);
```


To see the difference, try changing line 21 to a larger pixel size, like

`--spectrum-avatar-focus-indicator-gap: 6px;`

Before the change, this resulted in an off-center focus indicator:
Screenshot 2023-01-03 at 5 23 58 PM

After changing to the suggested calculation, the indicator centers even when the gap is a different amount than the border width:

Screenshot 2023-01-03 at 5 24 24 PM

Also, I think it _might_ be better to use margin-inline-start instead of "left".

components/avatar/index.css
```diff
@@ -10,107 +10,126 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-@import "../vars/css/components/spectrum-avatar.css";
+.spectrum-Avatar {
+  --spectrum-avatar-border-radius: 50%;
```
lunarfusion, line 14: _💭 Thought, not a necessary change:_

We could also set this equal to the component's width, like

`-spectrum-avatar-border-radius: var(--spectrum-avatar-size-100);`

components/avatar/index.css
```diff
@@ -10,107 +10,126 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  outline: none;
+}
+
 .spectrum-Avatar-image {
-  inline-size: var(--spectrum-avatar-width);
-  block-size: var(--spectrum-avatar-height);
+  inline-size: var(--mod-avatar-inline-size, var(--spectrum-avatar-inline-size));
+  block-size: var(--mod-avatar-block-size, var(--spectrum-avatar-block-size));
+
+  border-radius: var(--mod-avatar-border-radius, var(--spectrum-avatar-border-radius));
```
lunarfusion, line 134: _💭 Thought, not a necessary change:_

We could also achieve this by setting 

`boder-radius: inherit;`

on both .spectrum-Avatar-image and .spectrum-Avatar-link

components/avatar/index.css
```diff
@@ -10,107 +10,126 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-    content: '';
+  overflow: visible;
+
+  opacity: var(--mod-avatar-color-opacity, var(--spectrum-avatar-color-opacity));
+
+  &.is-disabled {
+    opacity: var(--highcontrast-avatar-color-opacity-disabled, var(--mod-avatar-color-opacity-disabled, var(--spectrum-avatar-color-opacity-disabled)));
+  }
+
+  &:focus-within:not(.is-disabled) .spectrum-Avatar-link::after {
```
lunarfusion, line 101: Adobe design indicated that we should not use :focus-within.
Slack link: https://adobedesign.slack.com/archives/G019JTYMT6H/p1667421055499529?thread_ts=1666626065.985779&cid=G019JTYMT6H

lunarfusion (changes requested):
Mostly suggestions, but I think the focus indicator position does need a change.

components/avatar/index.css
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Adobe. All rights reserved.
+Copyright 2022 Adobe. All rights reserved.
```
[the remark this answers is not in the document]
pfulton: Really, we should find a way to prepend this comment in every `dist` file when we run the build. I'll make a ticket for that :-)

Manually updating is fine for now, though!

components/avatar/index.css
pfulton: > Also, I think it might be better to use margin-block-start instead of "top".

Good call-out! I looked into this a little, and I think it's `inset-block-start`: https://developer.mozilla.org/en-US/docs/Web/CSS/inset-block-start

components/avatar/index.css
pfulton: > Also, I think it might be better to use margin-inline-start instead of "left".

Today I learned! Similar to the comment about `top`, I think this is `inset-inline-start`: https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-start

components/avatar/index.css
pfulton: I like this suggestion!

components/avatar/index.css
pfulton: Interesting suggestion. I looked into this briefly, and I think we _might_ not be able to safely do this and have it work like we want it to: https://css-tricks.com/shadow-roots-and-inheritance/

pfulton:
> Mostly suggestions, but I think the focus indicator position does need a change.

Thanks! I fixed up those things.

components/avatar/index.css
```diff
@@ -10,107 +10,128 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-    filter: color;
-    replace: --spectrum-avatar$1;
+  --spectrum-avatar-inline-size: var(--spectrum-avatar-size-700);
+  --spectrum-avatar-block-size: var(--spectrum-avatar-size-700)
+}
+
+@media (forced-colors: active) {
+  .spectrum-Avatar {
+    --highcontrast-avatar-color-opacity-disabled: 1;
+    --highcontrast-avatar-focus-indicator-color: Highlight;
```
lunarfusion, line 76: James Nurthen and I discussed what our focused component border color and focus indicator border color should be in WHCM. The verdict is:

**Component border (if there is one):** Highlight
**Focus indicator border color:** CanvasText

Since Avatar does not have a border, only the focus indicator color needs to be changed to CanvasText.

It may also be necessary to add "forced-color-adjust: none;" to the focus indicator. Here is the format:

```
  .spectrum-Avatar:not(.is-disabled) .spectrum-Avatar-link.focus-ring,
  .spectrum-Avatar:not(.is-disabled) .spectrum-Avatar-link:focus-visible {
    &::after {
      forced-color-adjust: none;
    }
  }
}
```

lunarfusion (changes requested):
Changes to calc format and to WHCM.

components/avatar/index.css
pfulton: It looks like we're ok without having to add the `forced-color-adjust: none;`

Here's what I'm seeing in AssitivLabs after updating to use `CanvasText`:

Screen Shot 2023-01-10 at 5 01 35 PM

pfulton:
@lunarfusion - pushed up some changes to address your most recent feedback. This includes the updates to WHCM and the reinstatement of `:focus-within`. Thanks!

lunarfusion (approved):
Looks good to me! 🚀

bernhard-adobe:
@pfulton @lunarfusion I am adding the new avatar here in a moment. Maybe please don't merge or release yet.

bernhard-adobe:
There we go:
Screen Shot 2023-01-11 at 17 28 54

I also added a 2x variant for future use.

Original file is: https://stock.adobe.com/images/attractive-fashionable-young-african-american-female-with-clean-healthy-skin-and-afro-haircut-looking-into-the-distance-with-thoughtful-face-dreaming-of-holidays-while-workin-on-laptop-at-a-cafe/113729455

Screen Shot 2023-01-11 at 17 29 48

pfulton:
Thanks @bernhard-adobe!!

pfulton:
This has been updated against the latest stuff in `main` and is ready for its beta release.

pfulton:
Beta released: `@spectrum-css/avatar@6.0.0-beta.0`

components/avatar/index.css
```diff
@@ -10,107 +10,123 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  overflow: visible;
+
+  opacity: var(--mod-avatar-color-opacity, var(--spectrum-avatar-color-opacity));
+
+  &.is-disabled {
+    opacity: var(--highcontrast-avatar-color-opacity-disabled, var(--mod-avatar-color-opacity-disabled, var(--spectrum-avatar-color-opacity-disabled)));
+  }
+}
+
+.spectrum-Avatar:focus-within:not(.is-disabled) .spectrum-Avatar-link {
```
Westbrook, line 103: This could also be: `.spectrum-Avatar:not(.is-disabled) .spectrum-Avatar-link:focus-visible` and have the same functional styles, right? The main difference that we've not discussed for `:focus-within` is that is doesn't match the keyboard heuristics of `:focus-visible`...

pfulton: Yeah, that'll work.

And yes, we definitely have a discussion about how best to serve `:focus-*` to y'all going forward.

Westbrook (approved):
Works for SWC! Graduate a stable release and we can get https://github.com/adobe/spectrum-web-components/pull/2844 closed out as well!

PR #1567, opened 2022-12-15
chore: hot-module reloading for storybook [CSS-279]

castastrophe:
Preview link: [https://pr-1567--spectrum-css.netlify.app/preview](https://pr-1567--spectrum-css.netlify.app/preview)

## Description

- [CSS-279](https://jira.corp.adobe.com/browse/CSS-279)

Added icon story & leveraged the new template to remove the generateSVG tooling from tools/preview package & instead, let stories import templates from each other. This ends up solving [CSS-315](https://jira.corp.adobe.com/browse/CSS-315) by optimizing how icons are being loaded.

Updated imports to point to the uncompiled assets rather than dist assets; this allows storybook's built-in functionality to watch for changes & automatically refresh.

Set up webpack to compile the css assets using the postcss rules defined in component-builder & component-builder-simple (determining which package to use based on the information in that component's package.json).

Package scripts:
 - Removed pre-start build step
 - Removed parallel watch command managed by gulp

Validated cross-story imports work by importing the icon template into button & helptext stories.

As the migration status was being collected for postcss processing, I took the opportunity to add the status badge plugin I mentioned in [CSS-309](https://jira.corp.adobe.com/browse/CSS-309) to identify migrated components *automagically*. 🪄

Squirreled a bit making an icon demo page & then making it pretty. 🐿️

 - **How this was tested:** 
    - [x] `yarn start` to spin up local development server
        - Expected: no pre-build step; local server opens
    - [x] All stories appear and have styles loaded correctly
    - [x] Make a change to a CSS asset that is included in one of the stories.
        - Expected: Webpack re-compiles, localhost refreshes
    - Chrome loads the full icon list demo very quickly while Safari crawls to a halt. @pfulton Should we add a note to the icon list page that suggests viewing in Chrome?

## To-do list


components/expressvars/package.json
```diff
@@ -10,7 +10,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/adobe/spectrum-css.git",
-    "directory": "components/vars"
+    "directory": "components/expressvars"
```
castastrophe, line 13: @pfulton Okay this is totally unrelated but I was in the file and noticed this and thought, that perhaps needs fixing.  Can open separate PR if you like!

pfulton: Wow, great catch. Thanks for fixing!

I think we can sneak this in with this PR, but, if possible: can you keep this change in its own commit, and it can be a `fix` or a `refactor` - it's probably more of a `fix`...but either way, it's just a patch.

THANKS.

tools/preview/preview.js
```diff
@@ -142,6 +159,15 @@ export const parameters = {
...
   },
+  status: {
+    statuses: {
+      migrated: {
+        background: '#f0f0f0',
+        color: '#444',
+        description: 'Migrated to the latest tokens.',
+      },
+    },
+  },
```
pfulton, line 170: LOVE this.

pfulton (approved):
This is great work, and is a huge step forward with the overall storybook effort. Thanks for taking the time on this and also thanks for the thoughtful "migrated" addition. That's fantastic.

I left one comment here about pulling the `expressvars/package.json` fix into its own commit, but that's the only thing that I noticed.

components/expressvars/package.json
castastrophe: Awesome - just pushed up that update.

PR #1568, opened 2022-12-22
feat(thumbnail)!: migrate tokens

yosevu:
## Description

Migrates Thumbnail to core tokens 
- uses Opacity Checkerboard 
- update use of Thumbnail in TreeView
- Add WHCM for focus and selected 
- Add variants for Layer (layer and layer, selected), [see slack convo ](https://adobedesign.slack.com/archives/GQ09RCBA5/p1680799351171439)

Variants Added: 
- Disabled 
- Layer 
- Layer, Selected

Variants Replaced: 
- Selected (replaced with Layer, Selected)
- T-shirt Sizes Replaced with Numeric Sizes 

Chrome 112 on macOS
Safari 16.4 on macOS
Firefox 112 on macOS

## To-do list


components/expressvars/css/components/index.css
pfulton, line 1: I think this is an old change from when Yosevu had started to work on this component. We should drop this change from the branch before we merge this in.

components/expressvars/css/components/spectrum-thumbnail.css
pfulton, line 1: I think this file needs to stay in the project and _should not_ be deleted. We should drop the commit responsible for this change or bring back the file in some other way.

components/vars/css/components/index.css
pfulton, line 1: Similar to my other comment about dropping the change to this file: I think this is a carry-over from when Yosevu had first started work on this component and that this needs to stay the same/shouldn't have any changes (even if it's a whitespace change)

pfulton (changes requested):
Great work! It'll be great to get this one in!

I left a few comments here related to what I think are some changes that were made when this work originally started, so we should take care of those things before merging this work in.

Additionally, can we update the Storybook story for this component? It looks like we'll want to update the Size knob to switch from t-shirt sizing to the numbered sizing, and also to account for some of the other options that we have on the docs page for the component.

components/expressvars/css/components/index.css
jenndiaz: got both of these removed!

components/expressvars/css/components/spectrum-thumbnail.css
jenndiaz: added back in!

components/thumbnail/metadata/thumbnail.yml
```diff
@@ -3,20 +3,8 @@ status: Contribution
...
-      | -------------------- | ----------------------- |
-      | `~.spectrum-Thumbnail--XXL~` | REMOVED |
-      | `~.spectrum-Thumbnail--XL~` | REMOVED |
-      | `~.spectrum-Thumbnail--L~` | `.spectrum-Thumbnail--sizeL` |
-      | `~.spectrum-Thumbnail--M~` | `.spectrum-Thumbnail--sizeM` |
-      | `~.spectrum-Thumbnail--S~` | `.spectrum-Thumbnail--sizeS` |
-      | | `.spectrum-Thumbnail--sizeXS` |
-      | | `.spectrum-Thumbnail--sizeXXS` |
+      ### Thumbnail sizing
+      Thumbnail sizes are specified by appending a number to the size class: `.spectrum-Thumbnail--size*`.
```
mlogsdon18, line 7: Should we specify what those number options are? That might be overkill but I don't know how knowledgable the developers using this are of their options

components/thumbnail/metadata/thumbnail.yml
```diff
@@ -26,77 +14,155 @@ examples:
     description: Rectangular images will fill the entire space.
     name: Thumbnail
     markup: |
-      <div class="spectrum-Thumbnail spectrum-Thumbnail--sizeM">
-        <img class="spectrum-Thumbnail-image" src="img/thumbnail.png" alt="Woman crouching">
+      <div class="spectrum-Thumbnail spectrum-Thumbnail--size500">
+        <img class="spectrum-Thumbnail-image"src="img/example-ava.jpg" alt="women resting head in hands">
```
mlogsdon18, line 18: super nit-picky but I just noticed it so might as well!
```suggestion
        <img class="spectrum-Thumbnail-image" src="img/example-ava.jpg" alt="women resting head in hands">
```

components/thumbnail/metadata/thumbnail.yml
```diff
@@ -26,77 +14,155 @@ examples:
...
+    markup: |
+      <div class="spectrum-Thumbnail spectrum-Thumbnail-layer spectrum-Thumbnail--size500">
+        <div class="spectrum-Thumbnail-layer-inner">
+          <img class="spectrum-Thumbnail-image" src="img/example-ava.jpg" alt="women resting head in hands"></div>
+      </div>
+  - id: thumbnail-selected
+    name: Thumbnail (layer, selected)
+    description: The thumbnail is given a thick blue border to indicate its selection when used in layer management.
+    markup: |
+      <div class="spectrum-Thumbnail spectrum-Thumbnail-layer spectrum-Thumbnail--size500 is-selected">
```
mlogsdon18, line 59: I noticed that you changed `is-selected` to `is-focused` for thumbnails that are not in a layer. Are you keeping this one as selected for a reason?

mlogsdon18 (commented):
One other thing I found in the Storybook is that when I am in the Layer story, if I decrease the size then the image starts disappearing towards the bottom. If I increase the size, the image starts moving toward the top. Is it supposed to do that? 
Screenshot 2023-04-18 at 12 22 52 PM
Screenshot 2023-04-18 at 12 24 08 PM

components/thumbnail/metadata/thumbnail.yml
jenndiaz: yes, the figma and [chatting with the designer](https://adobedesign.slack.com/archives/GQ09RCBA5/p1680799351171439), the thumbnail should only receive the selected styling when used within layer. So this variant is displaying the Thumbnail selected and with in layer

components/thumbnail/metadata/thumbnail.yml
jenndiaz: thats a good call, added language from design documentation.

jenndiaz:
> One other thing I found in the Storybook is that when I am in the Layer story, if I decrease the size then the image starts disappearing towards the bottom. If I increase the size, the image starts moving toward the top. Is it supposed to do that? Screenshot 2023-04-18 at 12 22 52 PM Screenshot 2023-04-18 at 12 24 08 PM

Nope, not supposed to do that. Pushed a fix.  Thanks for catching it!

components/thumbnail/metadata/thumbnail.yml
```diff
@@ -3,100 +3,205 @@ status: Contribution
...
-      </div>
-  - id: thumbnail-selected
-    name: Thumbnail (selected)
-    markup: |
-      <div class="spectrum-Thumbnail spectrum-Thumbnail--sizeM is-selected">
-        <img class="spectrum-Thumbnail-image" src="img/thumbnail.png" alt="Woman crouching">
+      <div class="spectrum-Thumbnail spectrum-Thumbnail--size500">
+        <div class="spectrum-Thumbnail-image-wrapper">
+          <img class="spectrum-Thumbnail-image"src="img/example-ava.jpg" alt="women resting head in hands">
+        </div>
```
jenndiaz, line 21: @Westbrook  Any concerns about this additional div from a SWC perspective? 

I have added the image-wrapper div to each of the variants to fix an issue with the focus state and the thumbnail image not maintaining border radius. This is an issues currently in place but the new design removed the double border on the focus state making is more visible.

pfulton:
Published: `@spectrum-css/thumbnail@3.0.0-beta.0`

components/thumbnail/metadata/thumbnail.yml
Westbrook: Shouldn't be an issue.

components/thumbnail/index.css
```diff
@@ -1,4 +1,4 @@
-/*!
+/*
```
castastrophe, line 1: Just wanted to flag this change - be sure to bring back the /*! before merging. It's a new standard we're using to ensure copyrights don't get stripped

pfulton (approved):

PR #1578, opened 2023-01-10
[DEMO DO NOT MERGE] fork: 11ty config proposal

castastrophe:
@Rajdeepc This is just an example PR to demonstrate some suggestions I had about the 11ty setup.

In this fork, I moved the `site` folder into `tools/site` along with its related dependencies so that it would function as a scoped workspace. That allows us to call it's scripts from root using `yarn workspace @spectrum-css/build-site serve` for example (assuming there's a `serve` script in the `tools/site/package.json`).

I also think we should keep with commonjs for now since most of the tooling in the repo leverages it. That prevents us from having to update a lot of files with `.cjs` and we can do a more concerted effort to migrate to modules later on.

Just a note - don't worry about trying to merge this into your branch. I just opened it as a fork against your branch so you could see the changes but it's just a demo and not functioning fully.


Rajdeepc:
@castastrophe Sure I can move site folder under tools to act as a scoped package. I changed few files to .cjs with type:"module" on the package since few of the packages needed them to run as modules once i changed its gulp dependency to vanilla nodejs. 
I will remove the type= "module" from package.json

PR #1614, opened 2023-02-07
refactor(button): combinator free layout rules

Westbrook:
## Description
The use of the `+` combinator to style icon/label layout in the Button requires SWC to leverage content observers in order to mimic the relationship across the slots leveraged to accept these two pieces of content. More heavily leveraging `display: inline-flex` and some alternate padding/margin applications should allow the same layout without requiring the use of a combinator.

**_Works best for SWC when the same can be done for Action Button (which used the `+` AND `:only-child`), will investigate further._**

## Screenshots
TK baring preview site build

## To-do list

### DIFF
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -50,9 +50,7 @@ governing permissions and limitations under the License.
   --spectrum-button-height: var(--spectrum-component-height-100);
 
   --spectrum-button-font-size: var(--spectrum-font-size-100);
-
-
-
+  
   --spectrum-button-edge-to-visual: calc(var(--spectrum-component-pill-edge-to-visual-100) - var(--spectrum-button-border-width));
   --spectrum-button-edge-to-text: calc(var(--spectrum-component-pill-edge-to-text-100) - var(--spectrum-button-border-width));
   --spectrum-button-padding-label-to-icon: var(--spectrum-text-to-visual-100);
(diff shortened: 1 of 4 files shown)


components/button/index.css
```diff
@@ -91,68 +89,39 @@ governing permissions and limitations under the License.
...
   @inherit: %spectrum-ButtonWithFocusRing;
 
-
   border-radius: var(--mod-button-border-radius, var(--spectrum-button-border-radius));
   border-width: var(--mod-button-border-width, var(--spectrum-button-border-width));
   border-style: solid;
-  block-size: auto;
+  font-size: var(--mod-button-font-size, var(--spectrum-button-font-size));
+  font-weight: var(--mod-bold-font-weight, var(--spectrum-bold-font-weight));
+  gap: var(--spectrum-button-padding-label-to-icon);
```
Westbrook, line 101: Ensures space between the icon and label.

components/button/index.css
```diff
@@ -91,68 +89,39 @@ governing permissions and limitations under the License.
...
     box-shadow: none;
   }
 
-  /* let staticColor variants inherit their color */
-  color: inherit;
-
   .spectrum-Icon {
-    inset-inline-end: calc(var(--mod-spectrum-button-edge-to-visual, var(--spectrum-button-edge-to-visual)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
-    inset-block-end: calc(var(--mod-spectrum-button-edge-to-visual, var(--spectrum-button-edge-to-visual)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
+    margin-inline-start: calc(
```
Westbrook, line 118: Ensures that the icon is closer to the edge of the Button (for now) that the label is by itself.

components/button/index.css
```diff
@@ -167,10 +136,15 @@ a.spectrum-Button {
   @inherit: %spectrum-ButtonLabel;
   padding-block-start: var(--mod-button-padding-label-top, var(--spectrum-button-padding-label-top));
   padding-block-end: var(--mod-button-padding-label-bottom, var(--spectrum-button-padding-label-bottom));
+  white-space: nowrap;
+  line-height: calc(
```
Westbrook, line 140: Prevents the label from being "taller" than the button, which I feel normalized the vertical positioning of the label, which moved _up_ in the Core Tokens refactor, but it moves up/down a lot and I'm never sure what design really wants here.

components/button/index.css
```diff
@@ -91,68 +89,39 @@ governing permissions and limitations under the License.
...
-  border-radius: var(--mod-button-border-radius, var(--spectrum-button-border-radius));
-  border-width: var(--mod-button-border-width, var(--spectrum-button-border-width));
   /* Start with text-only padding */
-  padding-inline-start: var(--mod-button-edge-to-visual, var(--spectrum-button-edge-to-visual));
-  padding-inline-end: var(--mod-button-edge-to-text, var(--spectrum-button-edge-to-text));
-
-
-  font-size: var(--mod-button-font-size, var(--spectrum-button-font-size));
-  font-weight: var(--mod-bold-font-weight, var(--spectrum-bold-font-weight));
+  padding-inline: var(--mod-button-edge-to-text, var(--spectrum-button-edge-to-text));
```
Westbrook, line 106: Ensures that padding is balances left/right when there is only a label. Currently, padding at the end of the Button is slightly larger than at the beginning.

components/button/index.css
```diff
@@ -91,68 +89,39 @@ governing permissions and limitations under the License.
...
+  gap: var(--spectrum-button-padding-label-to-icon);
   height: var(--mod-button-height, var(--spectrum-button-height));
-
-  min-width: var(--mod-button-min-width, var(--spectrum-button-min-width));
-  min-block-size: var(--mod-button-min-width, var(--spectrum-button-min-width));
-  position: relative;
-
-
-  padding-block-start: 0;
-  padding-block-end: 0;
```
Westbrook, line 112: Lots of stuff over here is either duplicated or has not active value. Much of it was likely removed at tooling time, but easier to understand what's going on when they're not here.

pfulton:
@Westbrook Thanks for taking a look at this and suggesting these refactors! Would you mind taking a look at how the buttons are rendering in Safari after these changes? It looks like the label (particularly in the medium size) is pushed upward a bit more, and is off-center.

Westbrook:
I _is_ really high, even with this correction that brings it down a couple of pixels. Descenders in the Button copy help it make more sense, I think. However, there was some padding I had missed.

By "off-center", do you mean vertically or horizontally? Horizontally seems fine, but 

Let me know if you catch anything else!

Westbrook:
@pfulton revisiting that vertical spacing, I have the distinct impression that the tokens are wrong for that `medium` button. For `small` the top and bottom paddings are 1px apart, and for `large` and `extra-large` they are each 2px apart, but in `medium` they are 3px apart, 6px on top and 9px on bottom, which leads to this difference. If the bottom padding is moved down to 8px, everything looks much nicer. What's the best way to get design to take another look at this with me?

Westbrook:
Got some good insight, and while there's likely some more thought to go into whether upstream tokens should be adjusted, this delivery should put the Button pattern on a nice footing regardless of how that plays out.

pfulton (approved):

PR #1616, opened 2023-02-09
chore: add support for console output in storybook; suppress autoprefixer

castastrophe:
 - **How this was tested:** 

## To-do list


tools/preview/preview.js
```diff
@@ -5,6 +5,14 @@ import {
   withReducedMotionWrapper
 } from './decorators/index.js';
 
+import '@storybook/addon-console';
+import { setConsoleOptions } from '@storybook/addon-console';
+
+const panelExclude = setConsoleOptions({}).panelExclude;
+setConsoleOptions({
+  panelExclude: [...panelExclude, /deprecated/, /TypeError/],
```
Rajdeepc, line 13: if panelExclude is spreading objects then /deprecated/ and /TypeError/ will be invalid types?
Do you want to keep a type check?

Also if panelExclude is returning null or undefined this line can throw error: Cannot spread on null

Rajdeepc (approved):
LGTM!!!

tools/preview/preview.js
castastrophe: A note for later, I got the implementation from here: https://github.com/storybookjs/storybook-addon-console#storybookaddon-consolesetconsoleoptionsoptionsorfn--addonoptions. The authors were on the storybook team so I suspect panelExclude always returns a value. it's a great point though.  We could add: `const panelExclude = setConsoleOptions({}).panelExclude || [];` just to be safe.

PR #1619, opened 2023-02-10
feat(colorwheel)!: migrate to core tokens

jenndiaz:
## Description
BREAKING CHANGE: This migrates the ColorWheel component to core tokens.
Jira Ticket [CSS 189](https://jira.corp.adobe.com/browse/CSS-189)

- refactored component to use a clip path, instead of a SVG with a mask. This more closely resembles how SWC is displaying the color wheel and takes a whole lot less code 
- the clip paths and container size are hard coded but include custom variables (`--track-width` and `--border-width`) that will allow JS access to the token values needed to calculate both of these 

 **How this was tested:** 
comparing local build of http://localhost:3000/docs/colorwheel.html with XD redline components for ColorWheel and with https://spectrum.adobe.com/page/color-wheel/ 
 
 **Browser(s) and OS(s) this was tested with:** 
Chrome 109 on macOS
Safari 16.2 on macOS
Firefox 109 on macOS 

### Medium with ColorArea
Screenshot 2023-03-08 at 12 34 34 PM

### Large
Screenshot 2023-03-08 at 12 34 10 PM

### WHCM (zoomed so not to scale)
whcm1

## To-do list


components/colorwheel/package.json
```diff
@@ -18,12 +18,14 @@
...
-    "@spectrum-css/vars": "^8.0.0"
+    "@spectrum-css/colorarea": "^1.0.29",
+    "@spectrum-css/tokens": "^7.0.0"
   },
   "devDependencies": {
-    "@spectrum-css/colorhandle": "^2.0.14",
-    "@spectrum-css/component-builder": "^4.0.0",
-    "@spectrum-css/vars": "^8.0.3",
+    "@spectrum-css/colorhandle": "^2.0.11",
+    "@spectrum-css/colorarea": "^1.0.29",
```
jenndiaz, line 26: I added ColorArea as a dependency so it could be displayed within the ColorWheel (see screenshot). This is included in the XD file and is closely related to the token (--spectrum-color-wheel-color-area-margin) but I was not sure if we wanted to have this added or not. Please let me know if we'd prefer to not include this.

pfulton:
Screenshot for disabled Windows High Contrast Mode:

Screen Shot 2023-02-15 at 11 19 08 AM

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
+  width: 100%;
+  margin: var(--mod-colorwheel-colorarea-margin, var(--spectrum-colorwheel-colorarea-margin));
 }
 
 .spectrum-ColorWheel-slider {
   @extend %spectrum-ColorControl-hiddenField;
 }
 
 .spectrum-ColorWheel-handle {
+  transform: translate(var(--spectrum-colorwheel-colorhandle-position), 0);
```
pfulton, line 120: Can you also add the `--mod-` prefixed custom property reference here?

```suggestion
  transform: translate(var(--mod-colorwheel-colorhandle-position, var(--spectrum-colorwheel-colorhandle-position)), 0);
```

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
+/* clip path set border-width wider than than spectrum-colorwheel-wheel to create appreance of a border*/
+.spectrum-ColorWheel-border{
+  position: relative;
+  background-color: var(--mod-colorwheel-border-color, var(--spectrum-colorwheel-border-color));
+  width: var(--mod-colorwheel-width, var(--spectrum-colorwheel-width));
+  height: var(--mod-colorwheel-height, var(--spectrum-colorwheel-height));
+  clip-path: path(evenodd, var(--spectrum-colorwheel-path-borders));
+
+  &.is-disabled{
+    background-color: var(--highcontrast-colorwheel-border-color-disabled, var(--spectrum-colorwheel-border-color));
```
pfulton, line 134:
```suggestion
    background-color: var(--highcontrast-colorwheel-border-color-disabled, var(--mod-colorwheel-border-color, var(--spectrum-colorwheel-border-color)));
```

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
   top: 50%;
 }
+
+/* clip path set border-width wider than than spectrum-colorwheel-wheel to create appreance of a border*/
+.spectrum-ColorWheel-border{
+  position: relative;
+  background-color: var(--mod-colorwheel-border-color, var(--spectrum-colorwheel-border-color));
+  width: var(--mod-colorwheel-width, var(--spectrum-colorwheel-width));
+  height: var(--mod-colorwheel-height, var(--spectrum-colorwheel-height));
+  clip-path: path(evenodd, var(--spectrum-colorwheel-path-borders));
```
pfulton, line 131: I'm on the fence about opening this one up for modification, but I think we probably _should_...

```suggestion
  clip-path: path(evenodd, var(--mod-colorwheel-path-borders, var(--spectrum-colorwheel-path-borders)));
```

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
+
+/* --track-width and --border-width to be used with JS in calculating the clip-path paths and colorarea-container-size */
+.spectrum-ColorWheel-wheel{
+  position: absolute;
+  background: conic-gradient(from 90deg, red, rgb(255, 128, 0), rgb(255, 255, 0), rgb(128, 255, 0), rgb(0, 255, 0), rgb(0, 255, 128), rgb(0, 255, 255), rgb(0, 128, 255), rgb(0, 0, 255), rgb(128, 0, 255), rgb(255, 0, 255), rgb(255, 0, 128), red);
+  top: 1px;
+  left: 1px;
+  width: 100%;
+  height: 100%;
+  clip-path: path(evenodd, var(--spectrum-colorwheel-path));
```
pfulton, line 152:
```suggestion
  clip-path: path(evenodd, var(--mod-colorwheel-path, var(--spectrum-colorwheel-path)));
```

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
+}
+
+/* --track-width and --border-width to be used with JS in calculating the clip-path paths and colorarea-container-size */
+.spectrum-ColorWheel-wheel{
+  position: absolute;
+  background: conic-gradient(from 90deg, red, rgb(255, 128, 0), rgb(255, 255, 0), rgb(128, 255, 0), rgb(0, 255, 0), rgb(0, 255, 128), rgb(0, 255, 255), rgb(0, 128, 255), rgb(0, 0, 255), rgb(128, 0, 255), rgb(255, 0, 255), rgb(255, 0, 128), red);
+  top: 1px;
+  left: 1px;
+  width: 100%;
+  height: 100%;
```
pfulton, line 151: Similar to my comment above about logical properties. I think we can convert these to [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties).

components/colorwheel/index.css
```diff
@@ -56,35 +94,68 @@
...
 .spectrum-ColorWheel-handle {
+  transform: translate(var(--spectrum-colorwheel-colorhandle-position), 0);
   left: 50%;
   top: 50%;
 }
+
+/* clip path set border-width wider than than spectrum-colorwheel-wheel to create appreance of a border*/
+.spectrum-ColorWheel-border{
+  position: relative;
+  background-color: var(--mod-colorwheel-border-color, var(--spectrum-colorwheel-border-color));
```
pfulton, line 128: Should we also provide a `--highcontrast-` prefixed custom property here? We've been trying to do that for everything that involves color-related properties/values.

```suggestion
  background-color: var(--highcontrast-colorwheel-border-color, var(--mod-colorwheel-border-color, var(--spectrum-colorwheel-border-color)));
```

pfulton (changes requested):
Thanks for all of the hard work on this one @jenndiaz! I left some feedback here. Let me know if you have any questions.

components/colorwheel/index.css
jenndiaz: I have questions on this one. Generally our pattern in the code base currently is to not use a high contrast borders in WHCM for the non disabled variants. I was using[ ColorSlider](https://opensource.adobe.com/spectrum-css/colorslider.html) as my example which has a similar "border" and have implemented ColorArea similarly but maybe this is not what we want moving forward. 

Should I change both ColorWheel and ColorArea (both of which do not directly receive focus, ColorHandle does) to have high contrast borders when in WHCM and if so what system color do we want for them?

pfulton: Oh, you may be correct here. I'm actually curious as to what this should look like in WHCM overall.

jenndiaz: We chatted with Accessibility team in slack, https://adobedesign.slack.com/archives/CSNDSSA5D/p1676488870249269, and have decided on not adding a high contrast border in this case.

components/colorwheel/index.css
```diff
@@ -56,35 +95,69 @@
...
 
 .spectrum-ColorWheel-slider {
   @extend %spectrum-ColorControl-hiddenField;
 }
 
 .spectrum-ColorWheel-handle {
-  left: 50%;
+  transform: translate(var(--mod-colorwheel-colorhandle-position, var(--spectrum-colorwheel-colorhandle-position)), 0);
   top: 50%;
+  left: 50%;
```
jenndiaz, line 126: I am using physical properties for the positioning of the ColorHandle. 
`top: 50%; left: 50%` Positions in the center regardless for direction, while ` inset-inline: 50%; inset-block: 50%;` or `inset-inline-start: 50%; inset-block-start` is only centered when the direction is `ltr`

jenndiaz: [codepen displaying what I was running into ](https://codepen.io/jenndiaz/pen/ZEMGOqg)

components/colorwheel/package.json
bernhard-adobe: yes this makes sense. They are both used for this component and I can't imagine an instance where we would have a color wheel without the handle.

components/colorwheel/index.css
```diff
@@ -1,27 +1,67 @@
...
 }
 
 %spectrum-ColorControl-handle--focused {
-  /* Bigger handle when focused */
+  /* Bigger colorhandle when focused */
   width: calc(var(--spectrum-colorhandle-size) * 2);
   height: calc(var(--spectrum-colorhandle-size) * 2);
 
   margin-left: calc(-1 * var(--spectrum-colorhandle-size));
   margin-top: calc(-1 * var(--spectrum-colorhandle-size));
```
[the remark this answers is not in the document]
bernhard-adobe: this article has an amazing graphic featuring logical properties. I look at it once in a while for this migration.

https://medium.com/@elad/new-css-logical-properties-bc694
Screen Shot 2023-02-21 at 10 56 03
5311ce7

components/colorwheel/index.css
```diff
@@ -1,27 +1,68 @@
...
-  --spectrum-colorwheel-focus-ring-size: var(--spectrum-global-dimension-static-size-25);
-  --spectrum-colorwheel-border-radius: 100%;
-  --spectrum-colorwheel-width: calc(var(--spectrum-global-dimension-size-125) * 16);
   --spectrum-colorwheel-height: var(--spectrum-colorwheel-width);
+  --spectrum-colorwheel-border-color: var(--spectrum-transparent-black-200);
+  --spectrum-colorwheel-border-width: var(--spectrum-border-width-100);
+  --spectrum-colorwheel-fill-color-disabled: var(--spectrum-disabled-background-color);
+  --spectrum-colorwheel-track-width: var(--spectrum-color-control-track-width);
+  --spectrum-colorwheel-colorarea-margin: var(--spectrum-color-wheel-color-area-margin);
+  --spectrum-colorhandle-size: 16px; /* TODO replace with new token for ColorHandle, placeholder for old token var(--spectrum-colorhandle-size) */
```
bernhard-adobe, line 19: does the color handle change size? I think so - at least that's what we did in SVG when we created the handle.
You might want to move it into the `.specturm-medium / large area`
Screen Shot 2023-02-21 at 11 06 46

components/colorwheel/index.css
```diff
@@ -56,35 +95,69 @@
...
-  box-sizing: border-box;
-
-  border-width: var(--spectrum-colorwheel-border-size);
-  border-style: solid;
-}
-
-.spectrum-ColorWheel-wheel {
-  position: relative;
-  z-index: 1;
+.spectrum-ColorWheel-internal{
```
bernhard-adobe, line 98: nit: space.

I am debating if we should call the inner area "internal". 
Spectrum React sets up the divs like this and call it "inner".

```
.spectrum-ColorWheel {
  .spectrum-ColorWheel-gradient {
    .spectrum-ColorHandle {
      .spectrum-ColorLoupe {
         .spectrum-ColorLoupe-inner-background {}
         .spectrum-ColorLoupe-inner-checker {}

```

reference: https://react-spectrum.adobe.com/react-spectrum/ColorWheel.html

components/colorwheel/index.css
bernhard-adobe: might need an over-write here like 
```
.spectrum-ColorWheel-handle {
 [dir="rtl"] & {
```

components/colorwheel/index.css
```diff
@@ -56,35 +95,69 @@
...
+  }
+}
+
+/* hack: make sure we don't complain about these variables being unused*/
+.🤫 {
+  width: var(--track-width);
+  border-width: var(--border-width);
+}
+
+/* --track-width and --border-width to be used with JS in calculating the clip-path paths and colorarea-container-size */
```
bernhard-adobe, line 148: do we still have JS calculating anything here? looks like we removed the previous JS from the YML file that was used for come calculations.

jenndiaz: I did remove a <script> from the .yml, this was doing some calculating for the cavnas element we are no longer displaying and was not related to the paths. 

Since we can't manipulate a string in CSS, I am hard coding the path and providing these values so they can be used in JS without hardcoding.

components/colorwheel/index.css
jenndiaz: good call and yes it does! Got this fixed.

components/colorwheel/index.css
jenndiaz: I also like inner better, changed and addressed the spacing inconsistencies.

components/colorwheel/index.css
jenndiaz: was able to use an over-write and get it working with logical properties.

bernhard-adobe:
Thanks for addressing the feedback @jenndiaz .
I think there is a minor conflict with package.json - easy to resolve, you could even use Gits web gui for it `    "@spectrum-css/colorhandle": "^2.0.16",`

@pfulton or @castastrophe could you please release a beta package for the Colorwheel so we could try it out in SWC? thank you.

castastrophe:
Did a rebase & beta release:

```
Successfully published:
 - @spectrum-css/colorwheel@2.0.0-beta.0
```

components/colorwheel/package.json
```diff
@@ -19,12 +19,13 @@
   },
   "peerDependencies": {
     "@spectrum-css/colorhandle": "^2.0.0",
-    "@spectrum-css/vars": "^8.0.0"
+    "@spectrum-css/tokens": "^7.0.0"
```
pfulton, line 22: Let's update this one, too.

```suggestion
    "@spectrum-css/tokens": "^7.1.0"
```

components/colorwheel/index.css
```diff
@@ -1,48 +1,72 @@
...
+  --spectrum-colorwheel-border-width: var(--spectrum-border-width-100);
+  --spectrum-colorwheel-fill-color-disabled: var(--spectrum-disabled-background-color);
+  --spectrum-colorwheel-track-width: var(--spectrum-color-control-track-width);
+  --spectrum-colorwheel-colorarea-margin: var(--spectrum-color-wheel-color-area-margin);
+
+  --spectrum-colorwheel-colorhandle-position: calc((var(--spectrum-colorwheel-width) / 2) - (var(--spectrum-colorwheel-track-width) / 2));
+
+  .spectrum--medium & {
+    --spectrum-colorwheel-width: 192px; /* TODO replace with incoming color-wheel-width token */
+    --spectrum-colorwheel-min-width: 175px; /* TODO replace with incoming color-wheel-minimum-width token */
```
pfulton, line 24: This is now available in the tokens package as: `--spectrum-color-wheel-minimum-width`, but as noted above, for consistency, we should probably do something like:

```css
--spectrum-colorwheel-min-width: var(--spectrum-color-wheel-minimum-width);
```

pfulton (changes requested):
Just a few notes here on the new color wheel tokens that are available.

components/colorwheel/index.css
jenndiaz: tokens for width and min width have been updated!

pfulton:
I just merged in and released the platform scale tokens additions. You'll be able to get those by bumping your `package.json` to point at: `@spectrum-css/tokens@7.2.0`

components/colorarea/index.css
```diff
@@ -74,8 +74,11 @@ governing permissions and limitations under the License.
 
 .spectrum-ColorArea-handle {
   transform: translate(calc(var(--mod-colorarea-width, var(--spectrum-colorarea-width)) - var(--spectrum-colorarea-border-width)), 0);
-  top: 0;
-  left: 0;
+  inset-block-start: 0;
+
+  [dir="rtl"] & {
+    inset-inline-end: 0;
+  }
```
jenndiaz, line 81: saw that these physical properties snuck though, updated them while looking at the ColorArea with in the ColorWheel

components/colorwheel/index.css
```diff
@@ -56,35 +63,97 @@
...
+  inset-block-start: 0;
+
+  [dir="ltr"] & {
+    inset-inline-end: 0;
+  }
+
+  [dir="rtl"] & {
+    inset-inline-start: 0;
+  }
+
```
jenndiaz, line 122: I am not sure why I needed this, I was observing that [the same properties coming](https://github.com/adobe/spectrum-css/pull/1619/files#r1129958692) from `spectrum-ColorArea-handle` were not having the same effect here and added this to adjust. Open to suggestions and thoughts why this might be the case.

components/colorwheel/stories/template.js
```diff
@@ -0,0 +1,41 @@
...
+      [rootClass]: true,
+      'is-disabled': isDisabled,
+      'is-focused': isFocused,
+      ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+  })}>
+  <div class="${rootClass}-inner">
+    <div class="${rootClass}-colorarea-container"></div>
+  </div>
+  <div class="${rootClass}-border ${isDisabled ? 'is-disabled' : ""}">
+    <div class="${rootClass}-wheel ${isDisabled ? 'is-disabled' : ""}"></div>
```
mlogsdon18, line 28: I don't know if this is any better or not but you could just do the `classMap` like you're doing for the div before this to conditionally add the `is-disabled` class

jenndiaz: Glad you send something, I had tried to use a `classMap` previously and couldn't get the syntax right, figured it out now!

pfulton (approved):
Awesome! Looks good to me!

PR #1622, opened 2023-02-13
chore: update storybook generator; fix syntax of import

castastrophe:
## Description

This PR tackles 2 small updates:
- Update generator to match the latest storybook syntax
- Update imports in files from `if-definedjs` to `if-defined.js`

 - **How this was tested:** 
 - [x] `yarn new`

## To-do list

### DIFF
--- a/tools/generator/templates/stories/template.js.hbs
+++ b/tools/generator/templates/stories/template.js.hbs
@@ -15,7 +15,7 @@ export const Template = ({
   return html`
     <div class=${classMap({
       [rootClass]: true,
-      [`${className}--size${size.toUpperCase()}`]: true,
+      [`${rootClass}--size${size.toUpperCase()}`]: true,
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
     })} id=${ifDefined(id)}>
       <!-- Component mark-up goes here -->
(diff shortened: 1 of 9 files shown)


tools/generator/templates/stories/template.js.hbs
```diff
@@ -1,22 +1,29 @@
...
+  rootClass = "spectrum-{{ pascalCase name }}",
+  size,
+  id,
+  customClasses = [],
+  ...globals
+}) => {
+  return html`
+    <div class=${classMap({
+      [rootClass]: true,
+      [`${className}--size${size.toUpperCase()}`]: true,
```
pfulton, line 18:
```suggestion
      [`${rootClass}--size${size.toUpperCase()}`]: true,
```

I think we want `rootClass` here. As it is, `className` is returning `undefined`

pfulton (changes requested):
Just one thing that I saw here @castastrophe. Otherwise, thanks for taking care of this.

tools/generator/templates/stories/template.js.hbs
castastrophe: Oh yes! Great catch.

pfulton (approved):

PR #1628, opened 2023-02-18
docs: script to add css mod variables to the corresponding markdown files

venkateshjo:
 - **How this was tested:** 


pfulton (approved):

scripts/mod-extractor.js
```diff
@@ -26,42 +37,34 @@ function writeToFile(file) {
...
-                    writeToFile(readFile);
-                } 
-            });
-        });
+        }
     }
 }
 
-searchFilesForString(rootDir, regex, fileExtension);
+searchFilesForString(rootDir, pattern, fileExtension);
```
Rajdeepc, line 69: How will you execute this script during the build process?

Rajdeepc (changes requested):

Rajdeepc:
@venkateshjo Can you also please make sure the description is well articulated, Jira tickets are mentioned, Images (if any) and also how to test this locally all commands should be noted in the description of the ticket. It would be easier for @pfulton  and @castastrophe to test this locally.
Thanks!!!

scripts/mod-extractor.js
venkateshjo: Currently, this script is not part of the build process. User can trigger the build with the following command
yarn docs:mod

Rajdeepc: Okay Nice!!! But this also needs to go into the build process eventually. Also please think on that part on how you can include this inside the component-builder build process.

venkateshjo:
Sure. 
JIRA: https://jira.corp.adobe.com/browse/CSS-306

The script can be triggered with the following command
"yarn docs:mod"

This will generate the corresponding markdown files if the script detects --mod variables in the .css files.

Screenshot 2023-02-21 at 8 54 36 PM

scripts/mod-extractor.js
castastrophe: @Rajdeepc True! Patrick & I had talked about making this manual for now until we can merge in your updates that way there is less likely to be a conflict.

castastrophe (approved):

PR #1631, opened 2023-02-22
docs: use tables to display mod custom props

pfulton:
## Description
Adds some formatting to the `mods.md` files

 - **How this was tested:** 

## To-do list


castastrophe:
@pfulton Awesome! I wonder if it makes sense to keep the mods in a data format so we can use them in other ways in the future. Maybe store them as json or yml? Then we can wrap them in the table via the docs template?

pfulton:
> @pfulton Awesome! I wonder if it makes sense to keep the mods in a data format so we can use them in other ways in the future. Maybe store them as json or yml? Then we can wrap them in the table via the docs template?

oh, great thought!

PR #1637, opened 2023-02-24
feat:adds storybook for colorarea and colorhandle

jenndiaz:
## Description
Adds storybook for ColorArea and ColorHandle

Jira Ticket[ CSS-397](https://jira.corp.adobe.com/browse/CSS-397)
Jira Ticket [CSS-398](https://jira.corp.adobe.com/browse/CSS-398)

## Screenshots

Screenshot 2023-02-24 at 4 26 26 PM
Screenshot 2023-02-24 at 4 26 45 PM
Screenshot 2023-02-24 at 4 26 57 PM

## To-do list

### DIFF
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -1,0 +1,35 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Colorarea",
+  description: "The Colorarea component is...",
+  component: "Colorarea",
+  argTypes: {
+    size: {
+      name: "Size",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["s", "m", "l", "xl"],
+      control: "select"
+    },
+  },
+  args: {
+    rootClass: "spectrum-",
+    size: "m",
+  },
+  parameters: {
+    actions: {
+      handles: []
+    },
+    status: {
+      type: process.env.MIGRATED_PACKAGES.includes('colorarea') ? 'migrated' : undefined
+    }
+  }
+};
+
+export const Default = Template.bind({});
+Default.args = {};
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -6,7 +6,8 @@ export default {
   description: "The Color area component allows users to visually select two properties of a color simultaneously. It's commonly used together with a color slider or color wheel.",
   component: "ColorArea",
   argTypes: {
-    customSize: { table: { disable: true } },
+    customWidth: { table: { disable: true } },
+    customHeight: { table: { disable: true } },
     handlePosition: { table: { disable: true } },
     isDisabled: {
       name: "Disabled",
@@ -30,6 +31,7 @@ export default {
   args: {
     rootClass: "spectrum-ColorArea",
     isDisabled: false,
+    isFocused: false,
   },
   parameters: {
     actions: {
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -46,6 +46,6 @@ Default.args = {};
 
 export const CustomSize = Template.bind({});
 CustomSize.args = {
-  customSize: "--mod-colorarea-height: 80px; --mod-colorarea-width: 80px;",
-  handlePosition: "transform: translate(80px, 0px);"
+  customWidth: "80px",
+  customHeight: "80px"
 };
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -1,8 +1,5 @@
 // Import the component markup template
-// stories is where the stories will be written
 import { Template } from "./template";
-import { default as ColorHandle } from "@spectrum-css/colorhandle"
-// how do I handle dependent components
 
 export default {
   title: "Color area",
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -25,7 +25,8 @@ export default {
         type: { summary: "boolean" },
         category: "State",
       },
-      control: "boolean"
+      control: "boolean",
+      if: { arg: 'isDisabled', truthy: false }
     },
   },
   args: {
--- a/components/colorarea/stories/colorarea.stories.js
+++ b/components/colorarea/stories/colorarea.stories.js
@@ -3,7 +3,7 @@ import { Template } from "./template";
 
 export default {
   title: "Color area",
-  description: "The Colorarea component is...",
+  description: "The Colorarea component allows users to visually select two properties of a color simultaneously. It's commonly used together with a color slider or color wheel.",
   component: "Colorarea",
   argTypes: {
     isDisabled: {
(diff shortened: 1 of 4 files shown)


components/colorarea/stories/template.js
```diff
@@ -0,0 +1,36 @@
...
+
+import '../index.css';
+
+export const Template = ({
+  rootClass = "spectrum-ColorArea",
+  customClasses = [],
+  isDisabled = false,
+  customSize,
+  handlePosition,
+  style = "--spectrum-picked-color: rgb(255, 0, 0);" + handlePosition,
```
castastrophe, line 16: Since this is being passed down to handle, maybe we can rename this property to indicate that so it doesn't get confused with the rootClass's custom styles?  Or maybe we can even pass those down as values to ColorHandle's function to render?

components/colorarea/stories/template.js
```diff
@@ -0,0 +1,36 @@
...
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { Template as ColorHandle } from "@spectrum-css/colorhandle/stories/template.js";
+
+import '../index.css';
+
+export const Template = ({
+  rootClass = "spectrum-ColorArea",
+  customClasses = [],
+  isDisabled = false,
+  customSize,
```
castastrophe, line 14: I like the idea of breaking these out as their own types (maybe string type so user's can put in units too?) - what do you think?  `width` & `height`? We can convert the values into the variables inside the template & optionally append a px unit if none is present.

components/colorarea/stories/colorarea.stories.js
```diff
@@ -0,0 +1,40 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Color area",
+  description: "The Colorarea component allows users to visually select two properties of a color simultaneously. It's commonly used together with a color slider or color wheel.",
+  component: "Colorarea",
+  argTypes: {
```
castastrophe, line 8:
```suggestion
  argTypes: {
    customSize: { table: { disable: true } },
    handlePosition: { table: { disable: true } },
```

Anything you add to args below will show up in the table but without types unless otherwise hidden.

components/colorarea/stories/template.js
```diff
@@ -0,0 +1,36 @@
...
+  // ...globals
+}) => {
+  return html`
+    <div class=${classMap({
+    [rootClass]: true,
+    'is-disabled': isDisabled,
+    ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+  })} ?disabled=${isDisabled}} style=${ifDefined(customSize)}>
+  <div class="spectrum-ColorArea-gradient" style="background: linear-gradient(to top, black 0%, rgba(0, 0, 0, 0) 100%), linear-gradient(to right, white 0%, rgba(0, 0, 0, 0) 100%), rgb(255, 0, 0);"></div>
+  ${ ColorHandle({
```
castastrophe, line 26: Awesome! Don't forget to pass globals down to the nested component's template so it is aware of theme contexts.

components/colorhandle/stories/colorhandle.stories.js
```diff
@@ -0,0 +1,36 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Color handle",
+  description: "The Color Handle component is used with ColorArea, ColorSlider and ColorWheel as the color selector",
+  component: "Colorhandle",
+  argTypes: {
+    isDisabled: {
```
castastrophe, line 9: Can we add `is-focused` here as well?

castastrophe (changes requested):

components/colorarea/stories/colorarea.stories.js
```diff
@@ -0,0 +1,40 @@
...
+    isDisabled: {
```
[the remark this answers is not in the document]
jenndiaz: Yes, I can add isFocused. 

I could not find a reason why `spectrum-ColorArea--small` is in the docs, and specifically why it would be only on the disabled variant. There are no t-shirt sizes for ColorArea and I do not see any properties coming in for it and I could not find anything in the git history that explained what it is there. Looking at it closely now I'm thinking it should probably be removed in the migration PR. Did you see something else that would suggest an importance for leaving it in?

components/colorarea/stories/colorarea.stories.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+  }
+};
+
+export const Default = Template.bind({});
+Default.args = {};
+
+export const CustomSize = Template.bind({});
+CustomSize.args = {
+  customSize: "height: 80px; width: 80px;",
```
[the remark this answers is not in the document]
jenndiaz: oh! I think this helps me understand how we are using `--mod`s more, I should also implement the same change in the [migration PR](https://github.com/adobe/spectrum-css/pull/1623/files#r1119373889 ) where we are implementing customer size?

components/colorarea/stories/template.js
jenndiaz: taking a look at this! I am understanding your suggestion as instead of using custom size and having `customSize: "--mod-colorarea-height: 80px; --mod-colorarea-width: 80px;",` to use separate variables for width and height  in the story and convert this to the string we need within the template. I can see how that is more flexible.

components/colorarea/stories/template.js
jenndiaz: Got this added to the template for ColorArea.
To make sure I am understanding, do we only use this on nested components?

components/colorarea/stories/template.js
jenndiaz: Ok got it implemented let me know if you have feedback on how I approached it!

components/colorarea/stories/template.js
castastrophe: Yes exactly - it's just ensuring that express theme, light/dark theme, right-to-left, etc. are passed to nested templates. They don't seem to inherit them on their own.

components/colorarea/stories/colorarea.stories.js
```diff
@@ -0,0 +1,51 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Color area",
+  description: "The Color area component allows users to visually select two properties of a color simultaneously. It's commonly used together with a color slider or color wheel.",
+  component: "ColorArea",
+  argTypes: {
+    customSize: { table: { disable: true } },
```
castastrophe, line 9:
```suggestion
    customWidth: { table: { disable: true } },
    customHeight: { table: { disable: true } },
```

^ will need to update the naming here since it's been updated below :D

components/colorarea/stories/template.js
```diff
@@ -0,0 +1,52 @@
...
+import '../index.css';
+
+export const Template = ({
+  rootClass = "spectrum-ColorArea",
+  customClasses = [],
+  isDisabled = false,
+  isFocused = false,
+  customWidth,
+  customHeight,
+  customSize,
```
castastrophe, line 16: Love it! You can remove `customSize` now too if you're not using it.

components/colorhandle/stories/template.js
```diff
@@ -0,0 +1,29 @@
...
+  return html`
+  <div class=${classMap({
+    [rootClass]: true,
+    'is-disabled': isDisabled,
+    'is-focused': isFocused,
+    ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+  })}
+  ?disabled=${isDisabled}
+  ?focused=${isFocused}
+  style=${colorHandleStyle}>
```
castastrophe, line 24:
```suggestion
  style=${ifDefined(colorHandleStyle)}>
```

^ I'd wrap this in an ifDefined (and import ifDefined from lit at the top) and then this is good to go!

castastrophe (approved):

PR #1638, opened 2023-02-26
feat: add additional storybook stories

pfulton:
## Description
- [X] InlineAlert
- [X] ProgressCircle
- [X] PickerButton
- [X] StatusLight

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/inlinealert/stories/inlinealert.stories.js
+++ b/components/inlinealert/stories/inlinealert.stories.js
@@ -1,0 +1,57 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Inlinealert",
+  description: "In-line alerts display a non-modal message associated with objects in a view. These are often used in form validation, providing a place to aggregate feedback related to multiple fields.",
+  component: "Inlinealert",
+  argTypes: {
+    headerText: {
+      name: "Header Text",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        disable: false,
+        category: "Component",
+      },
+      control: { type: "text" },
+    },
+    text: {
+      name: "Text",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        disable: false,
+        category: "Component",
+      },
+      control: { type: "text" },
+    },
+    variant: {
+      name: "In-line alert variants",
+      type: { name: "string" },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["neutral", "info", "positive", "notice", "negative", "closable"],
+      control: "select"
+    }
+  },
+  args: {
+    rootClass: "spectrum-InLineAlert",
+    headerText: "in-line alert header",
+    text: "This is an alert.",
+    variant: "neutral",
+  },
+  parameters: {
+    actions: {
+      handles: []
+    },
+    status: {
+      type: process.env.MIGRATED_PACKAGES.includes('inlinealert') ? 'migrated' : undefined
+    }
+  }
+};
+
+export const Default = Template.bind({});
+Default.args = {};
--- a/components/inlinealert/stories/inlinealert.stories.js
+++ b/components/inlinealert/stories/inlinealert.stories.js
@@ -35,13 +35,23 @@ export default {
       },
       options: ["neutral", "info", "positive", "notice", "negative", "closable"],
       control: "select"
-    }
+    },
+    isClosable: {
+      name: "Closable",
+      type: { name: "boolean" },
+      table: {
+        type: { summary: "boolean" },
+        category: "Advanced",
+      },
+      control: "boolean",
+    },
   },
   args: {
     rootClass: "spectrum-InLineAlert",
     headerText: "in-line alert header",
     text: "This is an alert.",
     variant: "neutral",
+    isClosable: false,
   },
   parameters: {
     actions: {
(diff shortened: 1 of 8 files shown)


components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -16,10 +16,91 @@ export default {
...
+        type: { summary: "boolean" },
+        category: "State",
+      },
+      control: "boolean",
+    },
   },
   args: {
-    rootClass: "spectrum-",
+    rootClass: "spectrum-PickerButton",
+    customClasses: ["spectrum-PickerButton--right"],
```
castastrophe, line 94:
```suggestion
    customClasses: ["spectrum-PickerButton--right"],
```

Oh interesting use-case here. I'd typically used customClasses solely for passing down contextual classes from parent elements but this is an interesting idea.  Would you ever want to offer the direction as a control above?

components/pickerbutton/stories/template.js
```diff
@@ -1,20 +1,67 @@
...
+  isDisabled = false,
+  label,
   // ...globals
 }) => {
+  const labelMarkup =
+    isWorkflowIconOnly === true || isUIIconOnly === true ? '' : html`<span class="${rootClass}-label is-placeholder">${label}</span>`;
+
+  const uiIconMarkup = html`
+    ${Icon({
+      iconName: "Chevron100",
```
castastrophe, line 29: Don't forget to pass globals down to nested components so they are aware of platform scale and context.

components/progresscircle/stories/progresscircle.stories.js
```diff
@@ -0,0 +1,55 @@
...
+  component: "Progresscircle",
+  argTypes: {
+    size: {
+      name: "Size",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["small", "medium", "large"],
```
castastrophe, line 16: Typically at this level, size is t-shirt sizing. `Small`, `medium`, and `large` here seem more aligned with platform scale. Is that accurate?  If so, you can call the platform scale values from the template inside the `...globals` spread property (it's defined in the parameters file at the storybook level.

castastrophe (changes requested):

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -0,0 +1,57 @@
...
+  description: "In-line alerts display a non-modal message associated with objects in a view. These are often used in form validation, providing a place to aggregate feedback related to multiple fields.",
+  component: "Inlinealert",
+  argTypes: {
+    headerText: {
+      name: "Header Text",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        disable: false,
+        category: "Component",
```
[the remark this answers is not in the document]
pfulton: Yeah, I like this, and I do think we can standardize on it. Thanks!

components/pickerbutton/stories/pickerbutton.stories.js
pfulton: Oh, you know: we don't have an example of `--left` on the current docs site, but there's CSS written for it, and it's something that SWC has on their docs. I think this should probably be a knob, and the class can toggle between the two directions, with `--right` being the default. I'll make that change.

components/progresscircle/stories/progresscircle.stories.js
pfulton: > Typically at this level, size is t-shirt sizing.

Yeah, that makes sense.

> Small, medium, and large here seem more aligned with platform scale. Is that accurate?

It's interesting: I think these are more akin to t-shirt sizes, but without an extra-large size. I was double-checking against the [design spec](https://spectrum.adobe.com/page/progress-circle/#Design-checklist) (see "all platform scales"). What do you think? Leave as is?

castastrophe: Oh interesting yeah - I think you're right. Should we use `s`, `m`, `l` to align with the other size values in storybook?

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -0,0 +1,57 @@
...
+      control: { type: "text" },
+    },
+    variant: {
+      name: "Variants",
+      type: { name: "string" },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["neutral", "info", "positive", "notice", "negative", "closable"],
```
castastrophe, line 36: Is closable in the same category as the other variants? I'm wondering if you can have a closable alert that is also neutral.

components/progresscircle/stories/progresscircle.stories.js
pfulton: Yeah, that's good for consistency's sake. I'll make that change.

components/inlinealert/stories/inlinealert.stories.js
pfulton: Oh, great catch. In our docs site we have "An in-line alert with a close button in the footer. Combine this strategy with any variant" as the definition of Closable, and the actual Spectrum site doesn't have Closable at all...

I'm going to switch it from a Variant to some other kind of option!

castastrophe (approved):

PR #1639, opened 2023-02-27
feat: cycle button story

mlogsdon18:
## Description

This adds a story for the Cycle Button component.

[Jira CSS-402](https://jira.corp.adobe.com/browse/CSS-402)

## To-do list

### DIFF
--- a/components/cyclebutton/stories/cyclebutton.stories.js
+++ b/components/cyclebutton/stories/cyclebutton.stories.js
@@ -1,0 +1,75 @@
+// Import the component markup template
+import { Template } from "./template";
+
+import { default as IconStories } from "../../icon/stories/icon.stories.js";
+
+export default {
+  title: "Cycle Button",
+  description: "The Cycle button component is an action button that cycles through two different icons, a play that then changes to a pause, for example.",
+  component: "Cyclebutton",
+  argTypes: {
+    size: {
+      name: "Size",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["s", "m", "l", "xl"],
+      control: "select"
+    },
+    initialIcon: {
+      name: "Initial icon",
+      type: { name: "string", required: true },
+      options: 
+        IconStories && IconStories.argTypes && IconStories.argTypes.iconName
+          ? IconStories.argTypes.iconName.options
+          : {},
+      control: { type: "select" },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+    },
+    selectedIcon: {
+      name: "Selected icon",
+      type: { name: "string", required: true },
+      options: 
+        IconStories && IconStories.argTypes && IconStories.argTypes.iconName
+          ? IconStories.argTypes.iconName.options
+          : {},
+      control: { type: "select" },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+    },
+    isSelected: {
+      name: "Selected",
+      type: { name: "boolean" },
+      table: {
+        type: { summary: "boolean" },
+        category: "State",
+      },
+      control: "boolean",
+    },
+  },
+  args: {
+    rootClass: "spectrum-CycleButton",
+    size: "m",
+    initialIcon: "Play",
+    selectedIcon: "Pause",
+    isSelected: false
+  },
+  parameters: {
+    actions: {
+      handles: []
+    },
+    status: {
+      type: process.env.MIGRATED_PACKAGES.includes('cyclebutton') ? 'migrated' : undefined
+    }
+  }
+};
+
+export const Default = Template.bind({});
+Default.args = {};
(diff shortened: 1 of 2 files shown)


components/cyclebutton/stories/template.js
```diff
@@ -2,20 +2,54 @@ import { html } from 'lit-html';
 import { classMap } from 'lit-html/directives/class-map.js';
 // import { ifDefined } from 'lit-html/directives/if-defined.js';
 
+import { Template as Icon } from "@spectrum-css/icon/stories/template.js"
```
castastrophe, line 5:
```suggestion
import { Template as Icon } from "@spectrum-css/icon/stories/template.js";
import { Template as ActionButton } from "@spectrum-css/actionbutton/stories/template.js";
```

Rather than re-define the markup below, we should pull in the template from ActionButton's story so that we can leverage that. That keeps our source of truth separate so if ActionButton's markup changes, we'll inherit that here automatically.

components/cyclebutton/stories/template.js
```diff
@@ -2,20 +2,54 @@ import { html } from 'lit-html';
...
+
 import '../index.css';
-import '../skin.css';
+import "@spectrum-css/actionbutton/index.css"
+import "@spectrum-css/actionbutton/themes/spectrum.css"
+import "@spectrum-css/actionbutton/themes/express.css"
```
castastrophe, line 10:
```suggestion
```

Leveraging the template above means you don't need to pull in the css assets (they'll come with the template automatically).

components/cyclebutton/stories/template.js
```diff
@@ -2,20 +2,54 @@ import { html } from 'lit-html';
...
-  // ...globals
+  initialIcon = "Play",
+  selectedIcon = "Pause",
+  isSelected,
+  ...globals
 }) => {
+  
   return html`
-    <div class=${classMap({
+    <button class=${classMap({
```
castastrophe, line 23: Can you update the markup here to leverage the Template function here:

https://github.com/adobe/spectrum-css/blob/b734251f0c2abbdc559e870d65fab7af2aa21d2d/components/actionbutton/stories/template.js#L1

Something like: 
```
ActionButton({
  ...globals,
  customClasses: [rootClass],
  isQuiet: true, ...
})
```

castastrophe (changes requested):
Awesome - thanks for tackling this. Since Cycle Button is just an opinionated Action Button, It think it's important that we are leveraging the ActionButton template directly.

components/cyclebutton/stories/cyclebutton.stories.js
```diff
@@ -1,10 +1,12 @@
 // Import the component markup template
 import { Template } from "./template";
 
+import { default as IconStories } from "../../icon/stories/icon.stories.js";
```
castastrophe, line 4:
```suggestion
import { default as IconStories } from "../../icon/stories/icon.stories.js";
import { default as ActionButtonStories } from "@spectrum-css/actionbutton/stories/actionbutton.stories.js";
```

components/cyclebutton/stories/cyclebutton.stories.js
```diff
@@ -1,10 +1,12 @@
...
+
 export default {
-  title: "Cycle button",
-  description: "The Cycle button component is...",
-  component: "Cyclebutton",
+  title: "Cycle Button",
+  description: "The Cycle button component is an action button that cycles through two different icons, a play that then changes to a pause, for example.",
+  component: "CycleButton",
   argTypes: {
     size: {
```
castastrophe, line 11:
```suggestion
    size: ActionButtonStories.argTypes.size,
```

components/cyclebutton/stories/cyclebutton.stories.js
```diff
@@ -16,10 +18,48 @@ export default {
...
+        IconStories && IconStories.argTypes && IconStories.argTypes.iconName
+          ? IconStories.argTypes.iconName.options
+          : {},
+      control: { type: "select" },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+    },
+    isSelected: {
```
castastrophe, line 47:
```suggestion
    isSelected: ActionButtonStories.argTypes.isSelected,
    isDisabled: ActionButtonStories.argTypes.isDisabled,
```

components/cyclebutton/stories/cyclebutton.stories.js
```diff
@@ -16,10 +18,48 @@ export default {
...
   },
   args: {
-    rootClass: "spectrum-",
+    rootClass: "spectrum-CycleButton",
     size: "m",
+    initialIcon: "Play",
+    selectedIcon: "Pause",
+    isSelected: false
   },
   parameters: {
```
castastrophe, line 64:
```
  handles: [ ...ActionButtonStories.parameters.actions.handles ],
```

This will pick up any supported actions from the ActionButton story.

components/cyclebutton/stories/cyclebutton.stories.js
castastrophe: This will pick up the definitions directly from the ActionButton story.

components/cyclebutton/stories/template.js
mlogsdon18: I thought about this but I didn't know how to add the functionality for Cycle Button. ActionButton's story doesn't allow me to pass in children... I can pass in an icon but I don't have a way to pass in the second icon. Unless I'm missing something there?

PR #1640, opened 2023-02-27
feat: drop indicator storybook

mlogsdon18:
## Description

This adds a story for Drop Indicator so that it displays and functions properly in storybook

## To-do list


components/dropindicator/stories/template.js
```diff
@@ -1,21 +1,30 @@
...
+    width: direction == 'horizontal' ? length + "px" : "",
+  };
+
   return html`
-    <div class=${classMap({
-      [rootClass]: true,
-      [`${rootClass}--size${size?.toUpperCase()}`]: typeof size !== "undefined",
-    ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
-  })}></div>
+    <div style="display: flex; align-items: center">
```
castastrophe, line 20:
```suggestion
    <div style="display: flex; align-items: center">
```

Can you pull this customization out? If we end up needing to embed the component in another one, we just want the template to contain a display tool.  If you like, you can add this to the story:

```
Default.parameters = {
  layout: 'centered'
};
```

components/dropindicator/stories/dropindicator.stories.js
```diff
@@ -3,23 +3,34 @@ import { Template } from "./template";
...
+    length: {
+      name: "Length",
+      description: "In pixels",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      control: { type: "text" },
+    }
```
castastrophe, line 28:
```suggestion
    size: {
      name: "Size",
      description: "Size of the drop indicator, requires a unit be provided; i.e., 200px or 100%.",
      type: { name: "string", required: false },
      table: {
        type: { summary: "string" },
        category: "Component",
      },
      control: { type: "text" },
    }
```

components/dropindicator/stories/template.js
```diff
@@ -1,21 +1,30 @@
...
 export const Template = ({
   rootClass = "spectrum-DropIndicator",
   customClasses = [],
-  size = "m",
-  // ...globals
+  direction = "vertical",
+  length,
 }) => {
+  const styles = {
+    height: direction == 'vertical' ? length + "px" : "",
```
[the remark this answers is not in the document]
mlogsdon18: Now that I'm looking at this I think you're right! I'll make that change and keep the control so it's still customizable but using the CSS variable

components/dropindicator/stories/template.js
mlogsdon18: I was trying to recreate the markup on the docs site but let me see if it works without!

components/dropindicator/stories/template.js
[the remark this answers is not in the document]
mlogsdon18: Actually I think that is the width of the indicator itself, not how tall it is. It's the border size of the line, not the other dimension. I didn't see anything in the CSS setting the height for the vertical drop indicator. Hopefully that makes sense

castastrophe (approved):

PR #1644, opened 2023-02-28
feat: field label storybook

mlogsdon18:
## Description

This adds a storybook instance for the Field Label component.

[Jira CSS-405](https://jira.corp.adobe.com/browse/CSS-405)

**Note: This does have a width style attribute hard coded and this is in order to demonstrate the alignment options, like how they do on [the docs site](https://opensource.adobe.com/spectrum-css/fieldlabel.html#left)**

### DIFF
--- a/components/fieldlabel/stories/template.js
+++ b/components/fieldlabel/stories/template.js
@@ -15,7 +15,7 @@ export const Template = ({
   alignment = "left",
   isDisabled,
   isRequired,
-  style = { width: "72px" },
+  style = {},
   ...globals
 }) => {
   if (!label) {
@@ -40,7 +40,7 @@ export const Template = ({
       "is-disabled": isDisabled,
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
     })}
-      style=${ifDefined(style)}
+      style=${ifDefined(styleMap(style))}
     >
       ${label}
       ${isRequired ?
--- a/components/fieldlabel/stories/template.js
+++ b/components/fieldlabel/stories/template.js
@@ -1,5 +1,7 @@
 import { html } from 'lit-html';
 import { classMap } from 'lit-html/directives/class-map.js';
+import { styleMap } from 'lit-html/directives/style-map.js';
+import { ifDefined } from "lit-html/directives/if-defined.js";
 
 import { Template as Icon } from "@spectrum-css/icon/stories/template.js";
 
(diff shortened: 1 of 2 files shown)


components/fieldlabel/stories/template.js
```diff
@@ -29,9 +33,20 @@ export const Template = ({
...
       [rootClass]: true,
       [`${rootClass}--size${size?.toUpperCase()}`]: typeof size !== "undefined",
+      [`${rootClass}--${alignment}`]: typeof alignment !== "undefined",
+      "is-disabled": isDisabled,
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
-  })}>
-    ${label}
-  </label>
+    })}
+      style="width: 72px;"
```
jenndiaz, line 40: Would a second story that displays it aligned make sense here?
 [Similar to how SWC does](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/field-label--side-align-start), but I think it could just be "aligned" and leave the control in there to flip it left to right. 

I do see that we are using this width for the left and right aligned on the docs site but not with the standard variant. 

I bring this up because I was locally adjusting the text length and if it is required or not and with the width fixed you get some unexpected text wrapping.

jenndiaz (commented):
I had some thoughts about the fixed width that I thought worth discussing. Not seeing any other concerns.

components/fieldlabel/stories/template.js
```diff
@@ -9,6 +10,9 @@ export const Template = ({
   customClasses = [],
   size = "m",
   label,
+  alignment = "left",
+  isDisabled,
+  isRequired,
```
castastrophe, line 15:
```suggestion
  isRequired,
  style = { width: '72px' },
```

I'd suggest using the styleMap (need to add it as an import at the top) and setting this as the default - that way we can overwrite it if necessary when nesting.

components/fieldlabel/stories/template.js
castastrophe, line 40:
```suggestion
      style=${ifDefined(styleMap(style))}
```

castastrophe (approved):

components/fieldlabel/stories/template.js
mlogsdon18: For styleMap this would need to be an array. Do I need to use styleMap for future nesting or do you think the object as you have it would be alright?

castastrophe: StyleMap uses objects keyed on the property name so it wouldn't need to be an array. If you need to toggle properties on or off, the object approach is nice but it doesn't have to be done that way.

castastrophe (approved):

PR #1647, opened 2023-03-01
feat(tokens): adds colorwheel custom express tokens

jenndiaz:
## Description
Adds ColorWheel clip path custom tokens for express. The border property applied to the ColorWheel increase from 1px to 2px when viewed in express. The clip paths also need to adjust accordingly. 

 - **How this was tested:** 

## To-do list


pfulton (approved):
Thanks for adding these!

PR #1649, opened 2023-03-01
feat: field group storybook

mlogsdon18:
## Description

This PR adds a storybook instance for Field Group. It contains two stories, one for vertical and one for horizontal layout. 

This Storybook instance utilizes Field Label, Radio Button, and Help Text to illustrate the use of Field Group


components/colorarea/stories/colorarea.stories.js
```diff
@@ -51,4 +51,4 @@ export const CustomSize = Template.bind({});
 CustomSize.args = {
   customWidth: "80px",
   customHeight: "80px"
-};
\ No newline at end of file
+};
```
jenndiaz, line 54: 😬 is my editor adding lines where it shouldn't be?

components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -1,35 +1,80 @@
...
+    id: "1",
+    label: "Radio 1",
+  },{
+    id: "2",
+    label: "Radio 2",
+  }]
+};
+
+export const Horizontal = Template.bind({});
+Horizontal.args = {
```
jenndiaz, line 71: I am noticing we have displayed horizontal and vertical as stories and within the controls table, do we want to remove them from the controls?

components/fieldgroup/stories/template.js
```diff
@@ -22,8 +29,33 @@ export const Template = ({
...
+      [`${rootClass}--${layout}`]: typeof layout !== "undefined",
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
-  })}></div>
+    })}>
+      ${FieldLabel({
+        ...globals,
+        size: "m",
+        label: "Field Group Label",
+        alignment: labelPosition === "side" ?? "right",
+      })}
```
jenndiaz, line 41: I am seeing on the docs site `spectrum-FieldLabel--right` applied to the field label which includes `padding-inline: 0 var(--mod-fieldlabel-side-padding-right, var(--spectrum-fieldlabel-side-padding-right));`

in the storybook when displayed with a side label this padding is not there and the label is touching the radio button. 
Screenshot 2023-03-01 at 12 23 34 PM

Looks like something is not working correctly where we are adding the `right` on, class is displaying as `spectrum-FieldLabel--true`

jenndiaz (changes requested):

components/fieldgroup/stories/fieldgroup.stories.js
mlogsdon18: I took a page out of Cassondra's book for the AssetCard. She has a few stories there but they are all also controls. I don't think it's bad to allow users to have controls, the stories just give them a quick way to visualize a main variation without having to remember to check it by changing it themselves

components/colorarea/stories/colorarea.stories.js
mlogsdon18: I have my editor set to add a new line at the end of files just because that's the typical sparkbox coding standard. It shouldn't have committed that though since I didn't edit these files...

PR #1650, opened 2023-03-01
chore: add storybook for colorslider

jenndiaz:
## Description
Adds storys for ColorSlider, Default and Vertical. 

## Screenshots

Screenshot 2023-03-01 at 4 07 38 PM
Screenshot 2023-03-01 at 4 07 51 PM

### DIFF
--- a/components/colorarea/stories/template.js
+++ b/components/colorarea/stories/template.js
@@ -3,7 +3,6 @@ import { classMap } from 'lit-html/directives/class-map.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { Template as ColorHandle } from "@spectrum-css/colorhandle/stories/template.js";
 
-
 import '../index.css';
 
 export const Template = ({
@@ -14,7 +13,10 @@ export const Template = ({
   customWidth,
   customHeight,
   handlePosition = "transform: translate(" + customWidth + ",  0px);",
-  colorHandleStyle = "--spectrum-picked-color: rgb(255, 0, 0);" + handlePosition,
+  colorHandleStyle = {
+    '--spectrum-picked-color': 'rgba(255, 0, 0)',
+    handlePosition,
+  },
   ...globals
 }) => {
 
--- a/components/colorarea/stories/template.js
+++ b/components/colorarea/stories/template.js
@@ -1,6 +1,6 @@
 import { html } from 'lit-html';
 import { classMap } from 'lit-html/directives/class-map.js';
-import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { styleMap } from 'lit-html/directives/style-map.js';
 import { Template as ColorHandle } from "@spectrum-css/colorhandle/stories/template.js";
 
 import '../index.css';
@@ -12,20 +12,12 @@ export const Template = ({
   isFocused = false,
   customWidth,
   customHeight,
-  handlePosition = "transform: translate(" + customWidth + ",  0px);",
-  colorHandleStyle = {
-    '--spectrum-picked-color': 'rgba(255, 0, 0)',
-    handlePosition,
+  styles = {
+    "--mod-colorarea-height":  customHeight,
+    " --mod-colorarea-width": customWidth,
   },
   ...globals
 }) => {
-
-  function handleCustomSize(customWidth, customHeight) {
-    if (customWidth && customHeight) {
-     return "--mod-colorarea-height:" + customHeight + "; --mod-colorarea-width:" + customWidth + ";"
-    }
-  }
-
   return html`
     <div class=${classMap({
     [rootClass]: true,
@@ -33,15 +25,16 @@ export const Template = ({
     'is-focused': isFocused,
     ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
   })}
-  ?disabled=${isDisabled}
-  ?focused=${isFocused}
-  style=${ifDefined(handleCustomSize(customWidth, customHeight))}>
+  style=${styleMap(styles)}>
   <div class="spectrum-ColorArea-gradient" style="background: linear-gradient(to top, black 0%, rgba(0, 0, 0, 0) 100%), linear-gradient(to right, white 0%, rgba(0, 0, 0, 0) 100%), rgb(255, 0, 0);"></div>
-    ${ ColorHandle({
+    ${ColorHandle({
       ...globals,
       isDisabled,
       customClasses: [`${rootClass}-handle`],
-      colorHandleStyle
+      colorHandleStyle : {
+        '--spectrum-picked-color': 'rgba(255, 0, 0)',
+        transfom:  `translate(${customWidth}, 0px)`
+      },
       })
     }
     <input type="range" class="spectrum-ColorArea-slider" name="x" aria-label="saturation and value" min="0" max="1" step="0.01">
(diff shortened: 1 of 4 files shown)


components/colorslider/stories/template.js
```diff
@@ -0,0 +1,42 @@
...
+  return html`
+    <div class=${classMap({
+      [rootClass]: true,
+      'spectrum-ColorSlider--vertical': vertical,
+      'is-disabled': isDisabled,
+      'is-focused': isFocused,
+      ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+  })}
+  ${isDisabled}
+  ${isFocused}>
```
mlogsdon18, line 27: What are these doing?

components/colorslider/stories/colorslider.stories.js
```diff
@@ -0,0 +1,51 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Color slider",
+  description: "The Color slider component lets users visually change an individual channel of a color..",
```
castastrophe, line 6:
```suggestion
  description: "The Color slider component lets users visually change an individual channel of a color.",
```

components/colorslider/stories/colorslider.stories.js
```diff
@@ -0,0 +1,51 @@
...
+  component: "Colorslider",
```
castastrophe, line 7:
```suggestion
  component: "ColorSlider",
```

components/colorslider/stories/template.js
castastrophe: I suspect these can be removed since the classes are being added above.

components/colorslider/stories/template.js
```diff
@@ -0,0 +1,42 @@
...
+import '../index.css';
+import '../skin.css';
+
+export const Template = ({
+  rootClass = "spectrum-ColorSlider",
+  customClasses = [],
+  isDisabled = false,
+  isFocused = false,
+  vertical,
+  colorHandleStyle = "--spectrum-picked-color: rgb(255, 0, 0);",
```
castastrophe, line 15:
```suggestion
  colorHandleStyle = {
    '--spectrum-picked-color': 'rgba(255, 0, 0, 0)',
  }
```

Love that this is passed through - can we define it as an object so we can render it through the `styleMap` utility?

components/colorslider/stories/template.js
```diff
@@ -0,0 +1,42 @@
...
+'right'}, rgb(255, 0, 0) 0%, rgb(255, 255, 0) 17%, rgb(0, 255, 0) 33%, rgb(0, 255, 255) 50%, rgb(0, 0, 255) 67%, rgb(255, 0, 255) 83%, rgb(255, 0, 0) 100%);"></div>
+  </div>
+  ${ColorHandle({
+    ...globals,
+    isDisabled,
+    customClasses: [`${rootClass}-handle`],
+    colorHandleStyle
+    })
+  }
+  <input type="range" class="${rootClass}-slider" min="0" max="100" step="1">
```
castastrophe, line 39:
```suggestion
  <input type="range" class="${rootClass}-slider" min="0" max="100" step="1" ?disabled=${isDisabled}>
```

Should the input field include the disabled attribute?

castastrophe (changes requested):

components/colorslider/stories/template.js
jenndiaz: I do not believe so. It is not disabled on the docs site and we handle all of the color selectors in the same way. 
There is no functionality to actually use the input and it is not visible.

components/colorslider/stories/template.js
jenndiaz: hmm yeah you both are right they can be removed. 
I was basing this off of how we implemented some of the button components but I can see they are not needed here.

components/colorslider/stories/template.js
jenndiaz: oh, yes! and I like using style map much more.  
This is passed to ColorHandle so implementation of styleMap it there. I also updated ColorArea to do the same.

components/colorhandle/stories/template.js
```diff
@@ -10,7 +10,11 @@ export const Template = ({
   customClasses = [],
   isDisabled = false,
   isFocused = false,
-  colorHandleStyle = "--spectrum-picked-color: rgba(255, 0, 0, 0.5); position: absolute; top: 50%; left: 50%;"
+  colorHandleStyle = {
+    '--spectrum-picked-color': 'rgba(255, 0, 0, 0.5)',
+    position: 'absolute; top: 50%; left: 50%;'
```
castastrophe, line 15:
```suggestion
    position: 'absolute',
    top: '50%',
    left: '50%',
```

components/colorslider/stories/template.js
```diff
@@ -0,0 +1,42 @@
...
+  style = {},
+  colorHandleStyle = {
+    '--spectrum-picked-color': 'rgba(255, 0, 0)',
+  },
+  ...globals
+}) => {
+  return html`
+    <div class=${classMap({
+      [rootClass]: true,
+      'spectrum-ColorSlider--vertical': vertical,
```
castastrophe, line 24:
```suggestion
      [`${rootClass}--vertical`]: vertical,
```

components/colorhandle/stories/template.js
castastrophe: ^ this format will not work; each css property needs to be it's own key-value pair

castastrophe (changes requested):

components/colorarea/stories/template.js
```diff
@@ -14,7 +13,10 @@ export const Template = ({
   customWidth,
   customHeight,
   handlePosition = "transform: translate(" + customWidth + ",  0px);",
-  colorHandleStyle = "--spectrum-picked-color: rgb(255, 0, 0);" + handlePosition,
+  colorHandleStyle = {
+    '--spectrum-picked-color': 'rgba(255, 0, 0)',
+    handlePosition,
```
castastrophe, line 18:
```suggestion
```

The handlePosition data is in a string format but needs to be used here as a key-value pair; it's also being passed in the same object so we can't access that data here.  I would maybe try adding it below on line 44 using a spread object:

```
colorHandleStyle: {
  ...colorHandleStyle,
  transform: `translate(${customWidth}px)`,
},
```

I also noticed on line 38 that the style assignment could be improved:

```
style=${ifDefined(styleMap({
  '--mod-colorarea-height': `${customHeight}px`,
  '--mod-colorarea-width': `${customWidth}px`,
}))}>
```

components/colorhandle/stories/template.js
```diff
@@ -10,7 +10,13 @@ export const Template = ({
   customClasses = [],
   isDisabled = false,
   isFocused = false,
-  colorHandleStyle = "--spectrum-picked-color: rgba(255, 0, 0, 0.5); position: absolute; top: 50%; left: 50%;"
+  colorHandleStyle = {
+    '--spectrum-picked-color': 'rgba(255, 0, 0, 0.5)',
```
castastrophe, line 14: Looks great!

components/colorslider/stories/template.js
```diff
@@ -0,0 +1,42 @@
...
+import '../index.css';
+import '../skin.css';
+
+export const Template = ({
+  rootClass = "spectrum-ColorSlider",
+  customClasses = [],
+  isDisabled = false,
+  isFocused = false,
+  vertical,
+  style = {},
```
castastrophe, line 15: Is style still being used here as an input to color slider? I see colorHandleStyle being used but not where the style object is applied.

castastrophe (changes requested):

castastrophe (approved):

PR #1654, opened 2023-03-03
chore: tabs storybook

mlogsdon18:
## Description

This PR adds the Tabs storybook instance. 

The only additional options on the documentation site that I did not add were "[Basic Tabs with Overflow](https://opensource.adobe.com/spectrum-css/tabs.html#basictabswithoverflow)". This was really going to muddy up the template in my opinion and it just implemented two additional components that already have their own stories (Picker and Popover). There wasn't anything specific for tabs being passed into the components that I could see. Because of this, I felt that it wasn't needed. I'm happy to add it if others have different opinions.

### DIFF
--- a/components/tabs/stories/tabs.stories.js
+++ b/components/tabs/stories/tabs.stories.js
@@ -7,6 +7,7 @@ export default {
   component: "Tabs",
   argTypes: {
     items: { table: { disable: true }},
+    selectorWidth: { table: { disable: true }},
     size: {
       name: "Size",
       type: { name: "string", required: true },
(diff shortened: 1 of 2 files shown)


castastrophe (approved):

castastrophe:
@mlogsdon18 This looks really great!  Just a note that `selectorStyle` is showing at the top of the controls panel - might want to hide that from the table.

mlogsdon18:
> @mlogsdon18 This looks really great! Just a note that `selectorStyle` is showing at the top of the controls panel - might want to hide that from the table.

Oh good catch! Done!

PR #1655, opened 2023-03-03
chore: add modal to storybook

jawinn:
## Description
Adds a story to Storybook for the **Modal** component, with markup and classes matching the documentation site. Includes controls for `is-open` and its variant classes (e.g. `fullscreenTakeover`).

>**Note**: The different background colors on [the documentation page for modal](https://opensource.adobe.com/spectrum-css/modal.html) comes from `.spectrum-CSSExample-example--overlay`, which is not present in the story.

## Screenshots
Screenshot 2023-03-03 at 3 54 01 PM

## To-do list

### DIFF
--- a/components/modal/stories/modal.stories.js
+++ b/components/modal/stories/modal.stories.js
@@ -24,7 +24,10 @@ export default {
       control: {
         type: "select",
       }
-    }
+    },
+    content: { 
+      table: { disable: true }
+    },
   },
   args: {
     isOpen: true,
@@ -41,4 +44,8 @@ export default {
 };
 
 export const Default = Template.bind({});
-Default.args = {};
+Default.args = {
+  content: [
+    "This is a modal. Don't use it like this; get yourself a Modal.",
+  ],
+};
--- a/components/modal/stories/template.js
+++ b/components/modal/stories/template.js
@@ -9,6 +9,7 @@ export const Template = ({
   customClasses = [],
   isOpen = true,
   variant,
+  content = [],
   ...globals
 }) => {
   return html`
@@ -18,7 +19,7 @@ export const Template = ({
           [`${rootClass}--${variant}`]: typeof variant !== "undefined",
           'is-open': isOpen,
           ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
-      })}>This is a modal. Don't use it like this; get yourself a Modal.</div>
+      })}>${content}</div>
     </div>
   `;
 }


jawinn:
@castastrophe I wasn't quite sure about the inclusion and implementation of the `variant` control on this.

1. Should it be included?
The `spectrum-Modal--fullscreen` and `spectrum-Modal--fullscreenTakeover` options are not on the modal docs page, but can be found on the Dialog component docs. The CSS also says "@unofficial" on these, though they do appear to be official within Dialog examples? The CSS also contained the `spectrum-Modal--responsive`, that I didn't see within the Dialog docs.
2. If it should be included, I was flip-flopping between including "undefined" as an option within the available control options, or leaving as it currently is (defaults to none selected; reset controls can be used to get back to default). Having that option would make it easier to swap back to the default when trying out all combination of controls, but the readability of a select option saying "undefined" might be a little undesirable.
3. I wasn't 100% that those variants couldn't be used in combination with each other.

components/modal/stories/modal.stories.js
```diff
@@ -3,23 +3,37 @@ import { Template } from "./template";
...
 export default {
   title: "Modal",
-  description: "The Modal component is...",
+  description: "A modal component used primarily by Dialog.",
   component: "Modal",
   argTypes: {
-    size: {
-      name: "Size",
-      type: { name: "string", required: true },
+    isOpen: {
```
castastrophe, line 9: Would you add a content input here so other components can leverage the Modal component and pass in their own inputs? You can hide it from the controls panel with: `content: { table: { disable: true } }`

components/modal/stories/modal.stories.js
```diff
@@ -3,23 +3,37 @@ import { Template } from "./template";
...
+      },
+      control: "boolean",
+    },
+    variant: {
+      description: "Controls how the modal fills the available space. <ul><li>\"responsive\" will fill the screen on small viewports.</li><li>\"fullscreen\" will fill almost all of the available screen space.</li><li>\"fullscreenTakeover\" will fill all of the available screen space.</li></ul>",
+      defaultValue: undefined,
       table: {
         type: { summary: "string" },
         category: "Component",
+        defaultValue: { summary: "undefined" },
```
castastrophe, line 25:
```suggestion
```

Can you leave off the defaultValues? Since it's a CSS library, nothing is really added automatically - it's all opt-in. You can set default values in the args section below.

components/modal/stories/modal.stories.js
```diff
@@ -3,23 +3,37 @@ import { Template } from "./template";
...
+      control: {
+        type: "select",
+      }
+    }
   },
   args: {
-    rootClass: "spectrum-",
-    size: "m",
+    isOpen: true,
+    variant: undefined,
```
castastrophe, line 35: It will be undefined by default so no need to add this here, just leave it off the list.

components/modal/stories/modal.stories.js
```diff
@@ -3,23 +3,37 @@ import { Template } from "./template";
 
 export default {
   title: "Modal",
-  description: "The Modal component is...",
+  description: "A modal component used primarily by Dialog.",
```
castastrophe, line 6:
```suggestion
  description: "A modal component is a dialog box/popup window that is displayed on top of the current page."
```

Just adding a little more context.

castastrophe (changes requested):
Awesome! Thanks for tackling this. Can you take a look at adding the content input so we can nest this in other components?

components/modal/stories/modal.stories.js
jawinn: Resolved in [fb67da1](https://github.com/adobe/spectrum-css/pull/1655/commits/fb67da1623b9335c38aafbaba085e361e0d5f238)

components/modal/stories/modal.stories.js
jawinn: Sure. Resolved in [fb67da1](https://github.com/adobe/spectrum-css/pull/1655/commits/fb67da1623b9335c38aafbaba085e361e0d5f238)

components/modal/stories/modal.stories.js
jawinn: I've added that in [b3f186c](https://github.com/adobe/spectrum-css/pull/1655/commits/b3f186c6fcd11943b462ebee9e1ccd6e754202cf), trying to match the format of some of the other stories that are using an array.

castastrophe (approved):

PR #1657, opened 2023-03-06
chore(splitbutton): Add Storybook story + template

pfulton:
## Description
Adds Storybook story + template for SplitButton

 - **How this was tested:** 

## To-do list


components/splitbutton/stories/splitbutton.stories.js
```diff
@@ -16,9 +16,19 @@ export default {
...
+    variant: {
+      name: "Variant",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["cta", "primary", "secondary"],
+      control: "select",
+    }
```
castastrophe, line 28: Does this need `iconName: { table: { disable: true } },` added?

components/splitbutton/stories/splitbutton.stories.js
```diff
@@ -16,9 +16,19 @@ export default {
       options: ["s", "m", "l", "xl"],
       control: "select"
     },
+    variant: {
+      name: "Variant",
+      type: { name: "string", required: true },
```
castastrophe, line 21:
```suggestion
      type: { name: "string" },
```

Q: Does split button render okay if no variant class exists on it? If so, let's remove the required flag.

components/splitbutton/stories/template.js
```diff
@@ -2,19 +2,66 @@ import { html } from 'lit-html';
...
+        iconName,
+        isOpen: false,
+        customClasses: ["spectrum-SplitButton-trigger"]
+      })}
+    </div>
+    
+    <br />
+    <br />
+    
+    <div class=${classMap({
```
castastrophe, line 44: I think right & left need to be in the controls so that we can only render 1 split button in the template. That will let us embed the component in others as needed.

castastrophe: Picker button & infield button are doing something similar.

castastrophe (changes requested):

components/splitbutton/stories/splitbutton.stories.js
pfulton: Yes! I can add it.

components/splitbutton/stories/template.js
pfulton: Oh, yep yep. I wasn't thinking in terms of being able to reuse within other components. Good call-out.

components/splitbutton/stories/template.js
```diff
@@ -0,0 +1,71 @@
...
+  customClasses = [],
+  size = "m",
+  variant = "cta",
+  iconName = "ChevronDown100",
+  position = "left",
+  ...globals
+}) => {
+
+  return html`
+    ${when(position === "right", () =>
```
castastrophe, line 20: The markup for these 2 when statements seems really similar - is there a way to combine them?

castastrophe (changes requested):

components/splitbutton/stories/splitbutton.stories.js
```diff
@@ -0,0 +1,96 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Splitbutton",
+  description: "The Splitbutton component is...",
+  component: "Splitbutton",
```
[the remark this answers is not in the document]
pfulton: Thank you. I know there are a few components in our repo that have no description and also don't exist on the Spectrum site.

PR #1659, opened 2023-03-06
chore(well): clean up story and template

misterbrownlee:
## Description
Adding storybook stuff for Well

## To-do list

### DIFF
--- a/components/well/stories/template.js
+++ b/components/well/stories/template.js
@@ -1,6 +1,5 @@
 import { html } from 'lit-html';
 import { classMap } from 'lit-html/directives/class-map.js';
-// import { ifDefined } from 'lit-html/directives/if-defined.js';
 
 import "../index.css";
 import "../skin.css";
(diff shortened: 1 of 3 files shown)


components/well/stories/template.js
```diff
@@ -14,25 +13,13 @@ export const Template = ({
...
     <span class=${classMap({
       [rootClass]: true,
-      [`${rootClass}--size${size?.toUpperCase()}`]: typeof size !== "undefined",
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
     })}>
       <em>Well done is better than well said.</em>
       <br>
       <strong>Benjamin Franklin</strong>
       <br><br>
       Well said Ben!
```
castastrophe, line 22: Can you pass the content in as a variable so that we can embed this story in others as needed? We've been using the `content` variable name and pass the data in as an array. Something like this here:

```suggestion
      ${content}
```

and this in the stories.js:

```
content: [
  html`<em>Well done is better than well said.</em><br><strong>Benjamin Franklin</strong><br><br>Well said Ben!`
]

castastrophe (changes requested):
Looking great! If you wouldn't mind updating the content to be passed in, then we can embed this in other stories if needed.

components/well/stories/template.js
misterbrownlee: makes sense, thanks for the example!

castastrophe (approved):
🚀

PR #1661, opened 2023-03-07
chore: migrate check peer dependencies script to node

castastrophe:
## Description

Pulls out the peer dependency check into its own command to be run by CI in future. Validates that peer dependencies and dependencies are not mismatched.

 - **How this was tested:** `yarn check:peerDeps` (I went in and edited a file to be incorrect to test that it would catch broken cases as well).

## To-do list


pfulton (approved):
Looks good. I ran this locally, and also made a change to see if it would catch it.

PR #1669, opened 2023-03-08
chore(treeview): storybook and template

mlogsdon18:
## Description

This PR creates a storybook instance with multiple stories for the TreeView component.


components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,241 @@
...
+export const Thumbnails = Template.bind({});
+Thumbnails.args = {
+  items: [
+    {
+      id: "group1",
+      label: "Group 1",
+      link: "#",
+      isOpen: true,
+      thumbnail: {
+        imageURL: "img/thumbnail.png",
```
jenndiaz, line 171: I am getting a broken image on each of these thumbnails. Are we expecting them to load locally? 
I was able to get `example-ava.png` to load but did not see why these were not.

components/treeview/stories/template.js
```diff
@@ -0,0 +1,117 @@
...
+  items,
+  ...globals
+}) => {
+
+  return html`
+    <ul class=${classMap({
+      [rootClass]: true,
+      [`${rootClass}--size${size?.toUpperCase()}`]: typeof size !== "undefined",
+      [`${rootClass}--quiet`]: isQuiet,
+        ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
```
jenndiaz, line 107: I am noticing on the doc site the parent most UL has `spectrum-TreeView--thumbnail` when viewing the thumbnail variant but this does not. This class brings in some spacing between the `img` and the text

jenndiaz (commented):
🌳 Over all looked good to me. 
I had two questions/observations around the thumbnails story

components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,241 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
+  title: "Treeview",
```
jawinn, line 5: On the docs site, it's called "Tree View", so this should probably be two words to match that.

components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,241 @@
...
+      name: "Size",
+      type: { name: "string", required: true },
+      table: {
+        type: { summary: "string" },
+        category: "Component",
+      },
+      options: ["s", "m", "l", "xl"],
+      control: "select"
+    },
+    isQuiet: {
```
jawinn, line 20: Could you add `description` properties to the control(s)? Especially for `isQuiet`, since its effect may not be obvious.

components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,241 @@
...
+        {
+          id: "label3",
+          label: "Label 3",
+          link: "#"
+        },
+        {
+          id: "group3",
+          label: "Group 3",
+          link: "#",
+          items: [
```
jawinn, line 87: It might be worth including a 4th tier and an empty group within the test data somewhere, to help catch any edge cases. I saw both on the docs site examples.

components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,241 @@
...
+          }
+        },
+      ]
+    },
+  ]
+};
+
+export const WithSections = Template.bind({});
+WithSections.args = {
+  items: [
```
jawinn, line 200: Maybe add a second section of data here like on the docs site, so we can test that they're spaced the same when within the same parent.

jawinn (changes requested):
Nice work! I've left a few comments on some minor things.

I noticed that when clicking on an item within the tree (link with a href of `#`), that Storybook will oddly go fullscreen. I guess it's an [open Storybook issue](https://github.com/storybookjs/storybook/issues/15934) and the comments mention that adding a `target="_self"` could fix it? Worth a try.

The other thing I saw is more a question for @castastrophe about the component itself, and doesn't have to do with this PR. The example markup ([docs site too](https://opensource.adobe.com/spectrum-css/treeview.html#sections)), uses a `div` as a child of `ul` for the section heading, which isn't valid HTML. Is that a known issue?

components/treeview/stories/treeview.stories.js
mlogsdon18: Probably because I didn't add them 🙃 they should be there now!

components/treeview/stories/template.js
mlogsdon18: Good catch! I noticed that too so it should be there now! I added a `variant` prop since there is another instance where `standalone` might be passed in

components/treeview/stories/treeview.stories.js
mlogsdon18: It's a common styling type for Spectrum and we haven't been adding descriptions to controls really... I honestly don't know what the proper description for it would be. It makes the styling the Quiet type? I think it should be alright but I'll ask @castastrophe

castastrophe: Yeah we can leave it off for now. What I'd love to do eventually is create a standard set of common controls and have every component just import that. Once we migrate to using schemas to populate the controls, that will be much easier. This look great for now!

components/treeview/stories/treeview.stories.js
```diff
@@ -0,0 +1,258 @@
+// Import the component markup template
+import { Template } from "./template";
+
+export default {
```
castastrophe, line 4: These are fantastic examples! Thank you for tackling this!

castastrophe (approved):
This look great! Love the asset additions too!

jawinn (approved):
All good now! The `target="_self"` fix works.

PR #1671, opened 2023-03-09
chore: update netlify settings

castastrophe:
## Description

Looking at the netlify documentation for 11ty integration, it suggests updating the .gitignore to use `**/node_modules/**` rather than `node_modules`.

I also updated the netlify config to include:

- Build commands so that if they change locally, we can update them via the repo
- An alias from storybook to preview b/c storybook is a common url pattern
- From a component name as a folder to the component html asset
- Post-processing support for non-deploy preview builds (minification & bundling)

## To-do list

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 dist
 dist-site
-node_modules
+
+# Recommended update re:https://docs.netlify.com/integrations/frameworks/eleventy/
+**/node_modules/**
+
 temp
 .npmrc*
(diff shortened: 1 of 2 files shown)


pfulton (commented):
I'm on board for these changes! Thanks for digging in.

When I hit the deployed URL and try to click-through to get to Storybook, it fails, though.

pfulton (approved):

PR #1685, opened 2023-03-22
fix(textfield): quiet - use underline focus indicator on click

jawinn:
## Description
On the quiet type of the textfield component, the focus indicator was showing the full ring instead of the single bottom border underline when clicking into the field. The underline was as expected when tabbing into the field and on examples where the indicator shows with a class.

Chrome and Firefox, MacOS

## Screenshots

After clicking into a quiet field on the docs page for textfield--
Before:
Screenshot 2023-03-22 at 5 28 54 PM

After:
Screenshot 2023-03-22 at 5 23 29 PM

## To-do list

### DIFF
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -337,13 +337,14 @@ governing permissions and limitations under the License.
       margin-inline-start: 0;
       margin-block-start: 0;
       border-radius: 0;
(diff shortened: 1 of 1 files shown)


mlogsdon18 (approved):
Looks great Josh! Thanks for putting this up so quickly and catching it!

PR #1686, opened 2023-03-22
feat(colorhandle)!: token migration

jenndiaz:
## Description
BREAKING CHANGE: This migrates the ColorHandle component to core tokens.
Additionally: 
- uses opacity checkerboard tokens for ColorHandle 
- refactors how we achieve the checkerboard pattern by using `repeating-conic-gradient`
- Updates storybook template
-  adds keyboard focus to ColorHandle

PR is pointing to a feature branch which will include the migration of ColorHandle and ColorLoupe. 

comparing local build of with figma for ColorArea and with https://spectrum.adobe.com/page/color-loupe/

Browser(s) and OS(s) this was tested with:
Chrome 109 on macOS
Safari 16.2 on macOS
Firefox 109 on macOS

Checked WHCM in assistive labs

## Screenshots
Screenshot 2023-03-22 at 4 03 10 PM
Screenshot 2023-03-22 at 4 03 19 PM

### WHCM 
Screenshot 2023-03-24 at 9 28 22 AM

### keyboard focus displayed with colorwheel: 
Screenshot 2023-03-22 at 4 04 14 PM

## To-do list


components/colorarea/metadata/colorarea.yml
```diff
@@ -32,8 +32,8 @@ examples:
           </svg>
         </div>
 
-        <input type="range" class="spectrum-ColorArea-slider" name="x" aria-label="saturation and value" min="0" max="1" step="0.01">
-        <input type="range" class="spectrum-ColorArea-slider" name="y" aria-label="saturation and value" min="0" max="1" step="0.01">
+        <input type="range" class="spectrum-ColorArea-slider" name="x" aria-label="saturation and value" min="0" max="1" step="0.01" tabindex="-1">
+        <input type="range" class="spectrum-ColorArea-slider" name="y" aria-label="saturation and value" min="0" max="1" step="0.01" tabindex="-1">
```
mlogsdon18, line 36: I'm assuming you have a reason for removing these inputs from the tab order and adding in the div for the handle instead? I know I can't currently drag it so it's hard to test but that of course removes those from being accessible

mlogsdon18 (commented):
Looks good! My main question is just about all the `tabindex` updates. I'm sure you have a good reason for them but just want to make sure we aren't losing keyboard functionality

components/colorarea/metadata/colorarea.yml
jenndiaz: Yes! Previously the ColorHandles did not receive any visible keyboard focus. I added a tabIndex of 0 to the ColorHandle to add it to Tab order so we can now see the keyboard focus styling . I removed the slider inputs from the tab order because they are hidden from view and were causing a pause or skipping when tabbing through the component. 

[Also a pattern SWC has in place too](https://github.com/adobe/spectrum-web-components/blob/d4e49f581d41720d92e9f786cda4af56dbafc8c5/packages/color-area/src/ColorArea.ts#L423)

I am curious if @pfulton has any concerns about this that I might now be aware of though.

components/colorhandle/index.css
```diff
@@ -54,26 +89,55 @@
...
     position: absolute;
     display: block;
-    width: var(--spectrum-colorhandle-hitarea-size);
-    height: var(--spectrum-colorhandle-hitarea-size);
+    width: var(--mod-colorhandle-hitarea-size, var(--spectrum-colorhandle-hitarea-size));
+    height: var(--mod-colorhandle-hitarea-size, var(--spectrum-colorhandle-hitarea-size));
   }
 
   &.is-focused,
   &.focus-ring {
```
jawinn, line 101: Since this div is now being made focusable through `tabindex="0"` in the examples, should we also use `&:focus` here?

components/colorarea/metadata/colorarea.yml
mlogsdon18: Ok what SWC has in place functions with a keyboard and with a screen reader, that was my main concern. The only thing is that they have one input labeled for saturation and one for luminosity. Should we keep that consistent here to differentiate which value the user is changing?

Issue #1054, opened 2020-10-15
[Card] subtitle style incorrectly uses `text-transform: uppercase;`


adixon-adobe:
## Description
The subtitle isn't all-uppercase in the spectrum design examples https://spectrum.adobe.com/page/cards/

## Steps to reproduce
1. Go to http://opensource.adobe.com/spectrum-css/card
2. Inspect the subtitle in the examples, and you'll see "text-transform: uppercase;"

## Expected behavior
We shouldn't transform to uppercase. In general this is a really bad pattern to use for some languages as well, so it's not something we should use css to enforce.

lazd:
Looks like a design change, we should fix it. Also good call on the issue with other languages.

PR #1693, opened 2023-03-28
feat(card)!: update cards tokens

mlogsdon18:
## Description

This PR updates the Card component to use the new tokens. It also addresses [Issue #1054](https://github.com/adobe/spectrum-css/issues/1054) removing `text-transform: uppercase` from Card component subtitle.

This was noted as not having design changes but the tokens did cause some design changes. Most notably: 
- Spacing between the title and content increased
- `min-width` of the card has decreased
- The Subtitle has been removed [per design request](https://adobedesign.slack.com/archives/GQ09RCBA5/p1680228713451179?thread_ts=1680104240.442689&cid=GQ09RCBA5) for this first iteration 

## Issues Addressed
- [Issue #1054](https://github.com/adobe/spectrum-css/issues/1054) 

 - Chrome 111.0.5563.110 on MacOS Monterey
 - Safari 16.2
 - Firefox 111.0

## To-do list

- [x] All tokens have been added or hardcoded


components/card/index.css
```diff
@@ -85,88 +193,106 @@ governing permissions and limitations under the License.
...
+  padding-inline-start: var(--mod-card-body-spacing, var(--spectrum-card-body-spacing));
+  padding-block-end: var(--mod-card-body-padding-bottom, var(--spectrum-card-body-padding-bottom));
 }
 
 .spectrum-Card-preview {
   overflow: hidden;
-  border-top-left-radius: calc(var(--spectrum-card-border-radius) - 1px);
-  border-top-right-radius: calc(var(--spectrum-card-border-radius) - 1px);
+  border-top-left-radius: var(--mod-card-corner-radius, var(--spectrum-card-corner-radius));
+  border-top-right-radius: var(--mod-card-corner-radius, var(--spectrum-card-corner-radius));
```
jenndiaz, line 235: same thing comment here too, should these be [logical properties for border radius](https://css-tricks.com/css-logical-properties-and-values/#aa-border-radius)

components/card/index.css
```diff
@@ -85,88 +193,106 @@ governing permissions and limitations under the License.
...
 
   display: flex;
   align-items: center;
   justify-content: center;
 
-  border-block-end: var(--spectrum-card-coverphoto-border-bottom-size) solid transparent;
-  border-top-left-radius: calc(var(--spectrum-card-border-radius) - 1px);
-  border-top-right-radius: calc(var(--spectrum-card-border-radius) - 1px);
+  border-top-left-radius: var(--mod-card-corner-radius, var(--spectrum-card-corner-radius));
+  border-top-right-radius: var(--mod-card-corner-radius, var(--spectrum-card-corner-radius));
```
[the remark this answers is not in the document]
mlogsdon18: Good call! I forget that border has logical properties

components/card/package.json
```diff
@@ -23,18 +23,18 @@
...
     "@spectrum-css/checkbox": "^3.1.3",
-    "@spectrum-css/component-builder": "^4.0.4",
-    "@spectrum-css/icon": "^3.0.36",
-    "@spectrum-css/quickaction": "^3.0.38",
-    "@spectrum-css/typography": "^4.0.30",
-    "@spectrum-css/vars": "^9.0.0",
+    "@spectrum-css/component-builder-simple": "^2.0.6",
+    "@spectrum-css/icon": "^3.0.35",
+    "@spectrum-css/quickaction": "^3.0.37",
+    "@spectrum-css/tokens": "^7.6.0",
```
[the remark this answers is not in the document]
mlogsdon18: whoops I did and didn't push... should be good to go now!

components/card/index.css
```diff
@@ -10,7 +10,79 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+
+}
+
+.spectrum-Card--sizeS {
+  /* Need less spacing for small cards since we've decreased min-width */
+  --spectrum-card-actions-spacing: calc(var(--spectrum-spacing-300) / 2);
+
+}
 
 .spectrum-Card--sizeS {
```
pfulton, line 87: We'll need to remove the @remapvars calls here and in `.spectrum-Card--sizeM`, too. This was functionality for the older "vars" tokens system that we're no longer using.

GitHub wasn't allowing me to highlight and call-out the code here (I guess because it wasn't changed?) but I'm thinking we should remove lines 87 - 101.

components/card/index.css
```diff
@@ -35,15 +107,57 @@ governing permissions and limitations under the License.
...
 
-  &:focus {
+  &::after {
+    content: '';
+    block-size: calc(100% + (var(--mod-card-focus-indicator-gap, var(--spectrum-card-focus-indicator-gap)) * 2));
+    inline-size: calc(100% + (var(--mod-card-focus-indicator-gap, var(--spectrum-card-focus-indicator-gap)) * 2));
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: 0;
```
pfulton, line 126: Would you mind changing these to use their [logical counterparts](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties#properties_for_floating_and_positioning)?

components/card/index.css
```diff
@@ -176,30 +304,28 @@ governing permissions and limitations under the License.
...
+    min-block-size: var(--mod-card-preview-minimum-height, var(--spectrum-card-preview-minimum-height));
     inline-size: 100%;
     flex: 1;
-    padding: var(--spectrum-card-quiet-preview-padding);
     margin: 0 auto;
     box-sizing: border-box;
-    border-radius: var(--spectrum-card-quiet-border-radius);
     position: relative;
-    transition: background-color var(--spectrum-global-animation-duration-100);
+    /*transition: background-color var(--spectrum-global-animation-duration-100);/* TODO need token */
```
pfulton, line 326:
```suggestion
    transition: background-color var(--spectrum-animation-duration-100);
```
I'm not sure when we'll see these in the `@adobe/spectrum-tokens` package, but we do have these available in [our custom tokens](https://github.com/adobe/spectrum-css/blob/e8502f3cfa7547b8b1e3f1dbf8b14083ceb5ba49/components/tokens/custom-spectrum/custom-vars.css#L19-L36).

pfulton (changes requested):
A few things here @mlogsdon18, but otherwise, it's looking pretty good.

I also noticed that there is a size toggle on the Storybook story for this component, but changing the size doesn't seem to do anything. Would you mind looking into that, as well?

components/card/index.css
```diff
@@ -85,88 +173,106 @@ governing permissions and limitations under the License.
...
+  padding-inline-end: var(--mod-card-subtitle-padding-right, var(--spectrum-card-subtitle-padding-right));
+}
+
+.spectrum-Card-description {
+  font-family: var(--mod-card-body-font-family, var(--spectrum-card-body-font-family));
+  font-size: var(--mod-card-body-font-size, var(--spectrum-card-body-font-size));
+  font-weight: var(--mod-card-body-font-weight, var(--spectrum-card-body-font-weight));
+  font-style: var(--mod-card-body-font-style, var(--spectrum-card-body-font-style));
+  line-height: var(--mod-card-body-line-height, var(--spectrum-card-body-line-height));
+  color: var(--highcontrast-card-body-font-color, var(--mod-card-body-font-color, var(--spectrum-card-body-font-color)));
```
pfulton, line 259: Do we also need to provide `.spectrum-Card-footer` with the same `color` definition? I ask, because when I switch to dark, the footer text is still coming through with its default color:

Screen Shot 2023-04-17 at 4 47 55 PM

pfulton (changes requested):

components/card/index.css
mlogsdon18: Yes I'll do that! Footer did not have any specific tokens so I was doing my best here to keep this backwards compatible so we'll definitely want some color

components/card/metadata/card.yml
```diff
@@ -8,10 +8,11 @@ sections:
 
       ### Change workflow icon size to medium
       Please replace `.spectrum-Icon--sizeS` with `.spectrum-Icon--sizeM`.
-      ### Use Spectrum Heading for title and Spectrum Detail for subtitle
-      To make the card component more flexible, products can have more fine control over the typography by using the Heading and Detail components with their corresponding sizes. To be more consistent with previous card designs you can use the following:
+      ### Use Spectrum Heading for title 
+      To make the card component more flexible, products can have more fine control over the typography by using the Heading component with its corresponding sizes. To be more consistent with previous card designs you can use the following:
         * Add the `spectrum-Heading` and `spectrum-Heading--sizeXS` or `spectrum-Heading--sizeXXS` to the `spectrum-Card-title` element.
```
castastrophe, line 13: I wonder if it would be valuable to extend the typography styles in the card so that users don't have to add the classes and pull in the typography styles separately since they're quite detailed. Or is the flexibility of using any class desirable?

components/card/metadata/card.yml
```diff
@@ -32,10 +33,11 @@ examples:
               </div>
             </div>
             <div class="spectrum-Card-content">
-              <div class="spectrum-Card-subtitle spectrum-Detail spectrum-Detail--sizeS">jpg</div>
+              <div class="spectrum-Card-description">Optional description should be kept to one or two lines</div>
```
castastrophe, line 36:
```suggestion
              <div class="spectrum-Card-description">Optional description; should be kept to one or two lines</div>
```

Minor nit - I was just thinking these are 2 distinct notes (that description is optional and that it should be 1 or 2 lines).

components/card/metadata/card.yml
```diff
@@ -32,10 +33,11 @@ examples:
...
             </div>
           </div>
           <div class="spectrum-Card-footer">
+            <hr class="spectrum-Divider spectrum-Divider--sizeS">
```
castastrophe, line 40: Instead of a divider component, should we take a similar tact of extending the styles from divider as a footer border-top maybe?

components/card/package.json
```diff
@@ -23,18 +23,18 @@
     "@spectrum-css/checkbox": "^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0",
     "@spectrum-css/icon": "^3.0.0",
     "@spectrum-css/quickaction": "^3.0.0",
-    "@spectrum-css/typography": "^4.0.0",
-    "@spectrum-css/vars": ">=9.0.0"
+    "@spectrum-css/tokens": "^8.0.1",
+    "@spectrum-css/typography": "^4.0.0"
```
castastrophe, line 27:
```suggestion
    "@spectrum-css/tokens": ">=8 <=9",
    "@spectrum-css/typography": ">=3 <=4"
```

Should we move to making peer dependencies more permissive to allow more cross-component compatibility?

components/card/package.json
```diff
@@ -23,18 +23,18 @@
...
     "@spectrum-css/actionbutton": "^3.0.31",
     "@spectrum-css/asset": "^3.0.35",
     "@spectrum-css/checkbox": "^6.0.19",
-    "@spectrum-css/component-builder": "*",
     "@spectrum-css/icon": "^3.0.38",
     "@spectrum-css/quickaction": "^3.0.43",
     "@spectrum-css/typography": "^4.0.31",
-    "@spectrum-css/vars": "^9.0.2",
+    "@spectrum-css/component-builder-simple": "^2.0.6",
+    "@spectrum-css/tokens": "^8.0.1",
```
castastrophe, line 37: This will likely need to be manually bumped. Thinking out loud but I should add a backlog story to add a utility to handle version bumps in WIP branches when things change during review.

components/card/stories/card.stories.js
```diff
@@ -132,15 +132,15 @@ export default {
...
-    },
+    // subtitle: {
+    //   name: "Subtitle",
+    //   type: { name: "string" },
+    //   table: {
+    //     type: { summary: "string" },
+    //     category: "Content",
+    //   },
+    //   control: { type: "text" },
+    // },
```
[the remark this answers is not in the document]
castastrophe: Would you mind adding a comment above it indicating that so we don't forget and remove it? Just something like: `// will likely be brought back by design in a future iteration`

castastrophe (approved):

components/card/metadata/card.yml
mlogsdon18: I assumed we were wanting the flexibility of being able to use any class but that's not a call for me to make!

components/card/metadata/card.yml
mlogsdon18: That was what I was doing originally but was told by design that this is supposed to be a Divider component...

components/card/package.json
mlogsdon18: I think that makes sense. We have a bunch of or's going on for the other peer dependencies so this could also clean that up a bit

components/card/themes/express.css
pfulton, line 1: Would you mind adding the empty container query here and in the `spectrum.css` file, too?

```css
@container (--system: express) {
  .spectrum-Card {
  }
}
```

pfulton (changes requested):
This looks awesome, @mlogsdon18. I noted one thing on the theme files, and also:

I noticed in dark and darkest that the horizontal variation's icon is _very_ difficult to see.

Screenshot 2023-05-16 at 4 55 33 PM

components/card/index.css
```diff
@@ -10,40 +10,168 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  --spectrum-card-subtitle-padding-right: var(--spectrum-spacing-100);
+
+  --spectrum-card-border-width: var(--spectrum-border-width-100);
+  --spectrum-card-corner-radius: var(--spectrum-corner-radius-100);
+  --spectrum-card-border-color: var(--spectrum-gray-200);
+  --spectrum-card-border-color-hover: var(--spectrum-gray-300);
+  --spectrum-card-border-color-selected: var(--spectrum-blue-700);
+
+  /* Typography */
+  --spectrum-card-title-font-family: var(--spectrum-sans-font-family-stack);
```
castastrophe, line 31: Woo! 🎉

components/card/index.css
```diff
@@ -10,40 +10,168 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+  }
 
-.spectrum-Card--sizeS {
-  @remapvars {
-    find: /--spectrum-card(.*)-s-/;
-    filter: color;
-    replace: --spectrum-card$1-;
+  /* Quiet/Gallery */
+  &--quiet,
+  &--gallery {
```
castastrophe, line 77: OMG so exciting to see this approach in action!

components/card/index.css
```diff
@@ -10,40 +10,168 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    inline-size: 100%;
+    position: absolute;
+    inset-inline-start: 0;
+    inset-block-start: 0;
+  }
 
-  &:focus {
+  /* Focus indicator - use after so change in border-width does not affect card width */
+  &::after {
+    content: '';
```
castastrophe, line 123: I feel like I've seen t his set of styles on other components - I might noodle on if a placeholder would be helpful for this in future. Maybe post-build rewrite

components/card/index.css
```diff
@@ -60,196 +188,282 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Card-actions {
+.spectrum-Card .spectrum-QuickActions {
```
castastrophe, line 191: Leveraging our new custom properties over nested class names approach, are there mods we can hook into on the Quick actions component rather than including the class here?

components/card/index.css
```diff
@@ -60,196 +188,282 @@ governing permissions and limitations under the License.
...
+    &::after {
+      border-width: 0;
+    }
+    .spectrum-Card-preview {
+      &::after {
+        border-width: var(--mod-card-focus-indicator-width, var(--spectrum-card-focus-indicator-width));
+        border-color: var(--mod-card-focus-indicator-color, var(--spectrum-card-focus-indicator-color));
+      }
+    }
+  }
```
castastrophe, line 329:
```suggestion
  &:focus::after,
  &.is-focused::after {
    border-width: 0;
  }
  
  &:focus .spectrum-Card-preview::after,
  &.is-focused .spectrum-Card-preview::after {
      border-width: var(--mod-card-focus-indicator-width, var(--spectrum-card-focus-indicator-width));
      border-color: var(--mod-card-focus-indicator-color, var(--spectrum-card-focus-indicator-color));
  }
```

I found this nesting a bit difficult to grok at first - do you think it would be beneficial to pull them out like this for readability/debugging purposes?

components/card/package.json
```diff
@@ -24,17 +24,17 @@
     "@spectrum-css/icon": ">=3",
     "@spectrum-css/quickaction": ">=3",
     "@spectrum-css/typography": ">=4 <=5",
-    "@spectrum-css/vars": ">=9"
+    "@spectrum-css/tokens": ">=8 <=9"
```
castastrophe, line 27:
```suggestion
    "@spectrum-css/tokens": ">=8 <=10"
```

I believe the token package is up to 10 now.

components/card/package.json
```diff
@@ -24,17 +24,17 @@
...
     "@spectrum-css/actionbutton": "^3.0.41",
     "@spectrum-css/asset": "^3.0.39",
     "@spectrum-css/checkbox": "^6.0.29",
-    "@spectrum-css/component-builder": "^4.0.8",
     "@spectrum-css/icon": "^3.0.42",
     "@spectrum-css/quickaction": "^3.0.54",
     "@spectrum-css/typography": "^5.0.15",
-    "@spectrum-css/vars": "^9.0.5",
+    "@spectrum-css/component-builder-simple": "^2.0.6",
+    "@spectrum-css/tokens": "^9.0.4",
```
castastrophe, line 37:
```suggestion
    "@spectrum-css/tokens": "^10.0.0",
```

components/card/index.css
castastrophe: Good point - I hadn't even thought about that factor!

components/card/index.css
mlogsdon18: I felt like it was a lot to. I'll play around with it! Your approach is already easier to read.

mlogsdon18: I added a comment and condensed the preview selector. I felt like since they were both using the focus state I didn't want to split them up entirely. Thoughts?

mlogsdon18:
@pfulton The issue with the icon should be fixed now, as well as the other issues that you and Cassondra mentioned!

components/card/index.css
mlogsdon18: @castastrophe question I just thought of. Should the variant variables be mods or not? 
`--spectrum-card-content-margin-top--quiet` vs `--mod-card-content-margin-top--quiet`? Since we are having them set mods normally, I'm wondering if these should also be mods to avoid confusion?

pfulton:
Released: `@spectrum-css/card@6.0.0-beta.1`

PR #1695, opened 2023-03-29
feat(inlinealert)!: updates tokens

jenndiaz:
## Description

Updates a number of tokens for Inline Alert

- Updated border from border-width-100 to border-width-200
- Updated typography style tokens for title and description (using typography style tokens)
- Changed Background from gray-50 to background-layer-2-color
- Changed Border from gray-600 to neutral-visual-color
- Changed Border and icon (informative) from Blue 800 and Blue 1000 to informative-visual-color
- Changed Border and icon (positive) from Green 800 and Green 1000 to positive-visual-color
- Changed Border and icon (notice) from Orange 600 and Orange 800 to notice-visual-color
- Changed Border and icon (negative) from Red 800 and Red 1000 to negative-visual-color
- Changed Title and description from Gray 900 to heading-color and body-color

Additionally:
- Updates story so `closable` it not a variant and is only an advanced boolean, since all variants can be closable
- Updates `mods.md`

comparing local build with XD file for In-line Alert

### Browser(s) and OS(s) this was tested with:
Chrome 109 on macOS
Safari 16.2 on macOS
Firefox 109 on macOS

## To-do list

### DIFF
--- a/components/inlinealert/package.json
+++ b/components/inlinealert/package.json
@@ -18,11 +18,11 @@
     "build": "gulp"
   },
   "peerDependencies": {
-    "@spectrum-css/icon": "^3.0.0",
-    "@spectrum-css/tokens": "^7.0.0"
+    "@spectrum-css/icon": "^3.0.35",
+    "@spectrum-css/tokens": "^7.7.0"
   },
   "devDependencies": {
-    "@spectrum-css/button": "^6.0.21",
+    "@spectrum-css/button": "^9.0.11",
     "@spectrum-css/component-builder-simple": "^2.0.7",
     "@spectrum-css/icon": "^3.0.35",
     "@spectrum-css/tokens": "^7.7.0",


components/inlinealert/index.css
```diff
@@ -12,10 +12,20 @@ governing permissions and limitations under the License.
 
 .spectrum-InLineAlert {
   /* Font */
-  --spectrum-inlinealert-font-weight: var(--spectrum-bold-font-weight);
+  --spectrum-inlinealert-heading-font-family: var(--spectrum-heading-sans-serif-font-family);
```
mlogsdon18, line 15: This is something that I just ran into with the card component that we need to ask Garth about. The `--spectrum-heading-sans-serif-font-family` and `--spectrum-body-sans-serif-font-family` tokens are incorrect which you can see in the screenshot that the font it's showing is a default serif font, not sans-serif. I think the issue is that the token is set to `Adobe Clean` and it needs to be set to `abode-clean`
Screenshot 2023-03-29 at 1 49 15 PM

jawinn: I've noticed this as well while working on other components. The Figma defined font-family Figma tokens do not translate to a usable font stack. I have been using `--spectrum-sans-font-family-stack`, and have been meaning to ask if this is correct.

pfulton: Oh interesting: this actually looks fine to me, but I also have Adobe Clean installed locally. Theoretically, it should be falling back to `adobe-clean` which would be provided by Typekit on the docs site. Might need to dig into this a little more.

pfulton: I wrote a bit more about this in [Slack](https://adobedesign.slack.com/archives/G019JTYMT6H/p1680114911681059), but: [we have font stacks](https://github.com/adobe/spectrum-css/blob/32efe27a3dbc89da7bb9882004eebe318e0ce060/components/tokens/custom-spectrum/custom-vars.css#L38-L47) that we'll need to use for `font-family` things. Sorry that we didn't discuss this earlier. Not all of our components define `font-family` and this is the first time I think it's come up with y'all.

pfulton (approved):

PR #1719, opened 2023-04-03
chore: bump supported node versions

castastrophe:
- Up supported node version to v16

## To-do list


package.json
```diff
@@ -77,6 +77,9 @@
     "through2": "^3.0.1",
     "yargs": "^17.6.2"
   },
+  "engines": {
+    "node": ">=14.0.0 <18"
```
pfulton, line 81: Should we:

```suggestion
    "node": ">=16.0.0 <18"
```

pfulton (commented):
One thought here @castastrophe

package.json
castastrophe: I tend to opt for permissive for flexibility. Our project does work with 14.

pfulton (approved):

PR #1752, opened 2023-04-05
fix(textfield, stepper) : textfield and stepper accessibility fixes

mlogsdon18:
This PR makes two updates to help with accessibility for the Textfield and Stepper components based on feedback from the SWC team:
- Updates the WHCM color for the focus ring for Textfield in both the default and quiet variants to be `HIghlight`
- Updates the Stepper button width to match the current wider implementation to allow for a larger target size

Previous WHCM Focus
Screenshot 2023-04-05 at 3 17 41 PM

Updated WHCM Focus
Screenshot 2023-04-05 at 3 19 45 PM

### DIFF
--- a/components/stepper/index.css
+++ b/components/stepper/index.css
@@ -60,8 +60,8 @@ governing permissions and limitations under the License.
   --spectrum-stepper-border-radius: var(--spectrum-corner-radius-100);
 
   /*** Buttons ***/
-  --spectrum-stepper-button-width: var(--spectrum-component-height-75); /* this height token matches the previous WIDTH token */
-  --spectrum-stepper-button-padding: var(--spectrum-component-edge-to-visual-50); /* 6px md + 7px lg - previously 6px md + 7.5px lg */
+  --spectrum-stepper-button-width: calc(var(--spectrum-spacing-400) - var(--spectrum-stepper-border-width)); /* this matches the previous WIDTH token */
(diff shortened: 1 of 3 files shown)


pfulton (approved):

pfulton:
@Westbrook would you mind taking a quick look at this one? Thanks!

mlogsdon18:
@Westbrook 
I just got it back to the original setup that was in the VRT. Since we want to give the Textfield a blue highlighted border, it was double layering the border and was looking strange keeping the focus ring a few pixels larger (see first screenshot). I can keep both blue borders like that or I can make an update to prevent WHCM from adjusting the border color and keep it white (second screenshot). 

Screenshot 2023-04-07 at 10 32 35 AM

or

Screenshot 2023-04-07 at 10 35 34 AM

Westbrook:
If design doesn't have specific direction for this, maybe @jnurthen would have some input from an accessibility angle?

jnurthen:
I vote for the 2nd, leave the outline white and add the larger focus indicator in Highlight (essentially the reverse of before)

mlogsdon18:
Ok @pfulton @Westbrook I've updated this to match James' recommendation so if you could take one last look that would be great!
Screenshot 2023-04-17 at 2 12 40 PM

pfulton:
This looks good to me. I'll defer to @Westbrook for the final approval.

pfulton (approved):

mlogsdon18:
@pfulton could we get this merged in and a new release of textfield and stepper? Josh's search field updates in SWC are blocked by the textfield work so I'd like to get it merged in if we can!

castastrophe (approved):

PR #1761, opened 2023-04-10
chore(search)!: core tokens migration

jawinn:
## Description

**Core tokens migration:** 
Updates the `search` component to use core tokens defined on the Figma design.
This also includes:
- Support for t-shirt sizes
- Updates to docs to include additional examples, and storybook updates
- Fix: inputted text no longer goes underneath the reset/clear button
- Fix: correct disabled color is used for magnifying glass icon
- Fix: round edges of button no longer appear on top of border in WHCM/forced-color mode

Also updates the `textfield` component to address the TODO comment "_reevaluate icon styles when search field is migrated_". The search icon related styles and story entry have been removed from the textfield component, and a couple lines of CSS brought over to `search`. These were not a part of the design of textfield and were specific to `search`.

CSS-384

Tested on MacOS Chrome, Firefox, and Safari. Tested in Edge WHCM via AssistivTunnel.

## To-do list


components/search/index.css
```diff
@@ -41,43 +169,168 @@ governing permissions and limitations under the License.
...
-        var(--spectrum-textfield-m-texticon-icon-gap) -
-        var(--spectrum-textfield-m-texticon-border-size)
+      var(--mod-search-edge-to-visual, var(--spectrum-search-edge-to-visual)) +
+      var(--mod-search-icon-size, var(--spectrum-search-icon-size)) +
+      var(--mod-search-text-to-icon, var(--spectrum-search-text-to-icon))
     );
+    padding-inline-end: var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size));
+  }
+
+  &:not(:has(.spectrum-Search-clearButton)) .spectrum-Search-input {
```
jawinn, line 294: I wasn't sure about this addition. The design shows spacing on the right of the input without the clear button using this token. I thought this might be a nice enhancement in the case of someone not wanting to use the clear button. Browsers without `:has` support would just have a little extra right padding. If that use case is not necessary to support, I could remove this CSS. If it is necessary to support and we want greater than `:has` support, then we'd probably need to add another parent class (e.g. has-reset).

components/search/metadata/search.yml
```diff
@@ -1,5 +1,5 @@
 name: Search
-description: 'The searchy, cleary field.'
+description: 'This component contains a single input field with both a magnifying glass icon and a \"reset\" button displayed within it.'
```
mlogsdon18, line 2: Hahaha way to write an actual description

components/search/themes/spectrum.css
```diff
@@ -0,0 +1,56 @@
...
+        --spectrum-search-border-radius: var(--spectrum-corner-radius-100);
+        --spectrum-search-edge-to-visual: var(--spectrum-component-edge-to-visual-300);
+        --spectrum-search-edge-to-text: var(--spectrum-component-edge-to-text-300);
+    }
+
+    .spectrum-Search--quiet {
+        --spectrum-search-border-radius: 0;
+        --spectrum-search-edge-to-text: var(--spectrum-field-edge-to-visual-quiet);
+        --spectrum-search-edge-to-visual: var(--spectrum-field-edge-to-visual-quiet);
+    }
```
pfulton, line 54: It looks like these custom properties are the same in Spectrum as they are in Express. Are we able to pull these out and place them into `index.css` rather than having them in these theme files?

components/search/index.css
pfulton: I think this bit:

> case of someone not wanting to use the clear button

is a good question for design folks. "Can people use this without the clear button?"

I don't mind the thought of this, though! And yes, you're right: we're not yet able to fully support/use `:has` for critical UI, but if folks who use browsers that don't yet support `:has` don't really see a difference, I think this is a fine progressive enhancement.

pfulton (changes requested):
Nice work, @jawinn - lookin' good!

Just a few things here + whatever comes out of the design review.

components/search/themes/spectrum.css
jawinn: I've moved these back with the most recent commit. I originally had them in index.css, but they were being overridden by the system specific files (that get tacked on at the end). The adjustment needed was to increase specificity for those particular quiet properties in index.css.

components/search/index.css
jawinn: I've removed this, per our discussion, and the discussion with the designer. The component shouldn't be used without the clear button, so that use case does not need to be supported right now.

pfulton (approved):

components/search/index.css
```diff
@@ -10,21 +10,173 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+
+  .spectrum-Search-textfield.is-keyboardFocused & {
+    color: var(--highcontrast-search-color-focus, var(--mod-search-color-key-focus, var(--spectrum-search-color-key-focus)));
+  }
+
+  .spectrum-Search-textfield.is-disabled &,
+  .spectrum-Search-textfield.is-disabled:hover & {
+    color: var(--highcontrast-search-color-disabled, var(--mod-search-color-disabled, var(--spectrum-search-color-disabled)));
+  }
 }
```
castastrophe, line 180:
```suggestion
.spectrum-Search-icon {
  --spectrum-search-color: var(--highcontrast-search-color-default, var(--mod-search-color-default, var(--spectrum-search-color-default)));
  color: var(--spectrum-search-color);

  .spectrum-Search-textfield:hover & {
    --spectrum-search-color: var(--highcontrast-search-color-hover, var(--mod-search-color-hover, var(--spectrum-search-color-hover)));
  }

  .spectrum-Search-textfield.is-focused & {
    --spectrum-search-color: var(--highcontrast-search-color-focus, var(--mod-search-color-focus, var(--spectrum-search-color-focus)));
  }

  .spectrum-Search-textfield.is-focused:hover & {
    --spectrum-search-color: var(--highcontrast-search-color-focus, var(--mod-search-color-focus-hover, var(--spectrum-search-color-focus-hover)));
  }

  .spectrum-Search-textfield.is-keyboardFocused & {
    --spectrum-search-color: var(--highcontrast-search-color-focus, var(--mod-search-color-key-focus, var(--spectrum-search-color-key-focus)));
  }

  .spectrum-Search-textfield.is-disabled &,
  .spectrum-Search-textfield.is-disabled:hover & {
    --spectrum-search-color: var(--highcontrast-search-color-disabled, var(--mod-search-color-disabled, var(--spectrum-search-color-disabled)));
  }
}
```

This seems like a great place to use a local variable approach with variable overrides based on the specificity of the selectors. What do you think?

components/search/index.css
```diff
@@ -41,43 +193,171 @@ governing permissions and limitations under the License.
...
+
+  &::placeholder {
+    color: var(--highcontrast-search-color-default, var(--mod-search-color-default, var(--spectrum-search-color-default)));
+  }
+
+  /* Hover */
+  .spectrum-Search-textfield:hover &,
+  &:hover {
+    color: var(--highcontrast-search-color-hover, var(--mod-search-color-hover, var(--spectrum-search-color-hover)));
+    border-color: var(--highcontrast-search-border-color-hover, var(--mod-search-border-color-hover, var(--spectrum-search-border-color-hover)));
```
castastrophe, line 226:
```suggestion
  &:hover {
    --spectrum-search-color: var(--highcontrast-search-color-hover, var(--mod-search-color-hover, var(--spectrum-search-color-hover)));
    --spectrum-search-border-color: var(--highcontrast-search-border-color-hover, var(--mod-search-border-color-hover, var(--spectrum-search-border-color-hover)));
```

^ this might be another great place for state-based variable override

components/search/package.json
```diff
@@ -20,15 +20,15 @@
   "peerDependencies": {
     "@spectrum-css/clearbutton": "^1.0.0",
     "@spectrum-css/icon": "^3.0.0",
-    "@spectrum-css/textfield": "^3.0.0 || ^4.0.0",
-    "@spectrum-css/vars": ">=9.0.0"
+    "@spectrum-css/textfield": "^4.0.0",
+    "@spectrum-css/tokens": ">=8.0.0"
```
castastrophe, line 24:
```suggestion
    "@spectrum-css/textfield": ">=4",
    "@spectrum-css/tokens": ">=8 <=9"
```

components/search/package.json
```diff
@@ -20,15 +20,15 @@
...
   },
   "devDependencies": {
     "@spectrum-css/clearbutton": "^1.2.25",
-    "@spectrum-css/component-builder": "*",
+    "@spectrum-css/component-builder-simple": "^2.0.5",
     "@spectrum-css/icon": "^3.0.38",
     "@spectrum-css/textfield": "^4.1.11",
-    "@spectrum-css/vars": "^9.0.2",
+    "@spectrum-css/tokens": "^8.1.1",
```
castastrophe, line 31: Need to bump this to whatever the latest version is

castastrophe (changes requested):
I noticed a note on textfield relating to search and how it handles icons in the input field - just wanted to check-in and see if that is addressed in this PR or being addressed elsewhere?

components/search/index.css
jawinn: I don't think I'd be able to do that here with all of these pseudos for the textfield input, because each one is overriding the color and border-color being set on the textfield subcomponent (unless I added some extra specificity to the selector that would set the `color:` to the local variable)

components/search/package.json
jawinn: I wanted to confirm that `@spectrum-css/tokens` change here. Did you mean for that to be `"@spectrum-css/tokens": ">=9"`? I've been meaning to ask about how we're generally doing these now. I think I've been seeing most of the devDependencies using the `>=` now for tokens (and vars) in recent changes. Are we using `>=` now generally for the components as well now? (`@spectrum-css/textfield` change suggested here, and I saw a few others in a recent rebase).

jawinn:
> I noticed a note on textfield relating to search and how it handles icons in the input field - just wanted to check-in and see if that is addressed in this PR or being addressed elsewhere?

Also, to this question; I'm not sure which note you're referring to with Textfield icons, so I don't know the answer. This PR doesn't address any changes to Textfield. It handles the left side search icon with its own styles.

components/search/index.css
jawinn: This whole section has been replaced with custom property passthroughs 🎉

components/search/index.css
castastrophe: There's a couple of benefits to updating custom properties instead of css properties for state:

- reusability: you can use the same updated variable to apply to multiple properties (i.e., calculations) or influence multiple elements (i.e., children element changes) but can update the value without having to redefine every place it was used.
- clarity: since the values cascade, we don't have to define them on the exact element they apply to, we can instead apply all state-based changes directly to the state's class and let those values influence any children they might need to.

jawinn: I've made your suggested change in the latest series of commits.

jawinn:
I've addressed the comment from textfield "_TODO - reevaluate icon styles when search field is migrated_" in the latest commit "refactor: search and textfield search icon leftovers". I think this is what you were referring to @castastrophe , with the "note on textfield". It ended up needing to be addressed sooner than later, as those styles were affecting search. This now removes those unused search specific styles from textfield; more details in the commit message.

pfulton:
Released: `@spectrum-css/search@6.0.0-beta.2`

PR #1769, opened 2023-04-11
docs(button): only show icons when appropriate in Storybook

pfulton:
* Fixes an issue where the `iconOnly` variants were not rendering as perfect circles at the `l` and `xl` t-shirt sizes
* Fixes an issue where icons were being set for each Button example in Storybook.
* Adds static background colors for static examples for better rendering.

Here's a list of stories that were affected and always showed an icon when they previously were not:
* Button Group (both default and vertical stories)
* Pagination
* Split Button (all stories)
* Toast

Examples of misshapen iconOnly buttons:

Before:
Screen Shot 2023-04-11 at 4 07 57 PM

After:
Screen Shot 2023-04-11 at 4 08 14 PM

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/button/stories/button.stories.js
+++ b/components/button/stories/button.stories.js
@@ -101,26 +101,28 @@ export default {
 
 const CustomButton = (args) => {
   return (
-    html`<div>
-      ${Template({
-        ...args,
-        iconName: undefined,
-      })}
-      ${Template({
-        ...args,
-        iconName: undefined,
-        treatment: 'outline'
-      })}
-      ${Template({
-        ...args,
-        iconName: args.iconName ?? 'Edit'
-      })}
-      ${Template({
-        ...args,
-        hideLabel: true,
-        iconName: args.iconName ?? 'Edit'
-      })}
-    </div>`
+    html`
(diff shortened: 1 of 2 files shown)


components/button/index.css
```diff
@@ -130,7 +130,7 @@ governing permissions and limitations under the License.
   }
 
   &.spectrum-Button--iconOnly {
-    min-inline-size: unset;
+    min-inline-size: var(--mod-button-height, var(--spectrum-button-height));
     padding: calc(var(--mod-button-edge-to-visual-only, var(--spectrum-button-edge-to-visual-only)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
     border-radius: 50%;
```
jenndiaz, line 135: Noticed a 2 pixel difference in height with the icon only buttons and the buttons with a label. 

I believe it is coming from 
`padding-block-end: calc(var(--mod-button-padding-label-bottom, var(--spectrum-button-padding-label-bottom)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));` 
on 
`.spectrum-Button-label`

pfulton: We'll look to fix this in a separate ticket. Thanks for catching it!

castastrophe (approved):

PR #1770, opened 2023-04-12
Ensure that all metadata/*.js assets are being loaded on the site into a bundle

Rajdeepc:
## Description
Added a loop on metadata/*.js and read all the js files and stored it in a variables which is parsed out in pages.njk

 After running 
``` yarn dev:docs ```

The following script tags are embedded in Accordion component's index.html

Screenshot 2023-04-12 at 8 43 46 PM

## Screenshots
Screenshot 2023-04-12 at 8 43 46 PM


castastrophe (approved):
I like it! Very elegant solution.

PR #1772, opened 2023-04-12
chore: add lerna caching

castastrophe:
## Description

Jira [CSS-454](https://jira.corp.adobe.com/browse/CSS-454)

This PR adds the necessary updates to support [Lerna caching](https://lerna.js.org/docs/concepts/how-caching-works).

- Add NX as a devDep
- Update all devDeps to local packages internally to use "*" instead of a specific version (nx will use this as a signifier to fetch the local version)
- Add config to support build caching
- Update a couple of the scripts that were using `yarn workspace` to run through `lerna --scope` instead (this allows it to connect to the cache where yarn workspace is unaware of it).

 - [x] `yarn build` x2 - validate that the second run uses the cached assets
 - [x] `yarn build`, `yarn build:preview` - preview build should use cached build assets
 - [x] `yarn build`, `git clean -dfX`, `yarn install`, `yarn build` - should not use cached assets

## To-do list


castastrophe:
- [ ] @pfulton going to validate "*" change via https://verdaccio.org

pfulton (approved):

pfulton (approved):

PR #1775, opened 2023-04-14
Image path fixes in YML Files

Rajdeepc:
## Description
Added filter to replace img paths

Chrome

## Screenshots
Screenshot 2023-04-14 at 5 56 32 PM


tools/documentation/_includes/meta-info.njk
```diff
@@ -11,31 +11,7 @@
...
-  'sidenav',
-  'site',
-  'statuslight',
-  'table',
-  'textfield',
-  'tooltip',
-  'typography',
-  'underlay'
-] %}
+{% for dep in components %}
```
castastrophe, line 14: I think this is a good patch for now. Originally that list of components was curated to match the styles being used in the website rendering separate from the documentation pages for the individual components. For a first pass, loading all the dependencies will work but I think eventually, we'll want to load components more minimally based on what's being used.

tools/documentation/.eleventy.js
```diff
@@ -71,7 +71,10 @@ module.exports = (config) => {
     return 'neutral';
   });
   config.addFilter('log', value => { console.log(value) });
-
+  config.addFilter('modifyImageSrc', function(value) {
```
castastrophe, line 74: I think instead of adding an 11ty filter, we should update the metadata *.yml files to reflect the URL change. I did a quick search for `src="img` in `components/*/metadata` and found about 79 examples. @pfulton Do you think that makes sense?

castastrophe (changes requested):

tools/documentation/.eleventy.js
Rajdeepc: I was thinking the same but I didn't want to touch the existing work. But yea changing it in the existing files will be more relevant as this will take some amount of memory to carry out the task every time.

tools/documentation/_includes/meta-info.njk
Rajdeepc: Yea but we need to do a thorough investigation on this since there are many adhoc component css are loaded in every pages.

tools/documentation/.eleventy.js
Rajdeepc: @pfulton Please let us know which one of the above fixes do you want me to add

pfulton: Yeah, let's go the route of updating the Yaml files!

Rajdeepc: @castastrophe  This is fixed now

PR #1785, opened 2023-04-17
Image and JS path fixes

Rajdeepc:
## Description
JS path fixes for store.json and index.json on root index.html and index.html under each component whose relative paths are both different to reach img folder also.
Added a regex to check for path and added the relative paths accordingly.

Chrome

## Screenshots

Screenshot 2023-04-17 at 2 47 16 PM
Screenshot 2023-04-17 at 2 47 21 PM
Screenshot 2023-04-17 at 2 47 26 PM

### DIFF
--- a/tools/documentation/assets/js/Search.js
+++ b/tools/documentation/assets/js/Search.js
@@ -122,14 +122,19 @@ function loadJSON(url, callback) {
   req.send();
 }
 
+function isComponentIndex() {
+  return /^\/docs\/[\w-]+\/$/.test(location.pathname);
+}
 Search.prototype.loadStore = function() {
-  loadJSON('./store.json', function(err, object) {
+  var storePath = isComponentIndex() ? '../store.json' : 'store.json';
+  loadJSON(storePath, function(err, object) {
     this.store = object;
   }.bind(this));
 }
 
 Search.prototype.loadIndex = function() {
-  loadJSON('./index.json', function(err, object) {
+  var indexPath = isComponentIndex() ? '../index.json' : 'index.json';
+  loadJSON(indexPath, function(err, object) {
     this.index = lunr.Index.load(object);
   }.bind(this));
 }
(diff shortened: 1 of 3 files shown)


Rajdeepc:
@castastrophe  for loadIcons as well as store.json and index.json the relative paths are different since all these files are needed in root index.html and also under components so I have created a check on url and added the paths accordingly. We might need to create a unified approach and remove the check.
Also for the url are we going ahead with showing docs/ in the url?

tools/documentation/assets/js/Search.js
```diff
@@ -122,14 +122,19 @@ function loadJSON(url, callback) {
   req.send();
 }
 
+function isComponentIndex() {
+  return /^\/docs\/[\w-]+\/$/.test(location.pathname);
+}
 Search.prototype.loadStore = function() {
-  loadJSON('./store.json', function(err, object) {
+  var storePath = isComponentIndex() ? '../store.json' : 'store.json';
```
castastrophe, line 129: @Rajdeepc It's not a good practice to alter pathing that is environment-based using front-end JS. What we want to do is make sure it renders locally and set up the appropriate redirects in netlify if necessary.

castastrophe (changes requested):

tools/documentation/content/get-started.md
```diff
@@ -100,7 +100,7 @@ eleventyNavigation:
     <img
       class="spectrum-CenteredImage"
       alt="Screenshot of the rendered CTA button in a browser window"
-      src="img/button-screen-shot.png"/>
+      src="../img/button-screen-shot.png"/>
```
castastrophe, line 103: Can we revert this update too?

Rajdeepc: Like we discussed with Patrick I have changed the img src in all yml and md files to reference from '../img' instead of 'img'. Please let me know if you have some other ideas too. Thanks

Rajdeepc: Done

PR #1787, opened 2023-04-18
feat: add stylelint

pfulton:
## Description
* Adds Stylelint to the project & borrows from some of the standard config.
* Adds a new `linting` GH Action to allow for linting in CI.

 - **How this was tested:** 
 ``` yarn lint:styles ```

Chrome and in Github workflow

## To-do list

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -1,0 +1,33 @@
+name: Linting
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+    paths:
+      - "**.css"
+
+jobs:
+  styles:
+    name: Lint CSS
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+      - name: Use Node LTS version
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - name: Install dependencies
+        run: yarn install
+      - name: Compile assets
+        run: yarn build
+      - name: Lint assets using stylelint
+        uses: reviewdog/action-stylelint@v1
+        with:
+          reporter: github-pr-review
+          fail_on_error: true
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,6 @@
+components/**/*-generated.css
+components/tokens/*
+components/vars/*
+components/express/*
+components/expressvars/*
+components/site/*
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,0 +1,37 @@
+{
+	"extends": ["stylelint-config-standard"],
+	"plugins": [
+		"stylelint-use-logical",
+		"stylelint-no-missing-parenthesis",
+		"stylelint-no-missing-var",
+		"stylelint-suit-naming-pattern"
+	],
+	"rules": {
+		"rule-empty-line-before": null,
+		"block-no-empty": null,
+		"at-rule-empty-line-before": null,
+		"at-rule-no-unknown": null,
+		"no-descending-specificity": null,
+		"selector-class-pattern": null,
+		"declaration-empty-line-before": null,
+		"custom-property-empty-line-before": null,
+		"comment-whitespace-inside": null,
+		"comment-empty-line-before": null,
+		"no-duplicate-selectors": null,
+		"property-no-vendor-prefix": null,
+		"max-nesting-depth": 2,
+		"csstools/use-logical": true,
+		"custom-rule/no-missing-parenthesis": true,
+		"custom-rule/no-missing-var": true,
+		"custom-rule/suit-naming-pattern": true,
+		"custom-property-pattern": "^[a-zA-Z0-9]+(-[a-z0-9]([a-zA-Z0-9]+)?)+$",
+		"alpha-value-notation": null,
+		"color-function-notation": null,
+		"import-notation": null,
+		"no-empty-source": null,
+		"property-no-unknown": true,
+		"declaration-property-value-no-unknown": true,
+		"value-keyword-case": null,
+		"selector-not-notation": null
+	}
+}
--- a/package.json
+++ b/package.json
@@ -1,109 +1,118 @@
 {
-	"private": true,
-	"name": "spectrum-css-monorepo",
-	"version": "0.0.0",
-	"license": "Apache-2.0",
-	"author": "Adobe",
-	"contributors": [
-		"Cassondra Roberts <castastrophe@users.noreply.github.com> (https://allons-y.llc)",
-		"Garth Braithwaite <garthdb@gmail.com> (https://garthdb.com)",
-		"Larry Davis <lazdnet@gmail.com>",
-		"Patrick Fulton <pfulton@users.noreply.github.com>",
-		"Steve Gill <stevengill97@gmail.com>"
-	],
-	"homepage": "https://opensource.adobe.com/spectrum-css/",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/adobe/spectrum-css.git"
-	},
-	"bugs": {
-		"url": "https://github.com/adobe/spectrum-css/issues"
-	},
-	"scripts": {
-		"build": "yarn build:components",
-		"build:all": "run-s build:components build:site build:preview",
-		"postbuild:all": "yarn mv:preview",
-		"build:clean": "run-s clean build",
-		"build:components": "lerna run --scope \"${SCOPE:-@spectrum-css/*}\" --ignore \"@spectrum-css/{*-builder*,preview,generator}\" build",
-		"postbuild:components": "yarn docs:mod",
-		"build:preview": "nx build @spectrum-css/preview",
-		"build:site": "gulp buildDocs -LLL",
-		"cache:clean": "nx reset",
-		"ci": "yarn build:all",
-		"ci:storybook": "nx storybook @spectrum-css/preview",
-		"clean": "yarn cache:clean && run-p clean:*",
-		"clean:components": "lerna exec --scope \"${SCOPE:-@spectrum-css/*}\" --ignore \"@spectrum-css/{*-builder*,preview,generator}\" -- rimraf dist",
-		"clean:docs": "rimraf dist",
-		"clean:preview": "nx clean @spectrum-css/preview",
-		"predev": "yarn build:components",
-		"dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
-		"docs:mod": "node tasks/mod-extractor.js",
-		"lint:components": "node ./tasks/packageLint.js",
-		"mv:preview": "rimraf dist/preview && mv -f tools/preview/storybook-static dist/preview",
-		"new": "nx run @spectrum-css/generator:new",
-		"precommit": "lint-staged",
-		"prepare": "husky install",
-		"prepare:site": "run-s clean:docs build:all",
-		"prerelease": "yarn version:build",
-		"release": "lerna publish",
-		"release:beta-from-package": "yarn release from-package --conventional-prerelease --preid beta --pre-dist-tag beta --no-private",
-		"prerelease:site": "yarn prepare:site",
-		"release:site": "gh-pages -d dist/ -f -e .",
-		"prestart": "yarn build:components",
-		"start": "NODE_ENV=development yarn workspace @spectrum-css/preview start",
-		"test": "nx test @spectrum-css/preview",
-		"test:scope": "nx test:scope @spectrum-css/preview",
-		"version:build": "lerna run --since origin/main build",
-		"watch": "gulp watch"
-	},
-	"workspaces": [
-		"components/*",
-		"tools/*"
-	],
-	"devDependencies": {
-		"@adobe/focus-ring-polyfill": "^0.1.5",
-		"@commitlint/cli": "^17.6.3",
-		"@commitlint/config-conventional": "^17.6.3",
-		"gh-pages": "^5.0.0",
-		"gulp": "^4.0.0",
-		"gulp-replace": "^1.0.0",
-		"gulplog": "^1.0.0",
-		"husky": "^8.0.3",
-		"lerna": "^6.6.2",
-		"lint-staged": "^13.2.2",
-		"loadicons": "^1.0.0",
-		"lunr": "^2.3.6",
-		"markdown-it": "^12.3.2",
-		"npm-run-all": "^4.1.5",
-		"nx": "^16.2.2",
-		"prettier": "^2.8.8",
-		"prettier-package-json": "^2.8.0",
-		"prismjs": "^1.23.0",
-		"rimraf": "^5.0.1",
-		"semver": "^7.5.1",
-		"through2": "^3.0.1",
-		"yargs": "^17.7.2"
-	},
-	"engines": {
-		"node": ">=14.0.0 <18"
-	},
-	"browserslist": [
-		"last 2 Edge versions",
-		"last 2 Chrome versions",
-		"last 2 Firefox versions",
-		"last 2 Safari versions",
-		"last 2 iOS versions"
-	],
-	"lint-staged": {
-		"*.{md,js,css,yml,hbs}": [
-			"prettier --write"
-		],
-		"components/*/package.json": [
-			"yarn lint:components",
-			"prettier-package-json --write"
-		],
-		"package.json": [
-			"prettier-package-json --write"
-		]
-	}
+  "private": true,
+  "name": "spectrum-css-monorepo",
+  "version": "0.0.0",
+  "license": "Apache-2.0",
+  "author": "Adobe",
+  "contributors": [
+    "Cassondra Roberts <castastrophe@users.noreply.github.com> (https://allons-y.llc)",
+    "Garth Braithwaite <garthdb@gmail.com> (https://garthdb.com)",
+    "Larry Davis <lazdnet@gmail.com>",
+    "Patrick Fulton <pfulton@users.noreply.github.com>",
+    "Steve Gill <stevengill97@gmail.com>"
+  ],
+  "homepage": "https://opensource.adobe.com/spectrum-css/",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/adobe/spectrum-css.git"
+  },
+  "bugs": {
+    "url": "https://github.com/adobe/spectrum-css/issues"
+  },
+  "scripts": {
+    "build": "yarn build:components",
+    "build:all": "run-s build:components build:site build:preview",
+    "postbuild:all": "yarn mv:preview",
+    "build:clean": "run-s clean build",
+    "build:components": "lerna run --scope \"${SCOPE:-@spectrum-css/*}\" --ignore \"@spectrum-css/{*-builder*,preview,generator}\" build",
+    "postbuild:components": "yarn docs:mod",
+    "build:preview": "nx build @spectrum-css/preview",
+    "build:site": "gulp buildDocs -LLL",
+    "cache:clean": "nx reset",
+    "ci": "yarn build:all",
+    "ci:storybook": "nx storybook @spectrum-css/preview",
+    "clean": "yarn cache:clean && run-p clean:*",
+    "clean:components": "lerna exec --scope \"${SCOPE:-@spectrum-css/*}\" --ignore \"@spectrum-css/{*-builder*,preview,generator}\" -- rimraf dist",
+    "clean:docs": "rimraf dist",
+    "clean:preview": "nx clean @spectrum-css/preview",
+    "predev": "yarn build:components",
+    "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
+    "docs:mod": "node tasks/mod-extractor.js",
+    "postdocs:mod": "prettier --cache --write components/*/metadata/*.md > /dev/null",
+    "lint:components": "node ./tasks/packageLint.js",
+    "lint:styles": "stylelint",
+    "mv:preview": "rimraf dist/preview && mv tools/preview/storybook-static dist/preview",
+    "new": "nx run @spectrum-css/generator:new",
+    "precommit": "lint-staged",
+    "prepare": "husky install",
+    "prepare:site": "run-s clean:docs build:all",
+    "prerelease": "yarn version:build",
+    "release": "lerna publish",
+    "release:beta-from-package": "yarn release from-package --conventional-prerelease --preid beta --pre-dist-tag beta --no-private",
+    "prerelease:site": "yarn prepare:site",
+    "release:site": "gh-pages -d dist/ -f -e .",
+    "prestart": "yarn build:components",
+    "start": "NODE_ENV=development yarn workspace @spectrum-css/preview start",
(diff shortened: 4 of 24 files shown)


.github/workflows/linting.yml
```diff
@@ -0,0 +1,24 @@
+name: Linting
+
+on:
+  pull_request:
```
Rajdeepc, line 4: Can we also add this on push?

.stylelintignore
```diff
@@ -0,0 +1 @@
+components/**/*-generated.css
```
castastrophe, line 1: Could this be done using `components/*/dist/*.css`?

.stylelintrc.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "stylelint-config-standard",
+  "rules": {
```
castastrophe, line 3: Nice, these are great customizations

package.json
```diff
@@ -34,6 +34,7 @@
     "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
     "docs:mod": "node tasks/mod-extractor.js",
     "lint:components": "node ./tasks/packageLint.js",
+    "lint:styles": "stylelint ./components/**/*.css",
```
castastrophe, line 37: +1

castastrophe (commented):
This looks pretty great, whenever you're ready to bring it in, let me know and I'll pull it down for testing

castastrophe (commented):
Can we rename the `custom-stylelint-plugins` folder to just `plugins`? We have several custom postcss plugins as well that I'd like to eventually move out of tools.

custom-stylelint-plugins/css-logical-properties.js
```diff
@@ -0,0 +1,73 @@
...
+      possible: [true],
+    });
+
+    if (!validOptions) {
+      return;
+    }
+
+    const isAutoFixing = Boolean(context.fix);
+
+    postcssRoot.walkDecls(decl => {
```
castastrophe, line 46: I'm not entirely clear on what this plugin is correcting for - it seems to be replacing logical properties with physical ones but isn't our best practice to use logical properties instead?

Would you mind adding some comments in here to clarify the goals?

package.json
castastrophe: Perhaps we should hook this in through postcss so that another command is not necessary: https://stylelint.io/user-guide/postcss-plugin?

custom-stylelint-plugins/no-missing-parenthesis.js
```diff
@@ -0,0 +1,22 @@
+const stylelint = require('stylelint');
+
+const ruleName = "custom-rule/no-missing-parenthesis";
```
castastrophe, line 3: I'm not sure this is needed - does the [custom-property-no-missing-var-function](https://stylelint.io/user-guide/rules/custom-property-no-missing-var-function) rule and the general error capturing not identify this?

custom-stylelint-plugins/no-missing-var.js
```diff
@@ -0,0 +1,23 @@
+const stylelint = require('stylelint');
+
+const ruleName = 'custom-rule/no-missing-var';
+
+const plugin = stylelint.createPlugin(ruleName, (enabled, options) => {
```
castastrophe, line 5: Same as above, this should be covered by https://stylelint.io/user-guide/rules/custom-property-no-missing-var-function

custom-stylelint-plugins/spectrum-class-name-rule.js
```diff
@@ -0,0 +1,37 @@
+const stylelint = require('stylelint');
+
+const ruleName = 'custom-rule/spectrum-class-name';
+const messages = stylelint.utils.ruleMessages(ruleName, {
```
castastrophe, line 4: A new plugin should not be needed for this - we can pass the customizations in to https://stylelint.io/user-guide/rules/selector-class-pattern/:

```
"selector-class-pattern": [
  "^spectrum(-[A-Z][a-zA-Z0-9]+)", {
    resolveNestedSelectors: true
  }
],
```

And as far as the excluded tags, this only validates classes so they should not be an issue.

package.json
Rajdeepc: We wanted to give the users to see lint warnings without doing the build.

custom-stylelint-plugins/css-logical-properties.js
Rajdeepc: I have replaced this with https://github.com/csstools/stylelint-use-logical module which lints for usage of non logical properties.

custom-stylelint-plugins/no-missing-parenthesis.js
Rajdeepc: This rule only reports custom properties that are defined within the same source.
It will not throw error for this issue:
.spectrum-Well {
  background-color: (--spectrum-well-background-color);
  border-color: var(--spectrum-well-border-color);
}

It will only throw error here:
  :root { --foo: red; }
    a { color: --foo; }

custom-stylelint-plugins/no-missing-var.js
Rajdeepc: This rule only reports custom properties that are defined within the same source.

Will not throw error here:
.spectrum-Well {
  background-color: (--spectrum-well-background-color);
  border-color: var(--spectrum-well-border-color);
}


But will only throw error here where the properties are defined on the same css source.
    :root { --foo: red; }
    a { color: --foo; }

custom-stylelint-plugins/spectrum-class-name-rule.js
Rajdeepc: I get it. I wanted to create separate plugins just in case if we need some more customizations on this. The one which you have mentioned is definately one way to go.

.stylelintignore
pfulton: I'm not sure. I included this so that we could ignore [these](https://github.com/adobe/spectrum-css/blob/main/components/pickerbutton/pickerbutton-generated.css) [two](https://github.com/adobe/spectrum-css/blob/main/components/infieldbutton/infieldbutton-generated.css) `-generated.css` files that are present in the source (not the `dist`). These were once generated by DSS and then included here. I imagine we'll remove them once these two components get migrated.

custom-stylelint-plugins/css-logical-properties.js
pfulton: @Rajdeepc once you push the commit containing `stylelint-use-logical`, it looks like we can remove this file?

package.json
pfulton: I think I'd prefer to leave it as a separate command for now. Additionally, I noticed this sentence in that link:

> However, we recommend using the CLI or Node.js API (directly or via an integration) as they provide better reporting.

I'm cool with keeping it as a separate command if you all are.

pfulton (commented):
@Rajdeepc things are starting to take shape here! Thanks for working on this. I left a few comments - also: how do you feel about writing some tests to cover the custom plugin functionality?

.stylelintignore
castastrophe: Oh! I thought this was related to a new asset being created by linting which seemed odd. Yes, that makes sense.

plugins/_tests_/no-missing-parenthesis.test.js
```diff
@@ -0,0 +1,25 @@
+const stylelint = require("stylelint")
```
castastrophe, line 1: These tests are great! I'm wondering if we can make the new stylelint plugins packages rather than individual files so we're not cluttering up the root package with dependencies only for these items. Can you look at https://github.com/castastrophe/postcss-splitinator for an idea of what to include? I like ava for lightweight testing and it plugs in nicely with c8 for code coverage reporting.

custom-stylelint-plugins/spectrum-class-name-rule.js
Rajdeepc: This rule is not sufficient to handle both Utlity and Component type SUIT pattern. Therefor a plugin was needed to incorporate both of them.
I have made the Component type SUIT to always start with .spectrum. Please let me know whether we can initialize a classname with Component Name only also.
@pfulton

Rajdeepc:
> @Rajdeepc things are starting to take shape here! Thanks for working on this. I left a few comments - also: how do you feel about writing some tests to cover the custom plugin functionality?

Got it. Tests are included now. Thanks

plugins/_tests_/no-missing-parenthesis.test.js
Rajdeepc: Got it. I checked out your amazing repo and updated the codebase by adding the plugins as packages and added ava as the testing tool. Added yarn lint:tests to trigger the stylelint plugins tests with lerna. 
Let me know how this looks now. Thanks @castastrophe

tools/stylelint-no-missing-var/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": "1.0.1",
+  "license": "MIT",
+  "author": "Rajdeep Chandra",
+  "main": "index.js",
+  "scripts": {
+    "lint": "stylelint index.js",
+    "prepublishOnly": "yarn test",
+    "test": "ava"
+  },
+  "devDependencies": {
```
castastrophe, line 12: Should stylelint be a peerDependency?

castastrophe (commented):
So fantastic! One note about the state-based class naming in the SUIT plugin. Would you also please move the new packages to the plugins folder? We'll need to add plugins as a workspace in the root package.json.

tools/stylelint-no-missing-var/package.json
Rajdeepc: Yes my bad. Added to package.json of every custom plugin.

castastrophe:
Oh and we might want to consider running this task on compiled assets b/c I noticed it's tripping over some of the postcss plugin syntax.

Rajdeepc:
> Can you dig into why the action is failing? It looks like it's throwing false errors around high contrast values.

Fixed

Rajdeepc:
> Oh and we might want to consider running this task on compiled assets b/c I noticed it's tripping over some of the postcss plugin syntax.

Done

package.json
```diff
@@ -39,7 +39,8 @@
     "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
     "docs:mod": "node tasks/mod-extractor.js",
     "lint:components": "node ./tasks/packageLint.js",
-    "mv:preview": "rimraf dist/preview && mv -f tools/preview/storybook-static dist/preview",
+    "lint:styles": "stylelint ./components/**/*.css",
```
castastrophe, line 42:
```suggestion
    "lint:styles": "stylelint ./components/*/dist/**/*.css",
```

^ Try this command - it's still running on pre-compiled assets b/c it's picking them up from the `components/*/*.css`. We can target dist specifically or we can look into ways to ignore the postcss specific syntax that's tripping up the linter.

castastrophe (changes requested):
This task is still failing and a lot of what I'm seeing in the logs are not incorrect so I think we need to adjust some of the settings.

Screenshot 2023-05-18 at 7 52 50 AM

This for example is capturing typography capitalization which is valid. We could set up an ignore rule for typography folder or see if there's a way to ignore certain keywords. Can you dig into the other failures and ask questions if you're not sure if it's a valid error or not?

castastrophe (changes requested):
This task is still failing and a lot of what I'm seeing in the logs are not incorrect so I think we need to adjust some of the settings.

Screenshot 2023-05-18 at 7 52 50 AM

This for example is capturing typography capitalization which is valid. We could set up an ignore rule for typography folder or see if there's a way to ignore certain keywords. Can you dig into the other failures and ask questions if you're not sure if it's a valid error or not?

Rajdeepc:
@castastrophe Please let me know if you think this is good to merge!!

pfulton:
> @castastrophe Please let me know if you think this is good to merge!!

I think things are looking pretty good with this, but I have two things:
1. I think we should be running this as a pre-compilation step (meaning, against the source files). It's definitely important that the CSS that we ship in `dist` is correct and syntactically sound, but I'd almost argue that it shouldn't even compile if the linter errors out.
2. I'd like to reorganize the order of our GH Actions so that the lint step runs in CI _before_ the build and deploy step (which sort of goes back to the first point above). I like the idea of failing quicker rather than having to wait for the entirety of the build and deploy step to finish before knowing that I have a linting error.

What do you two think, @castastrophe and @Rajdeepc?

Rajdeepc:
> > @castastrophe Please let me know if you think this is good to merge!!
> 
> I think things are looking pretty good with this, but I have two things:
> 
> 1. I think we should be running this as a pre-compilation step (meaning, against the source files). It's definitely important that the CSS that we ship in `dist` is correct and syntactically sound, but I'd almost argue that it shouldn't even compile if the linter errors out.
> 2. I'd like to reorganize the order of our GH Actions so that the lint step runs in CI _before_ the build and deploy step (which sort of goes back to the first point above). I like the idea of failing quicker rather than having to wait for the entirety of the build and deploy step to finish before knowing that I have a linting error.
> 
> What do you two think, @castastrophe and @Rajdeepc?

@pfulton You are right!!. I agree to your point that lint job should run before the build process so that no linted code is pushed to this repository from now on. I was working on this today morning and found out that it needs to add few branch protection rule checks here: https://github.com/orgs/community/discussions/25720
If you are the admin can you kindly disabled the required check in the related branch rules. More info on the above link. Unfortunately I didint have the admin access to this repo so cannot change it via settings. Thank you.

castastrophe:
> I think we should be running this as a pre-compilation step (meaning, against the source files). It's definitely important that the CSS that we ship in dist is correct and syntactically sound, but I'd almost argue that it shouldn't even compile if the linter errors out.

I agree that would be nice but it kept getting hung up on the postcss plug-in syntax that isn't valid CSS. We could run it as part of the PostCSS build so that it runs after those custom utilities are resolved but it's not that different from running on dist.

castastrophe:
> it shouldn't even compile if the linter errors out.

The challenge with this approach is that it is more likely to block developers on making progress until minor lint fixes are in place or (more often) until an issue in the linter logic is fixed. I find that making the linter a reporting tool until merging with main is ideal. It ensures nothing bad gets into production but keeps people unblocked.

pfulton:
@castastrophe @Rajdeepc ok. I think the three of us are in agreement that adding linting in any form is going to benefit the health of the project in a lot of ways.

Let's proceed with what @castastrophe is suggesting, which was where @Rajdeepc had things before I left my comments about the source vs. dist files and also run the steps in parallel.

It'll be great to add this into our process and see how the squad interacts with it and works with it. If we need to make adjustments, we can always make adjustments.

Thanks for your input, folks!

Rajdeepc:
> @castastrophe @Rajdeepc ok. I think the three of us are in agreement that adding linting in any form is going to benefit the health of the project in a lot of ways.
> 
> Let's proceed with what @castastrophe is suggesting, which was where @Rajdeepc had things before I left my comments about the source vs. dist files and also run the steps in parallel.
> 
> It'll be great to add this into our process and see how the squad interacts with it and works with it. If we need to make adjustments, we can always make adjustments.
> 
> Thanks for your input, folks!

Done. Good to merge

castastrophe (changes requested):
Awesome! Minor note on adding stylelint as a devDep on the packages but this is awesome! Can't wait!

plugins/stylelint-no-missing-parenthesis/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "lint": "stylelint index.js",
+    "prepublishOnly": "yarn test",
+    "test": "ava"
+  },
+  "peerDependencies": {
+    "stylelint": ">=15"
+  },
+  "devDependencies": {
+    "ava": "^3.9.0",
+    "stylelint": ">=15"
```
castastrophe, line 17:
```suggestion
    "stylelint": "^15.0.0"
```

I'm not sure but I don't think >= works in devDeps

Rajdeepc: Umm I think it works but I made the changes to its latest versions.

castastrophe (approved):

Rajdeepc (approved):

package.json
```diff
@@ -39,7 +39,8 @@
     "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
     "docs:mod": "node tasks/mod-extractor.js",
     "lint:components": "node ./tasks/packageLint.js",
-    "mv:preview": "rimraf dist/preview && mv -f tools/preview/storybook-static dist/preview",
+    "lint:styles": "stylelint ./components/*/dist/**/*.css",
```
pfulton, line 42: It looks like the path here is partially incorrect. As it's written, it's _only_ checking against CSS files that are in a sub-directory in each component's `dist`, and if I try to force a linting error, nothing is being detected by Stylelint. 

I think what we want is:

```suggestion
    "lint:styles": "stylelint ./components/*/dist/*.css",
```

pfulton:
@Rajdeepc @castastrophe what do y'all think about the possibility of adding [this other Stylelint plugin](https://www.npmjs.com/package/stylelint-csstree-validator) to check values? I'm on the fence: I wonder how much noise this will add and how accurate it would be, but it also might be handy if/when someone misspells something like `flex-end` — I'd expect that the engineer would _visually_ see that a rule wasn't working, but that's not always the case.

Rajdeepc:
> @Rajdeepc @castastrophe what do y'all think about the possibility of adding [this other Stylelint plugin](https://www.npmjs.com/package/stylelint-csstree-validator) to check values? I'm on the fence: I wonder how much noise this will add and how accurate it would be, but it also might be handy if/when someone misspells something like `flex-end` — I'd expect that the engineer would _visually_ see that a rule wasn't working, but that's not always the case.

I think we can tackle this with this out-of-the-box rule: https://stylelint.io/user-guide/rules/property-no-unknown/
Screenshot 2023-05-24 at 12 14 34 PM

package.json
castastrophe: Woah! Nice catch. I always forget if ** is 0+ or 1+.

.github/workflows/linting.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/checkout@v3
+      - name: Use Node LTS version
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - name: Install dependencies
+        run: yarn install
+      - name: Compile assets
+        run: yarn build
+      - name: Lint compiled CSS files
```
castastrophe, line 27: I'd suggest we run the lint command _only_ on the files edited by the PR: https://github.com/marketplace/actions/list-files-in-pull-request. It only works for pull requests so we could move the linting of the main branch over to the pr-build-and-deploy workflow

plugins/stylelint-suit-naming-pattern/test.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    },
+  });
+
+  return result;
+}
+
+test(`should pass valid class names`, async (t) => {
+  const validCode = `
+    .spectrum {}
```
castastrophe, line 23: let's add some of the context variants to this list: .spectrum--lightest

pfulton:
Ah, this is really great, @Rajdeepc.

Just a few things that I'm seeing as I look through the reports, post-run:

* We can ignore the CSS in the `components/site` dir for now. That's not a _real_ "component", and @castastrophe has moved those source files elsewhere in her 11ty work, so reporting things in there is not necessary at this time.
* The SUIT plugin looks mostly good. I think we're super close to it being what we want. There were a few legitimate things that are being caught by that rule: selector names that look like `.spectrum-Switch-input[disabled]` (where we're using attribute selectors), selector names that look like: `.spectrum:lang(ar)` (where we use the `:lang` pseudo-class selectors), selector names that look like `.spectrum-UIIcon-ChevronDown75` - icons use a mix of SUIT and camel casing, and then I also saw a few false positives on the "missing `var` before custom property - like [here](https://github.com/adobe/spectrum-css/blob/5bdf773b2592083da372f8066c624d5356feaaca/components/dropzone/index.css#L117), specifically in DropZone - I think it's getting tripped up on the `--`.

plugins/stylelint-suit-naming-pattern/index.js
```diff
@@ -0,0 +1,46 @@
...
+              message: messages.wrongUtilityClassName(selector),
+              node: rule,
+              result,
+              ruleName,
+            })
+          }
+        }
+      })
+    })
+  }
```
castastrophe, line 42:
```suggestion
    root.walkRules(async (rule) => {
      await parser(selectors => {
        selectors.walkClasses(selectorNode => {
          const selector = selectorNode.value;

          if (selector.startsWith('spectrum') && !suitClassPattern.test(selector)) {
            stylelint.utils.report({
              message: messages.wrongCapitalization(selector),
              node: rule,
              result,
              ruleName,
            });
          } else if ((selector.startsWith("u-") || selector.startsWith("is-")) && !utilityClassPattern.test(selector)) {
            stylelint.utils.report({
              message: messages.wrongUtilityClassName(selector),
              node: rule,
              result,
              ruleName,
            });
          } else {
            stylelint.utils.report({
              message: messages.wrongClassName(selector),
              node: rule,
              result,
              ruleName,
            });
          }
        });
      }).process(rule.selector);
    })
  }
```

This approach allows you to target each individual class name without having to worry about valid pseudos or tag usage. First the postcss utility breaks the string apart into distinct classes and then iterates over them to check them against our rules specifically.

castastrophe (changes requested):
Some feedback on how to optimize the class name parser

Rajdeepc:
> * .spectrum-UIIcon-ChevronDown75

```components/site``` is added to stylelintignore
Updated regex and the following cases are updated in the unit test:
```
.spectrum-ToolTip--size100{}
.spectrum-Switch-input[disabled]{}
.spectrum:lang(ar){}
.spectrum-UIIcon-ChevronDown75{}
    ```

plugins/stylelint-suit-naming-pattern/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+const messages = stylelint.utils.ruleMessages(ruleName, {
+  wrongCapitalization: (className) =>
+    `Component Class name "${className}" should follow the SUIT-style naming conventions with the ".spectrum" prefix.`,
+  wrongUtilityClassName: (className) =>  `Utility Class name "${className}" should follow the SUIT-style naming conventions with the ".u-camelCase" prefix.`,
+})
+const utilityClassPattern = /^\.u-[a-z]+[a-zA-Z0-9]+/;
+
+const suitClassPattern =
+/^spectrum((--|.)[a-z][a-z()]+)?((-[A-Z][a-zA-Z0-9:.()]+)([-]{1,2}[a-z][a-zA-Z-0-9():.]+)?(\s|,\s)?)?$/
```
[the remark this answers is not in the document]
Rajdeepc: You are right!!! I have made the regex to work on a single class. If we have added classnames it will split it and run the lint on it individually.

plugins/stylelint-suit-naming-pattern/index.js
Rajdeepc: Updated with postcss. I had to update the regex a bit to accomodate few patterns. Thanks for the idea.

castastrophe:
I validated this PR by applying a couple different types of errors inside of the accordion/index.css file and confirmed that the action failed and that comments were added to the files in the PR. ✅

PR #1793, opened 2023-04-19
feat(floatingactionbutton): add new component

mlogsdon18:
## Description

This adds a new FloatingActionButton component with two variants: primary and secondary.

This currently contains "hard coded" values for the `--spectrum-floating-action-button-drop-shadow-y` and `--spectrum-floating-action-button-drop-shadow-color` because we are waiting for these tokens to get added per this ticket

- Latest versions of Chrome, Firefox, and Safari

## To-do list

### DIFF
--- a/components/floatingactionbutton/index.css
+++ b/components/floatingactionbutton/index.css
@@ -101,7 +101,7 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-FloatingActionButton-icon {
+.spectrum-Icon.spectrum-FloatingActionButton-icon {
   block-size: var(--mod-floating-action-button-icon-size, var(--spectrum-floating-action-button-icon-size));
   inline-size: var(--mod-floating-action-button-icon-size, var(--spectrum-floating-action-button-icon-size));
   fill: var(--highcontrast-floating-action-button-icon-color, var(--mod-floating-action-button-icon-color, var(--spectrum-floating-action-button-icon-color)));
--- a/components/floatingactionbutton/index.css
+++ b/components/floatingactionbutton/index.css
@@ -16,8 +16,8 @@ governing permissions and limitations under the License.
   --spectrum-floating-action-button-padding: var(--spectrum-component-pill-edge-to-visual-only-200);
   --spectrum-floating-action-button-margin: var(--spectrum-spacing-200);
   --spectrum-floating-action-button-drop-shadow-x: var(--spectrum-drop-shadow-x);
-  --spectrum-floating-action-button-drop-shadow-y: var(--spectrum-drop-shadow-y);
-  --spectrum-floating-action-button-drop-shadow-color: var(--spectrum-transparent-black-300);
+  --spectrum-floating-action-button-drop-shadow-y: 4px; /* TODO: waiting for token */
+  --spectrum-floating-action-button-drop-shadow-color: var(--spectrum-transparent-black-300); /* TODO: waiting for token */
 
   --spectrum-floating-action-button-focus-ring-width: var(--spectrum-focus-indicator-thickness);
   --spectrum-floating-action-button-focus-ring-gap: var(--spectrum-focus-indicator-gap);
--- a/components/floatingactionbutton/index.css
+++ b/components/floatingactionbutton/index.css
@@ -47,12 +47,6 @@ governing permissions and limitations under the License.
 
 }
 
-@media (forced-colors: active) {
-  .spectrum-FloatingActionButton {
-
-  }
-}
-
 .spectrum-FloatingActionButton {
   cursor: pointer;
   block-size: var(--mod-floating-action-button-size, var(--spectrum-floating-action-button-size));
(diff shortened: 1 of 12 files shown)


components/floatingactionbutton/index.css
```diff
@@ -0,0 +1,125 @@
...
+    --spectrum-floating-action-button-icon-color-key-focus: var(--spectrum-neutral-subdued-content-color-key-focus);
+  }
+
+}
+
+.spectrum-FloatingActionButton {
+  cursor: pointer;
+  block-size: var(--mod-floating-action-button-size, var(--spectrum-floating-action-button-size));
+  inline-size: var(--mod-floating-action-button-size, var(--spectrum-floating-action-button-size));
+  border-radius: 50%;
```
jawinn, line 51: Should we add a `--spectrum-floating-action-button-border-radius` custom property and use it here and in the focus indicator? I could see someone wanting access to a `--mod-floating-action-button-border-radius`. I suppose it also could be done without adding a property too, e.g. `border-radius: var(--mod-floating-action-button-border-radius, 50%);`

components/floatingactionbutton/metadata/floatingactionbutton.yml
```diff
@@ -0,0 +1,36 @@
+name: Floating action button
```
jawinn, line 1: Maybe capitalize this and also add a description?

components/floatingactionbutton/index.css
```diff
@@ -0,0 +1,125 @@
...
+  }
+}
+
+.spectrum-Icon.spectrum-FloatingActionButton-icon {
+  block-size: var(--mod-floating-action-button-icon-size, var(--spectrum-floating-action-button-icon-size));
+  inline-size: var(--mod-floating-action-button-icon-size, var(--spectrum-floating-action-button-icon-size));
+  fill: var(--highcontrast-floating-action-button-icon-color, var(--mod-floating-action-button-icon-color, var(--spectrum-floating-action-button-icon-color)));
+}
+
+@media (forced-colors: active) {
```
jawinn, line 106: Should this @media block be closer to the top, under the other properties like they are on the other components?

jawinn (changes requested):
Looking really good! One thing I am seeing is an extra outline on focus in Firefox. It probably just needs an `outline: 0` when there is the custom focus indicator showing, to disable the browser default focus outline.
Screenshot 2023-04-26 at 11 21 30 AM

It also could use color `transition` styles similar to some of the other buttons. Some use some tokens like `--spectrum-animation-duration-100`.

components/floatingactionbutton/index.css
```diff
@@ -0,0 +1,125 @@
...
+
+    --highcontrast-floating-action-button-background-color: CanvasText;
+    --highcontrast-floating-action-button-background-color-hover: Highlight;
+    --highcontrast-floating-action-button-background-color-down: Highlight;
+    --highcontrast-floating-action-button-background-color-key-focus: Highlight;
+
+    --highcontrast-floating-action-button-icon-color: ButtonFace;
+    --highcontrast-floating-action-button-icon-color-hover: ButtonFace;
+    --highcontrast-floating-action-button-icon-color-down: ButtonFace;
+    --highcontrast-floating-action-button-icon-color-key-focus: ButtonFace;
```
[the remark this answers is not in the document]
mlogsdon18: I likely need to switch `CanvasText` for `ButtonText` but I based these on [what was going on with other buttons](https://github.com/adobe/spectrum-css/blob/b4a294a63c36f2be6c2230530738667cca68db34/components/button/index.css#L246). For the accent buttons that have a blue background, like this button, they use `ButtonText` as the background and `ButtonFace` as the text color. It allows for some of that color difference to stay even in WHCM. They also use `Highlight` as the hover/down/focus color, which is what I have done here

components/floatingactionbutton/index.css
mlogsdon18: So there are a few components that have this at the bottom of the file so that it has the highest priority. I ran into that with a different component, I can't think of which one now, but I started thinking these should be at the bottom if priority was going to be an issue. Basically if you have any CSS in this media query other than the variables, it gets overridden by the CSS below it. This ensures the high contrast styling conquers all

components/floatingactionbutton/index.css
mlogsdon18: Yeah I can add that in. These buttons are supposed to be perfect circles but since we're going super customizable I should probably add it just in case.

jawinn (approved):

pfulton (approved):

components/floatingactionbutton/index.css
```diff
@@ -0,0 +1,126 @@
...
+governing permissions and limitations under the License.
+*/
+
+.spectrum-FloatingActionButton {
+  --spectrum-floating-action-button-size: var(--spectrum-component-height-200);
+  --spectrum-floating-action-button-icon-size: var(--spectrum-workflow-icon-size-200);
+  --spectrum-floating-action-button-padding: var(--spectrum-component-pill-edge-to-visual-only-200);
+  --spectrum-floating-action-button-margin: var(--spectrum-spacing-200);
+  --spectrum-floating-action-button-drop-shadow-x: var(--spectrum-drop-shadow-x);
+  --spectrum-floating-action-button-drop-shadow-color: var(--spectrum-floating-action-button-shadow-color);
```
Rajdeepc, line 19: @mlogsdon18 --spectrum-floating-action-button-shadow-color this is undefined
Can you please check this. This is breaking the build

components/floatingactionbutton/index.css
```diff
@@ -0,0 +1,126 @@
...
+  cursor: pointer;
+  block-size: var(--mod-floating-action-button-size, var(--spectrum-floating-action-button-size));
+  inline-size: var(--mod-floating-action-button-size, var(--spectrum-floating-action-button-size));
+  border-radius: var(--mod-floating-action-button-border-radius, 50%);
+  padding-inline: var(--mod-floating-action-button-padding, var(--spectrum-floating-action-button-padding));
+  padding-block: var(--mod-floating-action-button-padding, var(--spectrum-floating-action-button-padding));
+  margin-inline-end: var(--mod-floating-action-button-margin, var(--spectrum-floating-action-button-margin));
+  margin-block-end: var(--mod-floating-action-button-margin, var(--spectrum-floating-action-button-margin));
+  border: none;
+  box-shadow: var(--mod-floating-action-button-drop-shadow-x, var(--spectrum-floating-action-button-drop-shadow-x)) var(--mod-floating-action-button-drop-shadow-y, var(--spectrum-floating-action-button-drop-shadow-y)) var(--mod-floating-action-button-drop-shadow-blur, var(--spectrum-floating-action-button-drop-shadow-blur)) var(--mod-floating-action-button-drop-shadow-color, var(--spectrum-floating-action-button-drop-shadow-color));
```
Rajdeepc, line 78: --spectrum-floating-action-button-drop-shadow-y is undefined

Rajdeepc (commented):
Please see inline comments.

components/floatingactionbutton/index.css
mlogsdon18: @Rajdeepc This token is defined and resolves to `4px`. Are you using the latest version of tokens?

PR #1794, opened 2023-04-19
fix(button): height and accessibility fixes

mlogsdon18:
## Description

- Update buttons to use custom tokens for padding to account for a change to 1.2 line-height. This allows for the buttons to be the minimum height specified for each size. 
- Added `aria-label` to the Icon Only variant of button
- Removed unneeded `aria-label` from icons that were marked as `aria-hidden="true"` since they won't be read
- Cleaned up storybook template by removing an unused prop `isOpen`

## To-do list


pfulton (approved):
This looks good to me, @mlogsdon18 - thanks for taking care of this.

@Westbrook how are things looking to you?

Westbrook:
LGTM!

PR #1798, opened 2023-04-21
feat(alertbanner): new component 

jenndiaz:
## Description
Creates new component and uses core tokens. 

 - **How this was tested:** 

## To-do list

- [x]  If my change(s) include visual change(s), a designer has reviewed and approved those changes.


components/alertbanner/stories/template.js
```diff
@@ -0,0 +1,82 @@
...
+    : "";
+
+  return html`
+  <div class=${classMap({
+    [rootClass]: true,
+    'is-open': isOpen,
+    [`${rootClass}--${variant}`]: typeof variant !== "undefined",
+    ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+  })}>
+    <div class="spectrum-AlertBanner-body">
```
mlogsdon18, line 42: We should probably replace all of these with `${rootClass}-body` for example

components/alertbanner/stories/template.js
```diff
@@ -0,0 +1,82 @@
...
+        treatment: "outline",
+        label: 'Action',
+        customClasses: [`${rootClass}-action-button`]
+      })}
+    </div>
+    <div class="spectrum-AlertBanner-end">
+    ${Divider({
+      vertical: true,
+      size: 's',
+      customClasses: [`${rootClass}-divider`, `${rootClass}--vertical`, `${rootClass}-divider--staticWhite`],
```
mlogsdon18, line 65: The vertical and staticWhite classes should be added based on the properties that you send through so they can be deleted from here

components/alertbanner/stories/template.js
```diff
@@ -0,0 +1,82 @@
...
+      tag: 'div',
+      style: {
```
mlogsdon18, line 67: You have this set as `verticalStyle` in the Divider template so we should probably pick one! I think style is probably best to just have it available whether the divider is vertical or not

mlogsdon18 (changes requested):
Nothing major but a few comments!

components/alertbanner/index.css
```diff
@@ -0,0 +1,111 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.spectrum-AlertBanner-icon {
+  inline-size: var(--mod-alert-banner-icon-size, var(--spectrum-alert-banner-icon-size));
+  block-size: var(--mod-alert-banner-icon-size, var(--spectrum-alert-banner-icon-size));
+  margin-block-start: var(--mod-alert-banner-top-icon, var(--spectrum-alert-banner-top-icon));
+  margin-inline-end: var(--mod-alert-banner-icon-to-text, var(--spectrum-alert-banner-icon-to-text));
+}
```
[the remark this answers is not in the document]
jenndiaz: hmm I am not seeing this. What browser were you seeing this on?

components/alertbanner/metadata/alertbanner.yml
```diff
@@ -0,0 +1,116 @@
...
+    description: Standard Alert Banners
+    markup: |
+        <div class="spectrum-AlertBanner is-open">
+          <div class="spectrum-AlertBanner-body">
+            <div class="spectrum-AlertBanner-content">
+              <p class="spectrum-AlertBanner-text">Your trial has expired</p>
+            </div>
+          </div>
+          <div class="spectrum-AlertBanner-end">
+            <div class="spectrum-Divider spectrum-Divider--sizeS spectrum-Divider--vertical spectrum-AlertBanner-divider spectrum-Divider--staticWhite" style="height: auto; align-self: stretch;"></div>
```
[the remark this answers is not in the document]
jenndiaz: ohh this is a good call. 

And has me thinking, if I can do this:
```
.spectrum-Divider--vertical {
  margin-block: var(--mod-alert-banner-edge-to-divider, var(--spectrum-alert-banner-edge-to-divider));
  height: auto;
  align-self: stretch;
}`
```

with in the index file for the Alert Banner component, then I don't need to be adding another class like I was with `.spectrum-AlertBanner-divider` which means I can refactor some of the other classes here too. 
Am I on the right track here?

components/alertbanner/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "main": "dist/index-vars.css",
+  "scripts": {
+    "build": "gulp"
+  },
+  "peerDependencies": {
+    "@spectrum-css/closebutton": "^3.0.0",
+    "@spectrum-css/divider": "^2.0.25",
+    "@spectrum-css/icon": "^3.0.0",
+    "@spectrum-css/tokens": ">=8.0.2"
```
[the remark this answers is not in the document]
jenndiaz: updated each of the peer dependencies!

components/alertbanner/stories/template.js
jenndiaz: my thought was that these styles should only be used when vertical and I was trying to make that clear. Do you think it is more confusing this way?

components/alertbanner/stories/template.js
jenndiaz: deleted!

components/alertbanner/themes/express.css
```diff
@@ -0,0 +1,20 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+@container (--system: express) {
+  .spectrum-AlertBanner {
+    --spectrum-alert-banner-close-button-spacing: var(--spectrum-spacing-200);
+    --spectrum-alert-banner-edge-to-divider: var(--spectrum-spacing-200);
+    --spectrum-alert-banner-edge-to-button: var(--spectrum-spacing-200);
```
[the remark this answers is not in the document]
jenndiaz: yep, thats correct, saw the color one was here and got carried away.

I saw that we used `.spectrum--large & {` within the main class for contextual help so I followed a similar pattern here too.

components/alertbanner/index.css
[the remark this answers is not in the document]
mlogsdon18: I'm in Chrome and it looks like it's only happening for me in Storybook

components/alertbanner/metadata/alertbanner.yml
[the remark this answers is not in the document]
mlogsdon18: Yes! That way we can simplify the HTML a bit.

components/alertbanner/themes/express.css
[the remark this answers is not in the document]
mlogsdon18: Yep that's look great! You can delete them from these files then :)

components/alertbanner/index.css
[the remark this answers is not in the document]
mlogsdon18: Looks like you fixed it!

mlogsdon18 (approved):

components/alertbanner/index.css
```diff
@@ -0,0 +1,128 @@
...
+  --spectrum-alert-banner-informative-background: var(--spectrum-informative-background-color-default);
+  --spectrum-alert-banner-negative-background: var(--spectrum-negative-background-color-default);
+  --spectrum-alert-banner-font-color: var(--spectrum-white);
+
+  .spectrum--large & {
+    --spectrum-alert-banner-close-button-spacing: var(--spectrum-spacing-200);
+    --spectrum-alert-banner-edge-to-divider: var(--spectrum-spacing-200);
+    --spectrum-alert-banner-edge-to-button: var(--spectrum-spacing-200);
+    --spectrum-alert-banner-text-to-button-vertical: var(--spectrum-spacing-200);
+  }
```
pfulton, line 41: We'll need to pull these out and place them into `tokens/custom-spectrum/custom-large-vars.css`, as SWC isn't currently set up to take in these scale-related custom properties from the components themselves.

pfulton (changes requested):
One thought here, @jenndiaz, otherwise, this looks awesome. Nice work!

components/alertbanner/index.css
```diff
@@ -0,0 +1,128 @@
...
+    margin-block: var(--mod-alert-banner-edge-to-button, var(--spectrum-alert-banner-edge-to-button));
+    margin-inline-end: var(--mod-alert-banner-text-to-divider, var(--spectrum-alert-banner-text-to-divider));
+    margin-inline-start: auto;
+  }
+
+  .spectrum-CloseButton {
+    margin-inline: var(--mod-alert-banner-close-button-spacing, var(--spectrum-alert-banner-close-button-spacing));
+    margin-block-start: var(--mod-alert-banner-close-button-spacing, var(--spectrum-alert-banner-close-button-spacing));
+    align-self: flex-start;
+  }
```
pfulton, line 84: I think we may want to consider using the custom properties approach that we discussed for the DropZone component instead of using these child selectors.

pfulton (changes requested):
Sorry, just one more thing!

components/alertbanner/index.css
jenndiaz: got this changed, are the custom props good to stay with in this PR or do they need to be moved into a separate one?

components/alertbanner/stories/template.js
jenndiaz: ok I think I am following you're suggesting that it should be `style` so it can be more flexible and other components can pass a style to the `divider`. 

In this case if the divider is vertical, as per the docs for the component, it needs the inline styling `align-self: stretch; height: auto`  So I don't really want it flexible. I refactored it to:
`style=${vertical === true ? "min-height: 20px; height: auto; align-self: stretch" : ""}`

This way I am also not passing around something that should always be the same if its vertical.
This might need more refactoring as divider is used more broadly but for now it seems like the simplest approach.

components/alertbanner/metadata/alertbanner.yml
```diff
@@ -0,0 +1,116 @@
+name: Alert Banner
```
castastrophe, line 1:
```suggestion
name: Alert banner
```

Minor nit - technically we're supposed to be using sentence case for headings rather than title case.

components/alertbanner/metadata/alertbanner.yml
```diff
@@ -0,0 +1,116 @@
...
+description: Alert banners show pressing and high-signal messages, such as system alerts. They're meant to be noticed and prompt users to take action.
+sections:
+  - name: Custom Properties API
+    description: |
+      This component can be modified via its `--mod-*` prefixed custom properties. A list of those prefixed custom properties can be found <a href="https://github.com/adobe/spectrum-css/tree/main/components/alertbanner/metadata/mods.md">here</a>.
```
castastrophe, line 6: You can remove this section - we'll be launching the new 11ty soon and I've gone through and removed these from the docs.

components/alertbanner/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "main": "dist/index-vars.css",
+  "scripts": {
+    "build": "gulp"
+  },
+  "peerDependencies": {
+    "@spectrum-css/closebutton": ">=3",
+    "@spectrum-css/divider": ">=2.0.25",
+    "@spectrum-css/icon": ">=3",
+    "@spectrum-css/tokens": ">=8.0.2"
```
castastrophe, line 24: I believe we're up to 10 on the tokens now

components/alertbanner/stories/template.js
```diff
@@ -0,0 +1,75 @@
...
+import { Template as Divider } from '@spectrum-css/divider/stories/template.js';
+import { Template as Button } from '@spectrum-css/button/stories/template.js';
+import { Template as CloseButton } from "@spectrum-css/closebutton/stories/template.js";
+import { Template as Icon } from '@spectrum-css/icon/stories/template.js';
+
+
+import '../index.css';
+
+export const Template = ({
+  rootClass = "${rootClass}",
```
castastrophe, line 13:
```suggestion
  rootClass = "${rootClass}",
```

Does this work? I don't think it will b/c it's trying to use itself to set itself.

components/button/index.css
```diff
@@ -112,6 +112,10 @@ governing permissions and limitations under the License.
   /* let staticColor variants inherit their color */
   color: inherit;
 
+  margin-block: var(--mod-button-margin-block);
```
castastrophe, line 115: NICE! #all-the-mods!!!

components/divider/stories/divider.stories.js
```diff
@@ -17,6 +17,16 @@ export default {
       options: ["s", "m", "l"],
       control: "select"
     },
+    staticColor: {
```
castastrophe, line 20: You are so freaking awesome

pfulton (approved):

PR #1803, opened 2023-04-24
fix: textfield and combobox fixes, adjust padding calculations 

jawinn:
## Description

Some fixes for Textfield and Combobox:

- **Textfield** and **Combobox**: Subtract border width from padding calculations because most of the "...to-edge" tokens include the border. The values were previously 1px larger than they should have been.
- **Textfield** and **Combobox**: fix corner radius of focus indicator when using a larger border radius or a larger indicator width (e.g. to test, try setting a  larger `--mod-textfield-border-width` or `--mod-textfield-focus-indicator-width`. Changed the calculation  to add the focus indicator width instead of the border width)
- **Combobox**: adjust styles so custom property for border width is correctly overriding everything from its sub-components. Previously, using `--mod-combobox-border-width` had no effect.
- **Combobox**: simplify/remove some custom properties related to those fixes and quiet variant.
- **Combobox**: border-radius should not increase for t-shirt sizes (per design feedback) 
- **Combobox**: Use a corrected calculation of the x-offset for the picker button on the quiet variant, based on design feedback and using the same method as the 'search' component.
- **Textfield**: Also adds "size" control to stories

Firefox and Chrome, Mac

## Screenshots

Padding issue before and after:
Screenshot 2023-04-24 at 11 03 05 AM
Screenshot 2023-04-24 at 11 08 07 AM

Issues with corner radius calculation on focus indicator when using a larger border width, or when changing the focus indicator width (now fixed):
Screenshot 2023-04-24 at 1 00 55 PM
Screenshot 2023-04-24 at 3 35 52 PM

## To-do list

### DIFF
--- a/components/combobox/index.css
+++ b/components/combobox/index.css
@@ -130,7 +130,10 @@ governing permissions and limitations under the License.
   --spectrum-combobox-spacing-inline-icon-to-button: var(--spectrum-combo-box-visual-to-field-button-quiet);
   --spectrum-combobox-spacing-inline-start-edge-to-text: var(--spectrum-field-edge-to-text-quiet);
   --spectrum-combobox-spacing-label-to-combobox: var(--spectrum-field-label-to-component-quiet-medium);
-  --spectrum-combobox-button-inline-offset: var(--spectrum-spacing-100);
+  --spectrum-combobox-button-inline-offset: calc(
+    (var(--mod-combobox-block-size, var(--spectrum-combobox-block-size)) / 2) - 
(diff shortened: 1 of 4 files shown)


mlogsdon18 (commented):
This looks pretty good to me Josh! Not something you did but I noticed that the sizing control isn't available for the Textfield storybook. Could you add that in for testing?

castastrophe:
Very cool! When this is ready, let's flag this for a vrt run to update baselines. 🎉

PR #1804, opened 2023-04-24
fix(picker): update spacing tokens and storybook

mlogsdon18:
## Description

This updates the Spacing (top/bottom and end edge to disclosure icon) tokens for the Picker component.

There were also some things that were not in Storybook or were not working in Storybook that I went ahead and fixed in this PR as well as styles that were not being used how the Figma was specifying
- Added a separate Quiet story
- Added a separate Loading story and the functionality required
- Added a FieldLabel 
- Updated icons based on size for PickerButton
- Fixed `min-inline-size` to use the multiplier correctly

 - **How this was tested:** 

## To-do list


components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -126,7 +135,7 @@ export default {
   },
   args: {
     rootClass: "spectrum-PickerButton",
-    label: "Select a Country",
+    placeholder: "Select a Country",
```
castastrophe, line 138:
```suggestion
    placeholder: "Select a country",
```

Minor nit - the styleguide says we are using sentence case rather than title case so I try to align the demos with that whenever I can.

castastrophe (approved):

PR #1808, opened 2023-04-25
chore(swatch): add functionality to storybook

castastrophe:
## Description

Bring in some of the functionality to the swatch storybook that exists for the site.

 - **How this was tested:**
 - [ ] Open storybook instance, click the swatch: expect the is-selected class to be toggled
 - [ ] Open storybook, key press the swatch: expect the is-selected class to be toggled

## To-do list

### DIFF
--- a/components/swatch/stories/swatch.stories.js
+++ b/components/swatch/stories/swatch.stories.js
@@ -4,7 +4,7 @@ import { Template } from "./template";
 export default {
 	title: "Components/Swatch",
 	description:
-		"A swatch shows a small sample of a fill—such as a color, gradient, texture, or material—that is intended to be applied to an object.",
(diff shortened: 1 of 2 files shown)


pfulton (changes requested):
This is a much-needed start to this component's entry in Storybook. However, it looks like there are several other variations that we might want to have represented in here (gradient, mixed-value, nothing, disabled needs the disabled icon).

We should probably audit this one and get a proper Jira ticket to work on this. What do you think?

castastrophe:
@pfulton That's a great point - I came across this trying to capture the JS from the site in the storybook entries and didn't want to lose that effort but you're right, there's a lot more to do to really cover this in testing

PR #1809, opened 2023-04-25
chore: remove unused tooling

castastrophe:
## Description

🚨 Lots of files alert! 🚨

- All copyrights in the repo updated to use the `/*!` syntax so they do not get stripped with general comments
- Peer dependencies in components updated to the permissive syntax `>=` to allow greater support in complex projects
- Remove bundle-builder's unused:
  - release, prepack, and publish commands
  - exec lib
- Component builders ->
  - use published postcss-discard-comments over the internal-only postcss-strip-comments
  - move browserlist support to the root package.json
  - unused legacy build file
  - unused mutate selectors plugin
  - unused mixins and functions plugins (was used by typography but not since migration)
- Generator reflects current state of repo builds

## To-do list


pfulton (approved):
This looks good to me. I was able to pull down, run a build locally and also able to run both the docs and the storybook environments.

I also checked the build artifacts to be sure that things were still compiling as expected, and those were fine too.

PR #1815, opened 2023-04-27
docs: add aria-label for actionbar close button (CSS-464)

pfulton:
## Description
Adds `aria-label="Clear selection"` to the Close Buttons found within the Action Bar component.

[CSS-464](https://jira.corp.adobe.com/browse/CSS-464)

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/actionbar/metadata/actionbar.yml
+++ b/components/actionbar/metadata/actionbar.yml
@@ -33,7 +33,7 @@ examples:
     markup: |
       <div class="spectrum-ActionBar is-open">
         <div class="spectrum-Popover spectrum-ActionBar-popover is-open">
-          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM">
+          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM" aria-label="Clear selection">
             <svg class="spectrum-CloseButton-UIIcon spectrum-Icon spectrum-UIIcon-Cross100" focusable="false" aria-hidden="true">
               <use xlink:href="#spectrum-css-icon-Cross100" />
             </svg>
@@ -58,7 +58,7 @@ examples:
 
       <div class="spectrum-ActionBar is-open">
         <div class="spectrum-Popover spectrum-ActionBar-popover is-open">
-          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM">
+          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM" aria-label="Clear selection">
             <svg class="spectrum-CloseButton-UIIcon spectrum-Icon spectrum-UIIcon-Cross100" focusable="false" aria-hidden="true">
               <use xlink:href="#spectrum-css-icon-Cross100" />
             </svg>
@@ -92,7 +92,7 @@ examples:
 
       <div class="spectrum-ActionBar is-open">
         <div class="spectrum-Popover spectrum-ActionBar-popover is-open">
-          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM">
+          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM" aria-label="Clear selection">
             <svg class="spectrum-CloseButton-UIIcon spectrum-Icon spectrum-UIIcon-Cross100" focusable="false" aria-hidden="true">
               <use xlink:href="#spectrum-css-icon-Cross100" />
             </svg>
@@ -127,7 +127,7 @@ examples:
     markup: |
       <div class="spectrum-ActionBar spectrum-ActionBar--emphasized is-open">
         <div class="spectrum-Popover spectrum-ActionBar-popover is-open">
-          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM spectrum-CloseButton--staticWhite">
+          <button class="spectrum-CloseButton spectrum-CloseButton--sizeM spectrum-CloseButton--staticWhite" aria-label="Clear selection">
             <svg class="spectrum-CloseButton-UIIcon spectrum-Icon spectrum-UIIcon-Cross100" focusable="false" aria-hidden="true">
               <use xlink:href="#spectrum-css-icon-Cross100" />
             </svg>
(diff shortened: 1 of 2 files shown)


mlogsdon18 (commented):
These look good! My only thought is that while you're adding aria labels to the close button, you could fix them for the other buttons in this component as well. I can't comment on the code since you didn't change it but there are unneeded aria labels on the SVGs that have an `aria-hidden="true"` attribute. These should be removed if the button has text, but added to the `<button>` element if they only contain the SVG. Currently they aren't read at all. 

```<svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Delete">```

**Comment by pfulton** [comment]
> These look good! My only thought is that while you're adding aria labels to the close button, you could fix them for the other buttons in this component as well. I can't comment on the code since you didn't change it but there are unneeded aria labels on the SVGs that have an `aria-hidden="true"` attribute. These should be removed if the button has text, but added to the `<button>` element if they only contain the SVG. Currently they aren't read at all.
> 
> `<svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Delete">`

Great suggestion. I just pushed up [a commit](https://github.com/adobe/spectrum-css/pull/1815/commits/aaf730b08777d47a57a8a4ccd5e0f72b58d2d8c2) to take care of those.

File: components/actionbar/metadata/actionbar.yml [hunk 1]
```diff
@@ -68,19 +68,19 @@ examples:
 
           <div class="spectrum-ActionGroup spectrum-ActionGroup--sizeM">
             <button class="spectrum-ActionButton spectrum-ActionButton--sizeM spectrum-ActionButton--quiet spectrum-ActionGroup-item">
-              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Edit">
+              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Edit">
```
mlogsdon18, line 71: Ok so it looks like you removed `aria-hidden` from the buttons that also have text in them. This causes this button to be read "Edit Edit" by the screen reader because it's reading the SVG label and the button text. So this is a place where we actually want to keep the `aria-hidden` so that doesn't happen. The buttons below this that only have icons in them (lines 103-118ish) are where you would want to remove the `aria-hidden` because they don't have button text and the screen reader can't read the aria-label because of the aria-hidden and it just reads "button"

components/actionbar/metadata/actionbar.yml
```diff
@@ -162,19 +162,19 @@ examples:
 
           <div class="spectrum-ActionGroup spectrum-ActionGroup--sizeM">
             <button class="spectrum-ActionButton spectrum-ActionButton--sizeM spectrum-ActionButton--quiet spectrum-ActionButton--staticWhite spectrum-ActionGroup-item">
-              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Edit">
+              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Edit">
```
mlogsdon18, line 165:
```suggestion
              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Edit" aria-hidden="true">
```

components/actionbar/metadata/actionbar.yml
```diff
@@ -162,19 +162,19 @@ examples:
...
                 <use xlink:href="#spectrum-icon-18-Edit"></use>
               </svg>
               <span class="spectrum-ActionButton-label">Edit</span>
             </button>
             <button class="spectrum-ActionButton spectrum-ActionButton--sizeM spectrum-ActionButton--quiet spectrum-ActionButton--staticWhite spectrum-ActionGroup-item">
-              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Copy">
+              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Copy">
```
mlogsdon18, line 171:
```suggestion
              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Copy" aria-hidden="true">
```

components/actionbar/metadata/actionbar.yml
```diff
@@ -162,19 +162,19 @@ examples:
...
                 <use xlink:href="#spectrum-icon-18-Copy"></use>
               </svg>
               <span class="spectrum-ActionButton-label">Copy</span>
             </button>
             <button class="spectrum-ActionButton spectrum-ActionButton--sizeM spectrum-ActionButton--quiet spectrum-ActionButton--staticWhite spectrum-ActionGroup-item">
-              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true" aria-label="Delete">
+              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Delete">
```
mlogsdon18, line 177:
```suggestion
              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-label="Delete" aria-hidden="true">
```

components/actionbar/metadata/actionbar.yml
mlogsdon18: Line 155 has a Close button on it that needs the `aria-label="Clear selection"` as well, I just can't comment on it because it wasn't changed

mlogsdon18 (changes requested):
Almost there!

mlogsdon18 (approved):
WOO! Looks great!

PR #1817, opened 2023-04-28
fix(contextualhelp): dark mode font color

mlogsdon18:
## Description

Adds `.spectrum-Heading` and `.spectrum-Body` classes to the heading and paragraph tags in order to get the global font colors needed since font color was not a specified token for this component. 

 - Latest versions of Chrome, Safari, Firefox

## To-do list

### DIFF
--- a/components/contextualhelp/metadata/contextualhelp.yml
+++ b/components/contextualhelp/metadata/contextualhelp.yml
@@ -13,7 +13,7 @@ examples:
         <div class="spectrum-Examples-item">
           <div class="spectrum-ContextualHelp">
             <button class="spectrum-ActionButton spectrum-ActionButton--sizeXS spectrum-ContextualHelp-button">
-              <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-ActionButton-icon" focusable="false" aria-hidden="true">
+              <svg class="spectrum-Icon spectrum-Icon--sizeXS spectrum-ActionButton-icon" focusable="false" aria-hidden="true">
                 <use xlink:href="#spectrum-icon-18-Info" />
               </svg>
             </button>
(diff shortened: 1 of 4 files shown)


Rajdeepc (approved):

components/contextualhelp/metadata/contextualhelp.yml
```diff
@@ -18,17 +18,17 @@ examples:
               </svg>
             </button>
             <div class="spectrum-Popover is-open spectrum-Popover--bottom-start spectrum-ContextualHelp-popover" role="presentation">
-              <h2 class="spectrum-ContextualHelp-heading">Permission required</h2>
-              <p class="spectrum-ContextualHelp-body">Cupcake ipsum dolor sit amet jelly beans. Chocolate jelly caramels. Icing soufflé chupa chups donut cheesecake. Jelly-o chocolate cake sweet roll cake danish candy biscuit halvah.</p>
+              <h2 class="spectrum-Heading spectrum-Heading--sizeM spectrum-ContextualHelp-heading">Permission required</h2>
```
castastrophe, line 21: I think ideally we'd simplify this such that the spectrum heading and body classes aren't needed and can be extended in the contextual help class. The more isolated a component is from other dependencies, the easier it is to convert to a web component.

castastrophe (approved):

components/contextualhelp/metadata/contextualhelp.yml
mlogsdon18: Right. I can go ahead and just use the tokens that the heading and body classes use even though they weren't specified for this component. Let me see about making that change!

mlogsdon18:
@castastrophe I was able to add in the tokens so that we could remove the `.spectrum-Heading` and `.spectrum-Body` classes. I also noticed that two of the tokens I was originally missing have since been added so I updated those as well

pfulton (changes requested):
Thanks for taking care of this, @mlogsdon18. Looks like this resolves the font color issue.

I do have one more thing related to this component: it looks like the button's top edge-to-icon spacing is off
Screen Shot 2023-05-02 at 10 42 39 AM
Screen Shot 2023-05-02 at 10 42 50 AM


Would you mind looking into that?

components/contextualhelp/stories/template.js
```diff
@@ -37,8 +37,8 @@ export const Template = ({
       })}
       ${Popover({
         content: [
-          title ? html`<h2 class="${rootClass}-heading">${title}</h2>` : "",
-          body ? html`<p class="${rootClass}-body">${body}</p>` : "",
+          title ? html`<h2 class="spectrum-Heading ${rootClass}-heading">${title}</h2>` : "",
+          body ? html`<p class="spectrum-Body ${rootClass}-body">${body}</p>` : "",
```
castastrophe, line 41: Are the header and body classes still needed?

pfulton (approved):

PR #1832, opened 2023-05-02
fix(icon): correct storybook icon sizing substitution

jawinn:
## Description
Fix issue where use of Icon in other stories would not change its sizing (scale) based on the value of the size arg. The conditional with uiIcons.includes was comparing against the icon name which still contained the sizing number, so was always failing.

Also adds some docs around this.

Chrome, Mac. Testing 'Table' in storybook with t-shirt sizes ("small" should now show smaller icon). Tested some of the other components in Storybook that are using Icon to make sure they still look the same: checkbox, closebutton, textfield

## To-do list


castastrophe (approved):
Looks great! https://pr-1832--spectrum-css.netlify.app/preview/?path=/story/table--default&args=size:s

PR #1856, opened 2023-05-09
fix(stepper): fix stepper button padding and borders

mlogsdon18:
## Description

This PR fixes the appearance of the Stepper buttons. They were slightly off center due to how the padding and borders were being applied. This PR updates the styling to more closely resemble what is currently shown in SWC and what was there previously. 

**NOTE:** The Stepper component is not yet migrated, so these changes are just a band-aid for now. The component was slightly migrated to allow for Textfield to get through.

## To-do list

### DIFF
--- a/components/stepper/index.css
+++ b/components/stepper/index.css
@@ -63,10 +63,8 @@ governing permissions and limitations under the License.
   --spectrum-stepper-button-width: calc(var(--spectrum-spacing-400) - var(--spectrum-stepper-border-width)); /* this matches the previous WIDTH token */
   --spectrum-stepper-button-padding: calc(var(--spectrum-spacing-200) / 2); /* 6px md + 7px lg - previously 6px md + 7.5px lg */
   --spectrum-stepper-button-gap: var(--spectrum-stepper-button-gap-reset);
-  --spectrum-stepper-button-border-radius: var(--spectrum-stepper-button-border-radius-reset);
 
   /* background same as textfield */
-  --spectrum-stepper-background-color: var(--spectrum-gray-50);
   --spectrum-stepper-background-color-disabled: var(--spectrum-disabled-background-color);
 
  /*** Quiet Variant ***/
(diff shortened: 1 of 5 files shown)


components/stepper/index.css
```diff
@@ -535,21 +511,24 @@ governing permissions and limitations under the License.
...
   padding-block-end: var(--mod-stepper-icon-nudge-end, var(--spectrum-stepper-icon-nudge-end));
+  padding-block-start: 0;
 
-  border-start-start-radius: 0;
-  border-start-end-radius: 0;
-  border-end-start-radius: var(--mod-stepper-button-border-radius, var(--spectrum-stepper-button-border-radius));
-  border-end-end-radius: var(--mod-stepper-button-border-radius, var(--spectrum-stepper-button-border-radius));
+  border-start-start-radius: var(--spectrum-stepper-button-border-radius-reset);
+  border-start-end-radius: var(--spectrum-stepper-button-border-radius-reset);
+  border-end-start-radius: var(--mod-stepper-button-border-radius-reset, var(--spectrum-stepper-button-border-radius-reset));
```
pfulton, line 530: Let's just drop the `--mod-` here. It looks like this is the only place we have it defined, and due to the temporary nature of these changes, I'd rather exclude this for now so that it doesn't require future breaking changes or support.

```suggestion
  border-end-start-radius: var(--spectrum-stepper-button-border-radius-reset);
```

pfulton (changes requested):
Just one thing here @mlogsdon18 - after you remove that `--mod`, please re-generate the mods file, too!

components/stepper/index.css
mlogsdon18: On it!

mlogsdon18: Ok this has been updated and the mods file regenerated

pfulton (approved):

pfulton (approved):

PR #1859, opened 2023-05-10
chore: harden mods script with more robust regex

castastrophe:
## Description

This PR is hardening the mod script by leveraging fast glob to pull out the necessary CSS assets directly without additional loops. This then reads in the content and runs a matchAll on a more expansive regex command in order to isolate mod properties being _used_. This is important because any mods being set in the component will likely be used to customized nested components. Before processing this runs a simple alphabetical sort so that results appear in the same order each time. Finally, it writes to the mods markdown file as well as adding a mods.json to the dist directory for downstream consumption.

 - **How this was tested:** `yarn docs:mod`

## To-do list


tasks/mod-extractor.js
```diff
@@ -10,62 +10,48 @@ governing permissions and limitations under the License.
...
+        const regex = /var\((--mod-(?:\w|-)+)/g;
+        const matches = await fsp.readFile(filepath, 'utf-8').then((content) => {
+            return [...content.matchAll(regex)].map((match) => match[1]);
+        }).catch((err) => {
+            console.log(err);
+        });
+
+        if (!matches || matches.length === 0) continue;
+
+        const found = [...new Set(matches)].sort();
```
pfulton, line 35: 🤩

pfulton (approved):

tasks/mod-extractor.js
Rajdeepc: Great!! A small suggestion. It would be great if you can add some code comments too?!!!

castastrophe: Just pushed up some more comments! Thanks for the suggestion!

PR #1863, opened 2023-05-11
fix(textfield, stepper): button padding and focus indicator

mlogsdon18:
## Description

- Removes `padding-block-end` to fix the Express version of the Stepper button.
- Switches Textfield and Stepper to use `outline` for the focus indicator rather than an `::after` pseudo element.

### To Review/Test

- [x] Verify button icons are 10x10 and centered on both Spectrum and Express
- [x] Verify that Quiet, Disabled and all other versions still look the way that they should

### DIFF
--- a/components/stepper/index.css
+++ b/components/stepper/index.css
@@ -43,7 +43,6 @@ governing permissions and limitations under the License.
 
   /*** Button Icons - appears to adjust position ***/
   --spectrum-stepper-icon-nudge-start: 1px; /* placeholder - was var(--spectrum-global-dimension-size-10) - 1px */
-  --spectrum-stepper-icon-nudge-end: 2px; /* placeholder -was var(--spectrum-global-dimension-size-25) - 2px */
 
  /*** :AFTER - this is for the :after element labeled below as hit area, but used as focus indicator in ActionButton ***/
   --spectrum-stepper-button-offset: calc(var(--spectrum-stepper-button-width) / 2 - var(--spectrum-stepper-icon-width) / 2);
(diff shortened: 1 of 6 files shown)


jawinn (changes requested):
By switching from the `::after` psuedo outline to the `outline` property, it looks like we've lost the focus indicator style for the quiet variant, which is a single line underneath. Was that change necessary for a fix? This also affects other components using textfield such as combobox and search (_also weirdly the [search docs page on this PR site](https://pr-1863--spectrum-css.netlify.app/search.html) isn't loading right for some reason?_).
Screenshot 2023-05-15 at 2 25 27 PM

mlogsdon18:
> By switching from the `::after` psuedo outline to the `outline` property, it looks like we've lost the focus indicator style for the quiet variant, which is a single line underneath. Was that change necessary for a fix? This also affects other components using textfield such as combobox and search (_also weirdly the [search docs page on this PR site](https://pr-1863--spectrum-css.netlify.app/search.html) isn't loading right for some reason?_). Screenshot 2023-05-15 at 2 25 27 PM

What in the world I specifically made sure this was fixed.... This component has been craziness. I'll take a look

components/stepper/index.css
```diff
@@ -315,6 +295,15 @@ governing permissions and limitations under the License.
     border-radius: 0;
     inline-size: var(--mod-stepper-quiet-width, var(--spectrum-stepper-quiet-width));
 
+    &::after {
+      content: '';
+      position: absolute;
+      bottom: calc(-1 * var(--mod-stepper-focus-indicator-gap, var(--spectrum-stepper-focus-indicator-gap)) * 2);
```
jawinn, line 301: I think this should be...
```suggestion
      bottom: calc(-1 * (var(--mod-stepper-focus-indicator-gap, var(--spectrum-stepper-focus-indicator-gap)) + var(--mod-textfield-focus-indicator-width, var(--spectrum-textfield-focus-indicator-width)));
```
(Mod the focus-indicator-width to a bigger value and you'll see; this should still maintain the gap).

jawinn:
The outline appearing for quiet is fixed now, thanks.

I just pushed up a PR with the update from the last comment, and the issue I was seeing with Combobox. I used the custom property passthroughs with the mods, and it worked great!

FYI this was the issue I fixed with Combobox, when tabbing into quiet fields:
Screenshot 2023-05-15 at 5 13 16 PM

jawinn (approved):

PR #1892, opened 2023-05-22
docs(alertbanner): add link to mods docs & fix capitalization

pfulton:
## Description
Small docs update to link out to the mods list + also update the capitalization in the name of the component.

 - **How this was tested:** 

## To-do list

### DIFF
--- a/components/alertbanner/metadata/alertbanner.yml
+++ b/components/alertbanner/metadata/alertbanner.yml
@@ -1,5 +1,9 @@
-name: Alert banner
+name: Alert Banner
 description: Alert banners show pressing and high-signal messages, such as system alerts. They're meant to be noticed and prompt users to take action.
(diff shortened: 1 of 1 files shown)


components/alertbanner/metadata/alertbanner.yml
```diff
@@ -1,5 +1,9 @@
-name: Alert banner
+name: Alert Banner
```
castastrophe, line 1: So I believe the content direction is to use sentence case and not title case for content on the site so Alert banner would be correct over Alert Banner.  I also suggested the link to the mods be left off since it was going to be accounted for in the 11ty rework which is impending (just another diff I have to resolve)

pfulton: aha! Ok.

PR #1898, opened 2023-05-25
fix(textfield): storybook valid icon size and rename mod

jawinn:
## Description
**fix(textfield): valid in storybook - display correct icon size**
Fixes the 'valid' checkmark icon not displaying as the right size in Storybook. Needed to add the sizing number, which is programattically replaced with the right size based on the t-shirt size passed to the icon.

**fix(textfield): mod name for keyboard focus color**
Mod custom property for one of the keyboard focus styles was using the focus mod instead of the keyboard focus mod.

Tested in Storybook, Chrome on Mac

## Screenshots
Storybook valid icon before:
Screenshot 2023-05-25 at 11 48 20 AM

Storybook valid icon after:
Screenshot 2023-05-25 at 11 48 55 AM

## To-do list

### DIFF
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -552,9 +552,8 @@ governing permissions and limitations under the License.
   /* keyboard focus */
   .is-keyboardFocused &,
   &:focus-visible {
-
     border-color: var(--highcontrast-textfield-border-color-keyboard-focus, var(--mod-textfield-border-color-keyboard-focus, var(--spectrum-textfield-border-color-keyboard-focus)));
-    color: var(--highcontrast-textfield-text-color-keyboard-focus, var(--mod-textfield-text-color-focus, var(--spectrum-textfield-text-color-keyboard-focus)));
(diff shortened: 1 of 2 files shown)


jenndiaz:
Both good catches! 

While we are in there, are we concerned about the spacing the around the search icon with in the storybook for text-field? I also didn't see an icon variant in the docs site for text-field should this even be here? 

Or is this something you're looking at it another PR? 

Screenshot 2023-05-25 at 10 27 00 AM

jawinn:
> While we are in there, are we concerned about the spacing the around the search icon with in the storybook for text-field? I also didn't see an icon variant in the docs site for text-field should this even be here?
> 
> Or is this something you're looking at it another PR?

This story and the styles for it are actually something I just removed as part of the `search` migration PR. It addressed the existing textfield TODO comment for when search is migrated: https://github.com/adobe/spectrum-css/pull/1761#issuecomment-1561952986

jenndiaz (approved):

Issue #1526, opened 2022-10-07
tabs: "quiet" lack of border is applied to vertical, compact tabs


hunterloftis:
## Description

Re: conversation w/ @pfulton and PJ, vertical compact should have a border, but currently doesn't:

https://opensource.adobe.com/spectrum-css/tabs.html#compactverticaltabs

## Screenshots

![image](https://user-images.githubusercontent.com/364501/194618029-a9ec0f81-957c-48c5-853b-2fefcf73d742.png)

I'm adding this issue to track, so I can override it temporarily in Spectrum Web Components & then delete the override once this is 👍

pfulton:
closed via #1902

PR #1902, opened 2023-05-26
feat(tabs)!: migrate tokens

mlogsdon18:
## Description

This PR updates the Tabs component to use the new tokens package. 

It also adds additional stories to Storybook. Since the components do not currently have an overarching parent directory it moved the Tabs to the bottom, this should be changed in a separate PR. But this one does contain a few changes to the base styling for Storybook to allow for nested stories. 

## Issues Addressed
- #1526 

## To-do list


jawinn (approved):
Excellent work! I compared the new and old and everything looks great. As does the CSS and WHCM.

My only questions are around the Storybook organization.

1- Do we need that many separate stories defined, when we have controls for them? It seems like there are too many tiers. What if we just had Horizontal and Vertical, with the 3 stories under each of those, and emphasized, quiet, and compact are handled by the controls?

 2- I was looking for tabs at first and couldn't find it, because it's been pushed to the bottom under Components. I looked at Storybook config options for a bit and can see how this is a bit tricky and there doesn't seem to be a way to push it into the root like the others. I know you had looked into this as well. I did find out that we can modify our main.js config to prefix all the components:
```module.exports = {
    stories: [
        {
            directory: "../../components",
            files: '*/stories/*.stories.@(js|mdx)',
            titlePrefix: 'Components'
        }
    ],  
```
Though this will change all the existing story URLs. So then if we wanted those URLs to not change, we'd have to change every component's story files to define the "id" ("_It is possible to manually set the story's id, which is helpful if you want to rename stories without breaking permalinks._"). And if we have to do that anyway, it might be best to just add "Components/" in each title anyway, rather than a global setting. 

I'm wondering if we should also do that organization in another PR?

mlogsdon18:
@jawinn Yeah my thought was that we could do the Storybook organization in another PR so I can get that up! I'll want Patrick to check that to make sure that it won't totally break Chromatic. 

And I was torn on the amount of stories as well. I felt like they were different enough that since we're running VRTs, we needed to have a visual representation of each of the ways represented in the docs so that they were all tested for regressions... it's a bit much but I didn't want to give up on testing Compact, Emphasized and Quiet with Chromatic because there were too many options. @pfulton what do you think?

pfulton:
Released a new beta: `@spectrum-css/tabs@4.0.0-beta.1`

pfulton:
Published a new beta: `@spectrum-css/tabs@4.0.0-beta.2`

PR #1907, opened 2023-05-30
chore: add prettier configuration

castastrophe:
## Description

A prettier config was required for a utility being used in the updated 11ty site so I thought we could pull the implementation out into it's own PR and evaluate if we want to add it as a linter during commits.

Implementation options:

- Merge this and let prettier do it's thing as files are updated in PRs
- Merge this and then run it across all our relevant files and commit that in one go

## To-do list

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,14 @@
+**/.git
+**/.svn
+**/.hg
+**/node_modules
+
+components/*/dist
+dist
+
+.DS_Store*
+ehthumbs.db
+Icon?
+Thumbs.db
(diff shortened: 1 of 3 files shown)


pfulton (approved):
Looks good!

pfulton:
> Implementation options:

    Merge this and let prettier do it's thing as files are updated in PRs
    Merge this and then run it across all our relevant files and commit that in one go

Any idea of how many files would be affected by doing it in one fell swoop? I think maybe just doing it all at once would be fine.

PR #1908, opened 2023-05-30
fix(divider): WHCM color

mlogsdon18:
## Description

The Divider currently does not display in WHCM because the background-color is being overwritten by the size-specific class styling. This makes the WHCM styling more specific so that it applies at the right time. 

## To-do list


pfulton (approved):
Thanks for catching and taking care of this!

PR #1912, opened 2023-06-01
feat(infieldbutton): migrate tokens

mlogsdon18:
## Description

This PR updates the In-Field Button component to use the new Spectrum tokens. This component currently is not being consumed by any other component and is not in Spectrum Web Components.

## To-do list


components/infieldbutton/stories/infieldbutton.stories.js
```diff
@@ -4,79 +4,94 @@ import { Template } from "./template";
...
+      },
+      control: "boolean"
+    },
+  },
+  args: {
+    rootClass: "spectrum-InfieldButton",
+    size: "m",
+    position: "left",
+    iconName: "Add",
+    isQuiet: false
```
jenndiaz, line 68: Noticed Disabled does not have a default and initially displays with `Set Boolean` do we want it to match isQueit and start off as `false`

components/infieldbutton/index.css
```diff
@@ -10,27 +10,352 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+				--mod-infield-button-border-radius-reset,
+				var(--spectrum-infield-button-border-radius-reset)
+			);
+			border-start-start-radius: var(
+				--mod-infield-button-border-radius-reset,
+				var(--spectrum-infield-button-border-radius-reset)
+			);
+		}
+	}
+
```
jenndiaz, line 241: Looking at the Figma file it looks like the express theme has all four corners rounded, currently the code is only rounding the right or left corners in express matching the spectrum theme.

components/infieldbutton/themes/express.css
```diff
@@ -0,0 +1,30 @@
...
+    --spectrum-infield-button-border-radius-reset: var(--spectrum-in-field-button-fill-stacked-inner-border-rounding);
+
+    --spectrum-infield-button-stacked-top-border-radius-start-start: var(--spectrum-corner-radius-75);
+    --spectrum-infield-button-stacked-bottom-border-radius-end-start: var(--spectrum-corner-radius-75);
+
+    --spectrum-infield-button-background-color: var(--spectrum-gray-200);
+    --spectrum-infield-button-background-color-hover: var(--spectrum-gray-300);
+    --spectrum-infield-button-background-color-down: var(--spectrum-gray-400);
+    --spectrum-infield-button-background-color-key-focus: var(--spectrum-gray-300);
+
```
jenndiaz, line 28: I am seeing a hover background color change on the disabled variant when looking at the express theme, should we prevent that in this file?

components/infieldbutton/metadata/infieldbutton.yml
```diff
@@ -22,7 +26,7 @@ examples:
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">S</h4>
           <button class="spectrum-InfieldButton spectrum-InfieldButton--sizeS spectrum-InfieldButton--right" aria-haspopup="listbox">
             <div class="spectrum-InfieldButton-fill">
-              <svg xmlns="http://www.w3.org/2000/svg" class="spectrum-Icon spectrum-InfieldButton-icon" focusable="false" aria-hidden="true" aria-label="Add">
+              <svg xmlns="http://www.w3.org/2000/svg" class="spectrum-Icon spectrum-Icon--sizeS spectrum-InfieldButton-icon" focusable="false" aria-hidden="true" aria-label="Add">
```
jenndiaz, line 29: noticing this will result in an empty button. do we want to do something similar as we did with [ActionButton](https://github.com/adobe/spectrum-css/pull/1815)?

components/infieldbutton/metadata/infieldbutton.yml
```diff
@@ -13,6 +13,10 @@ sections:
```
jenndiaz, line 13: while in here could we update the name to be `In-Field button`

components/infieldbutton/metadata/infieldbutton.yml
```diff
@@ -109,7 +112,7 @@ examples:
     markup: |
```
jenndiaz, line 112: Just curious what the Workflow icon only variant is displaying that is different than the other variants and if it could benefit from a description?

(not sure if this is the right line 😅)

jenndiaz (changes requested):
Nice work! Just a couple of questions/comments.

components/infieldbutton/metadata/infieldbutton.yml
```diff
@@ -22,7 +26,7 @@ examples:
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">S</h4>
           <button class="spectrum-InfieldButton spectrum-InfieldButton--sizeS spectrum-InfieldButton--right" aria-haspopup="listbox">
```
jenndiaz, line 27: A little hard to tell this abstracted from it's final use but is `aria-haspopup="listbox"` the correct aria to use in this case? 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-haspopup

components/infieldbutton/index.css
mlogsdon18: Oh good catch! Don't know how I missed that

components/infieldbutton/metadata/infieldbutton.yml
mlogsdon18: I'm honestly not sure. It was a separate option from the others so I left it there? I know we have two different types of icons but since we're already using a workflow icon maybe it should show with a UI icon only?

components/infieldbutton/metadata/infieldbutton.yml
mlogsdon18: So this was already there but I left it in because these buttons are typically used to display a select menu with a list of options of some sort, which would warrant the use of `aria-haspopup`. It looks like `listbox` should be used for "lists from which a user may select one or more items" so I think it makes sense here but that obviously depends on how the user implements it

components/infieldbutton/metadata/infieldbutton.yml
mlogsdon18: I have it as `In-field button` because I thought only the first letter was supposed to be capitalized... but this one is a weird one with the hyphen

components/infieldbutton/metadata/infieldbutton.yml
mlogsdon18: Oh good call out. I'll add some aria attributes here

pfulton (changes requested):
Two things here @mlogsdon18:
1. Can you add a Storybook story for the stacked variation?
2. Can you double-check the `small` stacked variant on the docs site? It looks like the icons are not rendering properly.

Screenshot 2023-07-26 at 9 57 00 AM

pfulton (approved):

PR #1924, opened 2023-06-05
refactor(colorslider)!: core tokens migration

jawinn:
## Description

**ColorSlider core tokens migration:**
Updates the `colorslider` component to use core tokens as defined in the design.

Other notable changes:
- Updates default border-color in forced-colors (high contrast) mode to use CanvasText, so border remains visible when using a dark high contrast theme (the default).
- Updates Storybook to include an option for the gradient color stops, and creates another story using the "Alpha" variant from the example docs, which shows the checkerboard background. This should be useful for future tracking of visual regressions.
- Handles flipping the gradient when using RTL base direction. A similar change was made [on the SWC side](https://github.com/adobe/spectrum-web-components/pull/3302) and that `transform: scaleX` style rule can likely be removed once this merges in there.
- Disabled no longer shows as 2px larger than enabled

~**ColorHandle**:~
~Includes minor adjustment to styles to add new mod `--mod-colorhandle-hitarea-border-radius`.~
_ColorHandle change has been cherry-picked into its own PR #2065 . This update to ColorSlider will require a minimum version of ColorHandle that includes that update when it merges._

~**OpacityCheckerboard**:~
~The OpacityCheckerboard story template is modified, so it can accept content as an arg and exclude the wrapper, so that it can be used in the ColorSlider story, and other components' stories. Also makes some other minor adjustments and adds a story.~
_These changes were also cherry-picked and included in an update that has merged into main ( #2056 )._

CSS-188

Chrome, Firefox, and Safari on Mac. And WHCM Edge through AssistivLabs.

## To-do list


components/colorslider/index.css
```diff
@@ -38,59 +60,140 @@
...
+
+	--spectrum-color-slider-border-color-local: var(
+		--highcontrast-color-slider-border-color,
+		var(
+			--mod-color-slider-border-color,
+			var(--spectrum-color-slider-border-color-rgba)
+		)
+	);
+
+	/* The floating inset box shadow (border) must be a separate element since <canvas> won't take it. */
```
jenndiaz, line 141: Since the canvas variant was removed do we need this comment? can the code be refactored now?

components/colorslider/stories/template.js
```diff
@@ -33,7 +33,7 @@ export const Template = ({
 					role="presentation"
 					style="background: linear-gradient(to ${vertical
 						? "bottom"
-						: "right"}, rgb(255, 0, 0) 0%, rgb(255, 255, 0) 17%, rgb(0, 255, 0) 33%, rgb(0, 255, 255) 50%, rgb(0, 0, 255) 67%, rgb(255, 0, 255) 83%, rgb(255, 0, 0) 100%);"
+						: "right"}, ${gradientStops});"
```
jenndiaz, line 36: Noticing that this change has resulted in the colorhandle no longer matching the color of the slider below it. 

Could we also add `--spectrum-picked-color` as an arg that could be set to match the gradient start color?

components/colorslider/metadata/colorslider.yml
```diff
@@ -81,40 +88,6 @@ examples:
         <input type="range" class="spectrum-ColorSlider-slider" min="0" max="100" step="1">
```
jenndiaz, line 88: adding the comment here since the related lines weren't change so it's not in the diff, but should we replace the inline style physical profiteers with logical ones? 

specifically the `ColorSlider-handle` has `left: 50%;`

components/colorslider/index.css
```diff
@@ -38,59 +60,140 @@
...
-  left: 0;
-  top: 50%;
-
-  &:after {
-    border-radius: var(--spectrum-colorslider-handle-hitarea-border-radius);
-    width: var(--spectrum-colorslider-handle-hitarea-width);
-    height: var(--spectrum-colorslider-handle-hitarea-height);
-  }
+	inset-inline-start: 0;
+	inset-block-start: 50%;
```
jenndiaz, line 103: This has resulted in the colorHandles being on the right when viewed from RTL which is a change from what is was of the colorHandles being on the left regardless of text direction.

jenndiaz (changes requested):
Nice work 🎨

Had a couple of questions/comments

components/colorslider/index.css
majornista: Changing the orientation of the slider is the correct behavior for RTL.

majornista: See https://github.com/adobe/spectrum-web-components/pull/3302

components/colorslider/index.css
jawinn: I've reworded this comment. It does look like it being semi-transparent and inset on top of the background is part of the design, so it seems like it should stay as a pseudo element, regardless of canvas. I'm keeping it as a `box-shadow` for now, though technically it could be refactored in the future to just use `border` (tried it and it threw off some z-index ordering; I'm going to err on the side of backwards compatibility by leaving this as-is for the migration).

components/colorslider/stories/template.js
jawinn: I've added this to the stories, and have also added a new story for the "With Image" variant.

components/colorslider/metadata/colorslider.yml
jawinn: Good catch, this is updated. Yeah, it has to be the logical property or else it shows up in the wrong place in RTL on the docs examples.

components/colorslider/index.css
```diff
@@ -38,59 +60,145 @@
...
+		left top /
+		calc(
+			var(
+					--mod-color-slider-checkerboard-size,
+					var(--spectrum-color-slider-checkerboard-size)
+				) * 2
+		)
+		calc(
+			var(
+					--mod-spectrum-color-slider-checkerboard-size,
```
jenndiaz, line 128:
```suggestion
					--mod-color-slider-checkerboard-size,
```

components/colorslider/index.css
jawinn: @majornista Does this look good to you now?

jenndiaz (approved):
Looks good!

jenndiaz (commented):
not related to your changes, should we also remove the `Focused` state from the storybook for ColorSlider? 

Everything else looks good!

jawinn:
> not related to your changes, should we also remove the `Focused` state from the storybook for ColorSlider?

Good question. Looking at the focus for color handle and the behavior of color slider in SWC, I'm thinking that it'd be a good idea to add the "is-focused" to the color handle element when this is toggled in Storybook. Which would show the scaled up handle.

pfulton:
Beta version released: `@spectrum-css/colorslider@5.0.0-beta.0`

pfulton:
New beta published: `@spectrum-css/colorslider@5.0.0-beta.1`

pfulton:
New beta released: `@spectrum-css/colorslider@5.0.0-beta.2`

PR #1938, opened 2023-06-14
chore: moving all stories under Components and updating title case

mlogsdon18:
## Description

This PR updates the titles for each story in Storybook to fall under `Components/`. While I was in there, I also updated any instances where we were using title case when we should not have been, to make sure we're following our design system standard.

## To-do list


pfulton:
One suggestion: do you think it'd be worth it for us to update our [generator script](https://github.com/adobe/spectrum-css/blob/44bc18af9521df4a520703da1bc683dd5aae69b0/tools/generator/templates/stories/%7B%7B%20folderName%20%7D%7D.stories.js.hbs#L4) so that when we run it to scaffold new stories for components, it already has the "Components/Component Name" in there?

mlogsdon18:
> One question: do you think it'd be worth it for us to update our [generator script](https://github.com/adobe/spectrum-css/blob/44bc18af9521df4a520703da1bc683dd5aae69b0/tools/generator/templates/stories/%7B%7B%20folderName%20%7D%7D.stories.js.hbs#L4) so that when we run it to scaffold new stories for components, it already has the "Components/Component Name" in there?

I thought about that but I didn't want to overstep since that's Cassondra's work! If you all are good with it then I can update the script. I think it would be a good update to make.

castastrophe:
> > One question: do you think it'd be worth it for us to update our [generator script](https://github.com/adobe/spectrum-css/blob/44bc18af9521df4a520703da1bc683dd5aae69b0/tools/generator/templates/stories/%7B%7B%20folderName%20%7D%7D.stories.js.hbs#L4) so that when we run it to scaffold new stories for components, it already has the "Components/Component Name" in there?
> 
> I thought about that but I didn't want to overstep since that's Cassondra's work! If you all are good with it then I can update the script. I think it would be a good update to make.

Oh the generator is totally open for edits! Go for it! I don't own it, I just live here. 😆

Issue #1875, opened 2023-05-15
Tooltip size change by placement makes calculations awkward


Westbrook:
This code:

https://github.com/adobe/spectrum-css/issues/new?permalink=https%3A%2F%2Fgithub.com%2Fadobe%2Fspectrum-css%2Fblob%2F0a74e16df1e0103cc9e2db2a3388527c3cbc4040%2Fcomponents%2Ftooltip%2Findex.css%23L152-L153

In conjunction with this code:

https://github.com/adobe/spectrum-css/blob/0a74e16df1e0103cc9e2db2a3388527c3cbc4040/components/tooltip/index.css#L259-L260

Means that the _tip_ of a Tooltip takes different size metrics depending on which direction is it tipping in. That means that is the direction is tips is altered by JS there could be a but of a cycle trying to calculate the position of the tip. This can be mitigated by changing the tip to always be square and then altering its clip path to suit. Something like:

```
.spectrum-Tooltip-tip {
    clip-path: polygon(0 -5%, 50% 50%, 100% -5%);
    width: var(--spectrum-tooltip-tip-inline-size) !important;
    height: var(--spectrum-tooltip-tip-inline-size) !important;
}

.spectrum-Tooltip--top .spectrum-Tooltip-tip {
    top: 100%;
}

.spectrum-Tooltip--bottom .spectrum-Tooltip-tip {
    bottom: 100%;
    clip-path: polygon(50% 50%, 0 105%, 100% 105%);
    top: auto;
}

.spectrum-Tooltip--left .spectrum-Tooltip-tip,
.spectrum-Tooltip--right .spectrum-Tooltip-tip {
    top: 50%;
    transform: translateY(-50%);
}

.spectrum-Tooltip--right .spectrum-Tooltip-tip {
    clip-path: polygon(50% 50%, 105% 100%, 105% 0);
    left: calc(
        var(
                --mod-tooltip-tip-block-size,
                var(--spectrum-tooltip-tip-block-size)
            ) * -2
    );
    right: 100%;
}

.spectrum-Tooltip--left .spectrum-Tooltip-tip {
    clip-path: polygon(-5% 0, -5% 100%, 50% 50%);
    left: 100%;
}
```

Westbrook:
It's highly likely that this situation would play out similarly in the Popover patter, as well.

PR #1939, opened 2023-06-14
feat(tooltip): square tip elements and modified hairline gap fix

jawinn:
## Description

:hammer: CSS-525 / Fixes Issue #1875 :

- Sets the element for the triangle tip of the tooltip to be square per request (issue 1875 & CSS-525). Modifies all clip-paths accordingly. Adds two new custom properties to make it more clear about what is going on and to avoid repetition.

:sparkles: Bonus Fix:
- Fixes issue with some tooltips still showing a hairline gap in Chrome (see production examples Directions: Top End and Bottom Right). The existing fix was adding 5% or -5% to all of the clip paths. By changing this to a calc that adds 0.5px, this fixes the instances where it can still be seen on the examples.

- Modifies two unnecessary right and left property values and sets them to auto instead. They are already positioned by the oposite property in the adjacent lines of CSS.

Chrome + Firefox, Mac

## Screenshots

Tooltip tip element is now square for all positioning variants:
Screenshot 2023-06-14 at 3 05 54 PM

Bonus Fix: Additional modified fix for hairline gap -- this screenshot shows the existing problem showing in [the production docs](https://opensource.adobe.com/spectrum-css/tooltip.html) in Chrome, that this PR fixes (200% scale and highlighted in blue):
tooltip-gap-200-percent-scale

## To-do list


components/tooltip/index.css
```diff
@@ -55,6 +55,10 @@ governing permissions and limitations under the License.
   /* tip */
   --spectrum-tooltip-tip-inline-size: var(--spectrum-tooltip-tip-width);
   --spectrum-tooltip-tip-block-size: var(--spectrum-tooltip-tip-height);
+  /* Percentage value of tip width divided by tip height, for calculating clip-path within a square tip. */
+  --spectrum-tooltip-tip-height-percentage: 50%;
```
jawinn, line 59: Note: I wish I could calculate this with a CSS calc() from `--spectrum-tooltip-tip-width` and `--spectrum-tooltip-tip-height`, but it's not possible to divide with both tokens having px units.

components/tooltip/index.css
```diff
@@ -55,6 +55,10 @@ governing permissions and limitations under the License.
...
+  /* Tip inset from container edges for clip-paths calculation, to fix hairline gap in Chrome cause by antialiasing. */
+  --spectrum-tooltip-tip-antialiasing-inset: 0.5px;
```
jawinn, line 61: To match the previous code, this value would be `5%`.

jawinn:
@Westbrook Is this accomplishing what you were intending?

My only concern is that the tooltip tip element square now extends beyond the bottom margin a little bit, which could _possibly_ interfere with the hover where this was implemented, if the `--mod-tooltip-animation-distance` was reduced. Though I tested that here and did not see a problem. Just something to test and be aware of and to test. It may not be an issue.

components/tooltip/index.css
```diff
@@ -149,7 +153,7 @@ governing permissions and limitations under the License.
 .spectrum-Tooltip-tip {
   position: absolute;
 
-  height: var(--mod-tooltip-tip-block-size, var(--spectrum-tooltip-tip-block-size));
+  height: var(--mod-tooltip-tip-inline-size, var(--spectrum-tooltip-tip-inline-size));
```
mlogsdon18, line 156: Should this be `block-size` rather than `height`?

mlogsdon18 (commented):
Just one comment about logical properties but looks great to me! I'll let Westbrook approve whether or not it solves his issue fully

components/tooltip/index.css
jawinn: The latest commit updates the usages of width and height in there with their logical properties.

jawinn:
Does this need to support the use case of the tip height being taller than its width? If so, it would complicate things a bit more; there would need to be another percentage custom property and then adding additional calcs() for the two points in all the polygon() functions that define the flat side of the tip. Along with doing something like `--spectrum-tooltip-tip-square-size: max(var(--spectrum-tooltip-tip-inline-size), var(--spectrum-tooltip-tip-block-size));`

components/tooltip/index.css
Westbrook: If this applies to both axises, should it be `var(--mod-tooltip-tip-size, var(--spectrum-tooltip-tip-size))`?

Westbrook (approved):
I'm not sure the benefit is all that high, but currently `var(--mod-tooltip-tip-antialiasing-inset, var(--spectrum-tooltip-tip-antialiasing-inset))` is used 12 times, I'm not sure there's a pattern for this, yet, but would it make sense to make a `--_tooltip-tip-anialiasing-inset: var(--mod-tooltip-tip-antialiasing-inset, var(--spectrum-tooltip-tip-antialiasing-inset))` and use it instead?

All my notes are non-blocking, so LGTM!

jawinn:
Please take a look at the latest commit. It hadn't been pushed when you added your suggestion.

I'm not sure about using that sort of pattern for a local variable. We haven't used them as such for the purpose of avoiding repeating a mod. Is there much benefit there?

Side thought: I wonder if the whole tip could be refactored such that it only needs one single clip-path property, and then uses `rotate` on the element, though it starts to get a little unintuitive with the positioning and `translate`. I didn't want to get into any significant changes like this here.

PR #1940, opened 2023-06-15
feat(pickerbutton)!: migrate tokens

mlogsdon18:
## Description

This PR updates the Picker Button component to use the new tokens. It is based on the [In-field Button PR](https://github.com/adobe/spectrum-css/pull/1912) so that one will need to go in first, as it uses a lot of the same tokens. 

## To-do list


components/pickerbutton/index.css
```diff
@@ -10,57 +10,431 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+	--spectrum-picker-button-font-weight: var(
+		--spectrum-body-sans-serif-font-weight
+	);
+	--spectrum-picker-button-font-size: var(--spectrum-font-size-100);
+
+	--spectrum-picker-button-background-animation-duration: var(
+		--spectrum-animation-duration-100
+	);
+
+	&:disabled {
```
jenndiaz, line 63: I am seeing a hover state on the disabled button when viewing the express theme

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -126,18 +99,18 @@ export default {
 	},
 	args: {
 		rootClass: "spectrum-PickerButton",
-		label: "Select a Country",
+		label: undefined,
 		size: "m",
 		isOpen: false,
 		isRounded: false,
-		isValid: false,
 		isInvalid: false,
```
jenndiaz, line 106: do we want to also delete the isInvalid from the args?

components/pickerbutton/stories/template.js
```diff
@@ -31,25 +28,29 @@ export const Template = ({
...
 				[`${rootClass}--uiicononly`]: !label && iconType === "ui",
-				[`${rootClass}--texticon`]: label && iconType !== "ui",
 				[`${rootClass}--icononly`]: !label && iconType !== "ui",
 				[`${rootClass}--${position}`]: typeof position !== "undefined",
 				[`${rootClass}--rounded`]: isRounded,
 				[`${rootClass}--size${size?.toUpperCase()}`]:
 					typeof size !== "undefined",
 				"is-disabled": isDisabled,
 				"is-focused": isFocused,
-				"is-keyboardFocused": isKeyboardFocused,
```
jenndiaz, line 49: saw that `isKeyboardFocused` is still in the controls but we removed the styling here.

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -104,15 +86,6 @@ export default {
...
-		isKeyboardFocused: {
-			name: "Keyboard Focused",
-			type: { name: "boolean" },
-			table: {
-				type: { summary: "boolean" },
-				category: "State",
-			},
-			control: "boolean",
-		},
 		position: {
```
jenndiaz, line 89: Is there an expected visual change with position? I know you didn't add this but it does not seem to have an effect.

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -153,3 +126,23 @@ export default {
...
+
+export const WithLabel = Template.bind({});
+WithLabel.args = {
+	label: "Select",
+};
+
+export const Disabled = Template.bind({});
+Disabled.args = {
+	isDisabled: true
+};
```
jenndiaz, line 138: Noticing we can toggle focused and open when it is disabled is true, do we want to not allow this?

jenndiaz (changes requested):
Looking really good, just a couple of questions/comments

components/pickerbutton/stories/template.js
mlogsdon18: Yeah the Figma file doesn't have any distinction between focused and keyboard focused like some other ones have so I left the control in so we could still test it but the styling should be the same as focused

components/pickerbutton/stories/pickerbutton.stories.js
mlogsdon18: It should have an affect on the Spectrum version. One side of the button does not have a border radius and that side should switch sides based on the position. Do you not see that?

Screenshot 2023-06-23 at 10 14 51 AM
Screenshot 2023-06-23 at 10 14 56 AM

jenndiaz: ahh ok yes I do see I think I might of been viewing it as express so was not seeing a difference.

pfulton:
Beta released: `@spectrum-css/pickerbutton@4.0.0-beta.0`

pfulton:
@mlogsdon18 I released the beta for this, but let's be sure that we remove the `pickerbutton-generated.css` file before this gets merged in—I don't think we're using it anymore, right?

pfulton:
New beta released: `@spectrum-css/pickerbutton@4.0.0-beta.1`

PR #1943, opened 2023-06-16
feat(clearbutton)!: migrate tokens

mlogsdon18:
## Description

This PR migrates the tokens for the Clear Button based on the [design file for In-field Button](https://www.figma.com/file/nchxla8gD4CxWYwt9eiRrz/Token-definition-library?type=design&node-id=814-8689&t=vlV21ud28D5m02I9-0)

## To-do list


components/clearbutton/index.css
```diff
@@ -10,99 +10,101 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-@import "../vars/css/scales/spectrum-large.css";
-@import "../vars/css/scales/spectrum-medium.css";
-@import "../commons/basebutton.css";
 @import "../vars/css/components/spectrum-clearbutton.css";
```
jenndiaz, line 13: Are we able to delete this import?

components/clearbutton/index.css
```diff
@@ -10,99 +10,101 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    --mod-clear-button-background-color-key-focus: var(--spectrum-clear-button-background-color-hover-over-background, var(--spectrum-transparent-white-300));
   }
 }
 
-.spectrum-ClearButton-fill {
-  background-color: var(--spectrum-clearbutton-fill-background-color);
-
-  inline-size: var(--spectrum-clearbutton-fill-size);
-  block-size: var(--spectrum-clearbutton-fill-size);
+.spectrum-ClearButton {
```
jenndiaz, line 67: we lost the `cursor: pointer;` with removing `%spectrum-BaseButton` so there is no longer a cursor on hover. I think it would make sense to add back the `cursor: pointer;` here.

components/clearbutton/stories/template.js
```diff
@@ -25,6 +27,7 @@ export const Template = ({
 				"is-disabled": isDisabled,
 				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
 			})}
+			style=${ifDefined(styleMap(styles))}
 			?disabled=${isDisabled}
 		>
 			<div class="${rootClass}-fill">
```
jenndiaz, line 33: I think it would be just below this line, we are hard coding `iconName: "Cross100",` within the icon which is resulting in the icon not changing size when you toggle the t-shirt size in the story controls.

components/clearbutton/metadata/clearbutton.yml
```diff
@@ -79,6 +79,16 @@ examples:
           </svg>
         </div>
       </button>
+  - id: clearbutton-quiet
+    name: Quiet
+    markup: |
+      <button class="spectrum-ClearButton spectrum-ClearButton--sizeM spectrum-ClearButton--quiet">
```
jenndiaz, line 85: Could we add the aria to prevent these buttons from being "empty", same as you did in `in-field button`. For each of the variants here.

components/clearbutton/stories/clearbutton.stories.js
```diff
@@ -55,3 +46,13 @@ export default {
...
 Default.args = {};
+
+export const OverBackground = Template.bind({});
+OverBackground.args = {
+	variant: "overBackground",
+	styles: {
+		"background-color": "rgb(15, 121, 125);", 
+		"color": "rgb(15, 121, 125);",
+		"padding": "15px 20px;"
+	}
```
jenndiaz, line 57: nice! this makes a lot of sense pulled out this way as opposed to how it was before

jenndiaz (changes requested):
Great work with each of these, 🙌 just couple of comments on this one!

components/clearbutton/index.css
mlogsdon18: Good call out!

pfulton:
Beta released: `@spectrum-css/clearbutton@2.0.0-beta.0`

castastrophe (approved):
Before we merge let's align the dependencies to current versions and we should be all set.

mlogsdon18:
Approval on the SWC side! https://github.com/adobe/spectrum-web-components/pull/3476#issuecomment-1677198034

jenndiaz (approved):
Looks great!

PR #1951, opened 2023-06-19
chore: optimize the storybook build with nx

castastrophe:
## Description

Updating storybook commands to optimize build (remove duplicate component build by leveraging nx for cache completion).

Alias nx @spectrum-css/preview commands to `storybook` for simplicity in command running; this change only affects commands running with `nx ...`.

Moved nx settings to it's own project.json file for readability. Leverage nx executors for more optimized caching in storybook package.

Add workflow icons package as a root-level dependency (needed for the site build).

Add Patterns & Utilities categorizations for storybook elements.

- [x] Generator updated

---

Update 2023-06-20

Due to the component build using lerna and the storybook build running through nx, there were some caching inconsistencies causing intermittent problems locally. To resolve this we:

- added build, clean, dev, and watch tasks to every component package.json so each is managing their own tasks; this makes parallelization less problematic
- added nx and rimraf dependencies to every package to support new scripts
- Add nx tags ("components", "tokens", or "utilities") to each package so they can be run separately
- Updated story templates to import stylesheets using a bare package import; this lets the main field in the package.json drive which assets are loaded and stops storybook from watching uncompiled assets and throwing errors when those assets go missing during a build.
- Added an index.css file to the commons package do we could provide a valid main pointer in the package.
- Removed clean tasks from the gulp build tasks to prevent it re-removing assets after the clean task has been handled by nx.
- 

- [x] `yarn nx build storybook`: builds static storybook assets (this validates the CI update)
- [x] `yarn build:preview`: same as above ^
- [x] `yarn clean:preview`: removes storybook static assets
- [x] `yarn start`: spins up local storybook and watches for changes
- [x] `yarn storybook`: clean alias for `yarn start` without node env or dev run
- [x] `yarn test`: runs test suite as expected
- [x] `yarn test:scope @spectrum-css/accordion`: runs only accordion test as expected

## To-do list


castastrophe:
- [x] `yarn build:all` to verify no regressions

PR #1958, opened 2023-06-21
fix(colorloupe): border bug

jenndiaz:
## Description

While working on migrating ColorSlider a design bug was observed on ColorLoupe where the right edge of the component was being slightly cut off. 

Example of issue:
Screenshot 2023-06-22 at 1 50 32 PM

Fix: 
- the issue was occurring originally because `overflow: visible` was added via html on the example for ColorLoupe but not on any of the components while also display colorloupe
- I refactored the SVG to no longer need the `overflow: visible` 

Bonus: 
- noticed the design regressed some and was no longer displaying the inner border, corrected the `xlink:href` for the mask which fixed this 

- **How this was tested:** 

## Screenshots

Screenshot 2023-06-22 at 2 01 13 PM

## To-do list


pfulton (approved):
This looks good. Thanks for catching it!

castastrophe (approved):

pfulton:
@jenndiaz (cc: @castastrophe) One _tiny_ request for this: would you mind updating the Story file to [adjust the diff threshold](https://www.chromatic.com/docs/threshold#setting-the-threshold), and then committing and pushing that change?

**Why?**
The first time that I ran Chromatic against this change, it _did not_ detect a diff, which was inaccurate. I pulled the branch down and adjusted the diff threshold locally, and re-ran VRT locally, and the change was detected. It seems like this is just one of those things that needs a special setting.

Setting it to `0.2` should do the trick.

castastrophe:
@pfulton Oh that's a great catch.

PR #1960, opened 2023-06-23
feat(stepper)!: stepper migrate tokens

mlogsdon18:
## Description

This utilizes the In-field button tokens and Figma file, along with the Textfield tokens, to migrate the Stepper component. This PR relies on the In-field button PR, so that will need to be merged first.

## To-do list


jenndiaz (changes requested):
Nice job, theres a lot of details in each of these. I had a couple of small questions. 

Do we also want to handle the button text in this component with an aria label on the button, like you did in in-field button ?

jenndiaz:
@mlogsdon18 It looks like something regressed with the keyboard focus indicator. 

It is not displaying on Storybook or when tabbing though the docs site.

components/stepper/stories/template.js
```diff
@@ -50,6 +67,7 @@ export const Template = ({
 		>
 			${Textfield({
 				...globals,
+				size,
 				placeholder: "1",
```
jawinn, line 71: Should we remove placeholder from this example? We don't have one in the docs and it makes it look like the field value is 1. Maybe set a default value of 1 or 0 in the field instead?

I also noticed that the input being rendered in Storybook isn't `type="number"` (allows letters to be inputted).

jawinn (changes requested):
I tried putting in a longer value and it can get obscured by going underneath the arrow buttons. The right padding probably needs an adjustment to add the width of the arrows. The prod version doesn't appear to do this currently.
Screenshot 2023-09-11 at 2 15 38 PM

components/stepper/stories/template.js
mlogsdon18: Good catches there! I removed placeholder and am passing the number type now to textfield.

components/stepper/themes/express.css
```diff
@@ -11,14 +11,23 @@ governing permissions and limitations under the License.
...
-  /* @todo remove this when stepper is migrated; this ensures stepper buttons inherit border color and width */
   .spectrum-Stepper {
-    --spectrum-stepper-border-width: var(--spectrum-border-width-200); /* express stepper has border */
+    --spectrum-stepper-border-width: var(--spectrum-border-width-200);
+
+    --spectrum-stepper-buttons-border-style: solid;
+    --spectrum-stepper-buttons-border-width: var(--spectrum-border-width-200);
+    --spectrum-stepper-buttons-background-color: var(--spectrum-gray-50);
+
+    --spectrum-stepper-button-border-width: 0;
```
pfulton, line 21: On this line, `-button-` is singular in the name, but in `themes/spectrum.css` it looks like it's plural `-buttons-` and I think `-buttons` is what we refer to in the `index.css`.

Can you take a look to make sure that the naming is correct?

pfulton (changes requested):
Just one thing I noticed here, @mlogsdon18

components/stepper/themes/express.css
mlogsdon18: You're right! It doesn't look like that variable is being used anymore (along with a few others) now that I'm using in-field button so I've removed them

pfulton:
Beta published: `@spectrum-css/stepper@5.0.0-beta.0`

pfulton (approved):

jenndiaz (approved):

jawinn (approved):

PR #1971, opened 2023-06-26
feat(dial)!: migrate to use `@adobe/spectrum-tokens`

pfulton:
## Description

Migrates the Dial component to use `@adobe/spectrum-tokens` instead of the legacy "vars" packages.

This component does not have an updated spec file from design, nor does it have new tokens defined for it. This is a "DIY migration".

- **How this was tested:** 

## To-do list


components/tokens/custom-spectrum/custom-large-vars.css
```diff
@@ -87,4 +87,12 @@ governing permissions and limitations under the License.
   --spectrum-treeview-item-indentation-large: 25px;
   --spectrum-treeview-item-indentation-extra-large: 30px;
   --spectrum-treeview-indicator-inset-block-start: 6px;
+
+  --spectrum-dial-border-radius: 20px;
+  --spectrum-dial-handle-position: 10px;
+  --spectrum-dial-handle-block-margin: 20px;
+  --spectrum-dial-handle-inline-margin: 20px;
+  --spectrum-dial-controls-margin: calc(20px / 2);
```
jenndiaz, line 95: the margin-inline-start `--spectrum-dial-controls-margin` is not correctly applying to the dial-control, I can see the token come through but it is not being applied. This is causing the dial to not be centered under the label. 

Screenshot 2023-09-22 at 9 08 35 AM

components/dial/stories/template.js
```diff
@@ -4,7 +4,7 @@ import { ifDefined } from "lit/directives/if-defined.js";
 import { when } from "lit/directives/when.js";
 
 import "../index.css";
-import "../skin.css";
+// import "../skin.css";
```
jenndiaz, line 7: Do we want to go ahead and delete this? 
```suggestion
```

jenndiaz (changes requested):
Couple of small questions and one issue with the custom token not being applied.

components/dial/index.css
```diff
@@ -77,86 +113,113 @@ governing permissions and limitations under the License.
...
+  inline-size: var(--mod-dial-width, var(--spectrum-dial-width));
+  block-size: var(--mod-dial-height, var(--spectrum-dial-height));
+  min-block-size: var(--mod-dial-controls-min-height, var(--spectrum-dial-controls-min-height));
+  margin-inline-start: var(--mod-dial-controls-margin, var(--spectrum-dial-controls-margin));
 
-  border-radius: var(--spectrum-dial-border-radius);
-  position: relative;
-  display: inline-block;
+  border-radius: var(--mod-dial-border-radius, var(--spectrum-dial-border-radius));
   margin: 0;
```
jenndiaz, line 122: oh I see what's going on with `--spectrum-dial-controls-margin` not being applied, the `margin: 0` is over writing it now. When previously it was pulled out, see below:
``` 
[dir="ltr"] .spectrum-Dial-controls {
    margin-left: var(--spectrum-dial-controls-margin);
}
``` 

Do we need the `margin: 0`? or could we switch to `margin-block: 0`?

components/dial/stories/template.js
jenndiaz: pushed a commit that removes this line

jenndiaz: I also addressed some weirdness with focus state not quite applying correctly. The component was just appearing white and I found we had a token  `--spectrum-dial-handle-marker-color-mouse-focus` that was not be used that made more sense to use for focused. Without design guidance this change made the most sense.

components/tokens/custom-spectrum/custom-large-vars.css
jenndiaz: more on this here: https://github.com/adobe/spectrum-css/pull/1971#discussion_r1334533780

components/dial/index.css
jenndiaz: I went with `margin-block: 0`

jawinn (changes requested):
This is looking pretty good.

My only questions are around focus and keyboard focus, which also apply to what is in prod. Right now there is no visible focus when tabbing to the dial. On the docs page, `is-keyboardFocused` is applied by the JS there, which does not trigger the `.is-focused` styles (the dark grey visible after click and drag). Rather than using `.is-focused`, should this be using `:focus` or `:focus-visible`? The docs currently are allowing focus on the handle with `tabindex='0'`, which allows those pseudo classes to be activated.

I'm also wondering about disabled showing a hover background color. It does it that way in prod, but it seems odd because it makes it look interactive. And it doesn't seem to match how other components' disabled states behave.

components/dial/index.css
```diff
@@ -11,30 +11,56 @@ governing permissions and limitations under the License.
...
 
   --spectrum-dial-handle-size: 100%;
   --spectrum-dial-min-height: 0;
   --spectrum-dial-controls-min-height: 0;
-  --spectrum-dial-controls-margin: calc(
-    var(--spectrum-global-dimension-size-200) / 2
-  );
+  
+  --spectrum-dial-min-max-tick-angles: 45deg;
+  --spectrum-dial-min-max-tick-angle: 45deg;
```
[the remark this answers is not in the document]
jenndiaz: I also only saw the plural used, I've removed it.

jenndiaz (approved):

jenndiaz:
> My only questions are around focus and keyboard focus, which also apply to what is in prod. Right now there is no visible focus when tabbing to the dial. On the docs page, `is-keyboardFocused` is applied by the JS there, which does not trigger the `.is-focused` styles (the dark grey visible after click and drag). Rather than using `.is-focused`, should this be using `:focus` or `:focus-visible`? The docs currently are allowing focus on the handle with `tabindex='0'`, which allows those pseudo classes to be activated.
> 
> I'm also wondering about disabled showing a hover background color. It does it that way in prod, but it seems odd because it makes it look interactive. And it doesn't seem to match how other components' disabled states behave.

@jawinn sorry I missed these comments previously. 

Disabled: 
I agree that the disabled dial should probably not have a hover. I pushed up a commit that removed the hover and active from disabled. I also adjusted the cursor to not display `grab` when the dial is disabled looks like that was intended but the css wasn't in the right spot. 

Didn't get to your focus comment but will look at it tomorrow.

jawinn:
The updates for disabled and focus look great!

One thing I am noticing is that tab focus disappears for one stop as it appears to go from the handle to the range input itself. I'm not sure if we should put a `tabindex="0"` on the input without some sort of keyboard control behavior/guidelines associated with it? This might also be a good case for `:focus-within`, to at least keep the focus style when the input is focused. To have a single stop in the tab order, we could use `:focus-within` and not have a tabindex on the handle at all, but I don't know if the browser support is enough to use it by default yet (can-i-use says 94.14% global support).

jawinn (approved):

PR #1982, opened 2023-06-28
fix(actionbutton): update action button color tokens

mlogsdon18:
## Description

This PR updates the Action Button component to use the new color alias tokens. The following changes were made:
- Selected (not emphasized) background color updated to new color alias tokens
- Selected (not emphasized) content color updated to `gray-50`. That is what is in the XD file and it was incorrectly using `spectrum-white`
- Added Quiet styling for Express that existed in the XD file but not in the CSS
- Moved color styling that applies to both Spectrum and Express into `index.css` and out of `spectrum.css` for organization purposes
- Added mods for variants such as `selected`
- Added additional Action Button stories to make sure all states and variants were accounted for
- Found a random issue with Checkbox where the selected Checkbox suddenly was not rendering due to some styles not be specific enough. I don't know why it's only compiling them in a different order in this PR, but I made a small update to ensure it overrides the default styles.

## To-do list


pfulton:
@mlogsdon18 Just one Chromatic catch here...which I think is actually correct. I'm not sure why it's catching this now, but the updated version matches what we have on the docs site.

mlogsdon18:
@pfulton I don't know why that's getting flagged now but I took a look and approved it!

pfulton (approved):

PR #1991, opened 2023-07-05
fix(picker): add side label variant

mlogsdon18:
## Description

This adds a side label variant and styling to Picker. 

- In order for this to work correctly, the Picker component must be wrapped in a `div` with `display: inline-block`. This allows for the picker to be placed next to the inline label. The styling cannot be added to the Picker itself as it throws off the positioning of the Menu. This is already being implemented in SWC. 
- Added a `spectrum-Picker--sideLabel` class to allow for easier targeting (and translating in SWC) for when additional styling needs to be applied, specifically when used with the quiet variant.
- `margin-block-start` is changed to match the label when side label is used with the Quiet variant 

### Before
Screenshot 2023-07-05 at 11 14 49 AM

### After
Screenshot 2023-07-05 at 11 16 47 AM

## To-do list

### DIFF
--- a/components/picker/metadata/picker.yml
+++ b/components/picker/metadata/picker.yml
@@ -73,7 +73,7 @@ examples:
               <use xlink:href="#spectrum-css-icon-Chevron100" />
             </svg>
           </button>
-          <div class="spectrum-Popover spectrum-Popover--bottom spectrum-Picker-popover is-open" style="width: 240px">
+          <div class="spectrum-Popover spectrum-Popover--bottom spectrum-Picker-popover is-open" style="width: 240px;">
             <ul class="spectrum-Menu" role="listbox">
               <li class="spectrum-Menu-item is-selected" role="option" aria-selected="true" tabindex="0">
                 <span class="spectrum-Menu-itemLabel">Donaudampfschifffahrtsgesellschaftskapitän</span>
@@ -110,22 +110,20 @@ examples:
         <div class="spectrum-Examples-item">
           <h4>Side Label</h4>
           <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left">Country</div>
-          <div style="display: inline-block">
-            <button class="spectrum-Picker spectrum-Picker--sizeM spectrum-Picker--sideLabel" aria-haspopup="listbox" style="width: 240px">
-              <span class="spectrum-Picker-label is-placeholder">Select a Country with a very long label, too long in fact</span>
-              <svg class="spectrum-Icon spectrum-UIIcon-ChevronDown100 spectrum-Picker-menuIcon" focusable="false" aria-hidden="true">
-                <use xlink:href="#spectrum-css-icon-Chevron100" />
-              </svg>
-            </button>
-          </div>
+          <button class="spectrum-Picker spectrum-Picker--sizeM spectrum-Picker--sideLabel" aria-haspopup="listbox" style="width: 240px">
+            <span class="spectrum-Picker-label is-placeholder">Select a Country with a very long label, too long in fact</span>
+            <svg class="spectrum-Icon spectrum-UIIcon-ChevronDown100 spectrum-Picker-menuIcon" focusable="false" aria-hidden="true">
+              <use xlink:href="#spectrum-css-icon-Chevron100" />
+            </svg>
+          </button>
         </div>
 
         <!-- Open with Thumbnails -->
         <div class="spectrum-Examples-item">
 
           <h4>Open with Thumbnails</h4>
-          <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left">Country</div>
-            <button class="spectrum-Picker spectrum-Picker--sizeM spectrum-Picker--sideLabel is-open" aria-haspopup="listbox" style="width: 240px">
+          <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM">Country</div>
+            <button class="spectrum-Picker spectrum-Picker--sizeM is-open" aria-haspopup="listbox" style="width: 240px">
               <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-Picker-icon" focusable="false" aria-hidden="true" aria-label="Image">
                 <use xlink:href="#spectrum-icon-18-Image" />
               </svg>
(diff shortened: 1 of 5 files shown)


components/picker/metadata/picker.yml
```diff
@@ -103,59 +106,76 @@ examples:
 
         <div class="dummy-spacing" style="block-size: 15rem;"></div>
 
+        <!-- Side Label -->
+        <div class="spectrum-Examples-item">
+          <h4>Side Label</h4>
+          <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Picker--sideLabel spectrum-FieldLabel--left">Country</div>
+          <div style="display: inline-block">
```
pfulton, line 113: You mentioned above that the Picker will need to be wrapped in a `div` that has `inline-flex` applied to it, but in these examples you're using `inline-block`. It looks like it works...but I was wondering if you had a typo?

pfulton (commented):
Just one question here @mlogsdon18!

pfulton:
@Westbrook, would you mind taking a quick look at this?

components/picker/metadata/picker.yml
mlogsdon18: Yep I did! It should be `inline-block` in the description. SWC uses `inline-flex` which is where I probably got confused, but they're adding some additional styling for the popover positioning. For it to look ok out of the box, it should be `inline-block`

Westbrook:
Should this variant (when `quiet`) make an appearance in the docs?

components/picker/metadata/picker.yml
Westbrook: "need" here feels like we missed something. If we intend to require the `spectrum-Picker--sideLabel` modifier, why can't this also add `display: inline-flex` as well?

Westbrook (commented):
We likely need to confirm the path to supporting this consuption in association with other patterns, e.g.: 
image

components/picker/metadata/picker.yml
mlogsdon18: "need" might be too strong of a word but `inline-flex` throws off the transform that is happening to get the Popover to display below the Picker since it no longer starts on a "new line" with flex. There's a way to do this without that div and only use the `spectrum-Picker--sideLabel` class but it would not be a pure CSS solution, I would have to throw in some javascript in order to get that menu to align properly with the Picker. Unless you all have an idea I didn't think of!

The `spectrum-Picker--sideLabel` class was added in order to have a way to identify when the side label layout was being used from a component level, specifically for the Quiet variant, since styling for the entire component needed to change based on it. 

Screenshot 2023-07-11 at 9 16 02 AM

mlogsdon18:
> We likely need to confirm the path to supporting this consuption in association with other patterns, e.g.: image

Field labels and pickers currently don't have any margin around them. Are you thinking we should provide documentation on how to add that if needed?

Westbrook:
> > We likely need to confirm the path to supporting this consuption in association with other patterns, e.g.: image
> 
> Field labels and pickers currently don't have any margin around them. Are you thinking we should provide documentation on how to add that if needed?

Margins are one thing, not sure if that's addressed by [Form](https://opensource.adobe.com/spectrum-css/form.html) or not. However, in this example, I've adversarially placed a new Field Label and a Text Field as siblings to the Field Label and Picker which cased the vertical alignment between the Field Label and Picker to return. If that's not something the CSS is supposed to support, then we need to document that, but it seems like something that it _should_, right?

mlogsdon18:
@Westbrook I'm not having that same issue... Could you send me the code you used?
Screenshot 2023-07-13 at 9 40 52 AM

Westbrook:
This is what I used:
```
<div>
        <label class=" spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left " style="">
			Country
		</label>
	
		<div style="display: inline-block">
	<button aria-haspopup="listbox" class=" spectrum-Picker spectrum-Picker--sizeM spectrum-Picker--quiet spectrum-Picker--sideLabel ">
			<span class="spectrum-Picker-label is-placeholder">Select a country</span><svg class="spectrum-Icon spectrum-UIIcon-ChevronDown100 spectrum-Icon--medium spectrum-Picker-menuIcon" focusable="false" aria-hidden="true" role="img" width="10" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><path d="M3 9.95a.875.875 0 01-.615-1.498L5.88 5 2.385 1.547A.875.875 0 013.615.302L7.74 4.377a.876.876 0 010 1.246L3.615 9.698A.872.872 0 013 9.95z"></path></svg>
		</button>
	
			</div>
		<label class=" spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left " style="">Country
		</label><div class="spectrum-Textfield spectrum-Textfield--sizeM" style=""><input type="text" autocomplete="" class=" spectrum-Textfield-input ">
		</div>
    </div>
```

mlogsdon18:
> This is what I used:
> 
> ```
> <div>
>         <label class=" spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left " style="">
> 			Country
> 		</label>
> 	
> 		<div style="display: inline-block">
> 	<button aria-haspopup="listbox" class=" spectrum-Picker spectrum-Picker--sizeM spectrum-Picker--quiet spectrum-Picker--sideLabel ">
> 			<span class="spectrum-Picker-label is-placeholder">Select a country</span><svg class="spectrum-Icon spectrum-UIIcon-ChevronDown100 spectrum-Icon--medium spectrum-Picker-menuIcon" focusable="false" aria-hidden="true" role="img" width="10" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><path d="M3 9.95a.875.875 0 01-.615-1.498L5.88 5 2.385 1.547A.875.875 0 013.615.302L7.74 4.377a.876.876 0 010 1.246L3.615 9.698A.872.872 0 013 9.95z"></path></svg>
> 		</button>
> 	
> 			</div>
> 		<label class=" spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-FieldLabel--left " style="">Country
> 		</label><div class="spectrum-Textfield spectrum-Textfield--sizeM" style=""><input type="text" autocomplete="" class=" spectrum-Textfield-input ">
> 		</div>
>     </div>
> ```

@Westbrook So the issue here is that your label for your Textfield is not inside the Textfield component. Because of that, the styling for the sideLabel variant for Textfield (which is already in place) is not being added. This is adding margin to the bottom which is throwing off that inline styling. You can see on [the docs page for Textfield](https://opensource.adobe.com/spectrum-css/textfield.html#textfieldwithsidelabel) that the label is inside a div with `spectrum-Textfield spectrum-Textfield--sideLabel` classes. If you set up your example this way then you'll get correct inline styles like the screenshot I sent.

mlogsdon18:
@pfulton @Westbrook Ok I have updated the approach a bit here so I'd appreciate it if you guys could give it another look. The Picker has been updated to include the `display: inline-flex; vertical-align: top` if the `.spectrum-Picker--sideLabel` class is used. This removes the need for a wrapping div and should fix the alignment issue with Westbrook's test case. Let me know what you think!

pfulton (approved):
I'm good with this one!

Pinging @Westbrook for the final say...

Westbrook (approved):

PR #1993, opened 2023-07-06
chore(popover): update popover stories

mlogsdon18:
## Description

- This updates the Popover stories to include an Action Button as the trigger source for the Popover to allow for a better visual. 
- Updates the Spectrum Website link for popover on the docs site from `popovers` to `popover`
- Adds an onclick event to activate the popover for the story
- Updates a few other components to pass through functions for content to account for an error that the Interaction add-on was causing (IllustratedMessage, Dialog, Card)

## To test

1. Run Storybook for this PR and navigate to the Popover story
2. Click the button to open the Popover
3. Change the position of the Popover and make sure that it updates accordingly (you will have to close and reopen the popover as the positioning is done with javascript on button click)

## To-do list


mlogsdon18:
Currently, adding the @storybook/addon-interactions causes an error when we are using other storybook templates as content for another story. For example, the Popover story has a Menu as the content and this cause an error. I did not have any luck finding a solution so any help would be greatly appreciated!

pfulton (changes requested):
Thanks for taking this on! I know it was a slightly larger chunk of work than we initially thought.

Regarding the position of the Popover after invoking it: I'm wondering if we should set some value other than `0` for `--spectrum-popover-cross-offset` and use that in conjunction with the calculation you use to position the Popover.

Without it, the tip overlaps the button
Screenshot 2023-07-27 at 4 38 09 PM

mlogsdon18:
@pfulton Alright I made some updates here and renamed the variables to [match the spec ](https://spectrum.adobe.com/page/popover/#Offset). The default for the offset is supposed to be 8px so I've updated that as well and the stories look better. Let me know what you think!

pfulton:
Beta release published: `@spectrum-css/popover@7.0.0-beta.0`

mlogsdon18:
@pfulton with the updates I made this had 0 VRT issues in SWC! So I think we should be good to get this merged!

PR #2007, opened 2023-07-11
feat(tokens)!: prefer lowercase, hyphenated font names in font stacks

pfulton:
BREAKING CHANGE: updates the font-family stacks to remove the provided value from @adobe/spectrum-tokens, and instead use the "older" naming syntax that Spectrum CSS had used previously.

[Slack discussion](https://adobedesign.slack.com/archives/C5N154FEY/p1688145744065719) for context.

- **How this was tested:** 

## To-do list


components/tokens/custom-spectrum/custom-vars.css
```diff
@@ -35,16 +35,13 @@ governing permissions and limitations under the License.
   --spectrum-animation-ease-out: cubic-bezier(0, 0, 0.40, 1);
   --spectrum-animation-ease-linear: cubic-bezier(0, 0, 1, 1);
 
-  --spectrum-sans-serif-font: var(--spectrum-sans-serif-font-family);
```
castastrophe, line 38: Do we want to add backwards compatibility here by remapping `--spectrum-sans-serif-font` to `--spectrum-sans-font-family-stack`? My thought was in case someone was using the other variable directly.

components/tokens/custom-spectrum/custom-vars.css
```diff
@@ -35,16 +35,13 @@ governing permissions and limitations under the License.
...
-  --spectrum-sans-font-family-stack: var(--spectrum-sans-serif-font), adobe-clean, 'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Ubuntu, 'Trebuchet MS', 'Lucida Grande', sans-serif;
+  --spectrum-sans-font-family-stack: adobe-clean, 'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Ubuntu, 'Trebuchet MS', 'Lucida Grande', sans-serif;
```
castastrophe, line 38: Should we leave the Adobe Clean in the mix? If we add it second, then at least the (probably internal) user would fallback to the local font instead of a different set entirely. Or is that too confusing for folks? It would only be used if the web font is not loaded.

castastrophe (approved):

components/tokens/custom-spectrum/custom-vars.css
pfulton: Yeah, this is a good suggestion. I'll update.

components/tokens/custom-spectrum/custom-vars.css
pfulton: Updated. Let me know what you think of this most recent commit.

castastrophe (approved):

castastrophe (approved):

PR #2044, opened 2023-07-21
fix(colorhandle): updates animation

jenndiaz:
## Description

Addresses [CSS-559](https://jira.corp.adobe.com/browse/CSS-559) and SWC issue [3443](https://github.com/adobe/spectrum-web-components/issues/3433) which reported issues with the animation to the focused state for the ColorHandle. 

Switching from targeting individual properties to `all` addressed the animation issue in safari

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

### Chrome (114.0.5735.198)
animation-chrome

### Firefox(115.0.2)
animation-firefox

### Safari(16.5.2)
animation-safari

## To-do list

### DIFF
--- a/components/colorhandle/index.css
+++ b/components/colorhandle/index.css
@@ -117,6 +117,8 @@ governing permissions and limitations under the License.
     border-radius: var(--mod-colorhandle-hitarea-border-radius, 100%);
   }
 
+  transition: all var(--mod-colorhandle-animation-duration, var(--spectrum-colorhandle-animation-duration)) var(--mod-colorhandle-animation-easing, var(--spectrum-colorhandle-animation-easing));
+
   &.is-focused,
   &.focus-ring {
       /* Bigger handle when focused */


castastrophe:
@jenndiaz Can you share more about the decision to replace the individual properties with `all`? How does this address the concern Westbrook raised about performance? Size & position transitions are high-cost as they cause repaint and I'm not sure how this might mitigate that. Could you do me 2 favors:

1. Would you please add an [interaction test](https://storybook.js.org/docs/react/writing-tests/interaction-testing#write-an-interaction-test) for this animation so we can prevent any possible regressions in this animation in future?
2. The interaction test above should automatically come with a profiling of the action. If that information is not there or not helpful, would you please leverage the [profiling tools in WebKit](https://webkit.org/web-inspector/timelines-tab/) to confirm the delta from the previous to this new method?

castastrophe:
@jenndiaz Oh and just for reference, I was reading through this article to confirm my understanding - https://codeburst.io/low-performance-css-transitions-and-some-alternatives-f84ff35bf07 - we could try replacing this with a scale transform instead which could be a more performant option.

castastrophe (changes requested):
I left a comment on the PR with a few suggestions and requests. I want to make sure we include a test for this as well so we don't risk accidentally regressing or breaking this in future.

jenndiaz:
> @jenndiaz Oh and just for reference, I was reading through this article to confirm my understanding - https://codeburst.io/low-performance-css-transitions-and-some-alternatives-f84ff35bf07 - we could try replacing this with a scale transform instead which could be a more performant option.

@castastrophe The [designs for the focus state of the ColorHandle](https://www.figma.com/file/nchxla8gD4CxWYwt9eiRrz/Token-definition-library?type=design&node-id=431-8226&mode=design&t=JLumZlQQdhd6I2oD-4) specify that the width and height increases but the border size(s) remaining the same. With the scale the border size increases. This was the main reason I did not go with scale. 
Here is the good example from the article with a border added: https://codepen.io/jenndiaz/pen/abQRpQY This displays the border increase which would be a design change in this case. 

`transition: all 130ms ease-in-out` addressed the safari issue. we could remove `all` since it is the default but I think it's helpful it is there in the case we change the styling of the ColorHandle in the future. 

I also put together a code pen displaying the animation using scale, using scale and transition, and as it is now transition with the width and height. The version with out the transition is not quite as smooth as the one with. 
https://codepen.io/jenndiaz/pen/QWJxZBj/3111a4c271dbaa21b585eebb016aa902?editors=1100 

As for the storybook testing, I can look closer at this, I know Megan is already working on adding the integration testing with #1993.

castastrophe:
@pfulton I would love to add a backlog story though to evaluate more performant ways to approach animations for CSS though in conjunction with designers.

PR #2052, opened 2023-07-26
feat(checkbox): add new color aliases and focus color fixes

jawinn:
### refactor(checkbox): add color alias tokens and fix some colors
    
- Adds new color alias tokens from CSS-472 and implements them. Focus token was previously commented out.
      * neutral-background-color-selected-default
      * neutral-background-color-selected-hover
      * neutral-background-color-selected-down
      * neutral-background-color-selected-key-focus
- Fix some focus and hover color related bugs in prod: 
  - non-emphasized, unchecked showing blue border on focus and hover instead of grey
  - emphasized, unchecked showing blue border on focus and hover
  - emphasized, indeterminate showing grey after click
- Changed psuedo `:focus-ring` from old spec to `:focus-visible` (a find replace for the generated .focus-ring class was already being done in the SWC version of the component, replacing it with :focus-visble). Checkboxes should now show their focus indicator when tabbed into in Storybook.

### fix(checkbox): high contrast mode updates and remove deprecated
    
Update to match previous WHCM behavior after styles update, and some needed system color updates noticed while looking at that code.
- Focus color update to match with modified styles.
- 'Background' system color is deprecated; replace with Canvas.
- Default text of label should be CanvasText, it is not on top of a background with ButtonFace.
- Focus indicator color custom property was repeated in the same style rule, and was using FieldText when not on top of a matching `Field` background pair. Replaced with CanvasText as this should be appearing on top of default Canvas.

There are some additional WHCM changes that should probably be made to align with the W3C spec on system color pair usage, which I will leave for a future update (guideline language may also need to change as it relates to that).

Some validation instructions:

WIP

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list


components/checkbox/index.css
```diff
@@ -352,7 +369,7 @@ governing permissions and limitations under the License.
     }
 
     /* Focus State */
-    &:focus-ring {
+    &:focus-visible {
```
jawinn, line 372: I'm wondering if we need this rule at all (`.spectrum-Checkbox--emphasized:focus-visible`). Is there a case where that could even get triggered? All the other focus rules are on the input. Clicking the parent label that wraps the input focuses the input.

mlogsdon18:
Just a few minor things that may or may not be issues:

- There was a TODO still in the index.css about using font-weight tokens. Can that be removed?
- In WHCM in prod, Emphasized Default does not change color on hover, but it does in your PR. I feel like it should change color like your PR is doing but wanted to point out that difference
- The emphasized checked and indeterminate states have a blue border on focus-visible and hover, not gray like you said, but again I think this is how it is supposed to be, just wanted to point out the difference

jawinn:
> Just a few minor things that may or may not be issues:
> 
>     * There was a TODO still in the index.css about using font-weight tokens. Can that be removed?

I'm not sure about that one. I don't see the weight or style set yet, so it may still have to be done? I don't have the context for why it was left as a TODO. Something to look into for a followup update perhaps (@pfulton).   

>     * In WHCM in prod, Emphasized Default does not change color on hover, but it does in your PR. I feel like it should change color like your PR is doing but wanted to point out that difference

An unintended fix :) Yeah, it does look like this is how it should behave. I see that the non-emphasized in prod does show the hover color in WHCM. 

>     * The emphasized checked and indeterminate states have a blue border on focus-visible and hover, not gray like you said, but again I think this is how it is supposed to be, just wanted to point out the difference

Yes, that's right. I updated the validation instructions for the two you hadn't checked yet; those had been worded wrong.

PR #2080, opened 2023-08-10
fix(meter): update layout tokens

jenndiaz:
## Description

1. Updates Meter to use new layout tokens:
`meter-minimum-width`
`meter-maximum-width`
`meter-default-width`
`meter-thickness-small`
`meter-thickness-large`
Large value from `component-top-to-text-200` to `component-top-to-text-75` 

2. Add `Meter` Stories 
3. Added controls to Progress Bar: value and label

Bit of context, for now `Meter` uses `ProgressBar` css. 

#### Test Outline 
 1. Open the [storybook](https://pr-2080--spectrum-css.netlify.app/preview/?path=/story/components-meter) for the Meter component:
- [x] Ensure Meter displays appropriately 
- [x] Toggle between fill colors 
- [x] toggle between meter sizing [@mlogsdon18]

2. Open the [storybook](https://pr-2080--spectrum-css.netlify.app/preview/?path=/story/components-progress-bar) for the Progress Bar component:
- [x] Ensure controls all work as they should [@mlogsdon18]

3. Open the [Docs site](https://pr-2080--spectrum-css.netlify.app/meter) for the Meter component. 
- [x] inspect for new tokens 

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

4. A migrated documentation page (such as [action group](https://pr-###--spectrum-css.netlify.app/actiongroup.html)), including:

- [ ] The page is accessible
- [ ] The page is responsive

### Before
Screenshot 2023-08-10 at 4 08 49 PM

### After
Screenshot 2023-08-10 at 4 09 18 PM

## To-do list


components/progressbar/index.css
```diff
@@ -89,12 +95,37 @@ governing permissions and limitations under the License.
...
 }
 
-@media (forced-colors: active) {
-  .spectrum-ProgressBar-track {
-    forced-color-adjust: none;
-    --highcontrast-progressbar-fill-color: ButtonText;
-    --highcontrast-progressbar-track-color: ButtonFace;
-    border: 1px solid ButtonText;
+.spectrum-Meter {
+  --spectrum-progressbar-size-default: var(--spectrum-meter-inline-size);
```
mlogsdon18, line 99: Since Meter is technically a separate variant, it would probably be a good idea for these to have separate mods. So updating these to something like `--spectrum-progressbar-size-default: var(--mod-meter-size-default, var(--spectrum-meter-inline-size));` where the name would match the progress bar naming but allow for separate modifying

components/progressbar/metadata/meter.yml
```diff
@@ -18,7 +18,7 @@ examples:
     markup: |
       <div class="spectrum-Examples" style="flex-direction: column;">
         <div class="spectrum-Examples-item">
-          <div class="spectrum-ProgressBar spectrum-Meter--sizeS is-positive" value="50" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" style="width: 200px;">
+          <div class="spectrum-ProgressBar spectrum-Meter spectrum-Meter--sizeS is-positive" value="50" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100">
```
mlogsdon18, line 21: So it looks like you've added the need for an additional class here so we should probably call that out in the Migration Guide

components/progressbar/stories/metertemplate.js
```diff
@@ -0,0 +1,55 @@
...
+		meterFill === "positive" ? "is-positive" :
+		meterFill === "notice" ? "is-notice" : "default";
+
+	return html`
+		<div
+			class=${classMap({
+				[rootClass]: true,
+				["spectrum-Meter"]: true,
+				[`spectrum-Meter--size${size?.toUpperCase()}`]: size,
+				[`${fillColor}`]: fillColor !== "default",
```
mlogsdon18, line 27: Rather than doing all of the conditionals above for this, could we just do something along the lines of 
```
[`is-${meterFill}`]: meterFill !== default
```

components/progressbar/stories/metertemplate.js
```diff
@@ -0,0 +1,55 @@
...
+			value="${value}%"
+			role="progressbar"
+			aria-valuenow="${value}%"
+			aria-valuemin="0"
+			aria-valuemax="100"
+		>
+			${FieldLabel({
+				...globals,
+				size: `${size}`,
+				label: `${label}`,
```
mlogsdon18, line 39:
```suggestion
				size,
				label,
```

components/progressbar/stories/metertemplate.js
```diff
@@ -0,0 +1,55 @@
...
+				alignment: "",
```
mlogsdon18, line 40: You can just remove this if you don't have anything to pass through

components/progressbar/stories/template.js
```diff
@@ -53,28 +54,28 @@ export const Template = ({
...
+				aria-valuenow="${value}%"
 				aria-valuemin="0"
 				aria-valuemax="100"
 			>
 				${FieldLabel({
 					...globals,
 					size: `${size}`,
-					label: "Loading",
+					label: `${label}`,
 					alignment: "",
```
mlogsdon18, line 67: I'm confused why this is passing through the variables as string interpolated variables, same comment as above that since it's named the same thing and the prop you can just pass the prop name through.

components/progressbar/stories/meter.stories.js
```diff
@@ -0,0 +1,102 @@
...
+		label: {
+			name: "Label",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: "text",
+		},
+		value: {
```
mlogsdon18, line 29: Love that this actually works in Storybook!

mlogsdon18 (changes requested):
Just a few comments/questions!

components/progressbar/index.css
jenndiaz: this makes a lot of sense!

components/progressbar/stories/meter.stories.js
```diff
@@ -0,0 +1,102 @@
+// Import the component markup template
+import { Template } from "./metertemplate";
+
+export default {
+	title: "Components/Meter",
+	description:
+		"The meter component is a visual representations of a quantity or an achievement. Their progress is determined by user actions, rather than system actions.",
+	component: "ProgressBar",
+	argTypes: {
+		labelPosition: {
```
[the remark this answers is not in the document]
jenndiaz: Side label is not an option for meter, got this removed

components/progressbar/stories/template.js
jenndiaz: I needed the string interpolation for the second FieldLabel label: `${value}%`, to get the `%` but not the in the other cases. Got it updated.

components/progressbar/stories/metertemplate.js
jenndiaz: Are you suggesting instead of  `const fillColor`? I think I do need that, it is adding the `is-` to the fill class.

jenndiaz: oh! Nice call, I like this a lot better.

mlogsdon18 (approved):

pfulton (approved):

PR #2099, opened 2023-08-16
chore: simplify gulp use by extracting dev & watch

castastrophe:

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list


mlogsdon18 (approved):
Looks great to me!

package.json
```diff
@@ -21,42 +21,40 @@
...
-    "build:components": "lerna run --scope \"${SCOPE:-@spectrum-css/*}\" --ignore \"@spectrum-css/{*-builder*,preview,generator}\" build",
-    "postbuild:components": "yarn docs:mod",
-    "build:preview": "nx build @spectrum-css/preview",
-    "build:site": "gulp buildDocs -LLL",
-    "cache:clean": "nx reset",
+    "build:all": "run-s build:preview build:site mv:preview",
+    "build:components": "nx run-many --all --exclude \"@spectrum-css/{*-builder*,preview,generator}\" --target build",
+    "postbuild:components": "node tasks/mod-extractor.js && prettier --cache --write components/*/metadata/*.md > /dev/null",
+    "build:preview": "nx run @spectrum-css/preview:build",
+    "build:site": "gulp build -LLL",
```
pfulton, line 28: Just to confirm: this command would not be run on its own, and instead, it'd be part of series where it'd run after a full component build, right?

I ask because running `yarn build:site` directly after a `yarn clean` results in an error, but maybe that was my ignorance in running it as its own command.

Running `yarn build:components` and then `yarn build:site` is successful.

package.json
```diff
@@ -21,42 +21,40 @@
...
-    "lint:components": "node ./tasks/packageLint.js",
-    "lint:styles": "stylelint",
-    "mv:preview": "rimraf dist/preview && mv tools/preview/storybook-static dist/preview",
+    "ci:storybook": "nx run @spectrum-css/preview:storybook",
+    "clean": "run-p clean:*",
+    "clean:cache": "nx reset",
+    "clean:components": "rimraf components/*/dist",
+    "clean:preview": "nx run @spectrum-css/preview:clean",
+    "clean:site": "rimraf dist",
+    "dev": "npm-run-all -s build:components -p dev:*",
```
pfulton, line 36: Running this command works well for serving the docs site, but I'm noticing that when I click over to view the Storybook instance, it looks like some of the styles are missing

Screenshot 2023-08-22 at 1 47 02 PM

package.json
```diff
@@ -21,42 +21,40 @@
...
-    "start": "NODE_ENV=development yarn workspace @spectrum-css/preview start",
-    "test": "nx test @spectrum-css/preview",
-    "test:scope": "nx test:scope @spectrum-css/preview",
-    "version:build": "lerna run --since origin/main build",
-    "watch": "gulp watch"
+    "release": "yarn build:components && lerna publish --no-private",
+    "release:beta": "yarn release:from-package --conventional-prerelease --preid beta --pre-dist-tag beta",
+    "release:from-package": "yarn build:components && lerna publish from-package --no-private",
+    "release:site": "yarn build:all && gh-pages -d dist/ -f -e .",
+    "start": "run-p storybook watch:components",
```
pfulton, line 51: Similar to my comment about running this command in isolation vs. running it as part of a series: if I run `yarn start` directly after `yarn clean`, Storybook attempts to start but fails due to missing `dist` assets. Maybe I shouldn't have been running it like that, but I wanted to be sure to mention it!

Running `yarn build:components` and then `yarn start` results in a successful Storybook launch and serve, but it looks like some styles assets may be missing:

Screenshot 2023-08-22 at 1 52 08 PM

package.json
```diff
@@ -21,42 +21,40 @@
...
-    "clean:preview": "nx clean @spectrum-css/preview",
-    "predev": "yarn build:components",
-    "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
-    "docs:mod": "node tasks/mod-extractor.js",
-    "postdocs:mod": "prettier --cache --write components/*/metadata/*.md > /dev/null",
-    "lint:components": "node ./tasks/packageLint.js",
-    "lint:styles": "stylelint",
-    "mv:preview": "rimraf dist/preview && mv tools/preview/storybook-static dist/preview",
+    "ci:storybook": "nx run @spectrum-css/preview:storybook",
+    "clean": "run-p clean:*",
```
pfulton, line 31: I had about 90% success with running this command, but there were a few occasions when it seemingly hung for a few moments and then failed. It might have gotten tripped up on cleaning the Preview stuff? I'm unsure, and it seemed a little fluke-y.

Screenshot 2023-08-22 at 1 29 37 PM

pfulton (changes requested):
A few things here, @castastrophe, but I think we're very close!

package.json
castastrophe: Yes, that's true for now. When build:site points to 11ty and it has a package of it's own, we can define build dependencies the same as the storybook site but for now, this would be run as part of the other steps - build:components, build:site, and optionally build:preview

package.json
castastrophe: Oh hmm, this looks related to the cache reset happening concurrent to the other commands, maybe if it's run first

package.json
castastrophe: Added an update for this. So the challenge here is that storybook was compiling assets for itself but we want to start using the assets as they are compiled by the build now. That means updating the pathing from relative ../index.css references to "@spectrum-css/package-name" with alias's to the local directory.

castastrophe: This should now be resolved

package.json
castastrophe: This should now be resolved 🥳

pfulton (approved):

PR #2119, opened 2023-08-22
refactor: remove is-focused and is-keyboardFocused

mlogsdon18:
## Description

Removes the `is-focused` and the `is-keyboardFocused` classes from the following components:

- Asset Card
- Floating Action Button
- In-field Button
- Logic Button
- Steplist

These components do not exist in SWC so there should not be any issues on that end. 

  1. Open the [docs site](url) for each component and ensure that the focus styling remained the same:
- [ ] Asset card
- [ ] Floating action button
- [ ] In-field button
- [ ] Logic button
- [ ] Steplist

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/assetcard/index.css
+++ b/components/assetcard/index.css
@@ -68,7 +68,6 @@ governing permissions and limitations under the License.
 
   outline: none;
 
-  &.is-focused,
   &:focus-visible {
     .spectrum-AssetCard-assetContainer {
       &:before {
@@ -329,7 +328,6 @@ governing permissions and limitations under the License.
     }
   }
 
-  &.is-focused,
   &.is-selected,
   &:focus,
   &:hover {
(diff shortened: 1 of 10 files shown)


pfulton (commented):
I'm on board for this (we need to do it!), and the changes here look correct. The only other thing I'm wondering is when/how we can also prevent these classes from being applied to elements in the docs site. On quick glance, it looks like the code that handles this in `enhancement.js` is kind of broad and applies those selectors to a lot of elements...

mlogsdon18:
@pfulton It is currently pretty broad in order to keep the functionality working for the components that need it. It's basically a duplicate of how the focus-ring javascript was running. We could try to add checks for specific components and only run the event handlers on those? Kind of modify the code we have for Textfield a bit and I think that could work. I'll have to look into that

mlogsdon18:
> Sounds good. We can do that as a separate ticket!

Ok I can write one up!

pfulton (approved):

PR #2130, opened 2023-08-29
feat(well)!: diy migration 

Rajdeepc:
## Description
The DIY migration for Well

Build system updated
implemented --mods
used custom tokens for padding, background color values

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/tokens/custom-spectrum/custom-light-vars.css
+++ b/components/tokens/custom-spectrum/custom-light-vars.css
@@ -35,4 +35,5 @@ governing permissions and limitations under the License.
   --spectrum-coach-indicator-ring-default-color: var(--spectrum-blue-800);
   --spectrum-coach-indicator-ring-dark-color: var(--spectrum-gray-900);
   --spectrum-coach-indicator-ring-light-color: var(--spectrum-gray-50);
+  --spectrum-well-border-color:var(--spectrum-black-rgb);
 }
--- a/components/well/index.css
+++ b/components/well/index.css
@@ -11,19 +11,18 @@ governing permissions and limitations under the License.
 */
 
 .spectrum-Well {
-  --spectrum-well-padding: var(--spectrum-global-dimension-size-200);
-  --spectrum-well-border-width: var(--spectrum-alias-border-size-thin);
-  --spectrum-well-margin-top: var(--spectrum-global-dimension-size-50);
-  --spectrum-well-min-width: var(--spectrum-global-dimension-size-3000);
+  --spectrum-well-border-width: var(--spectrum-border-width-100);
 }
 
 .spectrum-Well {
   text-align: start;
   display: block;
-  min-inline-size: var(--spectrum-well-min-width);
-  padding: var(--spectrum-well-padding);
-  margin-block-start: var(--spectrum-well-margin-top);
-  border-width: var(--spectrum-well-border-width);
+  min-inline-size: var(--mod-well-min-width, var(--spectrum-well-min-width));
+  padding: var(--mod-well-padding, var(--spectrum-well-padding));
+  margin-block-start: var(--mod-well-margin-top, var(--spectrum-well-margin-top));
+  border-width: var(--mod-well-border-width, var(--spectrum-well-border-width));
   border-style: solid;
-  border-radius: var(--spectrum-alias-border-radius-regular);
+  border-radius: var(--mod-well-border-radius, var(--spectrum-well-border-radius));
+  background-color: var(--mod-well-background-color, rgba(var(--spectrum-gray-800-rgb), 0.02));
+  border-color: var(--mod-well-border-color, var(--spectrum-well-border-color));
 }
--- a/components/tokens/custom-express/custom-dark-vars.css
+++ b/components/tokens/custom-express/custom-dark-vars.css
@@ -14,4 +14,5 @@ governing permissions and limitations under the License.
 .spectrum--express.spectrum--dark {
   /* Drop Zone background color rgb */
   --spectrum-drop-zone-background-color-rgb: var( --spectrum-indigo-900-rgb); /* var(--spectrum-accent-color-900);*/
+  --spectrum-well-border-color:rgba(var(--spectrum-white-rgb), 0.05);
 }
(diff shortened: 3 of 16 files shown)


components/well/index.css
```diff
@@ -11,19 +11,18 @@ governing permissions and limitations under the License.
...
-  margin-block-start: var(--spectrum-well-margin-top);
-  border-width: var(--spectrum-well-border-width);
+  padding: var(--mod-well-padding, var(--spectrum-well-padding));
+  margin-block-start: var(--mod-well-margin-top, var(--spectrum-well-margin-top));
+  border-width: var(--mod-well-border-width, var(--spectrum-well-border-width));
   border-style: solid;
-  border-radius: var(--spectrum-alias-border-radius-regular);
+  border-radius: var(--mod-well-border-radius, var(--spectrum-well-border-radius));
+  background-color: var(--mod-well-background-color, var(--spectrum-well-background-color));
+  border-color: var(--mod-well-border-color, var(--spectrum-well-border-color));
```
mlogsdon18, line 27: Rather than keeping these background colors custom, could we maybe utilize `--spectrum-gray-800` for the background color? For border, if we still need to use custom tokens for light/dark, they should no longer reference tokens in the vars folder. Could you add the custom tokens to the `components/tokens/custom-spectrum` folder in the correct color theme file? For example, adding `  --spectrum-well-border-color: rgba(255,255,255,0.05);` to  `components/tokens/custom-spectrum/custom-dark-vars.css`

components/well/index.css
```diff
@@ -11,19 +11,18 @@ governing permissions and limitations under the License.
...
-  --spectrum-well-border-width: var(--spectrum-alias-border-size-thin);
-  --spectrum-well-margin-top: var(--spectrum-global-dimension-size-50);
-  --spectrum-well-min-width: var(--spectrum-global-dimension-size-3000);
+  --spectrum-well-border-width: var(--spectrum-border-width-100);
 }
 
 .spectrum-Well {
   text-align: start;
   display: block;
   min-inline-size: var(--spectrum-well-min-width);
```
jenndiaz, line 20: I think we could add a mod for  the min inline size as well

components/well/index.css
```diff
@@ -11,19 +11,18 @@ governing permissions and limitations under the License.
...
-  padding: var(--spectrum-well-padding);
-  margin-block-start: var(--spectrum-well-margin-top);
-  border-width: var(--spectrum-well-border-width);
+  padding: var(--mod-well-padding, var(--spectrum-well-padding));
+  margin-block-start: var(--mod-well-margin-top, var(--spectrum-well-margin-top));
+  border-width: var(--mod-well-border-width, var(--spectrum-well-border-width));
   border-style: solid;
-  border-radius: var(--spectrum-alias-border-radius-regular);
+  border-radius: var(--mod-well-border-radius, var(--spectrum-well-border-radius));
+  background-color: var(--spectrum-gray-800);
```
jenndiaz, line 26: this is missing the opacity, 

```suggestion
  background-color: rgba(var(--spectrum-gray-800-rgb), 0.02);
```

components/tokens/custom-spectrum/custom-light-vars.css
```diff
@@ -31,4 +31,5 @@ governing permissions and limitations under the License.
   --spectrum-calendar-day-border-color-key-focus: var(--spectrum-blue-800);
 
   --spectrum-badge-label-icon-color-primary: var(--spectrum-white);
+  --spectrum-well-border-color:rgba(var(--spectrum-black-rgb), 0.05);
```
jenndiaz, line 34: Looks like we only need the opacity on the border color for express. you can handle this by adding these values to `components/tokens/custom-express/custom-light-vars.css` and `components/tokens/custom-express/custom-dark-vars.css` with the opacity and removing the opacity from where it is here and in `components/tokens/custom-spectrum/custom-dark-vars.css`

components/well/index.css
jenndiaz: could add a mod for the background-color too

jenndiaz (changes requested):
Nice work! I had a couple of comments. 

Then a nice to have would be adding typography classes to the text with in the well so it is visible when viewed in dark and darkest. 
`<p class="spectrum-Body">Well done is better than well said.</p>`

mlogsdon18 (commented):
This looks good to me! The only thing I'm seeing is that for Express with the Darkest color theme, the border is white, but the current docs site has the border as black. It's super hard to see so maybe black isn't the best but if we're keeping design the same then that should be updated to black

Rajdeepc:
> This looks good to me! The only thing I'm seeing is that for Express with the Darkest color theme, the border is white, but the current docs site has the border as black. It's super hard to see so maybe black isn't the best but if we're keeping design the same then that should be updated to black

Sorry cannot trace the difference on the border-color. FYI 
Screenshot 2023-09-01 at 2 17 55 PM

mlogsdon18:
@Rajdeepc oh interesting! Looks like the border is acting different on the docs site and on storybook. I see a white border for both dark and darkest on storybook but it looks fine on the docs site 
Screenshot 2023-09-01 at 1 54 54 PM

Rajdeepc:
> @Rajdeepc oh interesting! Looks like the border is acting different on the docs site and on storybook. I see a white border for both dark and darkest on storybook but it looks fine on the docs site Screenshot 2023-09-01 at 1 54 54 PM

Right! But there is something that needs to be confirmed from the design team. 
1. Are we keeping darkest theme?
2. If we are keeping darkest theme, there is a difference in border-color in docs site darkest and storybook darkest. Which one is the correct border-color?

mlogsdon18:
@Rajdeepc As far as I know darkest is staying, but I do think it's a good idea to ask the design team which border color is correct. And no, you shouldn't need to delete any variables in expressvars or vars folders. If the package.json for your component is referencing tokens and not vars, then those variables should not be affecting it.

Rajdeepc:
> @Rajdeepc As far as I know darkest is staying, but I do think it's a good idea to ask the design team which border color is correct. And no, you shouldn't need to delete any variables in expressvars or vars folders. If the package.json for your component is referencing tokens and not vars, then those variables should not be affecting it.

Thanks! As per discussions with Patrick and PJ, I have added rgba(255,255,255,0.5) to darkest. Please let me know if this looks good.

mlogsdon18 (approved):
This looks good to me!

Rajdeepc:
This is PR is closed accidentally. Reopening.

pfulton (approved):

PR #2132, opened 2023-08-30
chore: fix linter errors

mlogsdon18:
## Description

Fixes lintor errors for multiple components that have already been migrated. **This does not fix all linter errors.** There are components that have not been migrated yet that are not included in this PR. There are also a handful of components that has some remaining errors that require class name changes and therefore a change to the `spectrum-config` file in SWC. Those are not a part of this PR.

- [ ] Verify that there are no visual changes to any of the components

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/colorhandle/index.css
+++ b/components/colorhandle/index.css
@@ -57,56 +57,9 @@ governing permissions and limitations under the License.
     var(--mod-colorhandle-outer-border-width, var(--spectrum-colorhandle-outer-border-width))
     var(--mod-colorhandle-outer-border-color, var(--spectrum-colorhandle-outer-border-color));
 
-	/* Add a half-percent to fix diagonal line issue in Chrome on non-retina displays */
-	transition: inline-size
-			var(
-				--mod-colorhandle-animation-duration,
-				var(--spectrum-colorhandle-animation-duration)
-			)
-			var(
-				--mod-colorhandle-animation-easing,
-				var(--spectrum-colorhandle-animation-easing)
-			),
-		block-size
-			var(
-				--mod-colorhandle-animation-duration,
-				var(--spectrum-colorhandle-animation-duration)
-			)
-			var(
-				--mod-colorhandle-animation-easing,
-				var(--spectrum-colorhandle-animation-easing)
-			),
-		border-width
-			var(
-				--mod-colorhandle-animation-duration,
-				var(--spectrum-colorhandle-animation-duration)
-			)
-			var(
-				--mod-colorhandle-animation-easing,
-				var(--spectrum-colorhandle-animation-easing)
-			),
-		margin-inline
-			var(
-				--mod-colorhandle-animation-duration,
-				var(--spectrum-colorhandle-animation-duration)
-			)
-			var(
-				--mod-colorhandle-animation-easing,
-				var(--spectrum-colorhandle-animation-easing)
-			),
-		margin-block
-			var(
-				--mod-colorhandle-animation-duration,
-				var(--spectrum-colorhandle-animation-duration)
-			)
-			var(
-				--mod-colorhandle-animation-easing,
-				var(--spectrum-colorhandle-animation-easing)
-			);
-
   border-radius: 100%;
 
-  &:after {
+  &::after {
     content: '';
     inset-inline: calc(50% - calc(var(--mod-colorhandle-hitarea-size, var(--spectrum-colorhandle-hitarea-size)) / 2));
     inset-block: calc(50% - calc(var(--mod-colorhandle-hitarea-size, var(--spectrum-colorhandle-hitarea-size)) / 2));
--- a/components/dropzone/index.css
+++ b/components/dropzone/index.css
@@ -114,7 +114,7 @@ governing permissions and limitations under the License.
   }
 
   &.is-dragged {
-    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style--dragged, solid);
+    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style-dragged, solid);
     --mod-drop-zone-background-color: rgba(var(--spectrum-drop-zone-background-color), var(--mod-drop-zone-background-color-opacity, var(--spectrum-drop-zone-background-color-opacity)));
     --spectrum-drop-zone-border-color: var(--highcontrast-drop-zone-border-color-hover, var(--mod-drop-zone-border-color-hover, var(--spectrum-drop-zone-border-color-hover)));
 
@@ -143,7 +143,7 @@ governing permissions and limitations under the License.
 
 .spectrum-DropZone-content {
   display: var(--mod-drop-zone-content-display, none);
-  height: 100%;
+  block-size: 100%;
   align-items: center;
   justify-content: center;
   position: relative;
@@ -179,6 +179,6 @@ governing permissions and limitations under the License.
 
     /* Updated values for a nested illustrated message when in a high contrast state */
     --highcontrast-illustrated-message-illustration-color: var(--highcontrast-drop-zone-illustration-color);
-    --highcontrast-illustrated-message-illustration-color: var(--highcontrast-drop-zone-illustration-color-hover);
+    --highcontrast-illustrated-message-illustration-color-hover: var(--highcontrast-drop-zone-illustration-color-hover);
   }
 }
--- a/components/dropzone/index.css
+++ b/components/dropzone/index.css
@@ -174,11 +174,9 @@ governing permissions and limitations under the License.
   .spectrum-DropZone {
     --highcontrast-drop-zone-illustration-color: CanvasText;
 
-    --highcontrast-drop-zone-illustration-color-hover: Highlight;
     --highcontrast-drop-zone-border-color-hover: Highlight;
 
     /* Updated values for a nested illustrated message when in a high contrast state */
     --highcontrast-illustrated-message-illustration-color: var(--highcontrast-drop-zone-illustration-color);
-    --highcontrast-illustrated-message-illustration-color-hover: var(--highcontrast-drop-zone-illustration-color-hover);
   }
 }
--- a/components/dropzone/index.css
+++ b/components/dropzone/index.css
@@ -114,7 +114,8 @@ governing permissions and limitations under the License.
   }
 
   &.is-dragged {
-    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style-dragged, solid);
+    /* @deprecation --mod-drop-zone-border-style--dragged will be removed during the S2 migration; please update your code to --mod-drop-zone-border-style-dragged */
+    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style--dragged, var(--mod-drop-zone-border-style-dragged, solid));
     --mod-drop-zone-background-color: rgba(var(--spectrum-drop-zone-background-color), var(--mod-drop-zone-background-color-opacity, var(--spectrum-drop-zone-background-color-opacity)));
     --spectrum-drop-zone-border-color: var(--highcontrast-drop-zone-border-color-hover, var(--mod-drop-zone-border-color-hover, var(--spectrum-drop-zone-border-color-hover)));
 
--- a/components/fieldgroup/index.css
+++ b/components/fieldgroup/index.css
@@ -27,9 +27,7 @@ governing permissions and limitations under the License.
 /* field group */
 .spectrum-FieldGroup {
   display: flex;
-  align-items: top;
-  flex-wrap: wrap;
-  flex-direction: column;
+  flex-flow: column wrap;
 
   /* field group label top aligned */
   &--toplabel {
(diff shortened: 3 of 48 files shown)


jenndiaz (commented):
I made it through FieldGroup starting at the top of files changed. Did not find any issues!

components/slider/index.css
```diff
@@ -640,97 +641,72 @@ governing permissions and limitations under the License.
...
+  }
 
-    .spectrum-Slider-ramp {
-      /* Slider handle turns transparent in high contrast mode for ramp */
-      @media (forced-colors: active) {
-        & + .spectrum-Slider-handle {
-            fill: ButtonFace;
-            background-color: ButtonFace;
-          }
-      }
```
pfulton, line 695: Was this section intentionally removed?

components/slider/index.css
```diff
@@ -640,97 +641,72 @@ governing permissions and limitations under the License.
...
 
-
 @media (forced-colors: active) {
 
   .spectrum-Slider {
-    --spectrum-slider-track-color: GrayText;
-    --spectrum-slider-track-fill-color: GrayText;
-
-    --spectrum-slider-track-color-disabled: GrayText;
-    --spectrum-slider-track-fill-color-disabled: GrayText;
```
pfulton, line 728: Just calling these out to be sure they were intentionally removed.

components/textfield/index.css
```diff
@@ -97,8 +97,8 @@ governing permissions and limitations under the License.
 	--spectrum-textfield-icon-spacing-inline-end-override: 32px; /* TODO - placeholder - reevaluate after combobox migration */
 
 	/* TODO - icon spacing override for search - reevaluate after search is migrated */
-	--spectrum-Textfield-workflow-icon-width: 18px;
-	--spectrum-Textfield-workflow-icon-gap: 6px;
+	--spectrum-textfield-workflow-icon-width: 18px;
+	--spectrum-textfield-workflow-icon-gap: 6px;
```
pfulton, line 101: This is a good catch, but I wanted to call out that it's a change that we'll need to be aware of when I merge.

components/textfield/index.css
```diff
@@ -445,42 +444,42 @@ governing permissions and limitations under the License.
...
-			inline-size: 100%;
-			block-size: var(
-				--mod-textfield-focus-indicator-width,
-				var(--spectrum-textfield-focus-indicator-width)
+			background-color: var(
+				--highcontrast-textfield-focus-indicator-color,
+				var(
+					--mod-textfield-focus-indicator-color,
+					var(--spectrum-textfield-focus-indicator-color)
+				)
```
pfulton, line 480: Can I ask a favor for this section? Would you mind removing some of the whitespace and indenting that Prettier introduced? IMO it's a little _less pretty_ now...

components/toast/index.css
```diff
@@ -90,32 +90,26 @@ governing permissions and limitations under the License.
...
 
   &--positive {
     background-color: var(--highcontrast-toast-positive-background-color-default, var(--mod-toast-positive-background-color-default, var(--spectrum-toast-positive-background-color-default)));
     color: var(--highcontrast-toast-positive-background-color-default, var(--mod-toast-positive-background-color-default, var(--spectrum-toast-positive-background-color-default)));
 
-    .spectrum-Toast-closeButton {
-      &:focus-visible:not(:active) {
-        color: var(--highcontrast-toast-positive-background-color-default, var(--mod-toast-positive-background-color-default, var(--spectrum-toast-positive-background-color-default)));
-      }
+    .spectrum-Toast-closeButton &:focus-visible:not(:active) {
```
pfulton, line 111: I think this selector might be incorrect with the `&` in there.

components/tooltip/index.css
```diff
@@ -223,38 +223,28 @@ governing permissions and limitations under the License.
...
   /* logical property - tip is horizontal at end △ ▽ */
   .spectrum-Tooltip--top-end &,
   .spectrum-Tooltip--bottom-end & {
-    left: auto;
-    right: var(--mod-tooltip-pointer-corner-spacing, var(--spectrum-tooltip-pointer-corner-spacing));
-
-    [dir="rtl"] & {
-      left: var(--mod-tooltip-pointer-corner-spacing, var(--spectrum-tooltip-pointer-corner-spacing));
-      right: auto;
-    }
```
pfulton, line 257: Just checking to be sure that this was an intentional removal.

components/typography/index.css
```diff
@@ -353,7 +353,7 @@ governing permissions and limitations under the License.
 
 /* Windows high contrast mode */
 @media (forced-colors: active) {
-  .spectrum-body {
+  .spectrum-Body {
```
pfulton, line 356: Interesting! Great catch

pfulton (changes requested):
@mlogsdon18 thanks for taking this on! I called out several things that I want to be sure were intentional + a few other things that might have changed the CSS that I'd love for you to double-check before we merge.

components/colorhandle/index.css
```diff
@@ -57,56 +57,9 @@ governing permissions and limitations under the License.
     var(--mod-colorhandle-outer-border-width, var(--spectrum-colorhandle-outer-border-width))
     var(--mod-colorhandle-outer-border-color, var(--spectrum-colorhandle-outer-border-color));
 
-	/* Add a half-percent to fix diagonal line issue in Chrome on non-retina displays */
```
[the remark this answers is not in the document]
mlogsdon18: No it was removed on purpose. I chatted with Jenn and the transition property that was below this was overwriting this one so it wasn't even getting applied.

components/dropzone/index.css
```diff
@@ -179,6 +179,6 @@ governing permissions and limitations under the License.
 
     /* Updated values for a nested illustrated message when in a high contrast state */
     --highcontrast-illustrated-message-illustration-color: var(--highcontrast-drop-zone-illustration-color);
-    --highcontrast-illustrated-message-illustration-color: var(--highcontrast-drop-zone-illustration-color-hover);
+    --highcontrast-illustrated-message-illustration-color-hover: var(--highcontrast-drop-zone-illustration-color-hover);
```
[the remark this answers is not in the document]
mlogsdon18: Pushed up a change to actually remove this variable as it isn't used anywhere. But I double checked this against the current docs site and it matches the styling

components/fieldgroup/index.css
```diff
@@ -27,9 +27,7 @@ governing permissions and limitations under the License.
 /* field group */
 .spectrum-FieldGroup {
   display: flex;
-  align-items: top;
```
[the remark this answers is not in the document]
mlogsdon18: `align-items: top` is not valid CSS (it would need to be start or we would need to use `justify-content` here) so this property was not even applying, so I removed it. I can add it back in and change it to the proper value though if you'd like

components/rating/index.css
```diff
@@ -96,13 +94,11 @@ governing permissions and limitations under the License.
...
   /* When the entire component is hovered, show all solid icons */
   &:hover {
-    .spectrum-Rating-icon {
-      .spectrum-Rating-starActive {
-        display: block;
-      }
-      .spectrum-Rating-starInactive {
-        display: none;
-      }
+    .spectrum-Rating-starActive {
```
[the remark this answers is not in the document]
mlogsdon18: Yeah just to avoid the max depth error I removed it since there wasn't any styling specific to the `spectrum-Rating-icon` class and there isn't any instance of these being outside of it

components/slider/index.css
mlogsdon18: It wasn't removed but was rather just moved down to the bottom of the file with the other forced-colors

components/slider/index.css
mlogsdon18: They were intentionally removed because they were being overwritten anyway by the `highcontrast` variables below this

components/tag/index.css
```diff
@@ -304,10 +304,10 @@ governing permissions and limitations under the License.
...
+      inset-block-start: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
+      inset-block-end: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
+      inset-inline-start: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
+      inset-inline-end: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
```
jawinn, line 310: Could this be a single `inset` property instead? Looks like they're all set to the same value.

components/fieldgroup/index.css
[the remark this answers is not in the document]
pfulton: Ahhh right right. Now that you mention this, I've actually removed the same typo myself before!

jawinn (changes requested):

components/menu/index.css
```diff
@@ -506,7 +508,7 @@ governing permissions and limitations under the License.
   align-self: start;
 }
 
-.spectrum-menu-itemSelection {
+.spectrum-Menu-itemSelection {
```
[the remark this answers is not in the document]
jawinn: @pfulton AFAIK this shouldn't break anything. As you noticed, it doesn't appear to be used for anything currently. I checked in SWC as well, and it's converting this class to `menu-itemSelection` but then doesn't appear to be using that class. In a separate PR, let's remove it and double-check?

components/toast/index.css
mlogsdon18: Oh good catch!! I'll double check and see if I can get those fixed

components/tag/index.css
mlogsdon18: Yes but the last time I combined inset properties Westbrook had me split them back out because it was having some issues with their custom CSS, so that's why I left these as is

components/tooltip/index.css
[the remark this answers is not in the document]
mlogsdon18: Got it. I didn't mess with Popover for this reason so I should have known it would have affected Tooltip as well. I'll undo those changes

jawinn (approved):
Looking good to me. I looked over all changes on Friday and again today and didn't notice anything else.

components/textfield/index.css
castastrophe: What if we threw the misspelled variable into the varstack as a fallback with a deprecation note?

components/textfield/index.css
```diff
@@ -942,16 +945,16 @@ governing permissions and limitations under the License.
...
-			&::placeholder {
-				color: var(
-					--highcontrast-textfield-text-color-focus-hover,
-					var(
-						--mod-textfield-text-color-focus-hover,
-						var(--spectrum-textfield-text-color-focus-hover)
-					)
-				);
-			}
+		&:hover::placeholder {
```
castastrophe, line 950: Just curious - this was b/c it was nested too deep right?

components/typography/index.css
castastrophe: We could do the same here, basically keep both selectors and add a `/** @deprecation .spectrum-body will be removed during the S2 migration; please update your code to .spectrum-Body */`

mlogsdon18: I don't think we need to do this. `.spectrum-Body` is correct and is what is used everywhere else in this CSS file. This was the only place with the typo

pfulton: Yeah, I think we're safe on this one without the fallback.

components/textfield/index.css
pfulton: @mlogsdon18 would you mind making this change? Having it as a fallback would mean that we wouldn't have to selectively do a breaking change for this component.

components/dropzone/index.css
```diff
@@ -114,7 +114,7 @@ governing permissions and limitations under the License.
   }
 
   &.is-dragged {
-    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style--dragged, solid);
+    --mod-drop-zone-border-style: var(--mod-drop-zone-border-style-dragged, solid);
```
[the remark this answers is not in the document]
pfulton: Similar to the note that @castastrophe left below for TextField, https://github.com/adobe/spectrum-css/pull/2132#discussion_r1329047344, would you mind adding the double-dashed var to the stack so that we don't have to do a breaking change for this component?

castastrophe (approved):

pfulton (approved):

PR #2145, opened 2023-09-05
fix(button): button text centering on firefox

jenndiaz:
## Description

Addresses reported inconsistencies with text positioning with in the button component. 
Looking closer at it, this was observed only on Firefox. 

image
In the screenshot Firefox is on the left and you can see how it is not positioned with in the line height the same as it is in chrome on the right despite having the same padding. 

Fix: 
- Use separate values for button top padding (top-to-text) for Firefox to account for the different positioning. 

[Jira CSS-576](https://jira.corp.adobe.com/browse/CSS-576)

1. Open the [Docs site](https://pr-2145--spectrum-css.netlify.app/button-primary#outline) for one of the button components on Firefox 
- [x] Text should be centered with in the button component 
- [x] Text positioning with in the button component should match that of chrome and safari [@mlogsdon18 ]

2. Open the [Docs site](https://pr-2145--spectrum-css.netlify.app/button-primary#outline) for one of the button components on Chrome or Safari  
- [ ] Text should be centered with in the button component 

### Regression testing

Validate: @jawinn 

- [x] The page is accessible
- [x] The page is responsive

5. A migrated documentation page (such as [action group](https://pr-###--spectrum-css.netlify.app/actiongroup.html)), including:

- [x] The page is accessible
- [x] The page is responsive

## To-do list

### DIFF
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -155,7 +155,6 @@ a.spectrum-Button {
   @inherit: %spectrum-ButtonLabel;
   padding-block-start: calc(var(--mod-button-top-to-text, var(--spectrum-button-top-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
   padding-block-end: calc(var(--mod-button-bottom-to-text, var(--spectrum-button-bottom-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
-  white-space: nowrap;
   line-height: var(--mod-button-line-height, var(--spectrum-button-line-height));
   align-self: start;
 }
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -88,7 +88,22 @@ governing permissions and limitations under the License.
   --spectrum-button-padding-label-to-icon: var(--spectrum-text-to-visual-300);
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-extra-large);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-extra-large);
+}
 
+/* addresses differences in lineheight in firefox */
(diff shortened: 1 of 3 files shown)


jawinn (commented):
I looked into this as well, and found the same thing; that the 1 pixel vertical difference is caused by the difference in browser rendering of the font itself (setting the line-height to a specific pixel value makes no difference). The solution you have appears to be the only one that I'm aware of! I don't like the idea of Firefox specific values, but this does get things rendering the same between both Chrome and Firefox.

Safari seems to be rendering the docs site oddly for this PR site; I wonder if getting up to date with `main` and kicking off the build again would get that back to normal. Checking off the last validation step is just waiting on confirming it in Safari.

EDIT: After some additional testing we found that the 1 pixel shift in Firefox couldn't be adjusted with CSS as the rendering differs between operating systems (Windows and Mac render the text baseline differently). [This may be the related Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=480196).

components/button/index.css
```diff
@@ -157,7 +156,6 @@ a.spectrum-Button {
   @inherit: %spectrum-ButtonLabel;
   padding-block-start: calc(var(--mod-button-top-to-text, var(--spectrum-button-top-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
   padding-block-end: calc(var(--mod-button-bottom-to-text, var(--spectrum-button-bottom-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
-  white-space: nowrap;
```
jenndiaz, line 160: Removing this to allow text to wrap on buttons with a fixed width

PR #2165, opened 2023-09-12
feat(overlay): diy migration

Rajdeepc:
## Description

Migrating overlay component to diy core tokens.
Adding --mod values to `commons/overlay-coretokens.css`

### Regression testing

Validate:

- [x] The page is accessible
- [x] The page is responsive

- [x] The page is accessible
- [x] The page is responsive

## To-do list

### DIFF
--- a/components/commons/overlay-coretokens.css
+++ b/components/commons/overlay-coretokens.css
@@ -14,6 +14,7 @@ governing permissions and limitations under the License.
   /* TODO: replace with core tokens */
   --spectrum-overlay-animation-distance: 6px;
   --spectrum-overlay-animation-duration: var(--spectrum-animation-duration-100);
+  --spectrum-overlay-animation-duration-opened: var(--spectrum-animation-duration-0);
 
   visibility: hidden;
 
@@ -29,26 +30,26 @@ governing permissions and limitations under the License.
 %spectrum-overlay--open {
   visibility: visible;
   opacity: 1;
-  transition-delay: 0ms;
+  transition-delay: var(--mod-overlay-animation-duration-opened, var(--spectrum-overlay-animation-duration-opened));
   pointer-events: auto;
 }
 
 %spectrum-overlay--bottom--open {
   --spectrum-overlay-animation-distance: 6px;
-  transform: translateY(var(--spectrum-overlay-animation-distance));
+  transform: translateY(var(--mod-overlay-animation-distance, var(--spectrum-overlay-animation-distance)));
 }
 
 %spectrum-overlay--top--open {
   --spectrum-overlay-animation-distance: 6px;
-  transform: translateY(calc(-1 * var(--spectrum-overlay-animation-distance)));
+  transform: translateY(calc(-1 * var(--mod-overlay-animation-distance, var(--spectrum-overlay-animation-distance))));
 }
 
 %spectrum-overlay--right--open {
   --spectrum-overlay-animation-distance: 6px;
-  transform: translateX(var(--spectrum-overlay-animation-distance));
+  transform: translateX(var(--mod-overlay-animation-distance, var(--spectrum-overlay-animation-distance)));
 }
 
 %spectrum-overlay--left--open {
   --spectrum-overlay-animation-distance: 6px;
-  transform: translateX(calc(-1 * var(--spectrum-overlay-animation-distance)));
+  transform: translateX(calc(-1 * var(--mod-overlay-animation-distance, var(--spectrum-overlay-animation-distance))));
 }
--- a/components/overlay/index.css
+++ b/components/overlay/index.css
@@ -10,15 +10,17 @@ governing permissions and limitations under the License.
 */
 
 %spectrum-overlay {
-  --spectrum-overlay-animation-distance: var(--spectrum-picker-m-texticon-popover-offset-y);
+  --spectrum-overlay-animation-distance: 6px;
+  --spectrum-overlay-animation-duration: var(--spectrum-animation-duration-100);
+  --spectrum-overlay-animation-duration-opened: var(--spectrum-animation-duration-0);
 
   visibility: hidden;
 
   opacity: 0;
 
-  transition: transform var(--spectrum-global-animation-duration-100) ease-in-out,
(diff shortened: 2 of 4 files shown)


components/commons/overlay-coretokens.css
```diff
@@ -29,26 +28,22 @@ governing permissions and limitations under the License.
 %spectrum-overlay--open {
   visibility: visible;
   opacity: 1;
-  transition-delay: 0ms;
+  transition-delay: var(--mod-animation-duration-0, var(--spectrum-animation-duration-0));
```
mlogsdon18, line 31: We'll want to assign this to a component-specific variable before modding it. Just like you did at the top of this file you'll want to do something like 
`--spectrum-overlay-animation-duration-0: var(--spectrum-animation-duration-0)` and then using that variable you create in this spot

components/tokens/custom-spectrum/custom-vars.css
```diff
@@ -50,4 +50,5 @@ governing permissions and limitations under the License.
   /* static white / black background color for docs containers */
   --spectrum-docs-static-white-background-color: rgb(15, 121, 125);
   --spectrum-docs-static-black-background-color: rgb(206, 247, 243);
+  --spectrum-animation-distance: 6px;
```
mlogsdon18, line 53: If this variable does not change based on scale or theme, you can just add this in the index.css file for overlay.

components/commons/overlay-coretokens.css
```diff
@@ -29,26 +28,22 @@ governing permissions and limitations under the License.
...
   pointer-events: auto;
 }
 
 %spectrum-overlay--bottom--open {
-  --spectrum-overlay-animation-distance: 6px;
-  transform: translateY(var(--spectrum-overlay-animation-distance));
+  transform: translateY(var(--mod-animation-distance, var(--spectrum-overlay-animation-distance)));
```
mlogsdon18, line 36:
```suggestion
  transform: translateY(var(--mod-overlay-animation-distance, var(--spectrum-overlay-animation-distance)));
```

mlogsdon18: We keep the component name in there for the mods

components/overlay/index.css
mlogsdon18, line 1: @pfulton is there a reason that this file doesn't just `@inherit: %spectrum-overlay;` since it's the exact same styles?

mlogsdon18 (changes requested):

components/overlay/index.css
Rajdeepc: @pfulton Let me know your thoughts here! Thanks

components/overlay/index.css
```diff
@@ -10,15 +10,17 @@ governing permissions and limitations under the License.
 */
 
 %spectrum-overlay {
-  --spectrum-overlay-animation-distance: var(--spectrum-picker-m-texticon-popover-offset-y);
+  --spectrum-animation-distance: 6px;
+  --spectrum-overlay-animation-distance: var(--spectrum-animation-distance);
```
jawinn, line 14:
```suggestion
  --spectrum-overlay-animation-distance: 6px;
```
Is there a reason to have `--spectrum-animation-distance` here which is named like a global token? Otherwise I think it can be removed.

components/commons/overlay-coretokens.css
```diff
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 
 %spectrum-overlay {
-  /* TODO: replace with core tokens */
-  --spectrum-overlay-animation-distance: 6px;
+  --spectrum-overlay-animation-distance: var(--spectrum-animation-distance);
```
jawinn, line 14:
```suggestion
  --spectrum-overlay-animation-distance: 6px;
```
Isn't `--spectrum-animation-distance` only defined in the Overlay index.css? I don't see it as a global token. I see other components like Tooltip only importing the _commons_ file.

components/overlay/index.css
```diff
@@ -28,7 +30,7 @@ governing permissions and limitations under the License.
 
   opacity: 1;
 
-  transition-delay: 0ms;
+  transition-delay: var(--spectrum-animation-duration-0);
```
jawinn, line 33: This ties into the same question that Megan had earlier about whether these should be using the same styles. If so, this file would also have the `--spectrum-overlay-animation-duration-opened` (`--spectrum-overlay-animation-duration-0`) defined?

jawinn (changes requested):

components/overlay/index.css
jawinn: Good question @mlogsdon18 , I'm wondering the same thing. And also if they are the same, do we even need them in both places (the commons and as a component)? I see some other components currently importing Overlay in two different ways:
`@import "../overlay/index.css";` or `@import "../commons/overlay-coretokens.css";`

Rajdeepc: Also `overlay` doesnt have the `mods.md` defined. Do we need to define this or can we use `overlay-coretokens.css`?

Rajdeepc: @jawinn I am keeping both the file for now. Let me know if we can move this PR with your requested changes. Probably we can collate these two files into one in a separate PR?

Rajdeepc:
@jawinn As discussed yesterday can you please check whether this PR is good to go or not! We can keep both the css files for now.

components/commons/overlay-coretokens.css
jawinn: Won't `--spectrum-animation-distance` still be undefined here?

components/overlay/index.css
jawinn, line 1: Doesn't this file also need the same _--mod_ properties (and mods generated with `yarn docs:mod`)?

jawinn:
The suggestion was to set both instances of the animation distance to `--spectrum-overlay-animation-distance: 6px;`, in index.css and overlay-coretokens.css. It looks like --spectrum-animation-distance got added back in, which doesn't need to be in custom-vars.

jawinn (approved):
Looks good! Approved.

Rajdeepc:
@pfulton Will you take a final look at this PR and let me know if this is good to merge or not! Thank you

pfulton (approved):

PR #2166, opened 2023-09-12
feat(fieldlabel): required asterisk vertical alignment

jawinn:
## Description

Update the required asterisk icon (which is set to `display: inline-block`) to be vertically aligned with the baseline of the text using the `vertical-align` property, instead of using new snapshot tokens. Removes some margin-block from the asterisk that was not needed.

Also adds stories for required and wrapping + required, and updates the docs.

Note: Per confirmation with design about the original Slack discussion, the asterisk should stay inline at the end of the text. It's not currently achievable in Figma due to container/frame layouts in the tool. Because of that, the asterisk placement in Figma will not match the asterisk placement in code.

CSS-582

### Validation Steps

1. Design validation:
  - [x] Designer has confirmed that this alignment method looks good @jawinn 
2. Testing site validation:
  - [ ] Asterisk is aligned middle in example docs and 2 Storybook stories
  - [ ] Spacing between text and asterisk is no longer larger than it should be in examples and Storybook (markup space removed, token value should be only inline margin)
  - [ ] No other changes noticed, other than the required asterisk
3. VRT
  - [ ] VRT results are as expected and any changes accepted. There will be two new stories.

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/fieldlabel/index.css
+++ b/components/fieldlabel/index.css
@@ -91,14 +91,12 @@ governing permissions and limitations under the License.
 
   -webkit-font-smoothing: subpixel-antialiased;
   -moz-osx-font-smoothing: auto;
-  font-smoothing: subpixel-antialiased;
 
   color: var(--spectrum-fieldlabel-color);
 }
 
 /* international lang support */
 .spectrum-FieldLabel {
-
   &:lang(ja),
   &:lang(zh),
   &:lang(ko) {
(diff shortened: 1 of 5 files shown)


components/fieldlabel/index.css
```diff
@@ -176,7 +167,7 @@ governing permissions and limitations under the License.
     display: flex;
     flex-direction: column;
 
-    &+.spectrum-Form-item {
+    & + .spectrum-Form-item {
       margin-block-start: var(--mod-field-label-top-to-asterisk, var(--spectrum-field-label-top-to-asterisk));
```
jawinn, line 171: Note: all usages of `--spectrum-field-label-top-to-asterisk` can probably be removed from this file, as this margin within the _Labels Above_ form appears to be using the wrong token. I'll make a cleanup update for the `.spectrum-Form` related styles in a followup PR, as that was starting to escape the bounds of this PR's issue.

components/fieldlabel/metadata/mods.md
```diff
@@ -1,15 +1,16 @@
...
-| `--mod-spacing-300`                   |
-| `--mod-tableform-border-spacing`      |
-| `--mod-tableform-margin`              |
+| Modifiable Custom Properties                |
+| ------------------------------------------- |
+| `--mod-disabled-content-color`              |
+| `--mod-field-label-asterisk-vertical-align` |
+| `--mod-field-label-text-to-asterisk`        |
+| `--mod-field-label-top-to-asterisk`         |
+| `--mod-fieldlabel-font-size`                |
```
pfulton, line 7: I hadn't noticed it before, but our naming is inconsistent here, and I'm on the fence as to whether we should leave it alone or update it so that things are consistent. I think we might just want to leave it alone.

pfulton (approved):

PR #2169, opened 2023-09-14
fix(picker): picker-quiet-layout

blunteshwar:
## Description

Fixed the quiet layout of sp-picker by removing the extra margin

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list


pfulton (commented):
@blunteshwar I know we talked about this a little during standup on Thursday, but would you mind adding a little more context for this change?

It would be helpful for me to see a few screenshots of the problem(s) you were experiencing in Spectrum Web Components, and then also a before and after screenshot of what's changing here in Spectrum CSS. It doesn't look like our visual regression tests were able to catch the change (which might be a problem on our side).

For reference, I looked at our Figma and Xd specs, and this is the guidance (`quiet` is the second variation in this first image):

Screenshot 2023-09-15 at 1 44 33 PM

Screenshot 2023-09-15 at 1 45 49 PM

blunteshwar:
After removing the `margin-block-start: calc(var( --mod-picker-spacing-label-to-picker-quiet, var(--spectrum-picker-spacing-label-to-picker-quiet) ) + 1px);` picker renders correctly
before
after

blunteshwar:
@pfulton can you look at this and provide some feedback

components/picker/index.css
```diff
@@ -469,7 +469,7 @@ governing permissions and limitations under the License.
   border: none;
   border-radius: 0;
   padding-inline: var(--mod-picker-spacing-edge-to-text-quiet, var(--spectrum-picker-spacing-edge-to-text-quiet));
-  margin-block-start: calc( var(--mod-picker-spacing-label-to-picker-quiet, var(--spectrum-picker-spacing-label-to-picker-quiet)) + (1px));
+
```
Rajdeepc, line 472: can you remove the whitespace

Rajdeepc (changes requested):

components/picker/index.css
blunteshwar: done!

jenndiaz (commented):
I think this needs more discussion. 
I am understanding the issue as the `margin-block-start` is causing the layout to look incorrect when the label and picker are displayed side by side instead of on top. The most recent design guidance we have does have the negative margin top, like Patrick points out in the screenshots. So this is not added accidentally and removing it doesn't look like the correct solution. 

I do see that the [design page](https://spectrum.adobe.com/page/picker/#Label-position) does say that the label position can be top or side but this did not get displayed on the [figma we have for this component](https://www.figma.com/file/nchxla8gD4CxWYwt9eiRrz/Token-definition-library?type=design&node-id=496-7958&mode=design&t=FN2KHUCZJ8v8Wmeg-0).  

Could we make use of the `--mod-picker-spacing-label-to-picker-quiet` when displayed side by side? 
We could certainly add some guidance to the docs about doing so when displaying side by side. 

Of note, we do have a side by side variant [displayed in storybook](https://opensource.adobe.com/spectrum-css/preview/?path=/story/components-picker--side-label&args=isQuiet:!true).

alternatively I think we'd need to look into adding a side by side variant to our docs site and reach out to design concerning which tokens should be used for this and have separate mark up for a side by side variant that would not apply this margin

jenndiaz:
actually we do have markup for a side label: https://github.com/adobe/spectrum-css/blob/a0ff7502807b1b6e464a330574fa909024a18bf6/components/picker/index.css#L476

Do we know if the person reporting this issue is using this class?

blunteshwar:
This issue is being resolved in SWC repository

PR #2188, opened 2023-09-27
refactor(pagination)!: migrate to spectrum tokens

jawinn:
## Description

Migrates the **Pagination** component to use Spectrum tokens.
This also:
- Removes the `.spectrum-Pagination-pageButton` class styles, as this is no longer used anywhere in the repository
- Fixes the following bug with RTL arrows, as seen in prod:
  Screenshot 2023-09-27 at 4 48 53 PM

CSS-505

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/pagination/index.css
+++ b/components/pagination/index.css
@@ -10,6 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+.spectrum-Pagination {
+  --spectrum-pagination-counter-margin-inline-start: var(--spectrum-pagination-item-inline-spacing);
+  --spectrum-pagination-page-button-inline-spacing: var(--spectrum-pagination-item-inline-spacing);
+  --spectrum-pagination-counter-color: var(--spectrum-neutral-subdued-content-color-default);
+  --spectrum-pagination-counter-font-size: var(--spectrum-font-size-100);
+  --spectrum-pagination-counter-line-height: var(-spectrum-line-height-100);
+}
+
 .spectrum-Pagination--explicit,
 .spectrum-Pagination--listing {
   display: flex;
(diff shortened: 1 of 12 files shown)


components/pagination/metadata/mods.md
```diff
@@ -0,0 +1,8 @@
+| Modifiable Custom Properties                   |
+| ---------------------------------------------- |
+| `--mod-pagination-counter-color`               |
+| `--mod-pagination-counter-margin-inline-start` |
+| `--mod-pagination-page-button-inline-spacing`  |
+| `--mod-pagination-textfield-width`             |
+| `--mod-textfield-min-width`                    |
+| `--mod-textfield-width`                        |
```
pfulton, line 8: We'll probably want to merge in your update to the mods generation script before this one goes in.

components/pagination/index.css
```diff
@@ -10,33 +10,44 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  margin-inline-start: var(--spectrum-pagination-page-button-margin-horizontal);
+  margin-inline-start: var(--mod-pagination-page-button-inline-spacing, var(--spectrum-pagination-page-button-inline-spacing));
 }
 
-.spectrum-Pagination-prevButton,
-.spectrum-Pagination-nextButton {
-  .spectrum-Icon {
-    transform: logical rotate(0deg);
-  }
+/* Correct the direction of the arrows when viewing right-to-left */
```
pfulton, line 46: Thanks for catching this!

pfulton (approved):
LGTM

jenndiaz (commented):
Your changes look good. Nice catch on the RTL icon fix. 

I had a couple of storybook related observations which could be considered out of scope here but in case we want to address it now. 
- the default and Explicit variants have t-shirt sizing but not the docs site, are we sure we want to have t-shirt sizing for this component? 
- the default t-shirt sizing is changing the button sizes but not the action button sizing 
- the explicit t-shirt sizing is changing the button sizing but not the text-field 
- Do we want to add the button style variant to storybook?

jawinn:
> Your changes look good. Nice catch on the RTL icon fix.
> 
> I had a couple of storybook related observations which could be considered out of scope here but in case we want to address it now.
> 
> * the default and Explicit variants have t-shirt sizing but not the docs site, are we sure we want to have t-shirt sizing for this component?
> * the default t-shirt sizing is changing the button sizes but not the action button sizing
> * the explicit t-shirt sizing is changing the button sizing but not the text-field
> * Do we want to add the button style variant to storybook?

I've disabled the t-shirt sizing control after the recommendation today. Though the sizing is now properly passed along to the action button etc; the component just needs t-shirt size specific font sizing and perhaps inline margin adjustments to support t-shirt sizing fully.

I also added the "**Button style**" pagination to Storybook. I do think it's worth showing because the layout is a bit different than other SplitButton examples and it does have two pagination classes included in it that could affect things with future changes. I had to update the SplitButton and Button Storybook templates to have a few more options in order to display it (these changes are in their own commit).

Another change here was to remove the `spectrum-Body--secondary` class which doesn't seem to exist; I applied a font-size and line-height token there (to the "counter").

components/pagination/stories/template.js
```diff
@@ -48,6 +47,16 @@ export const Template = ({
 				})}
 			</nav>
 		`;
+	} else if (variant == "button") {
+		return SplitButton({
+			position: "left",
+			variant: "accent",
```
jenndiaz, line 53: So we can toggle the button style between the 3 variants, do we want to also add variant to the controls for pagination?

jawinn: It already exists as a disabled control. I'm not sure it should be enabled since these pagination types are so different and not really variation classes? The button one is just a SplitButton but with two classes added to the buttons.

jenndiaz (approved):

PR #2207, opened 2023-10-09
feat(menu): tray submenus

jawinn:
## Description

**Menu - Tray Submenus**
Adds the "Tray submenus" feature to the **Menu** component. [Guidelines](https://spectrum.adobe.com/page/menu/#Tray-submenus):
> When a menu is displayed within a tray, a submenu will replace the tray content when the parent menu item is selected. A submenu displays a back button (labeled by the title of the parent item) at the top of the tray to return the user to the previous level of the menu.

This adds some additional CSS, a new example on the docs, and a new story in Storybook.
Some additional markup was adjusted in the docs to rename some duplicate IDs.

The markup for this is based on the existing "Standard with section headers and dividers" example. As the back item functions like a "section header", but with an additional button for navigating back. I've included the button as an additional `role="menuitem"`, along with ARIA label text describing the function of the icon ("Back to previous menu").

CSS-303

### Validation steps

- [x] Validated by design [to be confirmed by @jawinn]
- [x] Menu tray submenu examples appear on docs and Storybook
- [x] Menu tray submenu looks correct, including at different t-shirt sizes. Note: the icon is now larger than appears on the Guidelines. The latest Xd design shows the updated larger arrow icon scale and usage of `navigational-indicator-top-to-back-icon-*` tokens. The icon and header text not lining up with the items below is also expected, per Slack discussion in the verification channel on Oct 9th.
- [x] Back button shows focus indicator on tab focus but not on click

### Regression testing

Validate:

## Screenshots

Screenshot 2023-11-15 at 12 55 35 PM

Updated design reference:
Screenshot 2023-11-15 at 1 00 57 PM

## To-do list

### DIFF
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -794,15 +794,43 @@ governing permissions and limitations under the License.
 }
 
 /* Tray submenu */
-.spectrum-Menu-item--back {
+.spectrum-Menu-back {
   display: flex;
   flex-flow: row wrap;
   align-items: center;
   padding-inline-start: var(--mod-menu-back-padding-inline-start, 0);
   padding-block: var(--mod-menu-back-padding-block, 0);
+
+  .spectrum-Menu-sectionHeading {
+    padding: 0;
+  }
+}
+
+.spectrum-Menu-backButton {
+  padding: 0;
+  margin: 0;
+  background: transparent;
+  border: 0;
+  display: inline-flex;
+  cursor: pointer;
+
+  &:focus-visible {
+    outline: var(--spectrum-focus-indicator-thickness) solid var(--spectrum-focus-indicator-color);
+    outline-offset: calc(-1 * (var(--spectrum-focus-indicator-thickness) + 1px));
+  }
+}
+
+.spectrum-Menu-backHeading {
+  display: block;
+  color: var(--highcontrast-menu-item-color-default, var(--mod-menu-back-heading-color, var(--spectrum-menu-section-header-color)));
+  font-size: var(--mod-menu-section-header-font-size, var(--spectrum-menu-section-header-font-size));
+  font-weight: var(--mod-menu-section-header-font-weight, var(--spectrum-menu-section-header-font-weight));
+  line-height: var(--mod-menu-section-header-line-height, var(--spectrum-menu-section-header-line-height));
 }
 
 .spectrum-Menu-backIcon {
   margin-block: var(--mod-menu-back-icon-margin-block, var(--spectrum-menu-back-icon-margin));
   margin-inline: var(--mod-menu-back-icon-margin-inline, var(--spectrum-menu-back-icon-margin));
+  fill: var(--highcontrast-menu-item-color-default, var(--mod-menu-back-icon-color-default));
+  color: var(--highcontrast-menu-item-color-default, var(--mod-menu-back-icon-color-default));
 }
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -91,6 +91,8 @@ governing permissions and limitations under the License.
   --spectrum-menu-checkmark-display-shown: block;
   --spectrum-menu-checkmark-display: var(--spectrum-menu-checkmark-display-shown);
 
+  --spectrum-menu-back-icon-margin: var(--spectrum-navigational-indicator-top-to-back-icon-medium);
+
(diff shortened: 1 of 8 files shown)


components/menu/index.css
```diff
@@ -431,7 +439,8 @@ governing permissions and limitations under the License.
   }
 }
 
-.spectrum-Menu-item:focus-visible {
+.spectrum-Menu-item:focus-visible,
+.spectrum-Menu-back:focus-visible {
```
mdt2, line 443: Do we need to include `.is-keyboardFocused` here or are we moving forward with only using `:focus-visible` since browser support is good for that now?

components/menu/metadata/menu.yml
```diff
@@ -1135,38 +1135,38 @@ examples:
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Without icons</h4>
           <div class="spectrum-Examples-itemGroup">
-            <ul class="spectrum-Menu spectrum-Menu--sizeM is-selectableMultiple" role="menu">
+            <ul class="spectrum-Menu is-selectableMultiple" role="menu">
               <li class="spectrum-Menu-item is-selected" role="menuitem" tabindex="0">
                 <label class="spectrum-Checkbox spectrum-Checkbox--sizeM spectrum-Checkbox--emphasized spectrum-Menu-itemCheckbox">
-                  <input type="checkbox" class="spectrum-Checkbox-input" aria-labelledby="ms-label1" checked>
```
mdt2, line 1141: Curious about the removal of `aria-labelledby` and their associated `id`s.

components/menu/metadata/menu.yml
```diff
@@ -1338,3 +1338,48 @@ examples:
           </ul>
         </div>
       </div>
+  - id: tray-submenus
+    name: Tray submenus
+    description: "When a menu is displayed within a tray, a submenu will replace the tray content when the parent menu item is selected. A submenu displays a back button (labeled by the title of the parent item) at the top of the tray to return the user to the previous level of the menu.
```
mdt2, line 1343: Excellent description 👍

components/menu/metadata/menu.yml
```diff
@@ -1338,3 +1338,48 @@ examples:
...
+      The back arrow size scale used with the various menu sizes are small: `200`, medium: `300`, large: `400`, and extra large: `500`."
+    markup: |
+      <div class="spectrum-Tray-wrapper spectrum-CSSExample-dialog" style="background: rgba(0,0,0,0.4);">
+        <div class="spectrum-Modal spectrum-Tray is-open spectrum-Tray">
```
mdt2, line 1347: double class here for `spectrum-Tray`

components/menu/metadata/menu.yml
```diff
@@ -1338,3 +1338,48 @@ examples:
...
+          <ul class="spectrum-Menu is-selectableMultiple" role="menu" style="--mod-menu-inline-size: 100%;">
+            <li role="presentation">
```
mdt2, line 1349: Tell me about this, why do we use role of presentation here?

mdt2 (commented):
Looking good in Storybook and Docs site 🎉 Just a few questions/comments.

components/menu/index.css
jawinn: AFAIK `is-keyboardFocused` should only be needed where `:focus-visible` does not get us the desired effect. Mostly it's just needed on text input fields that use the keyboard, because they show `:focus-visible` on click per the spec.

components/menu/metadata/menu.yml
jawinn: The input and its label text are already wrapped inside the `<label>`, so they are not needed.

> _Alternatively, you can nest the `<input>` directly inside the `<label>`, in which case the for and id attributes are not needed because the association is implicit [[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)]_

components/menu/metadata/menu.yml
jawinn: This was the existing markup used for section headers and I think I understand now why it was done. It's because the component is using the `menu` role, and the `li` no longer belongs to that role. The `<li>` is just used to house the heading w/ its list of menu items and the dividers. And it doesn't really function as a list item (_but needs to be one of course because other elements can't be children of a_ `ul`).

mdt2 (approved):
LGTM!

PR #2209, opened 2023-10-11
refactor(radio): remove deprecated use of invalid

jawinn:
## Description

The invalid state of radios is now defined on the Field group (with `role="radiogroup"`):
> In updating documentation on properly showing invalid state in radio groups in SWC. We discussed removing the invalid attribute from the Radio component—favoring the Radio Group instead. In WAI-ARIA 1.2, the aria-invalid state is deprecated on role="radio"...

Some of the Spectrum CSS documentation was updated previously, but there were specific styles covering the invalid state in the Radio component's CSS. This `is-invalid` class has been removed in this PR. This does not to appear to have any visual effect, as the color was already a default. Per the guidelines and existing Field group example, "the error is indicated with negative Help text, along with an Icon."

The `aria-invalid` attribute mentioned in the writeup for this issue has also been added to the example markup for FieldGroup.

This PR also includes some docs and Storybook updates noticed while making updates:
- **FieldGroup**: existing stories now set `isInvalid` to true, as they were previously showing the neutral colors with the invalid icon, which didn't match with the guidelines.  
:framed_picture: _Expected VRT Change:_ Both stories will now show negative (red) text color

- **HelpText**: storybook templates had a mismatched arg name between the stories file and template file ("treatment" vs "variant"), and the "treatment" control did not work. This is now "variant" for both, and the control works . The template will also no longer show the invalid icon for the "neutral" variant, to conform with the guidelines. There was previously only one story in Storybook. I've added another story so both the "neutral" and "negative" (+ wrapping text) variants are represented in VRTs. The missing "size" control has also been added.
:framed_picture: _Expected VRT Change:_ Default story will now show neutral text color and no icon. New story is added that shows negative text color and invalid icon.

CSS-583

### Validation Steps

- [ ] Invalid radio examples still display the same (default, emphasized, focus, etc)
- [ ] Field group invalid radios example has `aria-invalid="true"` in docs and Storybook
- [ ] Confirm and test the Storybook changes for Field group and Help text as described in the PR description

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/radio/metadata/radio.yml
+++ b/components/radio/metadata/radio.yml
@@ -21,7 +21,7 @@ sections:
       - Read-only radio buttons do not have a focus ring, but the button should be focusable.]
 
       ### Invalid/Error State
-      - Invalid radio buttons are signified by including [HelpText](helptext.html) in a [FieldGroup](fieldgroup.html).
+      - Invalid radio buttons are signified by including [Help text](helptext.html) in a [Field group](fieldgroup.html). The `.is-invalid` class has been removed. See Field group for an example with an invalid radio group. 
 
       ### Remove focus-ring class
       We've migrated away from the focus-ring class in favor of the native `:focus-visible` pseudo-class due to changes in browser support.
--- a/components/fieldgroup/metadata/fieldgroup.yml
+++ b/components/fieldgroup/metadata/fieldgroup.yml
@@ -1,20 +1,20 @@
 name: Field group
 description: |
   - A group of fields, usually Radios or Checkboxes.
-  - Fieldgroup incorporates the help text component which may appear below a fieldgroup.
+  - Field group incorporates the Help text component which may appear below a Field group.
   - Help text is necessary to denote invalid checkbox fields, invalid radio button fields, and required fields.
-  - Invalid radio buttons are signified only by alert help text.
-  - Invalid checkboxes are signified by alert help text and alert input box color.
+  - Invalid radio buttons are signified only by negative Help text.
+  - Invalid checkboxes are signified by negative Help text and the negative/invalid color on the input box.
 sections:
   - name: Migration Guide
     description: |
-      ### Field Group Now Includes Label and Helptext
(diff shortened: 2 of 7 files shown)


pfulton (approved):
I think this looks good. Thanks for making the update!

components/radio/metadata/radio.yml
```diff
@@ -21,7 +21,7 @@ sections:
       - Read-only radio buttons do not have a focus ring, but the button should be focusable.]
 
       ### Invalid/Error State
-      - Invalid radio buttons are signified by including [HelpText](helptext.html) in a [FieldGroup](fieldgroup.html).
+      - Invalid radio buttons are signified by including [Help text](helptext.html) in a [Field group](fieldgroup.html). The `.is-invalid` class has been removed. See Field group for an example with an invalid radio group.
```
jnjosh, line 24: Super minor and maybe not worth changing, but I'm curious if the bit about removing `.is-invalid` class is needed in this documentation. Seems like a release note comment not documentation?
```suggestion
      - Invalid radio buttons are signified by including [Help text](helptext.html) in a [Field group](fieldgroup.html). See Field group for an example with an invalid radio group. 
```

jnjosh (approved):
One minor comment, otherwise looks great.

components/radio/metadata/radio.yml
jawinn: I'll defer to @pfulton on that one, but his part is under the "Migration Guide" which often lists things that have been changed or removed within other components.

This is a good point, and I have been wondering if there is a better way to present the notes under Migration Guide, across all components, since they keep getting longer and it is starting to become unclear as to what version what notes apply to---some migration notes may apply to a much older version that would be irrelevant to someone upgrading, but a more recent change may be mixed in the notes. I think more of a release notes style with versions that shows changes and any migration notes for that version might be an improvement.

Issue #2200, opened 2023-10-05
Slider focused style should support forced-colors: active / WHCM


najikahalsema:
## Description

Passing along a [bug report](https://github.com/adobe/spectrum-web-components/issues/3328) filed in SWC by @majornista. 

 The slider's `focused` style does not match the Spectrum specification.

From [Spectrum's docs site](https://spectrum.corp.adobe.com/page/slider/#Keyboard-focus) (expected): 
![image](https://github.com/adobe/spectrum-css/assets/44980010/28bb3376-707a-4eab-99d4-1d9f3c6c3463)

From [Spectrum CSS](https://opensource.adobe.com/spectrum-css/slider.html) (actual): 
![image](https://github.com/adobe/spectrum-css/assets/44980010/2fbb8945-658d-42fc-aa21-33aec9404fe0)

Additionally, the focus ring does not render with `forced-colors: active` or WHCM. We should also use `highlight` color appropriately for `track`, `fill-color`, and `focused` state. 

In Spectrum's specification for WHCM/`forced-colors: active`: 
![image](https://github.com/adobe/spectrum-css/assets/44980010/52cfd254-1817-47e6-98ee-c73d332bcb24)

In Spectrum CSS for WHCM/`forced-colors: active`:
![image](https://github.com/adobe/spectrum-css/assets/44980010/a5083f8e-aedf-43a2-8ed6-e09677977a51)

castastrophe:
This appears to be corrected in the latest release: https://opensource.adobe.com/spectrum-css/?path=/story/components-slider--default&globals=testingPreview:!true

<img width="925" alt="Screenshot 2024-12-10 at 5 50 13 PM" src="https://github.com/user-attachments/assets/27f614aa-d126-494f-8d84-2b75bc1d4dfd">

PR #2217, opened 2023-10-17
fix(slider): focus state matches spec, supports forced-colors

mdt2:
## Description

This PR addresses issue #2200.

- Adjust focus state to match [the spec](https://spectrum.corp.adobe.com/page/slider/#Keyboard-focus) (previously it looked like the active state which was incorrect)
- Standard variant works in high contrast mode. See Xd file in spec for reference.
- Disabled slider has the default cursor on hover. This was already set in a couple places (`.spectrum-Slider` and `.spectrum-Slider-handle`) but needed to be explicitly set on `.spectrum-Slider-controls` to take effect since this class has a specified `cursor: pointer` when the slider isn't disabled.

### Validation

- [x] On focus, the slider handle should match the spec (it shouldn't be filled in and should have a focus ring). Confirm this for high contrast mode as well
- [x] On hover in high contrast mode, the slider track and handle are the `Highlight` color
- [x] On active/dragged in high contrast mode, the handle fills in with `Highlight` to match the spec
- [x] Disabled sliders should get the default cursor on hover
- [x] Focus and disabled states are documented on the docs site

### Regression testing

Validate:

- [x] The page is accessible
- [x] The page is responsive

- [x] The page is accessible
- [x] The page is responsive

## To-do list

### DIFF
--- a/components/slider/index.css
+++ b/components/slider/index.css
@@ -613,15 +613,9 @@ governing permissions and limitations under the License.
 .spectrum-Slider--ramp {
   .spectrum-Slider-handle {
     box-shadow: 0 0 0 var(--spectrum-slider-handle-gap) var(--highcontrast-slider-ramp-handle-border-color-active, var(--mod-sectrum-slider-ramp-handle-border-color-active, var(--spectrum-slider-ramp-handle-border-color-active)));
-    background: var(--mod-slider-ramp-handle-background-color, var(--spectrum-slider-ramp-handle-background-color));
-
-    @media (forced-colors: active) {
-      /* forced-color-adjust required to ensure the "circle" around the handle is transparent */
-      forced-color-adjust: none;
-      box-shadow: 0 0 0 var(--spectrum-slider-handle-gap) ButtonFace;
-      border-color: ButtonText;
-      background-color: ButtonFace;
-    }
+    background: var(--mod-slider-ramp-handle-background-color, var(--highcontrast-slider-ramp-handle-background-color, var(--spectrum-slider-ramp-handle-background-color)));
+    /* forced-color-adjust required to ensure the "circle" around the handle is transparent */
+    forced-color-adjust: none;
   }
 }
 
@@ -707,17 +701,20 @@ governing permissions and limitations under the License.
 @media (forced-colors: active) {
 
   .spectrum-Slider {
-
     /*  over-writes */
     --highcontrast-slider-track-color: ButtonText;
     --highcontrast-slider-track-fill-color: ButtonText;
     --highcontrast-slider-ramp-track-color: ButtonText;
     --highcontrast-slider-ramp-track-color-disabled: GrayText;
     --highcontrast-slider-tick-mark-color: ButtonText;
+    --highcontrast-slider-handle-border-color: ButtonText;
     --highcontrast-slider-handle-border-color-hover: Highlight;
     --highcontrast-slider-handle-border-color-down: Highlight;
     --highcontrast-slider-handle-border-color-key-focus: Highlight;
     --highcontrast-slider-handle-focus-ring-color-key-focus: CanvasText;
+    --highcontrast-slider-handle-background-color: ButtonFace;
+    --highcontrast-slider-ramp-handle-border-color-active: ButtonFace;
+    --highcontrast-slider-ramp-handle-background-color: ButtonFace;
 
     --spectrum-slider-track-color: ButtonText;
     --spectrum-slider-track-fill-color: ButtonText;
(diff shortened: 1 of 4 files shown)


jawinn (changes requested):
It's looking much closer to the intended spec now. A few questions and suggestions below.

There doesn't seem to be any indication in the WHCM design spec about how to handle the "Filled", , "Filled-Offset", and "Range". I think perhaps that at least for focus and hover, the darker portion should use the Highlight color (in the PR you can see the reverse of this on "Filled-Offset"). 
Ideally there'd be a way to differentiate the darkened part of the range by default in WHCM as well, but I'm not sure about the best way; possibly a thicker track since there isn't another appropriate system color to use. But that would need some followup design/a11y involvement/approval.

I also noticed that there are some existing high contrast styles for the ramp that we may want to move next to the rest of the high contrast styles.

components/slider/index.css
```diff
@@ -742,6 +746,36 @@ governing permissions and limitations under the License.
 
   }
 
+  /* Slider track background color doesn't show in high contrast mode, but border does */
+  .spectrum-Slider:not(.is-disabled) {
+    .spectrum-Slider-track {
+      &::before {
+        border-block-start: 2px solid transparent;
+      }
+    }
```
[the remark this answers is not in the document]
mdt2: We found out this was because I was using "Enable automatic dark mode" instead of "prefers-color-scheme" so now I'm seeing it too 👍

components/slider/index.css
```diff
@@ -742,6 +746,36 @@ governing permissions and limitations under the License.
...
+
+    /* We only need these states in high contrast mode */
+    .spectrum-Slider-controls:hover,
+    .spectrum-Slider-controls:active,
+    .spectrum-Slider-controls:focus-within {
+      .spectrum-Slider-track::before {
+        background: var(--highcontrast-slider-track-color-hover, var(--mod-slider-track-color, var(--spectrum-slider-track-color)));
+        border-block-start: none;
+      }
```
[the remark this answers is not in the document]
mdt2: Hahah yes this is much better. Thank you!

The handle border-color is a bit different. We only have a hover effect set up on the handle itself, not the track. Which is fine without high contrast turned on because the track doesn't need a hover effect until we have WHCM.
So if I only update the `--highcontrast-slider-handle-border-color-hover` custom property, it does work, but we don't get the hover effect on the handle unless we hover it specifically. So the track turns the `Highlight` color but the handle is still white unless you're hovering the handle. I think this new solution matches the spec better, but I'm open to thoughts!

components/slider/index.css
```diff
@@ -516,22 +515,6 @@ governing permissions and limitations under the License.
...
-    .spectrum-Slider-handle {
-      cursor: default;
-      pointer-events: none;
-    }
-
-    .spectrum-Slider-tickLabel {
-      color: var(--highcontrast-slider-label-text-color-disabled, var(--mod-slider-label-text-color-disabled, var(--spectrum-slider-label-text-color-disabled)));
-    }
-  }
-}
```
mdt2, line 533: Relocated this to an existing `spectrum-Slider.is-disabled` selector on new line 628

components/slider/index.css
```diff
@@ -609,15 +593,9 @@ governing permissions and limitations under the License.
...
     box-shadow: 0 0 0 var(--spectrum-slider-handle-gap) var(--highcontrast-slider-ramp-handle-border-color-active, var(--mod-sectrum-slider-ramp-handle-border-color-active, var(--spectrum-slider-ramp-handle-border-color-active)));
-    background: var(--mod-slider-ramp-handle-background-color, var(--spectrum-slider-ramp-handle-background-color));
-
-    @media (forced-colors: active) {
-      /* forced-color-adjust required to ensure the "circle" around the handle is transparent */
-      forced-color-adjust: none;
-      box-shadow: 0 0 0 var(--spectrum-slider-handle-gap) ButtonFace;
-      border-color: ButtonText;
-      background-color: ButtonFace;
-    }
```
mdt2, line 620: We can rely on the `--highcontrast` custom properties to do this instead.

components/slider/index.css
```diff
@@ -609,15 +593,9 @@ governing permissions and limitations under the License.
...
+    background: var(--mod-slider-ramp-handle-background-color, var(--highcontrast-slider-ramp-handle-background-color, var(--spectrum-slider-ramp-handle-background-color)));
+    /* forced-color-adjust required to ensure the "circle" around the handle is transparent */
+    forced-color-adjust: none;
```
jawinn, line 598: I would move the two `forced-color-adjust` declarations into the `@media (forced-colors: active)` block, as they only apply there.

jawinn (commented):
Looks good to me! Thanks for those updates. 
I only have one small recommendation to move a few lines of CSS.

components/slider/index.css
mdt2: Sounds good! I put them in media queries inline, but do you think it's better to move them down to the big forced-colors block and repeat the selector instead?

pfulton:
@najikahalsema @majornista This work should address https://github.com/adobe/spectrum-css/issues/2200. Would you mind taking a look to be sure that we've fixed everything? PR deploy is here: https://pr-2217--spectrum-css.netlify.app/

najikahalsema:
Looks good to me! Thanks for doing this work.

pfulton:
@castastrophe Noting that this one will have some Docs changes!

PR #2229, opened 2023-10-26
refactor(assetcard)!: token migration

jenndiaz:
## Description

Migrates AssetCard
- Addresses empty checkbox label in docs site and storybook 
- adds WHCM values for selectionIndicatorOrder 
- adds alt text to the card image 
- Replace `visibility` with `display`

1. Open the docs site for the asset card component: 
- [ ] No visual differences between production and this branch 
- [ ] when emulating WHCM, highlight color border should appear on hover for the card border and the checkbox 

### Regression testing

Validate:

- [x] The page is accessible
- [x] The page is responsive

3. A migrated documentation page (such as [action group](https://pr-###--spectrum-css.netlify.app/actiongroup.html)), including:

- [x] The page is accessible
- [x] The page is responsive

## To-do list


components/assetcard/stories/template.js
```diff
@@ -52,15 +53,15 @@ export const Template = ({
...
+				<img class="${rootClass}-asset" alt="assetcard example image" src="${image ?? exampleImage}" />
 				<div class="${rootClass}-selectionOverlay"></div>
 			</div>
 			${when(
 				title,
 				() => html`<div class="${rootClass}-header">
 					${when(
 						title,
-						() => html`<div class="${rootClass}-title">${title}</div>`
+						() => html`<div class="${rootClass}-title" id=${camelCase(title)}>${title}</div>`
```
jenndiaz, line 64: @pfulton  I'm using `camelCase()` here to use the title as the ID, remove the spacing in multi word titles, and avoid needing to pass a separate ID. I know we're not using Lodash a lot currently. Any concerns about using it here?

pfulton: I think that's fine for now. It shouldn't increase the bundle size/affect performance too much, and it's just for Storybook anyway.

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
 /* outer container, unstyled */
 .spectrum-AssetCard {
   /* todo: this isn't quite the size from the XD file as 232px is not a size token, so we use 224px */
-  --spectrum-assetcard-asset-size: var(--spectrum-global-dimension-static-size-2800);
+  --spectrum-assetcard-asset-size: 224px;
```
castastrophe, line 16: @pfulton Should we go ahead and use 232px like the original design called for since we have to hard-code this value anyway?

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
...
+  --spectrum-assetcard-background-color: var(--spectrum-gray-200);
+  --spectrum-assetcard-asset-animation-duration: var(--spectrum-animation-duration-100);
+  --spectrum-assetcard-assetcontainer-border-size: 1px;
```
castastrophe, line 19:
```suggestion
  --spectrum-assetcard-asset-container-border-size: 1px;
```

Since this is a new property, should we use the dash in the name to keep it human-readable?

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
...
-  --spectrum-assetcard-selectionindicator-margin: var(--spectrum-global-dimension-size-150);
-  --spectrum-assetcard-selectionindicator-size: var(--spectrum-global-dimension-static-size-500);
-  --spectrum-assetcard-selectionindicator-border-radius: var(--spectrum-alias-border-radius-regular);
-
-  --spectrum-assetcard-heading-text-color: var(--spectrum-global-color-gray-900);
-  --spectrum-assetcard-content-text-color: var(--spectrum-global-color-gray-700);
+  --spectrum-assetcard-border-color-hover: var(--spectrum-gray-500);
+  --spectrum-assetcard-border-color-down: var(--spectrum-gray-600);
+
+  --spectrum-assetcard-focus-ring-gap: 5px;
```
castastrophe, line 27: Would this map well to: `var(--spectrum-focus-indicator-gap)`?

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
...
+  --spectrum-assetcard-focus-indicator-thickness: var(--spectrum-focus-indicator-thickness);
+
+  /* selection indicator - checkbox or number */
+  --spectrum-assetcard-selectionindicator-background-color-default: rgba(var(--spectrum-gray-100-rgb), 0.9);
+  --spectrum-assetcard-selectionindicator-size: var(--spectrum-card-selection-background-size);
+  --spectrum-assetcard-selectionindicator-border-radius: var(--spectrum-corner-radius-100);
```
castastrophe, line 33: Should this be 100 or 75? Just wanted to double check.

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
...
 
+.spectrum-AssetCard {
   /* contain selection indicator */
   position: relative;
 
   display: flex;
   flex-direction: column;
 
-  inline-size: var(--spectrum-assetcard-asset-size);
+  inline-size: var(--mod-assetcard-asset-size, var(--spectrum-assetcard-asset-size));
```
castastrophe, line 74: I noticed a token for minimum width - should we account for that here too? `var(--spectrum-card-minimum-width)`

components/assetcard/index.css
```diff
@@ -136,89 +175,56 @@ governing permissions and limitations under the License.
   align-items: center;
 
   /* space between the asset and the header, place it on the header so header can be optional */
-  margin-block-start: var(--spectrum-global-dimension-static-size-200);
+  margin-block-start: var(--mod-assetcard-header-margin-block-start, var(--spectrum-assetcard-header-margin-block-start));
 }
 
 /* bold title */
 .spectrum-AssetCard-title {
```
castastrophe, line 182: I took a different approach to setting up the typography here - https://github.com/adobe/spectrum-css/pull/2250/files#diff-9a31086a7f778cd08f7edbddfa8a949a9d8cf3871d6fe3d4eb63b5aa92de334cR63 - did you want to take a look and see if this would make sense to use? It allows for more typography cascading and contextual updating as well as including the language modifiers we might need.

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
 /* outer container, unstyled */
 .spectrum-AssetCard {
   /* todo: this isn't quite the size from the XD file as 232px is not a size token, so we use 224px */
-  --spectrum-assetcard-asset-size: var(--spectrum-global-dimension-static-size-2800);
+  --spectrum-assetcard-asset-size: 224px;
+  --spectrum-assetcard-background-color: var(--spectrum-gray-200);
+  --spectrum-assetcard-asset-animation-duration: var(--spectrum-animation-duration-100);
+  --spectrum-assetcard-assetcontainer-border-size: 1px;
+  --spectrum-assetcard-header-margin-block-start: var(--spectrum-spacing-300);
```
castastrophe, line 20: I wonder if this alias is slightly more accurate here than pointing directly to the spacing var? `var(--spectrum-text-to-visual-300)`

components/assetcard/index.css
```diff
@@ -13,52 +13,84 @@ governing permissions and limitations under the License.
...
+
+  /* content */
+  --spectrum-assetcard-content-text-color: var(--spectrum-gray-700);
+  --spectrum-assetcard-content-font-family: var(--spectrum-sans-font-family-stack);
+  --spectrum-assetcard-content-font-weight: var(--spectrum-body-sans-serif-font-weight);
+  --spectrum-assetcard-content-font-style: var(--spectrum-default-font-style);
+  --spectrum-assetcard-content-line-height: var(--spectrum-line-height-200);
+  --spectrum-assetcard-content-letter-spacing: 0;
+
+  --spectrum-assetcard-content-margin-block-start: var(--spectrum-spacing-75);
```
castastrophe, line 64:
```suggestion
  --spectrum-assetcard-content-margin-block-start: calc(
		var(--spectrum-assetcard-title-font-size) *
			var(--spectrum-heading-margin-bottom-multiplier)
	);
```

I'd like to see us capture the logic here which is pulled from the updated typography styles.

components/assetcard/index.css
```diff
@@ -227,37 +233,38 @@ governing permissions and limitations under the License.
...
 
-  color: var(--spectrum-global-color-static-white);
-  font-weight: var(--spectrum-global-font-weight-bold);
-  font-size: var(--spectrum-global-dimension-font-size-400);
+  color: var(--highcontrast-assetcard-selectionindicator-color, var(--mod-assetcard-selectionindicator-color, var(--spectrum-assetcard-selectionindicator-color)));
+  font-weight: var(--mod-assetcard-selectionindicator-font-weight, var(--spectrum-assetcard-selectionindicator-font-weight));
+  font-size: var(--mod-assetcard-selectionindicator-font-size, var(--spectrum-assetcard-selectionindicator-font-size));
 
-  visibility: hidden;
+  display: none;
```
castastrophe, line 263: Is this overwriting the display: flex above on 248?

components/assetcard/index.css
```diff
@@ -305,18 +312,18 @@ governing permissions and limitations under the License.
...
   }
 
   &.is-selected,
-  &:focus,
+  &:focus-visible,
   &:hover {
     .spectrum-AssetCard-selectionIndicator {
-      visibility: visible;
+      display: flex;
       opacity: 1;
```
castastrophe, line 327: We can remove opacity now that we're using display: none

components/assetcard/index.css
```diff
@@ -227,37 +233,38 @@ governing permissions and limitations under the License.
...
-  font-weight: var(--spectrum-global-font-weight-bold);
-  font-size: var(--spectrum-global-dimension-font-size-400);
+  color: var(--highcontrast-assetcard-selectionindicator-color, var(--mod-assetcard-selectionindicator-color, var(--spectrum-assetcard-selectionindicator-color)));
+  font-weight: var(--mod-assetcard-selectionindicator-font-weight, var(--spectrum-assetcard-selectionindicator-font-weight));
+  font-size: var(--mod-assetcard-selectionindicator-font-size, var(--spectrum-assetcard-selectionindicator-font-size));
 
-  visibility: hidden;
+  display: none;
 
   opacity: 0;
```
castastrophe, line 265: We can remove opacity now that we're using display: none

components/assetcard/index.css
```diff
@@ -335,9 +342,24 @@ governing permissions and limitations under the License.
 
   &.is-selected {
     .spectrum-AssetCard-selectionIndicator {
-      visibility: visible;
+      display: flex;
       opacity: 1;
```
castastrophe, line 346: Same here - this can be removed

components/assetcard/package.json
```diff
@@ -20,15 +20,15 @@
   "peerDependencies": {
     "@spectrum-css/checkbox": ">=7",
     "@spectrum-css/icon": ">=3",
-    "@spectrum-css/typography": ">=4 <=5",
-    "@spectrum-css/vars": ">=9"
+    "@spectrum-css/tokens": ">=10",
```
castastrophe, line 23:
```suggestion
    "@spectrum-css/tokens": ">=12",
```

components/assetcard/package.json
```diff
@@ -20,15 +20,15 @@
...
+    "@spectrum-css/typography": ">=4 <=5"
```
castastrophe, line 24: Are we still using typography here?  It seems like we're hooking into the variables directly and not using the classes from that package anymore.

components/assetcard/stories/template.js
castastrophe: Just a heads up @pfulton lodash-es is loaded in storybook for every story already - so not a new dependency

castastrophe (changes requested):

components/assetcard/index.css
jenndiaz: I did look at Card, with the context that this component will be updated to be more like a quiet card, and card does not have a set width and instead has a min-width of 100px.

components/assetcard/index.css
jenndiaz: I wanted it to! But the values are different `--spectrum-focus-indicator-gap: 2px;`

components/assetcard/index.css
jenndiaz: 100 is correct. the values for `--spectrum-corner-radius-100` do align with the border radius on the component currently.

components/assetcard/index.css
jenndiaz: yes it is, I'll remove the `display: flex` above this

components/assetcard/index.css
jenndiaz: The `--spectrum-card-minimum-width` token is used in the separate card component. I think It would make sense to move to that but I was not sure if that's with in scope of this migration.

components/assetcard/index.css
jenndiaz: I added in the cjk custom properties (in looking at it noticed we do not have size tokens for cjk body text just cjk heading).

Could you explain more about what other changes you're suggesting?

components/assetcard/index.css
jenndiaz: Currently this value is 4px regardless of text size, changing to the suggested syntax would change the margin for Large scale to 4.25px. Do we want to introduce this change?

castastrophe (approved):

castastrophe (approved):

PR #2230, opened 2023-10-26
docs: storybook - fix duplicate ids for checkbox and switch

jawinn:
## Description

This PR includes some minor Storybook updates to fix duplicate ID messages noticed in the console:

**Checkbox**
The same value was being used for the ID attribute on the checkbox label and checkbox itself. This update makes sure they are both given different IDs based on the passed in ID, by adding a suffix to the ID used on the checkbox input.

This also fixes the markup previously showing an empty ID attribute on the checkbox when the ID was not set.

**Switch**
The switch was using a hardcoded ID for the input and its label, so showing multiple switches resulted in duplicate IDs. This now sets unique IDs based on the passed in ID, along with a fallback.

### Validation steps

Open the [storybook](https://pr-2230--spectrum-css.netlify.app/preview/) for the PR site:
  - [x] Storybook for _**Menu - With Actions**_ no longer shows Chrome console message ("Issues" tab) "Duplicate form field id in the same form"
  - [x] Storybook for **_Menu - Multiple Selection_** no longer shows Chrome console message ("Issues" tab) "Duplicate form field id in the same form"
  - [x] Storybook for **_Checkbox - Default_** no longer shows Chrome console message ("Issues" tab) "A form field element should have an id or name attribute"

## To-do list

### DIFF
--- a/components/checkbox/stories/template.js
+++ b/components/checkbox/stories/template.js
@@ -73,7 +73,7 @@ export const Template = ({
 					if (isDisabled) return;
 					updateArgs({ isChecked: !isChecked });
 				}}
-				id=${id}
+				id=${ifDefined(id ? `${id}-checkbox` : undefined)}
 			/>
 			<span class="${rootClass}-box">
 				${Icon({
--- a/components/checkbox/stories/template.js
+++ b/components/checkbox/stories/template.js
@@ -73,7 +73,7 @@ export const Template = ({
 					if (isDisabled) return;
 					updateArgs({ isChecked: !isChecked });
 				}}
-				id=${ifDefined(id ? `${id}-checkbox` : undefined)}
+				id=${ifDefined(id ? `${id}-input` : undefined)}
 			/>
 			<span class="${rootClass}-box">
 				${Icon({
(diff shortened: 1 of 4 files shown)


pfulton:
@castastrophe want to sneak this one into the build branch?

components/checkbox/stories/template.js
```diff
@@ -73,7 +73,7 @@ export const Template = ({
 					if (isDisabled) return;
 					updateArgs({ isChecked: !isChecked });
 				}}
-				id=${id}
+				id=${ifDefined(id ? `${id}-checkbox` : undefined)}
```
castastrophe, line 76: This is resulting in `default-checkbox-checkbox` which is good because it's not duplicate but seems a tiny bit silly so maybe we can alter the customization a bit? 
```suggestion
				id=${ifDefined(id ? `${id}-input` : undefined)}
```

castastrophe (changes requested):
One minor nit about an id of checkbox-checkbox but everything else is looking great.

castastrophe (approved):

PR #2246, opened 2023-10-31
fix(popover): offset matches spec

mdt2:
## Description

This has been a difficult component to test without implementing it. What I landed on is that the [8px offset distance in the spec](https://spectrum.adobe.com/page/popover/#Offset) should be accomplished by the animation.

For popovers without a tip, we can simply customize the animation-distance value to align with the spec.

For popovers with a tip, we need to add an offset margin equal to the width or height of the tip (depending on the position of the popover) so that the animation-distance is applied against the popover + tip.

- Popover (or it's tip if there is one) is offset 8px from the source [per the spec](https://spectrum.adobe.com/page/popover/#Offset)
- Docs have been updated to reflect the change to default
- Storybook examples have been updated to show the correct default offset

### Validation
**Tip for Docs Site:** If you're having trouble seeing if the animation is actually moving 8px like I was, you can turn off the visibility changes in `components/commons/overlay-coretokens.css` by changing `visibility: hidden` to `visibility: visible` and `opacity: 0;` to `opacity: 1;`. This will let you see the popover move up and down 8px on click.

**Note about positioning:** We're forcing position of the popover in both the Docs site and in Storybook, but we do it in different ways on each. The Docs site uses a `dummy-spacing` class on a `div` that we can put right up next to the dummy source which gives the popover the space to move the 8px animation-distance on open/close. In Storybook there are some extra calculations. I've adjusted the offset value for each situation to give it 8px distance visually, but the values applied here weren't a uniform solution.

- [ ] Popover is 8px from the source at all positions (you'll have to close between changing positions or you'll see weird behavior)
    - [x] [Default Storybook example](https://pr-2246--spectrum-css.netlify.app/preview/?path=/story/components-popover--default&args=isOpen:!true)
        - [x] LTR
        - [x] RTL
    - [ ] [With Tip Storybook example](https://pr-2246--spectrum-css.netlify.app/preview/?path=/story/components-popover--with-tip&args=isOpen:!true)
        - [ ] LTR
        - [ ] RTL
    - [x] [Nested Storybook example](https://pr-2246--spectrum-css.netlify.app/preview/?path=/story/components-popover--nested&args=isOpen:!true)
        - [x] LTR
        - [x] RTL
- [x] [Docs site examples](https://pr-2246--spectrum-css.netlify.app/popover#popover-offsetandanimation) show popover 8px from source and they animate in and out @jenndiaz 

### Regression testing

Validate:

- [ ] The page is accessible
- [ ] The page is responsive

- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

### DIFF
--- a/components/popover/index.css
+++ b/components/popover/index.css
@@ -19,8 +19,8 @@ governing permissions and limitations under the License.
     --flow-direction: -1;
   }
 
-  /* change to offset from source */
-  --spectrum-popover-offset: 0px;
+  /* animation distance is equal to and responsible for popover offset */
+  --spectrum-popover-animation-distance: var(--spectrum-spacing-100);
 
   --spectrum-popover-background-color: var(--spectrum-background-layer-2-color);
   --spectrum-popover-border-color: var(--spectrum-gray-400);
@@ -111,9 +111,15 @@ governing permissions and limitations under the License.
 .spectrum-Popover--top-right,
 .spectrum-Popover--top-start,
 .spectrum-Popover--top-end {
-  margin-bottom: var(--mod-popover-offset, var(--spectrum-popover-offset));
+  /* spacing to include tip in calculation of offset from source */
+  &.spectrum-Popover--withTip {
+    /* tip size minus where it overlaps with popover border */
+    margin-bottom: calc(var(--mod-popover-pointer-height, var(--spectrum-popover-pointer-height)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+  }
+
   /* popover animates upward ⬆ */
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance));
     @inherit: %spectrum-overlay--top--open;
   }
 }
@@ -124,9 +130,15 @@ governing permissions and limitations under the License.
 .spectrum-Popover--bottom-right,
 .spectrum-Popover--bottom-start,
 .spectrum-Popover--bottom-end {
-  margin-top: var(--mod-popover-offset, var(--spectrum-popover-offset));
+  /* spacing to include tip in calculation of offset from source */
+  &.spectrum-Popover--withTip {
+    /* tip size minus where it overlaps with popover border */
+    margin-top: calc(var(--mod-popover-pointer-height, var(--spectrum-popover-pointer-height)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+  }
+
   /* popover animates downward ⬇ */
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance));
     @inherit: %spectrum-overlay--bottom--open;
   }
 }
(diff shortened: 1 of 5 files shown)


components/popover/stories/template.js
```diff
@@ -72,31 +71,37 @@ export const Template = ({
...
-				} 
+				}
 				if (position.includes("left") || position.includes("right")) {
 					y = triggerYCenter - (popHeight > 0 ? popHeight / 2 : popHeight);
 				}
 				if (position.includes("top") && !position.includes("-top")) {
 					y = rect.top - popHeight;
-					yOffset = "- var(--spectrum-popover-offset)";
+					yOffset = withTip
+						? "- (var(--spectrum-popover-pointer-height) + var(--spectrum-popover-animation-distance))"
```
jenndiaz, line 81: When looking at the storybook variant with tip, it appears as if the tip to source button is measuring 9px instead of 8px. Is it possible we need to also be accounting for the border on the tip?

Without tip, measures 8px from source to border:
Screenshot 2023-11-03 at 9 12 05 AM

With Tip, source button to popover tip is 10px but source button to menu border is 16px (pink box)

Screenshot 2023-11-03 at 9 12 49 AM

jenndiaz (changes requested):
The docs site looks good. I had some questions about the storybook with tip values.

components/popover/stories/template.js
mdt2: Good catch! I'll look into it, you're probably onto something with the border consideration

jenndiaz (approved):
looks good to me!

components/popover/index.css
```diff
@@ -111,9 +111,15 @@ governing permissions and limitations under the License.
...
-  margin-bottom: var(--mod-popover-offset, var(--spectrum-popover-offset));
+  /* spacing to include tip in calculation of offset from source */
+  &.spectrum-Popover--withTip {
+    /* tip size minus where it overlaps with popover border */
+    margin-bottom: calc(var(--mod-popover-pointer-height, var(--spectrum-popover-pointer-height)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+  }
+
   /* popover animates upward ⬆ */
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance));
```
jawinn, line 122: Won't this immediately get overwritten by the same variable that's defined in the `%spectrum-overlay--top--open`? Also, I'm wondering if it makes more sense to just use the one `transform` line it's importing with new values rather than needing the `@inherit` and probably needing to move the custom property after the `@inherit`.

Initially I was thinking that this variable could be defined once in `.spectrum-Popover.is-open` since it appears to be defined the same way for is-open on all (?) the variants.

mdt2: Yeah, per [that convo](https://github.com/adobe/spectrum-css/pull/2266#discussion_r1383879047) in the tooltip PR (#2266), I added a commit that reworks this to define the transforms in `popover/index.css` instead 👍

components/popover/stories/template.js
jawinn: For the "With Tip" story, I'm measuring 9px between the tip and the button:
Screenshot 2023-11-08 at 3 33 59 PM

mdt2: Ah I see the confusion! Those super light gray 2 pixels outside of what you've selected are part of the tip (the path inside the svg), so if we measure from that it's 8px:

Your image pointing at the pixels I'm talking about:
measure-these-too

The path selected:
Screen Shot 2023-11-08 at 4 17 34 PM

The popover parent element selected:
Screen Shot 2023-11-08 at 4 19 17 PM

mdt2: Hmm, you're sure that was latest @jawinn? I'm seeing 8px on the PR deploy:

Screen Shot 2023-11-08 at 4 19 17 PM

jawinn (approved):

castastrophe:
@mdt2 When you get a chance, can you rebase this one and resolve the conflicts?

PR #2248, opened 2023-10-31
fix(button): support for wrapping text

jawinn:
## Description

This update allows text wrapping within the **Button** component and updates the alignment of the label and icon when wrapping to conform with the design specs.

- Fixes a previously introduced bug that disabled wrapping of lengthy text inside a Button. This removes the white-space property that prevented wrapping.
- BREAKING CHANGE: Updates styles so that buttons that contain a workflow icon have the correct icon and text alignment. Previously everything was vertically aligned center and the text was aligned center, which did not match the design in the case of wrapping text. See screenshots for reference to the design.

  This is a significant change as the vertical flex alignment of the icon is now to the top/start instead of the center, and the design defined _--spectrum-component-top-to-workflow-icon-_* tokens are used for the spacing between the icon and the top of the component. While button is intended to be used with workflow icons, there were not previously any guidelines around _not_ using UI icons and even SplitButton was using one---in order to support vertical alignment of any existing usage of UI icons, some additional CSS calculations were included to add some margin based on the difference between the height of the intended workflow icon size and the actual icon size used (now accessible through `--spectrum-icon-block-size` since the 6.0 core tokens migration of _Icon_).
- Adds a "Wrapping" example to Storybook. This also has a Chromatic-only version with additional variations displayed.

CSS-621

### Validation steps

  - [x] Verify changes with design team [@jawinn]
  - [x] Button wrapping example exists in Storybook.
  - [x] Button without icon wraps and has centered text (like [guidelines](https://spectrum.adobe.com/page/button/#Text-overflow)).
  - [x] Button with icon has icon aligned to the top of the component and text aligned to the left/start of the component. Matching "Design spec for wrapping behavior with icon" in screenshots.
  - [x] Wrapping behavior looks correct in right-to-left text direction.
  - [ ] Non-wrapping examples of the button with an icon have no visual changes.

### Regression testing

Validate:

- [x] The page is accessible
- [x] The page is responsive

- [x] The page is accessible
- [x] The page is responsive

## Screenshots

Updated wrapping behavior (Storybook):
Screenshot 2023-10-31 at 3 48 35 PM

Chromatic view of Wrapping story that includes showing backwards compatible alignment of differently sized UI icons
Screenshot 2024-02-08 at 11 57 21 AM

Reference: design spec for wrapping behavior with icon
Screenshot 2023-10-31 at 4 02 40 PM
Previous behavior:
Screenshot 2024-02-08 at 2 14 28 PM

Reference: design spec for wrapping behavior without icon
button_behaviors_text-overflow@2x_1649295070805

## To-do list

- [ ] VRTs have been run and reviewed.

### DIFF
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -146,6 +146,7 @@ governing permissions and limitations under the License.
       var(--mod-button-edge-to-text, var(--spectrum-button-edge-to-text))
     );
     color: inherit;
+    flex-shrink: 0;
   }
 
   /* correct focus indicator radius for t-shirt sizing */
@@ -176,7 +177,6 @@ a.spectrum-Button {
   @extend %spectrum-ButtonLabel;
   padding-block-start: calc(var(--mod-button-top-to-text, var(--spectrum-button-top-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
   padding-block-end: calc(var(--mod-button-bottom-to-text, var(--spectrum-button-bottom-to-text)) - var(--mod-button-border-width, var(--spectrum-button-border-width)));
-  white-space: nowrap;
   line-height: var(--mod-button-line-height, var(--spectrum-button-line-height));
   align-self: start;
 }
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -155,10 +155,12 @@ governing permissions and limitations under the License.
       var(--spectrum-icon-block-size, var(--spectrum-button-intended-icon-size))
     );
 
-    margin-block-start: max(0px,
-      var(--mod-button-top-to-icon, var(--spectrum-button-top-to-icon)) - 
-      var(--mod-button-border-width, var(--spectrum-button-border-width)) +
-      (var(--_icon-size-difference, 0px) / 2)
+    margin-block-start: var(--mod-button-icon-margin-block-start,
+      max(0px,
+        var(--mod-button-top-to-icon, var(--spectrum-button-top-to-icon)) - 
+        var(--mod-button-border-width, var(--spectrum-button-border-width)) +
+        (var(--_icon-size-difference, 0px) / 2)
+      )
     );
 
     margin-inline-start: calc(
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -23,6 +23,7 @@ governing permissions and limitations under the License.
   --spectrum-button-focus-ring-border-radius: calc(var(--spectrum-button-border-radius) + var(--spectrum-button-focus-ring-gap));
   --spectrum-button-focus-ring-thickness: var(--spectrum-focus-indicator-thickness);
   --spectrum-button-focus-indicator-color: var(--spectrum-focus-indicator-color);
+  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-50);
 }
 
 .spectrum-Button--sizeS {
@@ -40,6 +41,7 @@ governing permissions and limitations under the License.
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-small);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-small);
   --spectrum-button-top-to-icon: var(--spectrum-component-top-to-workflow-icon-75);
+  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-75);
 }
 
 .spectrum-Button--sizeM {
@@ -57,6 +59,7 @@ governing permissions and limitations under the License.
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-medium);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-medium);
   --spectrum-button-top-to-icon: var(--spectrum-component-top-to-workflow-icon-100);
+  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-100);
 
   &[pending],
   &.is-pending {
@@ -79,6 +82,7 @@ governing permissions and limitations under the License.
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-large);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-large);
   --spectrum-button-top-to-icon: var(--spectrum-component-top-to-workflow-icon-200);
+  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-200);
 
   &[pending],
   &.is-pending {
@@ -101,6 +105,7 @@ governing permissions and limitations under the License.
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-extra-large);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-extra-large);
   --spectrum-button-top-to-icon: var(--spectrum-component-top-to-workflow-icon-300);
+  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-300);
 
   .spectrum--medium &[pending],
   .spectrum--medium &.is-pending {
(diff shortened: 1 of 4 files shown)


jenndiaz (changes requested):
One comment about the static color variant background colors. 

I looked at it on Safari, Firefox, and WHCM and didn't see any other issues.

components/button/index.css
```diff
@@ -170,26 +183,22 @@ a.spectrum-Button {
...
-
     transition: box-shadow var(--mod-button-animation-duration, var(--spectrum-button-animation-duration)) ease-in-out;
-
     pointer-events: none;
     content: '';
   }
 
   &:focus-visible {
-    /* kill the default ring */
+    /* Remove the default focus outline */
```
mdt2, line 201: Ahaha much more tactful, good job

components/button/stories/button.stories.js
```diff
@@ -78,6 +80,26 @@ export default {
 			options: ["white", "black"],
 			control: "select",
 		},
+		backgroundColor: {
```
mdt2, line 83: Nice, thanks for adding this option 💯

mdt2 (commented):
Looking good! +1 to Jenn's feedback and then I think this should be good

components/button/stories/button.stories.js
```diff
@@ -99,31 +123,45 @@ export default {
...
 };
 
+/**
+ * Multiple button variations displayed in one story template.
+ * Used as the base template for the stories.
+ */
 const CustomButton = (args) => {
+	const wrapperStyles = {
+		'padding': '1rem',
+		...(args.staticColor && args.backgroundColor ? {'background-color:': args.backgroundColor} : {}),
```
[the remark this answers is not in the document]
jawinn: Thanks. This has been fixed.

pfulton (approved):
Thanks for taking care of this one!

jenndiaz (approved):

components/button/index.css
```diff
@@ -121,11 +123,17 @@ governing permissions and limitations under the License.
   }
 
   .spectrum-Icon {
+    margin-block-start: max(0px,
+      var(--mod-button-top-to-icon, var(--spectrum-button-top-to-icon)) - 
+      var(--mod-button-border-width, var(--spectrum-button-border-width))
+    );
```
castastrophe, line 129:
```suggestion
    margin-block-start: max(0px,
      var(--mod-button-top-to-icon, calc(var(--spectrum-button-top-to-icon)) - 
      var(--mod-button-border-width, var(--spectrum-button-border-width)))
    );
```

This part doesn't need to be wrapped in a calc?

components/button/index.css
```diff
@@ -236,7 +245,16 @@ a.spectrum-Button {
...
 
-/* windows high contrast mode over-writes for accent variant */
+/* Static color variants */
+.spectrum-Button--staticWhite {
+  --spectrum-button-focus-indicator-color: var(--mod-static-black-focus-indicator-color, var(--spectrum-static-black-focus-indicator-color));
+}
+
+.spectrum-Button--staticBlack {
+  --spectrum-button-focus-indicator-color: var(--mod-static-black-focus-indicator-color, var(--spectrum-static-black-focus-indicator-color));
+}
```
castastrophe, line 255: Do these 2 need increased specificity by incorporating the .spectrum-Button selector as well?

components/button/stories/button.stories.js
```diff
@@ -79,6 +80,16 @@ export default {
...
+		showIconOnlyButton: {
+			table: {
+				disable: true,
+			},
+		},
+		showOneButtonPerLine: {
+			table: {
+				disable: true,
+			},
+		},
```
castastrophe, line 92:
```suggestion
		layout: {
		    name: "Layout",
		    description: "How the buttons align in the preview",
		    type: { name: "string" },
			table: { 
			    type: { summary: "string" },
			    category: "Advanced"
			},
			options: ["stacked","inline"],
			control: "radio"
		},
```

This might be nice to open up to developers for live-testing

components/button/stories/button.stories.js
```diff
@@ -87,6 +98,8 @@ export default {
 		variant: "accent",
 		treatment: "fill",
 		isDisabled: false,
+		showIconOnlyButton: true,
+		showOneButtonPerLine: false,
```
castastrophe, line 102:
```suggestion
		layout: "inline",
```

components/button/stories/button.stories.js
```diff
@@ -100,42 +113,57 @@ export default {
...
 };
 
+/**
+ * Multiple button variations displayed in one story template.
+ * Used as the base template for the stories.
+ */
 const CustomButton = ({
 	iconName,
 	staticColor,
+	showOneButtonPerLine,
```
castastrophe, line 123:
```suggestion
	layout,
```

components/button/stories/button.stories.js
```diff
@@ -100,42 +113,57 @@ export default {
...
 	customStyles = {},
 	...args
 }) => {
+	// Optional wrapper for each button, to assist with the testing of wrapping text.
+	const ButtonWrap = (content) => {
+		const buttonWrapStyles = {
+			'margin-block': '15px',
+			'max-width': '480px',
+		};
+		return showOneButtonPerLine ? html`<div style=${styleMap(buttonWrapStyles)}>${content}</div>` : content;
```
castastrophe, line 134:
```suggestion
		return layout === "stacked" ? html`<div style=${styleMap(buttonWrapStyles)}>${content}</div>` : content;
```

components/button/stories/button.stories.js
```diff
@@ -184,3 +212,11 @@ WithForcedColors.parameters = {
 WithForcedColors.args = {
 	iconName: "Actions",
 };
+
+export const Wrapping = CustomButton.bind({});
+Wrapping.args = {
+	showOneButtonPerLine: true,
```
castastrophe, line 218:
```suggestion
	layout: "stacked",
```

castastrophe (commented):
A couple notes but looking great!

components/button/stories/button.stories.js
jawinn: `layout` is a much better name, thanks! I've made all these updates to rename that control and make it visible.

components/button/index.css
jawinn: I don't think so, unless there's a scenario you're thinking of? I just moved these up a tiny bit above the high contrast styles. They are still at the end of the file and override the same specificity of their definitions in `.spectrum-Button`.

jawinn:
This is ready for re-review. Storybook is updated and has a Chromatic-only template. I've updated the styles to support existing use of UI icons instead of Workflow icons to address my [last comment](https://github.com/adobe/spectrum-css/pull/2248#issuecomment-1821303733), and have added some more explanation in the PR description.

pfulton:
> This is ready for re-review. Storybook is updated and has a Chromatic-only template. I've updated the styles to support existing use of UI icons instead of Workflow icons to address my [last comment](https://github.com/adobe/spectrum-css/pull/2248#issuecomment-1821303733), and have added some more explanation in the PR description.
> 
> I'm wondering if the alignment change should be marked as a breaking change. Or if it would be better to save that part of the change for Spectrum 2 ( @pfulton )?

I'm inclined to make it a breaking change and introduce it now.

components/button/index.css
castastrophe: We can't rely on order for apply styles appropriately because a lot of processing moves things around. Just raising the question because I've seen a few components where insufficient specificity was causing styles not to be applied after processing.

components/button/package.json
```diff
@@ -15,7 +15,7 @@
   },
   "main": "dist/index-vars.css",
   "peerDependencies": {
-    "@spectrum-css/icon": ">=4",
+    "@spectrum-css/icon": ">=6",
```
castastrophe, line 18: 🥳 Thanks for capturing this requirement update!

components/button/stories/button.stories.js
```diff
@@ -82,14 +82,31 @@ export default {
...
-		}
+		},
+		showIconOnlyButton: {
+			table: {
+				disable: true,
+			},
+		},
+		layout: {
+		    name: "Layout",
+		    description: "How the buttons align in the preview (Storybook only).",
```
castastrophe, line 101: Just noting here that this will probably no longer be necessary after the storybook layout decorator merges - plus side, I might steal this approach for that PR #2420!

castastrophe (approved):

lazd:
FYI this one caught us by surprise over in Project X, we got unexpectedly wrapping buttons and misaligned icons!
image

image

image

We're disabling this wrapping behavior by default as it introduces bugs and putting it behind a `[wrap]` attribute for the fix. see PR https://git.corp.adobe.com/Horizon/hz/pull/89334

Issue #2156, opened 2023-09-08
[Tooltip] SASS mixin overriding expected output?


Westbrook:
Would you expect this value:

https://github.com/adobe/spectrum-css/blob/9205ac4c07b2b93270b08707d3f2c2b89490b370/components/tooltip/index.css#L19

Or this value: https://unpkg.com/browse/@spectrum-css/tooltip@5.1.11/dist/index-vars.css#L14

<img width="481" alt="image" src="https://github.com/adobe/spectrum-css/assets/1156657/aa32836f-12c7-43e3-889a-af05829625a7">

To be the delivered value of `--spectrum-overlay-animation-distance`?

It seems like the `@inherit: %spectrum-overlay;` in `@import "../commons/overlay-coretokens.css";` is normalizing this value in a way that is not expected.

What's more, it seems like the non-mixin value is applied in the [docs site](https://opensource.adobe.com/spectrum-css/tooltip.html#showonhover), which is even more confusing:

<img width="514" alt="image" src="https://github.com/adobe/spectrum-css/assets/1156657/fb84e693-310c-4df5-bb5a-f0a6b44ec56a">

PR #2266, opened 2023-11-06
fix(tooltip): custom animation distance overrides overlay value

mdt2:
## Description

Per issue #2156, Tooltips (all variants except the hover variant) were inheriting the wrong animation distance (6px instead of the 4px that is [noted in the spec](https://spectrum.adobe.com/page/tooltip/#Offset)). This work adjusts the animation distance for the component.

### Validation steps
- [ ] Use dev tools to inspect a tooltip. You should see the animation distance transform at the top level of the component being set with the `--spectrum-tooltip-animation-distance` token.
    - [x] Confirmed in [the docs site](https://pr-2266--spectrum-css.netlify.app/)
    - [ ] Confirmed in [storybook](https://pr-2266--spectrum-css.netlify.app/preview/)

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/tooltip/index.css
+++ b/components/tooltip/index.css
@@ -15,9 +15,6 @@ governing permissions and limitations under the License.
 .spectrum-Tooltip {
   --spectrum-tooltip-animation-duration: var(--spectrum-animation-duration-100);
 
-  /* animation distance is equal to spacing to source */
-  --spectrum-tooltip-animation-distance: var(--spectrum-spacing-75);
-
   /* override if additional spacing to source is required */
   --spectrum-tooltip-margin: 0px;
 
--- a/components/tooltip/index.css
+++ b/components/tooltip/index.css
@@ -122,6 +122,7 @@ governing permissions and limitations under the License.
   }
 
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-tooltip-animation-distance, var(--spectrum-tooltip-animation-distance));
     @inherit: %spectrum-overlay--open;
   }
 
@@ -399,6 +400,7 @@ governing permissions and limitations under the License.
 
   /* tooltip animates upward ⬆ */
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-tooltip-animation-distance, var(--spectrum-tooltip-animation-distance));
     @inherit: %spectrum-overlay--top--open;
   }
 }
@@ -414,6 +416,7 @@ governing permissions and limitations under the License.
 
   /* Tooltip animates downward ⬇ */
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-tooltip-animation-distance, var(--spectrum-tooltip-animation-distance));
     @inherit: %spectrum-overlay--bottom--open;
   }
 }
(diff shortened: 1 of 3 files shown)


components/tooltip/index.css
```diff
@@ -122,6 +122,7 @@ governing permissions and limitations under the License.
   }
 
   &.is-open {
+    --spectrum-overlay-animation-distance: var(--mod-tooltip-animation-distance, var(--spectrum-tooltip-animation-distance));
```
mdt2, line 125: I'm a bit torn on this approach. Popover and Tooltip are the only components that import `"../commons/overlay-coretokens.css"`. And I'm overriding the animation distance value for each of them (popover changes happening in #2246). On one hand, it's nice to have the overlay transform in one place to keep the code DRY...on the other hand, if each use their own values and we have to override the custom property each time, is it really all that DRY? Definitely open to opinions and thoughts on this.

mdt2: Yeah, I'm also leaning towards the direction of just defining the transforms within tooltip and popover. That change will make things more readable as well.

QuickAction appears to be inheriting from `"../overlay/index.css"` so it should be unaffected.

Westbrook (commented):
It would be useful to compare the actual output here for review, to ensure we're not seeing any values forced on us from the mixins leveraged here?

Is there benefit from using `--spectrum-overlay-animation-distance` or a similar value in the various directions rather than all of `var(--mod-tooltip-animation-distance, var(--spectrum-tooltip-animation-distance))` each time?

mdt2:
@Westbrook hello! Yes, below is the dist output on prod vs this PR. Before this PR, we were relying on some transform values in `"../commons/overlay-coretokens.css"`, which is what caused the wonkiness you were seeing in the issue you posted. You can see this in the prod output that it's falling back to the 6px instead of the `spacing-75` 4px animation distance value that we set for Tooltip.

Instead of trying to force the animation-distance to be something different than in `overlay-coretokens.css`, I've refactored the Tooltip to define it's own transforms so that we're no longer inheriting those. This eliminates the connection to `--spectrum-overlay-animation-distance` and hopefully also eliminates the confusion caused by the inheritance.

Let me know what other questions you have.

tooltip-dist

components/tooltip/index.css
```diff
@@ -122,6 +119,7 @@ governing permissions and limitations under the License.
...
```
Westbrook, line 122: Overwriting this value in the `.is-open` state means that JS application will not have access to this value while the Tooltip is closed. I can't remember right off why someone was asking about that recently, but would it be that different to move this onto the default rule?

mdt2: I either didn't send this earlier or it disappeared, so for posterity:

Good catch! We don't need this overwrite since we've [refactored the transforms](https://github.com/adobe/spectrum-css/pull/2266/commits/51f31f5a703c9dd75c283cec35ceb239bc380604) so I removed it.

Westbrook (approved):
LGTM!

components/tooltip/index.css
jawinn: The changes to refactor this look great. Nice work.

pfulton (approved):

PR #2269, opened 2023-11-06
refactor(fieldlabel): form - replace table layout with grid

jawinn:
### Use CSS grid in Form component
Refactors `.spectrum-Form` (which currently exists within the FieldLabel component) to modernize its layout. This changes the display to use CSS `grid` instead of `table`, which helps get rid of the need for negative margins and different ways of handling spacing between the two variants. 

### Mod property Changes
Deprecates some Form --mod properties:
- `--mod-tableform-border-spacing` has been renamed to `--mod-form-item-block-spacing` and has a temporary fallback for the old name.
- `--mod-tableform-item-block-spacing-labels-above` has been renamed to `--mod-form-item-block-spacing-labels-above` and has a temporary fallback for the old name.
- `--mod-tableform-margin`
- `--mod-tableform-margin-labels-above`

Additional mods have been added for modifying the size of the the grid tracks.

### Docs updates
**FieldGroup**: Removes the `.spectrum-Form` class from the Field group examples, as this currently is not being used for alignment or with the correct markup for Form (e.g. there is no .spectrumForm-item used, etc). It is not needed here  to demonstrate the Field group markup and CSS, which has its own flex alignment used.

**Form**: General cleanup of the example docs, including
 - Use updated component markup for Stepper
 - Use updated component markup for Picker
 - Replace duplicate IDs with unique IDs

CSS-591

### Validation steps

- [x] Form in the docs site still has the same spacing and layout as production (Chrome, FF, Safari, and Edge). _There will be a slight improvement in the appearance of Stepper, as its markup was previously outdated._
- [x] Form in Storybook still has the same spacing.
- [x] Confirm that Field group docs examples look the same as production.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/fieldgroup/metadata/fieldgroup.yml
+++ b/components/fieldgroup/metadata/fieldgroup.yml
@@ -25,8 +25,7 @@ examples:
       <div class="spectrum-Examples">
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Radio</h4>
-
-          <form class="spectrum-Form">
+          <form>
             <div class="spectrum-FieldGroup" role="radiogroup" aria-labelledby="radiogroup-label-1">
               <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM" id="radiogroup-label-1">Radio Group Label</div>
 
@@ -53,8 +52,7 @@ examples:
 
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Radio</h4>
-
-          <form class="spectrum-Form">
+          <form>
             <div class="spectrum-FieldGroup spectrum-FieldGroup--toplabel spectrum-FieldGroup--vertical" role="radiogroup" aria-labelledby="radiogroup-label-1">
               <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM" id="radiogroup-label-1">Radio Group Label</div>
 
@@ -81,8 +79,7 @@ examples:
 
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Checkbox</h4>
-
-          <form class="spectrum-Form">
+          <form>
             <div class="spectrum-FieldGroup spectrum-FieldGroup--toplabel spectrum-FieldGroup--vertical" role="group" aria-labelledby="checkboxgroup-label-1">
               <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM" id="checkboxgroup-label-1">Checkbox Group Label</div>
 
@@ -124,8 +121,7 @@ examples:
       <div class="spectrum-Examples">
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Radio</h4>
-
-          <form class="spectrum-Form">
+          <form>
             <div class="spectrum-FieldGroup spectrum-FieldGroup--toplabel spectrum-FieldGroup--horizontal" role="radiogroup" aria-labelledby="radiogroup-label-2">
               <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM" id="radiogroup-label-2">Radio Group Label</div>
 
@@ -152,8 +148,7 @@ examples:
 
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Checkbox</h4>
-
-          <form class="spectrum-Form">
+          <form>
             <div class="spectrum-FieldGroup spectrum-FieldGroup--toplabel spectrum-FieldGroup--horizontal" role="group" aria-labelledby="checkboxgroup-label-2">
               <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM" id="checkboxgroup-label-2">Checkbox Group Label</div>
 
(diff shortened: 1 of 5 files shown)


components/fieldlabel/index.css
```diff
@@ -108,6 +104,7 @@ governing permissions and limitations under the License.
 .spectrum-FieldLabel--left,
 .spectrum-FieldLabel--right {
   display: inline-block;
+  vertical-align: top;
```
jawinn, line 107: Note: this is moved from the main `.spectrum-FieldLabel` class because it only applies with inline-block or previously in a table cell; on the main class it does not have an affect—this removes the postcss warning `Property is ignored due to the display. With 'display: block', vertical-align should not be used.`

components/fieldgroup/metadata/fieldgroup.yml
```diff
@@ -25,8 +25,7 @@ examples:
       <div class="spectrum-Examples">
         <div class="spectrum-Examples-item">
           <h4 class="spectrum-Heading spectrum-Heading--sizeXS spectrum-Examples-itemHeading">Radio</h4>
-
-          <form class="spectrum-Form">
+          <form>
```
mdt2, line 28: Good catch on this.

components/fieldlabel/index.css
```diff
@@ -116,54 +113,57 @@ governing permissions and limitations under the License.
...
-  border-spacing: var(--mod-tableform-border-spacing, var(--spectrum-tableform-border-spacing));
-  margin: var(--mod-tableform-margin, var(--spectrum-tableform-margin));
+  margin: 0;
+  display: grid;
+  grid-template-columns: var(--mod-form-grid-template-columns, auto auto);
+  inline-size: var(--mod-form-inline-size, fit-content);
+  justify-content: start;
+
+  /* @deprecation --mod-tableform-item-block-spacing has been renamed to
+     --mod-form-item-block-spacing. The fallback will be removed in a future version. */
```
mdt2, line 127: A helpful comment. Is this how we usually let consumers know about deprecations?

components/fieldlabel/metadata/form.yml
```diff
@@ -7,25 +7,25 @@ examples:
     markup: |
       <form class="spectrum-Form">
         <div class="spectrum-Form-item">
-          <label for="fieldLabelExample-lifestory" class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Form-itemLabel spectrum-FieldLabel--left">Company Title</label>
+          <label for="fieldLabelExample-company-1" class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Form-itemLabel spectrum-FieldLabel--left">Company Title</label>
           <div class="spectrum-Form-itemField">
             <div class="spectrum-Textfield spectrum-Textfield--multiline">
-              <textarea id="fieldLabelExample-lifestory" placeholder="Enter your life story" name="field" class="spectrum-Textfield-input"></textarea>
```
mdt2, line 13: Is there any risk in removing the placeholders?

components/fieldlabel/metadata/form.yml
```diff
@@ -7,25 +7,25 @@ examples:
...
           <div class="spectrum-Form-itemField">
             <div class="spectrum-Textfield">
-              <input class="spectrum-Textfield-input" aria-invalid="false" type="text" placeholder="Enter your e-mail address" id="fieldLabelExample-emailaddress">
+              <input class="spectrum-Textfield-input" aria-invalid="false" type="text" id="fieldLabelExample-emailaddress-1">
             </div>
           </div>
         </div>
         <div class="spectrum-Form-item">
-          <label class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Form-itemLabel spectrum-FieldLabel--left" for="spectrum-textinput-instance">Country</label>
+          <div class="spectrum-FieldLabel spectrum-FieldLabel--sizeM spectrum-Form-itemLabel spectrum-FieldLabel--left">Country</div>
```
mdt2, line 26: Wondering if this should still be `label` and if we need a `for` and `id` pair on these for the example to be accessible?

mdt2 (commented):
Looking good in the docs site and Storybook. Just a few questions.

components/fieldlabel/metadata/form.yml
jawinn: Not that I see. If we need to capture in VRTs, it should be within Textfield. In general they were removed previously as they're [not intended to be used](https://spectrum.adobe.com/page/text-field/#Don%E2%80%99t-use-placeholder-text) or demonstrated.

components/fieldlabel/index.css
jawinn: Yes, I'm following this suggestion as was mentioned in a Slack discussion. To add to the fallback stack, include a comment about deprecation, and a note in the commit message. Based on [an linked example from Dropzone](https://github.com/adobe/spectrum-css/blob/a0cf37bf5fa1093ab6fd6ae8244c9b93352907aa/components/dropzone/index.css#L117).

components/fieldlabel/metadata/form.yml
jawinn: There's no actual form field in the picker (though it feels like there should be a hidden select there!), and the label element `for` is [only for certain "labelable" elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#for) , which is why I changed this after seeing a warning, to match the picker docs which use a div.

mdt2: Ahh, gotcha. Thanks for explaining!

jawinn: 👍 It does look like the `button` is considered "labelable" by `for` on `label`; on the country fields I've set this back to labels and have connected the two with IDs. I had based this on the picker docs markup, but maybe those should change as well!

mdt2 (approved):
LGTM! Good work Josh.

components/fieldlabel/index.css
```diff
@@ -116,54 +113,57 @@ governing permissions and limitations under the License.
...
+  row-gap: var(--mod-form-item-block-spacing, var(--mod-tableform-item-block-spacing, var(--spectrum-tableform-item-block-spacing)));
 }
 
+/* Row */
 .spectrum-Form-item {
-  display: table-row;
+  display: contents;
+}
+
+@supports(grid-template-columns: subgrid){
```
pfulton, line 136: 😍

pfulton (approved):
Looks great. Thanks for taking care of this!

Issue #2235, opened 2023-10-27
[Stepper]: When "Express" in-field buttons in a Stepper expect no border


Westbrook:
## Description
See this code: https://github.com/adobe/spectrum-css/blob/main/components/stepper/themes/express.css#L15

This adds a fatter border to the Express output rather than a thinner one, as seen in the design specs for this pattern

## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/stepper.html
2. Switch the "Vars Version" to "Express"
3. Observe the far border on the in-field buttons within the Stepper.

## Expected behavior
Express should exhibit no border on these elments.

## Screenshots

Current:
<img width="161" alt="image" src="https://github.com/adobe/spectrum-css/assets/1156657/e8c4404c-44e9-4532-b04f-b5b45d256d45">

Expected:
<img width="185" alt="image" src="https://github.com/adobe/spectrum-css/assets/1156657/22075cfe-1e00-4cef-82b8-2036c3ece815">

pfulton:
Tracking this in Jira at: [CSS-625](https://jira.corp.adobe.com/browse/CSS-625).

PR #2272, opened 2023-11-07
fix(stepper): express theme button border

mdt2:
## Description
Per issue #2235, in the Express theme the infield buttons used in stepper shouldn't have a border. This work adjusts the theming variables so that Express theme [matches the spec](https://www.figma.com/file/nchxla8gD4CxWYwt9eiRrz/Token-definition-library?type=design&node-id=814-8857&mode=design&t=qK8REyzZc27iySls-0), and Spectrum theme work remains unchanged since it already matches the spec.

### Validation steps

Note that if you refresh the page, the theme will automatically reset to Spectrum.

- [ ] Open the Stepper [in the docs site](). It will be handy to also have [the prod docs site](https://opensource.adobe.com/spectrum-css/stepper.html) open for comparison.
    - [ ] Spectrum theme hasn't changed (including hover, focus, keyboard focus)
    - [ ] Express theme has been updated so that infield buttons don't have a border on any variant
    - [ ] Express theme hover, focus, and keyboard focus states still work (they should be the same as prod just without the border around the buttons)

### Regression testing

Validate:

## To-do list


jawinn (changes requested):
Nice work on this. The Express version is looking more like the spec now.

I did notice one small thing on the Large and Extra Large sizes on the docs. The buttons are a different height, which I think can be fixed by removing the transparent top border on the bottom button's `. spectrum-InfieldButton-fill`? I tried it and it seems to improve the spacing around the buttons on those two sizes and make the heights the same.
Screenshot 2023-11-15 at 4 14 53 PM

This is also a problem in prod, but the Express version in high contrast mode is a little weird in places. I don't know what the best fix might be as there isn't a design to go off of. The disabled state border in cuts off a little, and the small size buttons get a little harder to read. There's also the gap / missing border between the buttons. It almost makes me wonder if WHCM for Express should show like WHCM Spectrum, but with the 2px border.

mdt2:
> Nice work on this. The Express version is looking more like the spec now.
> 
> I did notice one small thing on the Large and Extra Large sizes on the docs. The buttons are a different height, which I think can be fixed by removing the transparent top border on the bottom button's `.spectrum-InfieldButton-fill`? I tried it and it seems to improve the spacing around the buttons on those two sizes and make the heights the same. Screenshot 2023-11-15 at 4 14 53 PM
> 
> This is also a problem in prod, but the Express version in high contrast mode is a little weird in places. I don't know what the best fix might be as there isn't a design to go off of. The disabled state border cuts off a little, and the small size buttons get a little harder to read. There's also the gap / missing border between the buttons. It makes me wonder if WHCM for Express should show like WHCM Spectrum, but with the 2px border.

Great catches, thanks @jawinn! You should see the height difference in L and XL fixed in #2300. It does look like there's still some weirdness on S and M sizes there though, so I'll plan to address that as part of PR 2300.

As far as WHCM for Express, I'll make a card for this since it sounds like there might need to be some convo with the design team.

jawinn (approved):

pfulton (approved):

PR #2275, opened 2023-11-09
fix(tabs): adds rules for not showing a focus indicator on tap

jnjosh:
## Description

Exploring options to fix a bug on tabs where 

TBD

### Regression testing

TBD

## Screenshots

Screenshot 2023-11-08 at 10 51 16 PM

Note the lack of a focus indicator

## To-do list

### DIFF
--- a/components/menu/metadata/menu.yml
+++ b/components/menu/metadata/menu.yml
@@ -1344,7 +1344,7 @@ examples:
       The back arrow size scale used with the various menu sizes are small: `200`, medium: `300`, large: `400`, and extra large: `500`."
     markup: |
       <div class="spectrum-Tray-wrapper spectrum-CSSExample-dialog" style="background: rgba(0,0,0,0.4);">
-        <div class="spectrum-Modal spectrum-Tray is-open spectrum-Tray">
+        <div class="spectrum-Modal spectrum-Tray is-open">
           <ul class="spectrum-Menu is-selectableMultiple" role="menu" style="--mod-menu-inline-size: 100%;">
             <li role="presentation">
               <div class="spectrum-Menu-back">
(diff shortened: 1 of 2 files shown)


jnjosh:
Part of this issue was also to audit for other focus-indicator-on-click issues. After a very quick review these seem at least worth investigation or design discussion:

* **Search Within** - Shows a focus indicator around the drop down when clicking into the text field.
searchwithin

* **Floating Action Button** - Shows an indicator when clicking the button.
floatingactionbutton

* **Drop Zone** - Shows focus indicator when clicking into the zone.
dropzone

* **Breadcrumbs** - Seem to have default with a focus indicator?
breadcrumbs

FYI @pfulton

components/tabs/index.css
```diff
@@ -408,6 +408,12 @@ governing permissions and limitations under the License.
 		}
 	}
 
+	&:focus:not(:focus-visible) {
+		&::before {
+			border-color: transparent;
+		}
+	}  
+
 	&:focus,
```
Westbrook, line 417: I _think_ you can achieve the same by simply removing this line, instead?

jnjosh: I tried that but it didn't seem to work as I expected. I may have done something incorrect though, so I'll try again!

Rajdeepc: chrome leaves an ```outline``` property maybe you can try with that.

jawinn: Yeah, removing both instances of `:focus` within the CSS should be the way to go here. `focus-visible` will not trigger on click of the divs that have `tabindex="0"`.

jnjosh:
@Rajdeepc I think we have a duplicate, closing in favor of #2306

PR #2299, opened 2023-11-13
fix(menu)!: reimplement text truncating

jenndiaz:
## Description

BREAKING CHANGE

"Previously the spectrum-Menu-itemLabel--wrapping class could be added to the spectrum-Menu-itemLabel element in a Menu Item to prevent it from wrapping and force it to use an ellipsis instead."

Fix: 
- the css to truncate was not able to be applied because `spectrum-Menu-itemLabel` had `display: flex;`
[read more about flex causing issues with truncate](https://css-tricks.com/flexbox-truncated-text/)
- I was able to remove this by changing the mark up of the Multi Selection with icons example which had the SVG with in the label 
- Rename classes to indicate truncating and not wrapping
- removed any html that was making it appear that the text with truncating `...`
 
Docs: 
- added example to docs site 
- added example to Storybook 
- added control to Storybook to toggle on all variants

### Validation steps

- [ ] truncating example visible on docs page 
- [ ] storybook has truncate story with each menu item over 150px truncates and displays `...`
- [ ] each story for menu has a truncate control which adds a max width to the menu and adds `spectrum-Menu-itemLabel--truncate` to each menu item or heading

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -537,8 +537,6 @@ governing permissions and limitations under the License.
 
 .spectrum-Menu-itemLabel {
   grid-area: labelArea;
-  display: flex;
-  align-items: flex-start;
   font-size: var(--mod-menu-item-label-font-size, var(--spectrum-menu-item-label-font-size));
   color: var(--highcontrast-menu-item-color-default, var(--mod-menu-item-label-content-color-default, var(--spectrum-menu-item-label-content-color-default)));
 }
@@ -641,7 +639,6 @@ governing permissions and limitations under the License.
 }
 
 .spectrum-Menu-itemLabel--truncate {
-  display: block;
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -537,7 +537,6 @@ governing permissions and limitations under the License.
   grid-area: iconArea;
 }
 
-
 .spectrum-Menu-checkmark {
   grid-area: checkmarkArea;
   align-self: start;
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -377,7 +377,7 @@ governing permissions and limitations under the License.
   display: grid;
   /* stylelint-disable declaration-block-no-redundant-longhand-properties */
   grid-template-areas:
-    ".            chevronAreaCollapsible .             iconArea sectionHeadingArea  .         .           .                 "
+    ".            chevronAreaCollapsible .             headingIconArea  sectionHeadingArea  .         .           .                 "
     "selectedArea chevronAreaCollapsible checkmarkArea iconArea labelArea        valueArea actionsArea chevronAreaDrillIn"
     ".            .                      .             .        descriptionArea  .         .           .                 "
     ".            .                      .             .        submenuArea      .         .           .                 ";
@@ -525,6 +525,18 @@ governing permissions and limitations under the License.
   grid-area: iconArea;
   align-self: start;
 }
+
+.spectrum-Menu-item--collapsible .spectrum-Menu-itemIcon {
+  grid-area: headingIconArea;
+}
+
+.is-selectableMultiple .spectrum-Menu-item {
+  align-items: start;
+}
+.is-selectableMultiple .spectrum-Menu-itemCheckbox {
+  grid-area: iconArea;
+}
+
 
 .spectrum-Menu-checkmark {
   grid-area: checkmarkArea;
--- a/components/menu/index.css
+++ b/components/menu/index.css
@@ -16,7 +16,7 @@ governing permissions and limitations under the License.
   --spectrum-menu-item-icon-width: var(--spectrum-workflow-icon-size-100);
   --spectrum-menu-item-label-font-size: var(--spectrum-font-size-100);
   --spectrum-menu-item-label-text-to-visual: var(--spectrum-text-to-visual-100);
-  
+
   --spectrum-menu-item-label-inline-edge-to-content: var(--spectrum-component-edge-to-text-100);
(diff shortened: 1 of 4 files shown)


components/menu/index.css
```diff
@@ -369,11 +369,11 @@ governing permissions and limitations under the License.
   display: grid;
   /* stylelint-disable declaration-block-no-redundant-longhand-properties */
   grid-template-areas:
-    ".            chevronAreaCollapsible .             iconArea sectionHeadingArea  .         .           .                 "
+    ".            chevronAreaCollapsible .             headingIconArea  sectionHeadingArea  .         .           .                 "
```
jenndiaz, line 372: this previously had two `iconAreas` which causes some issues with menu item icons and labels not being on the correct row with in the grid.

components/menu/index.css
```diff
@@ -528,8 +540,6 @@ governing permissions and limitations under the License.
 
 .spectrum-Menu-itemLabel {
   grid-area: labelArea;
-  display: flex;
-  align-items: flex-start;
```
jenndiaz, line 532: Flex was causing `  text-overflow: ellipsis;` to not work, was able to refactor the itemLabel to not need to rely on flex. This is what appeared to of broken the truncating functionality.

mdt2 (commented):
This is looking good! A couple notes:

- On line 1333 in `components/menu/metadata/menu.yml` we have one more html `...` that can be removed.
- The Storybook control doesn't do anything except for on the Truncate page, I'm wondering if we should either limit the control to only show on that Truncate page or maybe set one of the menu items on the other stories to be able to show truncating? Let me know what you think!

jenndiaz:
> * The Storybook control doesn't do anything except for on the Truncate page, I'm wondering if we should either limit the control to only show on that Truncate page or maybe set one of the menu items on the other stories to be able to show truncating? Let me know what you think!

@mdt2 Could you say more about this? It is defiantly working for me, although some are more subtle, If you adjust the `Max Inline Size*` it makes it more obvious. I have it setting to 150px, I could adjust that if we think it's needed. 

I do think it's helpful to have the toggle on all of them, about half of them were not working previously and I think having the ability to see it in all variants could prevent it from regressing again.

mdt2:
Oh my mistake! I had only tested a few that weren't long enough for truncation. Thanks for pointing me toward the `Max Inline Size`, it helped me realize my error quickly.

mdt2 (approved):
Great work Jenn!

components/menu/stories/template.js
```diff
@@ -158,7 +163,8 @@ export const MenuGroup = ({
       role="presentation">
       ${heading
         ? html`<span
-            class="spectrum-Menu-sectionHeading"
+            class="spectrum-Menu-sectionHeading ${shouldTruncate ? "spectrum-Menu-itemLabel--truncate" : "" }"
+            style=${maxInlineSize ? `max-inline-size: ${maxInlineSize}px;` : ""}
```
jawinn, line 167: Could the `px` be removed and added to the values, so this is more open and could be set to other valid widths like a percentage?

components/menu/stories/menu.stories.js
```diff
@@ -23,6 +23,25 @@ export default {
...
+    maxInlineSize: {
+			name: "Max Inline Size",
+			type: { name: "text", required: true },
+			table: {
+				type: { summary: "text" },
+				category: "Component",
+			},
+			control: "number",
+      if: { arg: "shouldTruncate", truthy: true },
+		},
```
jawinn, line 44: A little formatting adjustment to indentations is needed around this section

components/menu/metadata/menu.yml
```diff
@@ -1330,7 +1330,29 @@ examples:
...
+            </li>
+            <li class="spectrum-Menu-item" role="menuitem" tabindex="0">
+              <span class="spectrum-Menu-itemLabel">Save Selection</span>
+            </li>
+          </ul>
+        </div>
+      </div>
+  - id: menu-truncate
+    name: Menu with truncating text
+    description: Use the class `spectrum-Menu-itemLabel--truncate` on the item label or section heading that should truncate with in a menu with a set `inline-size` or `max-inline-size`
```
jawinn, line 1343:
```suggestion
    description: Use the class `spectrum-Menu-itemLabel--truncate` on the item label or section heading that should truncate within a menu. When text would typically overflow beyond the available horizontal space and wrap (default behavior), ellipsis will appear instead. This is demonstrated here by setting an `inline-size` on the menu.
```

components/menu/metadata/menu.yml
```diff
@@ -1279,10 +1279,10 @@ examples:
                 </div>
               </li>
               <li class="spectrum-Menu-item" role="menuitem" tabindex="0">
+                <svg class="spectrum-Icon spectrum-Icon--sizeM spectrum-Menu-itemIcon spectrum-Menu-itemIcon--workflowIcon" focusable="false" aria-hidden="true" aria-label="Select Add Icon">
+                  <use xlink:href="#spectrum-icon-18-SelectAdd"></use>
+                </svg>
                 <label class="spectrum-Switch-label spectrum-Menu-itemLabel" for="switch-onoff-4">
```
jawinn, line 1285: Should this change in markup be noted in the migration guide?

components/menu/stories/template.js
jenndiaz: yes, I switch to not using a number input so I could make it more flexible. 
In looking at this I also caught that my changes had gotten rid of some custom styling on other stories, addressed this too.

components/menu/metadata/menu.yml
jenndiaz: I've added to the migration guide and update the PR description to reflect breaking change.

PR #2302, opened 2023-11-14
background layers "open props" approach

jenndiaz:
## Description

Experimental PR to look at implementing Background Layers 

This approach is using an [open props](https://open-props.style/) inspired approach, where specific tokens are accessible by adding `@import "../../tools/background-layer-props.css";` with in the css file and using the custom properties defined with in `../../tools/background-layer-props.css` 

changes: 
- added a file in `/tools` which defines the custom properties for implementing background layers
- within the CSS file for where these custom properties are needed, added @import and the file path 
- use the defined custom properties within the CSS for the component where you want to use them 

One con from a development stand point, I needed to rebuild after each change to the `background-layer-props.css` 

Look at the [docs site for Alert Dialog ](https://pr-2302--spectrum-css.netlify.app/alertdialog) to see the elevated custom property (`var(--bg-color-elevated)` used with in `spectrum-Modal` 

[Storybook ](https://pr-2302--spectrum-css.netlify.app/preview/?path=/docs/importprops-background-layers--docs)

## Screenshots

Screenshot 2023-11-16 at 11 00 04 AM

Screenshot 2023-11-16 at 11 00 22 AM

### DIFF
--- a/tools/background-layer-props.css
+++ b/tools/background-layer-props.css
@@ -1,0 +1,16 @@
+* {
+  --bg-color-elevated: var(--spectrum-gray-25);
+  --bg-color-layer-2: var(--spectrum-gray-25);
+  --bg-color-layer-1: var(--spectrum-gray-50);
+  --bg-color-pasteboard: var(--spectrum-gray-100);
+  --bg-color-base: var(--spectrum-gray-25);
+
+  --drop-shadow-color-1: var(--spectrum-gray-600);
+
+  --drop-shadow-horiztonal: 0;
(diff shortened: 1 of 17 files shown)


tools/background-layer-props.css
jawinn, line 1: To me, this begs the question of whether these could all just be stored in the global tokens package, rather than needing to be in an imported file. Or am I missing something here?

(also these should probably live under `:root` or `html` rather than applying them to every element with `*`. Interestingly, I peaked at Open Props and they put them in `:where(html)`.)

PR #2322, opened 2023-11-28
fix: run mod-extractor on dist and support for multiple files

jawinn:
## Description

An update for the`yarn docs:mod` utility:
If there were multiple CSS files in the root, such as with the icon component and commons, only the mods from the last file found were being included. Basically the mods.md and mods.json were being overwritten each time a CSS file was processed. During migration changes to Icon (CSS-511), some mods were being left out of the metadata when the utility was run. This fixes that issue.

The utility now bundles all parsed mods per component directory into an array before exporting them. It also splits the utility into two functions `collectMods` and `exportMods` as it was quite large and the functionality was already very independent. This also makes it easier to add unit tests in the future.

_Per PR discussions, this also now runs the utility on dist files instead of source files to make sure all mods are being included, such as those coming in through inherits or extends. It also improves console logging output (see screenshot), and allows the utility to be run on a single component._

### Validation steps
- [x] Pull down the branch locally and run `docs:mod`. There should not be any changes.
- [x] Delete all markdown and json files locally `rm -R ./components/*/metadata/mods.md && rm -R ./components/*/dist/mods.json`, see that files have been deleted and run `docs:mod` again. All files should be regenerated and there are no changes except for searchwithin mods.md being deleted because it has no mods in its CSS.
- [x] Add the following CSS to the icon component's **workflow-icons.css** file .spectrum-Icon `color: var(--mod-test);` and run `docs:mod`. The mods.md for icon should show both `--mod-test` and `--mod-icon-color`

### Regression testing

Validate:

## Screenshots

Screenshot 2023-12-07 at 6 49 05 PM

## To-do list

### DIFF
--- a/components/commons/metadata/mods.md
+++ b/components/commons/metadata/mods.md
@@ -1,4 +1,8 @@
 | Modifiable Custom Properties              |
 | ----------------------------------------- |
+| `--mod-animation-duration-100`            |
+| `--mod-focus-indicator-gap`               |
+| `--mod-line-height-100`                   |
 | `--mod-overlay-animation-distance`        |
 | `--mod-overlay-animation-duration-opened` |
+| `--mod-sans-font-family-stack`            |
--- a/tasks/mod-extractor.js
+++ b/tasks/mod-extractor.js
@@ -14,8 +14,16 @@ const { writeFile, readFile } = require("fs").promises;
 const path = require("path");
 const fg = require("fast-glob");
 
-async function main(dir = path.join(__dirname, "../components")) {
-	/* Loop over the directories in the components folder and find all the first-level css files */
+/**
+ * Parse first-level CSS files to extract "--mod" custom properties.
+ *
+ * @param {string} dir Parent components directory. 
+ * @returns Array of objects containing the component directory and the list of mod names.
+ */
+async function collectMods(dir = path.join(__dirname, "../components")) {
+	let modsPerPath = [];
+
+	// Loop over the directories in the components folder and find all the first-level CSS files.
 	for (const filepath of await fg("*/*.css", {
 		cwd: dir,
 		absolute: true,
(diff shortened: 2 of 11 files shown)


components/commons/metadata/mods.md
```diff
@@ -1,4 +1,8 @@
 | Modifiable Custom Properties              |
 | ----------------------------------------- |
+| `--mod-animation-duration-100`            |
+| `--mod-focus-indicator-gap`               |
+| `--mod-line-height-100`                   |
 | `--mod-overlay-animation-distance`        |
 | `--mod-overlay-animation-duration-opened` |
+| `--mod-sans-font-family-stack`            |
```
jawinn, line 8: This uncovers the fact that these previously unlisted mod names within `commons` may need to be renamed (not as part of this PR).

mdt2: Tell me more?

tasks/mod-extractor.js
```diff
@@ -37,53 +45,88 @@ async function main(dir = path.join(__dirname, "../components")) {
...
+	await writeFile(
+		`${destPath}/mods.md`,
+		formattedResults.join("\n"),
+		(err) => {
+			if (err) throw err;
+		}
+	);
+
+	/* -- JSON Output -- */
+	destPath = `${directory}/dist`;
```
mdt2, line 110: This is a nice simplification

mdt2 (approved):
Code looks good and is very readable, and everything's working as described locally 👍

components/commons/metadata/mods.md
jawinn: These are in the _basebutton_ file I think. And the mod names are referencing the global token name which is generally something we have been avoiding.

castastrophe: Oh yes, there's also nothing surfacing this information to consumers at this time. I wonder if we should be grepping the *compiled* css instead of src assets...

jawinn: > Oh yes, there's also nothing surfacing this information to consumers at this time. I wonder if we should be grepping the _compiled_ css instead of src assets...

Probably a good idea too. Currently, anything that uses an `@inherit` of some of those % prefixed "placeholder classes" may not have those mods listed. For example, modal should probably be showing `--mod-overlay-animation-duration-opened`.

jawinn: I've updated this to run on the _dist/index.css_ files with the latest commit.

castastrophe (changes requested):
This looks great! Thanks for tackling this. I have just two general thoughts, perhaps we need some logging to report status and what is happening in the script and I'd like to see more error capturing.

tasks/mod-extractor.js
```diff
@@ -14,15 +14,23 @@ const { writeFile, readFile } = require("fs").promises;
...
+ *
+ * @param {string} dir Parent components directory. 
+ * @returns Array of objects containing the component directory and the list of mod names.
+ */
+async function collectMods(dir = path.join(__dirname, "../components")) {
+	let modsPerPath = [];
+
+	// Loop over the directories in the components folder and find the built
+	// CSS file(s) that might contain mods.
+	for (const filepath of await fg("*/dist/index.css", {
```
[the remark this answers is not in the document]
jawinn: Please take a look at the latest commit and see what you think! I changed it so it runs on each component at a time, and gives better logging output (screenshot added to PR description). Good call; the logging output actually turned up that I needed to parse different files for a few exceptions (commons and overlay) that don't have a dist/index.css.

castastrophe:
@jawinn Looks great!

jawinn:
Closing this as the bulk of its changes are already covered in the recent refactoring that is in main.

PR #2324, opened 2023-11-30
fix: docs watch unable to find package.json

mdt2:
## Description

Fixes the below described bug:

When `yarn dev` was already running and changes were made to any component .yml file, we would get a console error: `'buildDocs_forDep' errored after 9.74 ms. Error: Cannot find module '...component-name/package.json'`. The issue was that on initial docs site build, `buildDocs_forDep` was receiving the correct path `@spectrum-css/component-name/package.json`, but on handleFileChanged, the path would come back without `@spectrum-css`, causing the error because it couldn't find the package.json, which is one of the first things the function needs.

### Validation steps
- [ ] Pull down the branch and run `yarn dev`
- [ ] Once the docs site spins up, make an html change to a component's .yml file. The change shouldn't cause an error and the docs site should be updated, either by hot reload or manual reload (you might see this change hot reload instead of having to manually reload, I've seen it work intermittently. We've always had to manually reload on html change before so it's fine if that's still the case)
- [ ] Now make an easily visible change to a component index.css. The change should be applied to the docs site and it should hot reload

### Regression testing

Validate:

## To-do list

### DIFF
--- a/tools/bundle-builder/dev/index.js
+++ b/tools/bundle-builder/dev/index.js
@@ -191,7 +191,8 @@ function watch() {
 					},
 					function buildDocs_forDep() {
 						logger.debug(`Building docs for ${package}`);
-						return docs.buildDocs_forDep(package);
+						let packageDir = path.dirname(require.resolve(`@spectrum-css/${package}/package.json`));
+						return docs.buildDocs_forDep(packageDir);
 					}
 				)();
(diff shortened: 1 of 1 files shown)


pfulton (approved):
Pulled it down locally and ran through the steps. Works as you described! Thanks for fixing!

PR #2338, opened 2023-12-05
chore(build): support tagged component builds

castastrophe:
## Description

This PR focuses on updating the dependencies for all component packages and migrates the build scripts into a separate project.json for each component. The build is still leveraging gulp but it's firing it from the root of the package, thus removing the need for a local component dependency on gulp and making it easier to update the build in a central location.

The mod-extractor was updated to run on an individual package rather than the entire components folder so that it could be chained with the build task.

No regressions for the following commands @mdt2:

- [x] `yarn dev`
- [x] `yarn start`
- [x] `yarn build:preview`
- [x] `yarn build`
  - my understanding is that this replaces `yarn build:components` - mdt2
- [x] `yarn clean`
  - my understanding is that this replaces `yarn clean:components` - mdt2

## To-do list


package.json
```diff
@@ -56,11 +52,11 @@
     "release:beta-from-package": "yarn release from-package --conventional-prerelease --preid beta --pre-dist-tag beta --no-private",
     "prerelease:site": "yarn prepare:site",
     "release:site": "gh-pages -d dist/ -f -e .",
-    "prestart": "yarn build:components",
+    "prestart": "yarn build",
     "start": "NODE_ENV=development yarn workspace @spectrum-css/preview start",
```
castastrophe, line 56:
```suggestion
    "prestart": "yarn build",
    "start": "NODE_ENV=development yarn workspace @spectrum-css/preview start",
```
```suggestion
    "start": "NODE_ENV=development nx start @spectrum-css/preview",
```

castastrophe:
Debugging the vrt regressions in the most recent build.

components/commons/package.json
```diff
@@ -13,7 +13,7 @@
   "bugs": {
     "url": "https://github.com/adobe/spectrum-css/issues"
   },
-  "main": "dist/index-vars.css",
+  "main": "index.css",
```
mdt2, line 16: Can you tell me about this change?

components/logicbutton/metadata/mods.md
```diff
@@ -1,5 +1,8 @@
 | Modifiable Custom Properties                             |
 | -------------------------------------------------------- |
+| `--mod-animation-duration-100`                           |
```
mdt2, line 3: My assumption is that we're seeing changes to the mods file for some of these because these may have been accidentally excluded before, is that right? Should these be included?

components/menu/package.json
```diff
@@ -14,27 +14,27 @@
...
+    },
+    "@spectrum-css/divider": {
+      "optional": true
+    },
+    "@spectrum-css/switch": {
+      "optional": true
+    },
+    "@spectrum-css/tray": {
+      "optional": true
+    }
```
mdt2, line 37: What makes these optional and `@spectrum-css/icon` not optional?

mdt2 (commented):
The commands are running well. Mostly just have some context questions for now 🙂

pfulton (approved):
I was able to pull this down and run through the verification steps. Big improvement on being able to run the different commands without first having to run `yarn cache:clean`!

components/actionbutton/metadata/mods.md
```diff
@@ -47,4 +47,8 @@
 | `--mod-actionbutton-icon-size`                                    |
 | `--mod-actionbutton-label-color`                                  |
 | `--mod-actionbutton-min-width`                                    |
+| `--mod-actionbutton-static-content-color`                         |
 | `--mod-actionbutton-text-to-visual`                               |
+| `--mod-animation-duration-100`                                    |
+| `--mod-line-height-100`                                           |
+| `--mod-sans-font-family-stack`                                    |
```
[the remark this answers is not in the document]
castastrophe: Yeah, I didn't want to mess too much with the script because I knew Josh's was going to merge soon but I needed to update it enough to run on individual packages so it could be a part of the build flow.

components/actionbutton/package.json
```diff
@@ -14,19 +14,17 @@
...
-  },
   "peerDependencies": {
     "@spectrum-css/icon": ">=4",
     "@spectrum-css/tokens": ">=13"
   },
+  "peerDependenciesMeta": {
+    "@spectrum-css/icon": {
+      "optional": true
+    }
+  },
```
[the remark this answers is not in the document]
castastrophe: Since none of our packages include `dependencies`, none of these changes will impact them downstream. All our packages list `devDependencies` (those required for the build process) and `peerDependencies` (expected to be installed by the downstream consumer). We're clarifying the existing `peerDependencies` in this PR to highlight elements that are optional to rendering the component so that downstream users not using those variations don't get unnecessary warnings for missing peer dependencies.

components/actiongroup/package.json
```diff
@@ -14,20 +14,14 @@
...
-    "clean": "rimraf dist"
-  },
   "peerDependencies": {
     "@spectrum-css/actionbutton": ">=5",
     "@spectrum-css/tokens": ">=13"
   },
-  "devDependencies": {
-    "@spectrum-css/actionbutton": "^5.0.16",
-    "@spectrum-css/component-builder-simple": "^3.0.1",
-    "@spectrum-css/icon": "^4.0.5",
```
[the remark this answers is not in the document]
castastrophe: None of our packages need `devDependencies` anymore because the build work happens at root now. The only reason we had them previously was for storybook but storybook doesn't need them either in a monorepo environment.

components/commons/package.json
castastrophe: Yes, so commons doesn't actually have a compiled output because it's a pre-compilation utility with extends in it. When someone tries to access `@spectrum-css/commons` it was sending them to a non-existent package and so would always throw an error. Now if you want to use our extends in another project, you can `require("@spectrum-css/commons")` and load up all the extends we define here.

components/logicbutton/metadata/mods.md
castastrophe: These are similar to what Josh's update was seeing because we weren't running on compiled results.

components/menu/package.json
castastrophe: I think in the markup I saw it included Icon and it wasn't wrapped in an if or when so I left it off optional.  I'd defer after merge to the developer who worked on menu last to make that final call.

components/actiongroup/package.json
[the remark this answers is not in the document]
mdt2: Oh ok, and since this was never a peer dependency it doesn't need to be one now 👍

components/commons/package.json
mdt2: Cool cool, thanks for explaining 🙂

jawinn (commented):
This LGTM. I was able to run all the commands, and tested making changes to CSS and docs while dev and start were running. I did notice one warning in the console output after `yarn build:preview`: 

`(node:46827) [DEP_WEBPACK_STATS_WARNINGS_FILTER] DeprecationWarning: config.stats.warningsFilter is deprecated in favor of config.ignoreWarnings`

Issue #2139, opened 2023-09-01
Color Handle delivery sensitive to CSS inclusion order.


Westbrook:
The mark up for a Color Handle is as follows:
```html
<div class="spectrum-ColorHandle spectrum-OpacityCheckerboard" is-disabled" style="--spectrum-picked-color: rgba(255, 0, 0, 0.5); position: absolute; inset-block: 50%; inset-inline: 50%">
  <div class="spectrum-ColorHandle-inner"></div>
</div>
```
With both the `spectrum-ColorHandle` and the `spectrum-OpacityCheckerboard` classes applied to the host element the rules within those selectors will race depending on how CSS is deployed in a consuming application. Specifically, the `size` rules for `.spectrum-OpacityCheckerboard`:
```css
.spectrum-OpacityCheckerboard {
    inline-size: 100%;
    block-size: 100%;
}
```
Can overwright those of `.spectrum-ColorHandle`:
```css
.spectrum-ColorHandle {
    inline-size: var(--mod-colorhandle-size, var(--spectrum-colorhandle-size));
    block-size: var(--mod-colorhandle-size, var(--spectrum-colorhandle-size));
}
```

pfulton:
Tracking this in Jira here: https://jira.corp.adobe.com/browse/CSS-587

castastrophe:
This appears to be resolved with the latest code.

PR #2350, opened 2023-12-07
refactor(opacity checkerboard)!: refactor from component into commons

jenndiaz:
## Description

BREAKING CHANGE - would require removing a component 

Refactors Opacity Checkerboard from a component to be with in `components/commons/opacity-checkerboard.css` this addresses concerns around inclusion order overwriting css values (#2139) and simplifies opacity checkerboard. 

**Alternative solution** to the GitHub issue: move css unnecessary to implement the background pattern, specifically `inline-size`, ` block-size` from the `index.css` and add to inline styling for the component docs site and storybook. 

### Validation steps

**Docs Site**
Opacity Checkerboard should appear correctly on: 
- [ ] colorhandle 
- [ ] mods with in opacity checkerboard still work as expected (look at `--mod-opacity-checkerboard-position` with in colorhandle) 
- [ ] colorslider 
- [ ] thumbnail
- [ ] swatch 

**Storybook**
Opacity Checkerboard should appear correctly on: 
- [ ] colorhandle 
- [ ] colorslider 
- [ ] thumbnail
- [ ] swatch 

### Regression testing

Validate:

## To-do list


jenndiaz:
mod file is currently not generating mods of opacity checkerboard, this is being addressed in #2322

components/colorslider/index.css
```diff
@@ -105,7 +120,7 @@
 
 /* Visually hidden "range" input field */
 .spectrum-ColorSlider-slider {
-	opacity: 0;
+	opacity: 0%;
```
jenndiaz, line 123: style lint wants opacities in %

pfulton:
@Westbrook I'd be curious to get your thoughts on this refactor. I know this was implemented differently in SWC (not a component) vs. how we have it in our repo (a "component"). What level of churn (if any) would this cause for you all?

Westbrook (commented):
I'm not sure the right path for continued consumption of this approach in SWC without seeing the actual published code. For us to ship it as we have it needs to be staic _somewhere_ in the output, separate from the consuming patterns.

Things that I do know:
- using CSS preprocessing means that there are _multiple_ definitions of the CSS that can be shipped to users, rather than a single copy.
- we do have users that leverage the Opacity Checkerboard outside of the Spectrum elements
- the stand alone class was a useful way to support normalization across the consumption of this in both Spectrum patterns and Adobe apps and features at large.

jenndiaz:
> I'm not sure the right path for continued consumption of this approach in SWC without seeing the actual published code. For us to ship it as we have it needs to be staic _somewhere_ in the output, separate from the consuming patterns.
> 
> Things that I do know:
> 
> * using CSS preprocessing means that there are _multiple_ definitions of the CSS that can be shipped to users, rather than a single copy.
> * we do have users that leverage the Opacity Checkerboard outside of the Spectrum elements
> * the stand alone class was a useful way to support normalization across the consumption of this in both Spectrum patterns and Adobe apps and features at large.

@pfulton and @Westbrook 

I think based on the fact that people are using this outside of Spectrum elements steers me towards a more simple approach. I'll look at simplifying the CSS if I can and documenting inclusion order more clearly.

I think this could be a good option to be more of "utility class" like we are looking at with background layers.

PR #2362, opened 2023-12-12
chore(tokens): move tasks to nx workflow

castastrophe:
## Description

Migrate tokens workflow from package.json into the project.json for nx to manage.

Validate that the rendered tokens output is identical.

## To-do list

### DIFF
--- a/.github/actions/file-diff/index.js
+++ b/.github/actions/file-diff/index.js
@@ -56,8 +56,8 @@ async function run() {
 		// --------------- End evaluation  ---------------
 
 		/** Split the data by component package */
-		const COMPONENTS = splitDataByComponent(pathOutput, diffOutput);
-		const sections = makeTable(COMPONENTS);
+		const { filePath, PACKAGES } = splitDataByPackage(pathOutput, path, diffOutput);
+		const sections = makeTable(PACKAGES, filePath, path);
 
 		const overallSize = [...pathOutput.values()].reduce(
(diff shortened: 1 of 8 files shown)


.github/actions/file-diff/index.js
```diff
@@ -56,8 +56,8 @@ async function run() {
 		// --------------- End evaluation  ---------------
 
 		/** Split the data by component package */
-		const COMPONENTS = splitDataByComponent(pathOutput, diffOutput);
```
castastrophe, line 59: @pfulton Needed to update this to understand where the assets live - it was hardcoded to look for them in components and tokens and ui-icons don't live there. I also updated this to capture the "main" value from the package instead of assuming we want to compare the index.css file.

tasks/compare-compiled-output.js
```diff
@@ -44,10 +41,12 @@ env.addFilter("print", (data) => JSON.stringify(data, null, 2));
 const log = {
 	error: (err) => process.stderr.write(`${err}\n\n`),
 	write: (msg) => process.stdout.write(msg),
-	writeTable: (data, padding = 30) => {
```
castastrophe, line 47: Updated this to find the package rather than assume where it lives in the repo and to better pad the console output for longer dist paths like the tokens use.

mdt2 (approved):
This looks good!

pfulton (approved):

PR #2375, opened 2023-12-18
refactor(combobox): use sub component mod properties

jawinn:
## Description

Combobox was migrated to core tokens before Textfield's migration was finished and we started using passthrough mod properties for the sub-components. This is a broad stroke refactor of the CSS to simplify much of it by using the mods available in Textfield and Picker Button. Removing mods was avoided in this PR to avoid breaking changes.

- Refactors much of the CSS to use mod properties for its sub components Textfield and Picker Button. 
- Includes some additional selectors to fix the border styles around the picker button so they match when interacting with the textfield, which is in more alignment with the gifs shown in the Guidelines. Previously when the textfield was focused for example, the focus border color did not extend around the whole combobox because the picker button border color was not changing as well. This was hard to notice because of the subtle gray border color changes; tips for testing are included in the validation steps.
- Improves WHCM appearance of quiet _Picker Button_ to not show a border (aligns with _Picker_ WHCM quiet design):
Screenshot 2023-12-18 at 3 03 13 PM

CSS-602

### Validation steps

Overall validation
- [x] Docs variations look and act like production
- [x] Storybook variations look and act like production
- [ ] No mod properties have been lost from list of available mods

Border color fix validation

Try setting the follow mod values on the body element in your inspector, to better test how the border colors change. Note: if you try this on prod, you'll see the difference where only the left side of the combobox shows the color change.
```
--mod-combobox-border-color-default: blue;
--mod-combobox-border-color-hover: red;
--mod-combobox-border-color-focus-hover: cyan;
--mod-combobox-border-color-key-focus: seagreen;
--mod-combobox-border-color-focus: orange;
```
- [x] Hovered textfield shows red border
- [x] Focused and hovered textfield shows cyan border
- [x] Focused textfield shows orange border
- [x] Keyboard focused (tab) textfield shows green border

WHCM validation
- [x] High contrast styles have not regressed and show improvement to the quiet variation as noted in description.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/combobox/index.css
+++ b/components/combobox/index.css
@@ -36,25 +36,8 @@ governing permissions and limitations under the License.
 
   --spectrum-combobox-spacing-label-to-combobox: var(--spectrum-field-label-to-component);
 
-  --spectrum-combobox-font-family: var(--spectrum-sans-font-family-stack);
-  --spectrum-combobox-font-weight: var(--spectrum-regular-font-weight);
   --spectrum-combobox-font-style: var(--spectrum-default-font-style);
   --spectrum-combobox-line-height: var(--spectrum-line-height-100);
-  --spectrum-combobox-font-color-default: var(--spectrum-neutral-content-color-default);
-  --spectrum-combobox-font-color-hover: var(--spectrum-neutral-content-color-hover);
-  --spectrum-combobox-font-color-focus: var(--spectrum-neutral-content-color-focus);
-  --spectrum-combobox-font-color-focus-hover: var(--spectrum-neutral-content-color-focus-hover);
-  --spectrum-combobox-font-color-key-focus: var(--spectrum-neutral-content-color-key-focus);
-  --spectrum-combobox-font-color-placeholder: var(--spectrum-gray-600);
-
-  --spectrum-combobox-background-color-default: var(--spectrum-gray-50);
-  --spectrum-combobox-background-color-hover: var(--spectrum-gray-50);
-  --spectrum-combobox-background-color-focus: var(--spectrum-gray-50);
-  --spectrum-combobox-background-color-focus-hover: var(--spectrum-gray-50);
-  --spectrum-combobox-background-color-key-focus: var(--spectrum-gray-50);
-
-  --spectrum-combobox-background-color-disabled: var(--spectrum-disabled-background-color);
-  --spectrum-combobox-font-color-disabled: var(--spectrum-disabled-content-color);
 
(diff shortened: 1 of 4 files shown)


mdt2 (commented):
This is looking good to me. Express does have some differences from prod and main, here's one example where focus is different and the button sizing looks different:

combobox-express-focus

Also wrote down some buggy things that also exist in prod so they're probably out of scope for this work/would be less complicated to review in a separate PR:

- On combobox hover, picker button border is a different color than the rest
- Picker button focus state sticks after click
- Picker button might need a `cursor: pointer` to be accessible on focus (two changes is the norm, we have a background color change on focus but nothing else)
- We have a similar problem to ActionButton in Combobox where when it’s disabled, the background-color is transparent, when it should probably be some gray variant (a token probably exists for this).

mdt2 (commented):
I also wonder if we could add some Chromatic coverage using that `isChromatic()` method, either in this PR or in a followup. Might give us more confidence in fixing some of those bugs later on

jawinn:
> This is looking good to me. Express does have some differences from prod and main, here's one example where focus is different and the button sizing looks different:

This one is correct; focus also sets the color around the button as I saw in the Guidelines. This did bring to the light the issue with the Storybook pseudos add-on affecting the CSS, as this currently is displaying differently in the docs vs Storybook. The size of the button is correct now; it looks like this inadvertently also got fixed 👍 

> Also wrote down some buggy things that also exist in prod so they're probably out of scope for this work/would be less complicated to review in a separate PR:
> 
> * On combobox hover, picker button border is a different color than the rest

This is another one that appears to be related to the Storybook add-on's injected CSS. It does this in Storybook, but not on the docs site.

> * Picker button focus state sticks after click
> * Picker button might need a `cursor: pointer` to be accessible on focus (two changes is the norm, we have a background color change on focus but nothing else)

I've created Jira issue 666 for Picker Button in the backlog to address these.

> * We have a similar problem to ActionButton in Combobox where when it’s disabled, the background-color is transparent, when it should probably be some gray variant (a token probably exists for this).

I'm not seeing this one. I'm seeing `--spectrum-disabled-background-color` already being used?

> I also wonder if we could add some Chromatic coverage using that isChromatic() method, either in this PR or in a followup. Might give us more confidence in fixing some of those bugs later on

Yes, this is a good idea. I've made a followup Jira issue 667.

mdt2 (approved):
LGTM

PR #2379, opened 2023-12-18
feat(storybook): custom decorators for chromatic coverage

mdt2:
## Description

Overall, this work will increase our Chromatic coverage without increasing our number of snapshots by showing LTR and RTL, all three background colors, both platform sizes, and every available size of component to Chromatic by default.

- Implement custom decorators for background color, platform scale, and text direction
- Turn off controls when custom decorators are in use in the toolbar (because the controls don't work unless the global value in the toolbar is set to "default," so removing them to avoid confusion of which to use)
- Turn on size decorator so Chromatic tests components at each size in one snapshot

### Validation steps

**Text direction** @jenndiaz 
- [x] When `ltr` is active (the default), the component has the ltr direction
- [x] When `rtl` is active, the component have the rtl direction
- [x] When `ltr + rtl` is active, there are two components showing, one is ltr and the other is rtl

**Background color** @jenndiaz 
- [x] When `default bg color` is active, the background is Light
- [x] When `default bg color` is active, the background color can be changed in the control panel
- [x] When `stacked bg colors` is active, the background color cannot be changed in the control panel (it shouldn't even show as a control)
- [x] When `stacked bg colors` is active, the component shows three times with our three different background colors, stacked one on top of the other
- [x] When `side-by-side bg colors` is active, the background color cannot be changed in the control panel (it shouldn't even show as a control)
- [x] When `side-by-side bg colors` is active, the component shows three times with our three different background colors, side by side
- [x] If you go back to the default `default bg color`, the control should show back up in the control panel. If you adjust it, the component adjusts accordingly.

**Platform sizes** @jenndiaz 
- [x] When `medium - desktop` is active, the component shows at the medium platform size
- [x] When `medium - desktop` is active, the platform size can be changed in the control panel
- [x] When `large - mobile` is active, the component shows at the large platform size (usually a bit bigger visually)
- [x] When `large - mobile` is active, the platform size cannot be changed in the control panel (it shouldn't even show as a control)
- [x] When `medium + large` is active, two components show, with a heading of "Platform Size: Size", and one is medium and the other is large (in alignment with their heading)
- [x] When `medium + large` is active, the platform size cannot be changed in the control panel (it shouldn't even show as a control)
- [x] If you go back to the default `medium - desktop`, the control should show back up in the control panel. If you adjust it, the component adjusts accordingly.

**Testing together!**
Try some combinations of the controls:
- [ ] Set `ltr + rtl`, `side-by-side bg colors`, and `medium + large` to see what Chromatic will see
- [ ] Try all sorts of combinations. Try to break it. Adjust control panel controls and then their toolbar counterparts, etc.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.storybook/decorators/index.js
+++ b/.storybook/decorators/index.js
@@ -2,6 +2,7 @@ import { makeDecorator, useEffect } from "@storybook/preview-api";
 import isChromatic from "chromatic/isChromatic";
 import { html } from "lit";
 import { styleMap } from "lit/directives/style-map.js";
+import { isFullscreen, isPopover } from "./utilities.js";
 
 export { withContextWrapper } from "./contextsWrapper.js";
 
@@ -28,7 +29,10 @@ export const withTextDirectionWrapper = makeDecorator({
 		};
 		const showBothDirections = textDirection === DIRECTIONS.both;
 
-		if (showBothDirections) {
+		const componentIsFullscreen = isFullscreen(context);
+		const componentHasPopover = isPopover(context);
+
+		if (showBothDirections && !componentIsFullscreen && !componentHasPopover) {
 			return html`
 				<div style="margin-bottom: 4rem" dir=${DIRECTIONS.left}>
 					${StoryFn(context)}
@@ -36,6 +40,14 @@ export const withTextDirectionWrapper = makeDecorator({
 				<div dir=${DIRECTIONS.right}>
 					${StoryFn(context)}
 				</div>
+			`;
+		};
+
+		if (showBothDirections && (componentIsFullscreen || componentHasPopover)) {
+			return html`
+				<h4 class="spectrum-Detail spectrum-Detail--sizeXS spectrum-Examples-itemHeading" style="text-transform: none; background-color: var(--spectrum-negative-color-900); color: var(--spectrum-white); padding: 1rem;">
+					⚠️ Sorry, feature is unavailable for this component. Choose a single-setting option in the toolbar to view the component, or try changing the setting in the control panel.
+				</h4>
 			`;
 		};
 
@@ -106,6 +118,9 @@ export const withPlatformScaleWrapper = makeDecorator({
 	parameterName: "context",
 	wrapper: (StoryFn, context) => {
 		const { args, globals, argTypes } = context;
+		const componentIsFullscreen = isFullscreen(context);
+		const componentHasPopover = isPopover(context);
+
 		const bodyEl = document.body;
     const bodyElClasses = bodyEl.classList;
 		const mediumClass = 'spectrum--medium';
@@ -116,7 +131,7 @@ export const withPlatformScaleWrapper = makeDecorator({
 			both: 'both',
 		};
 		const defaultScale = SCALES.medium;
-		const scale = globals.platformScale || argTypes.scale || args.scale || defaultScale;
+		let scale = globals.platformScale || argTypes.scale || args.scale || defaultScale;
 		const showBothScales = scale === 'both';
 
 		const updateBodyClasses = () => {
@@ -132,7 +147,7 @@ export const withPlatformScaleWrapper = makeDecorator({
 			};
 		}, [scale]);
 
-		if (showBothScales) {
+		if (showBothScales && !componentIsFullscreen && !componentHasPopover) {
 			return html`
 				<div style="margin-bottom: 1rem" class=${mediumClass}>
 					<h4 class="spectrum-Detail spectrum-Detail--sizeXS spectrum-Examples-itemHeading">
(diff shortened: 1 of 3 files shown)


.storybook/decorators/index.js
```diff
@@ -93,6 +101,57 @@ export const withLanguageWrapper = makeDecorator({
...
+			if (scale) {
+				setTimeout(() => {
+					updateBodyClasses();
+        }, 100);
+			};
+		}, [scale]);
+
+		if (showBothScales) {
+			return html`
+				<div style="margin-bottom: 1rem" class=${mediumClass}>
```
jenndiaz, line 137: When displaying both scales on component where a popover is displayed, this space is not enough to also display the associated popover or causes some weirdness with how the components are displayed. 


Date Picker
Screenshot 2023-12-18 at 6 13 33 PM

Also, Alert Dialog, Combobox and Popover

.storybook/decorators/index.js
```diff
@@ -123,20 +182,83 @@ export const withSizingWrapper = makeDecorator({
...
+      }
+    }, [bodyEl]);
+
+    // is it the "default" display option? Return early
+    if (isDefault) return StoryFn(context);
+
+    // prep the display areas with the correct classes
+    const isExpress = args.express;
+
+    const backgroundColorClassNames = ['spectrum--light', 'spectrum--dark', 'spectrum--darkest'];
```
jenndiaz, line 237: will including darkest cause issues when we move to s2 and no longer support darkest?

jenndiaz (commented):
I got through BG colors and text direction, will plan to continue tomorrow.  

Really excited for this! I think it will help a lot to have this coverage.

.storybook/decorators/index.js
mdt2: Good callout! I can probably make this dynamic.

jenndiaz (commented):
With everything on, ltr + rtl, stacked, and medium + large, I was surprised how well everything displayed. 

The popover/modal/is-open components consistently do not display well with any of the multiple or both. 
(Alert Dialog, Alert Banner, CoachMark, Color Loupe, ComboBox, Contextual help, Datepicker, Dialog, Menu: tray submenu, Modal, Picker, Popover, Search within, Tray, Underlay)

jenndiaz:
I am also seeing some issues with text colors not adjusting for color theme when displaying multiple background colors. 

Well: if you toggle the individual background layers in controls the font adjusts correctly but if you display all 3 the font is displaying incorrectly on dark and darkest. 

Same thing with: 
- Radio read only variant 
- Treeview with section, the section headers

mdt2:
> With everything on, ltr + rtl, stacked, and medium + large, I was surprised how well everything displayed.
> 
> The popover/modal/is-open components consistently do not display well with any of the multiple or both. (Alert Dialog, Alert Banner, CoachMark, Color Loupe, ComboBox, Contextual help, Datepicker, Dialog, Menu: tray submenu, Modal, Picker, Popover, Search within, Tray, Underlay)

Good catch. I've been researching and trying different things to fix this. Here's what I'm finding:

- Unfortunately, toolbar items [can't be added/removed based on the page yet](https://github.com/storybookjs/storybook/issues/22081) (see also [a duplicate, related issue](https://github.com/storybookjs/storybook/issues/22023)). This would've been my ideal solution, to just turn off the multiple/both option for popover and modal components. Sad.
  - We did find [a plugin](https://storybook.js.org/addons/storybook-conditional-toolbar-selector) that could help with conditional toolbar items, but the last update was three years ago, and that's a bit too long for it to really be considered maintained.
- Then I thought, what if we don't use the toolbar for this and instead we can have dynamic argTypes in the control panel. But I found out that [argTypes are immutable](https://github.com/storybookjs/storybook/issues/12571#issuecomment-698695040), so that won't work unless we were to set the control on every story 😬 An option I suppose, but kinda icky.

So I'm about to push up a bandaid fix with logic that says if the component is a popover/fullscreen component, instead of rendering the component, we render an error that the feature is unavailable for that component. Not my favorite...

Jenn and I also chatted a bit about how if popovers were static, the issues around them in this context would be more manageable. But making them static means losing the example of interaction and animation that are CSS based. It's almost like we need an extra story for each popover instance that shows it statically to be able to show a bunch of them on a single page for Chromatic. Which makes me wonder more about a Chromatic-only Storybook instance like react-spectrum has.

Our hope, or at least my hope, with using custom decorators was that it wouldn't be much to manage moving forward compared to a separate Storybook instance for Chromatic, but the way things are heading I wonder if the management level might be more similar than initially expected.

castastrophe:
This is some really fantastic work!  I'm sorry I didn't realize it was going on. One thing I wanted to note is that in the rework of the build systems and architecture that I've been doing, they each include pulling out decorators into distinct files to make maintaining them easier (it also gives us a chance to publish them separately at some point for the greater Storybook community to leverage).  See this PR for an example: https://github.com/adobe/spectrum-css/pull/2398

One thing I've done in that PR is tried to create a utility for styling the wrapper element to facilitate groupings for Chromatic - I did this by setting flex box layout on the storybook wrappers but allowing components to opt-out or alter it using a `customStorybookStyles` object in the args of the story.  Let me know what you think of this approach. It might help us handle grouped layouts consistently and allow for dynamically setting the background colors (it also ensures all components contain padding for focus/hover state capturing). My big question mark here is if flexbox is too wonky a layout to use as a default (it can cause weird side effects on components that are designed for use in block for example that might not be super clear until you dig into the styles above the component).

castastrophe:
Here's some more readable documentation on the wrapper styles approach I mentioned above: https://github.com/adobe/spectrum-css/pull/2398/files?short_path=de90dbe#diff-de90dbe4186cda8c8dfb22049ae90250d47d1d38c8b61f151d9787cf49186b6b

PR #2314, opened 2023-11-21
fix(popover): nested storybook example is open and static

mdt2:
## Description

The Nested story for Popover has never quite worked. I added some logic to the existing Storybook template to turn off implementation for the Nested example in favor of a static and open example.

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

[Nested story](https://pr-2314--spectrum-css.netlify.app/preview/?path=/story/components-popover--nested):
- [ ] The story shows a nested popover that is open
- [ ] The position and withTip controls aren't shown because they're not set up to do anything to this instance

[Default story](https://pr-2314--spectrum-css.netlify.app/preview/?path=/story/components-popover--default):
- [ ] Has not been broken or affected in any way

[With Tip story](https://pr-2314--spectrum-css.netlify.app/preview/?path=/story/components-popover--with-tip&args=isOpen:!true):
- [ ] Has not been broken or affected in any way

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [ ] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [ ] VRTs have been run and looked at.
- [ ] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

## To-do list

- [x] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [x] I have updated relevant storybook stories and templates.
- [ ] I have tested these changes in Windows High Contrast mode.
- [ ] If my change impacts **other components**, I have tested to make sure they don't break.
- [ ] If my change impacts **documentation**, I have updated the documentation accordingly.
- [ ] ✨ This pull request is ready to merge. ✨

PR #2398, opened 2024-01-02
feat(storybook): add a shared decorator for group rendering and static color settings

castastrophe:
## Description

Many storybook stories require adding layout and/or padding in order to view variants together or snapshot focus states. The goal of this PR is to unify this effort in a shared decorator that can be customized using a new `customStorybookStyles` object that is leveraged in a story's args settings.

The default storybook styles applied by the decorator are:

```js
const customStorybookStyles = {
  /** @todo: do we really want flex as our default display type? does this make development harder b/c we have to consider how it impacts a component or is it helpful so we can stack components in a view easily? */
  display: "flex",
  // automatically set the background color for static color settings
  backgroundColor: staticColor === "white" ? "rgb(15, 121, 125)" : staticColor === "black" ? "rgb(181, 209, 211)" : undefined,
  // Gap is only supported for flex and grid so we don't need to set it for other display types
  gap: customStyles.gap ?? (!customStyles.display || customStyles.display && ["flex", "grid"].some(d => customStyles.display.endsWith(d))) ? "10px" : undefined,
  padding: customStyles.padding ?? (customStyles.display && ["flex", "grid", "contents"].every(d => !customStyles.display.endsWith(d))) ? "10px 0" : undefined,
  ...customStyles,
};
```

Any settings added by the story will override these values. You can unset a value completely using `undefined`, i.e.:

```js
customStorybookStyles: {
  display: undefined,
}
```

### Validation steps

- [ ] Accordion: overrides padding settings to 0
- [ ] Action bar: overrides display to use "contents" (instead of flex)
- [ ] Alert banner: sets flex-direction to "column" (instead of default value of row)
- [ ] Checkbox:
  - flex-direction: column
  - gap is unset (should not appear in stylesheet)
  - padding: 1rem
- [ ] Opacity checkerboard: inline & block size is 100px
- [ ] Radio: flex-direction is "column" and align-items is "flex-start" (instead of default of stretch)
- [ ] Stepper: flex-direction is "column" and align-items is "flex-start"
- [ ] All other components should default to:
  - display: flex
  - background-color: transparent (unless static-white or static-black settings are active)
  - gap: 1rem

**Note**: All values set by this decorator should appear as inline-styles assigned to the #root-inner element.

## To-do list


tools/component-builder-simple/css/index.js
castastrophe, line 1: Removed this as it's no longer in use and very out-of-date with our current set-up

tasks/mod-extractor.js
castastrophe, line 1: A few reporting upgrades brought in from the build rework branch.

tasks/compare-compiled-output.js
```diff
@@ -136,7 +136,7 @@ async function processComponent(
 	if (!component) return Promise.reject("No component specified.");
 
 	cleanAndMkdir(join(output, "diffs", component));
-	cleanAndMkdir(join(output, "latest"));
+	cleanAndMkdir(join(pathing.latest, component));
```
castastrophe, line 139: This minor fix corrects parallelization errors that were occurring in the compiled output tool

tools/component-builder-simple/package.json
```diff
@@ -29,7 +29,7 @@
     "postcss-dropdupedvars": "file:../../plugins/postcss-dropdupedvars",
     "postcss-dropunusedvars": "file:../../plugins/postcss-dropunusedvars",
     "postcss-import": "^10.0.0",
-    "postcss-inherit": "^4.0.3",
+    "postcss-inherit": "^4.1.0",
```
castastrophe, line 32: Update to use the latest inherit version

tools/component-builder/css/lib/varUtils.js
```diff
@@ -169,7 +169,9 @@ function getAllVars() {
 				`${varDir}/css/globals/*.css`,
 				`${varDir}/custom.css`,
 				coreTokensFile,
-			])
+			], {
+				allowEmpty: true,
```
castastrophe, line 173: Allows us to delete empty themes assets from components that don't need express styles

tools/component-builder/css/vars.js
```diff
@@ -12,18 +12,12 @@ governing permissions and limitations under the License.
 
 const gulp = require("gulp");
 const through = require("through2");
-const postcss = require("postcss");
```
castastrophe, line 15: All these variables are unused in the file

package.json
```diff
@@ -27,7 +27,7 @@
     "clean:docs": "rimraf dist",
     "clean:preview": "nx clean storybook",
     "cleaner": "nx run-many --target clean --projects",
-    "compare": "nx run-many --verbose --target compare --projects",
+    "compare": "node ./tasks/compare-compiled-output.js",
```
castastrophe, line 30: The script itself allows for better reporting on multiple component sets - the largest use-case of the script is verifying changes on the entire repo so better to use it directly than running it individually on each package (see the results when run `node ./tasks/compare-compiled-output.js` vs. `nx run-many --verbose --target compare --projects`

components/accordion/metadata/mods.md
```diff
@@ -1,4 +1,4 @@
-| Modifiable Custom Properties                          |
+| Modifiable custom properties                          |
```
castastrophe, line 1: This is super minor and a tiny bit annoying but the content style guide for the site requires sentence case headings so this gets us ready to publish this content on the pages.

components/actionmenu/stories/actionmenu.stories.js
```diff
@@ -1,20 +1,21 @@
-import { within, userEvent } from '@storybook/testing-library';
-import { html } from "lit";
+import { userEvent, within } from '@storybook/testing-library';
```
mdt2, line 1: ActionMenu has some visual differences from prod:

actionmenu-visual-difference

components/alertdialog/stories/template.js
```diff
@@ -69,32 +70,29 @@ export const Template = ({
...
+      onclick: () => {
+        if (isOpen) updateArgs({ isOpen: false });
+      },
     })}
     ${Button({
       ...globals,
       size: "m",
       variant: "secondary",
-      label: "Click to open Alert Dialog",
+      label: "Click to open",
```
mdt2, line 85:
```suggestion
      label: "Click to open Alert Dialog",
```

I always prefer more explicit labels for accessibility reasons, just a suggestion

components/assetcard/stories/assetcard.stories.js
```diff
@@ -161,3 +160,6 @@ DropTarget.args = {
 	isDropTarget: true,
 	isSelected: true,
 };
+
+export const Express = Template.bind({});
```
mdt2, line 164: This PR is adding a lot of Express stories. Won't these all need to be removed when we move to S2? It's reasonable to expect that users can use the control panel to see Express, and we can add Chromatic coverage for Express to the default story using `isChromatic()`.

components/dial/stories/dial.stories.js
```diff
@@ -1,11 +1,12 @@
-// Import the component markup template
 import { Template } from "./template";
 
 export default {
 	title: "Components/Dial",
 	description: "A dial is an input control used for selecting a value within a range, similar to a slider. It's often used in audio and video mixing and editing applications, where horizontal space is limited.",
 	component: "Dial",
 	argTypes: {
+		/* No theme styles for express available */
+		express: { table: { disable: true } },
```
mdt2, line 9: There is a different look for this one [on prod for Express](https://opensource.adobe.com/spectrum-css/preview/?path=/story/components-dial--default-express), I'm a bit confused on that

mdt2 (commented):
Here's some thoughts I have so far! Review is still in progress but I need a brain break.

.storybook/README.md
```diff
@@ -205,6 +205,75 @@ label: {
...
+If the display property is updated to something other than `*flex` or `*grid`, the gap property will be removed (`*flex` can equal `flex` or `inline-flex`). As long as the display property is not `contents`, the gap value will be assigned to `padding` instead.
+
+For example, if the display value is set to `block`, the following styles will be applied:
+
+```css
+display: block;
+padding: 10px;
+```
+
+Leveraging an argType in your component that matches `staticColor` with allowed values of `white` or `black` will automatically set the background color of the preview to an appropriate value (white will use `rgb(15, 121, 125)`; black will use `rgb(181, 209, 211)`). If you need to override this value, you can do so by setting the `customStorybookStyles` property to an object with a `backgroundColor` property. i.e.,
```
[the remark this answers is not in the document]
castastrophe: These are not tokens because it's a 1-off for showing the static black/static white settings. I think the idea was for it to be kinda funky.

components/alertdialog/stories/template.js
castastrophe: Sure but context is relevant too. Screen readers offer context and we don't have to be too redundant, in part because it's kind of annoying when the reader keeps repeating itself. My thought here is, we're in the context of an Alert Dialog so click to open being the only action available is going to apply to this context.

components/assetcard/stories/assetcard.stories.js
castastrophe: It's adding about 8 I believe - I can check on that number - and yes it's expanding coverage but it's pretty important because we're going to start removing Express for S2 and we want to ensure no regressions as we go through that process. Express is used in a pretty high visibility project. I think it's worth it

components/alertdialog/stories/template.js
castastrophe: I'll also note, it's a theme that impacts every state and variant so adding it as just another item in the collection would make snapshots unmanageable to debug in my opinion.

components/assetcard/stories/assetcard.stories.js
castastrophe: I'll also note, it's a theme that impacts every state and variant so adding it as just another item in the collection would make snapshots unmanageable to debug in my opinion.

components/dial/stories/dial.stories.js
castastrophe: Probably influenced by the global token values. My goal was adding express for components where we are adding local overrides since that's what we're maintaining.

components/assetcard/stories/assetcard.stories.js
mdt2: Ok. In that case, can we add the variants (wrapped with `isChromatic()`) to the Express story so that we get VRT coverage of the variants with Express turned on?

components/menu/stories/menu.stories.js
```diff
@@ -1,4 +1,6 @@
-// Import the component markup template
+// import isChromatic from "chromatic/isChromatic";
+const isChromatic = () => true;
```
mdt2, line 2: Ope I think this is a holdover from testing, we want to use the flag from Chromatic right?

mdt2 (commented):
Continuing review on this one! Here's a few more comments. Also, once I've clicked on a static color story, I'm seeing that background color persist after I click away to a different, non-static-color story. 🤔

components/popover/stories/template.js
```diff
@@ -4,9 +4,9 @@ import { ifDefined } from "lit/directives/if-defined.js";
 import { styleMap } from "lit/directives/style-map.js";
```
mdt2, line 4: Seems like the visual differences I was seeing in ActionMenu are probably due to these changes in popover. This has been a notoriously tricky one. Recently [I did some work](https://github.com/adobe/spectrum-css/pull/2246) to address an issue with Storybook not having the correct offset which seems to be reverted with this refactor (though I agree a refactor is a good idea for this component). And we had so much [trouble with nested popovers](https://adobedesign.slack.com/archives/G019JTYMT6H/p1693924472908079) that we chose to [show them statically](https://github.com/adobe/spectrum-css/pull/2314). Hopefully you can find a way to use this refactored code and still have the offset reflect the spec.

popover-difference

components/popover/stories/popover.stories.js
```diff
@@ -188,17 +215,17 @@ Nested.args = {
...
 			isOpen: true,
 			customStyles: {
-				"margin-inline-start": "136px",
-				"margin-block-start": "32px"
+				"margin-inline-start": "2px",
+				"margin-block-start": "2px"
 			},
 			trigger: (passthroughs) => ActionButton({
-				label: "Hop on pop(over) 2",
+				label: "Grand-pop(over)",
```
mdt2, line 228: hahaha excellent

mdt2 (commented):
Ok I've officially looked at all the code changes 🎉 Between my reviews, it looks like there's a few things to be addressed, so let me know when you need a re-review from me.

components/assetcard/stories/assetcard.stories.js
castastrophe: No need - the express stories are leveraging the groupings which include isChromatic() so it's inheriting the full sets of tests where they are available.

castastrophe: Asset card doesn't have groupings available which is why you're not seeing it here but check the [ActionButtons](https://github.com/adobe/spectrum-css/pull/2398/files#diff-b965736461e30accd067d0f9fe09aa0e60ac58b4910b6dfb39c194bbb5200f5a)

components/coachmark/stories/coachmark.stories.js
```diff
@@ -54,8 +58,9 @@ export default {
 		status: {
 			type: process.env.MIGRATED_PACKAGES.includes("coachmark")
 				? "migrated"
-				: undefined,
+				: "legacy",
 		},
+		chromatic: { delay: 2000 },
```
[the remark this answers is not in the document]
castastrophe: Oui!

components/menu/stories/menu.stories.js
castastrophe: Yes thank you!  Good catch.

PR #2405, opened 2024-01-05
chore: update compare script to fix an issue when runs are parallelized

castastrophe:
## Description

During parallelization, the latest files are being unpacked into the same folder and so compare scripts are sometimes being run against another component's assets as a baseline resulting in invalid results.

This ensures each run is unpacked into a distinct folder.

This is also correcting an issue when running the compare script against multiple components using the yarn shortcut as they each spin up separate HTML summary pages rather than combining results into 1 summary page as expected. By running the script directly, this ensures results are collated.

### Setup

1. `yarn install`
2. `yarn build`

### Validation

- [x] `yarn compare`: runs the compare task against all components by default and reports the results in 1 summary document that spins up on completion.
- [x] `yarn compare accordion`: runs the compare task against only the accordion component
- [x] `yarn compare accordion, assetlist`: runs the compare task against both the accordion and assetlist components 

## To-do list

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "clean:docs": "rimraf dist",
     "clean:preview": "nx clean storybook",
     "cleaner": "nx run-many --target clean --projects",
-    "compare": "nx run-many --verbose --target compare --projects",
+    "compare": "node ./tasks/compare-compiled-output.js",
     "predev": "nx run-many --projects ui-icons,tokens --target build",
     "dev": "NODE_ENV=development BROWSERSYNC_OPEN=true gulp dev",
     "preinstall": "command -v nvm >/dev/null 2>&1 && nvm use || exit 0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "clean:docs": "rimraf dist",
     "clean:preview": "nx clean storybook",
     "cleaner": "nx run-many --target clean --projects",
-    "compare": "nx run-many --verbose --target compare --projects",
+    "compare": "node ./tasks/compare-compiled-output.js",
```
mdt2, line 30: Is this not configurable with Nx?

castastrophe: It is not - nx is focused solely on parallelization. In this, we want to run the compare in parallel but report on the results only once. That last step is not easily set-up with nx but the script already has this ability.

mdt2 (approved):

PR #2406, opened 2024-01-05
chore: update stylelint config settings

castastrophe:
## Description

This PR updates the stylelint settings and adds a listing task to the build.

New stylelint plugins:

- **stylelint-no-missing-var**: ensures custom properties are not missing the var() function wrapper.
- **stylelint-no-unknown-custom-properties**: reports on custom properties that cannot be found locally or in the token package.
- **stylelint-no-unused-custom-properties**: reports on custom properties defined in a package but not used by the component.

Removes **stylelint-suit-naming-pattern** in favor of the existing selector-bem-pattern plugin:

```
"plugin/selector-bem-pattern": [
  {
    preset: "suit",
    presetOptions: { namespace: "spectrum" },
    utilitySelectors: /^\.(is|u)-[A-z0-9]+$/,
    componentName: /^[A-Z][A-z0-9]+$/,
  },
  {
    severity: "warning",
  },
],
```

- [x] `yarn lint`: runs and reports linting errors to the console
- [x] `yarn linter accordion`: runs linter against accordion and reports errors to the console
- [x] `yarn linter accordion --fix`: runs linter and fixes what it can in the pre-compiled assets in the accordion package. **Note** do not commit these updated assets at this time (at least not as part of this PR).
- [ ] Expect none of the above commands to block commits or merges if there are linting errors.
- [x] `yarn test:plugins`: this new command runs the existing test suite for the stylelint plugins and reports their success (or failure).

Tested by @jawinn -- note: `--verbose` is needed to see warnings. I also saw that the automatic `--fix` can create duplicate declarations, though I'm assuming this is expected to not be perfect and won't be run without human intervention of the output before commit.

## To-do list


plugins/stylelint-no-unknown-custom-properties/index.js
```diff
@@ -0,0 +1,148 @@
...
+        root.walkDecls(/^--/, ({ prop }) => {
+            localDefinitions.add(prop);
+        });
+
+        const sourceFile = root.source.input.file;
+        const parts = sourceFile ? sourceFile.split(path.sep) : [];
+        const rootIdx = parts.indexOf("components");
+        const componentRoot = parts.slice(0, rootIdx + 2).join(path.sep);
+
+        function fetchResolutions(depName) {
```
jawinn, line 61: Could you add a JSDoc style comment to this function, as well as some additional comments in this file for some of the larger blocks of code (e.g. line 73, 86, 107, etc)?

jawinn (changes requested):
I was able to verify that the commands in the validation instructions are working, with `--verbose` needed to see the warnings in the output. I tested out the plugins by modifying CSS and running linter on Accordion:

### Plugins testing

no-unknown-custom-properties - **working**
`--spectrum-accordion-test-2: var(--spectrum-test-undefined);`

stylelint-no-unused-custom-properties - **not working?**
`--spectrum-test: red;`
`--spectrum-accordion-test: 10px;`
I'm not seeing a warning for either of these added test properties.

stylelint-no-missing-parenthesis - **working**
`--spectrum-accordion-item-width: --spectrum-accordion-minimum-width;` (also tested w/ line break)

stylelint-no-missing-var - **is this still needed?**
I already get a regular syntax error from the lint command when testing out the missing bracket examples: "_Unclosed bracket  CssSyntaxError_". Is this now covered by the linter by default? I wasn't able to create an example where I could see a warning from this plugin because of the syntax error being raised first.

stylelint.config.js
```diff
@@ -1,97 +1,169 @@
...
+    "spectrum-tools/no-unknown-custom-properties": [
+      true,
+      {
+        /** @note this is a list of custom properties that are allowed to be unknown */
+        ignore: [
+          /^--mod-/,
+          /^--highcontrast-/,
+          /^--system-/,
+          /^--spectrum-(global|alias|component)-/,
+          /^--spectrum-animation-/,
```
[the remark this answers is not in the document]
castastrophe: When we remove the last @spectrum-css/vars package, I think that's the time to remove these entries rather than in advance. While I like the idea that we'd catch them in migrated components, it's a nice-to-have vs. throwing a false error on un-migrated assets.

plugins/stylelint-no-unknown-custom-properties/index.js
castastrophe: I can but can we make that a post-merge to-do since it's not affecting functionality? Only ask because this is a step toward simplifying the build migration and the build migration is my main focus. I can add a backlog story to ensure this gets returned to.

README.md
```diff
@@ -345,7 +345,9 @@ This project includes several small scripts to help with common tasks.
   - **Note** that you must run `yarn build` before running `yarn compare` to ensure that the latest build is being compared.
 - `yarn refresh:env`: This copies values for the project's `.env` file (an asset never committed to the repo as it contains login secrets) by using the `.env.example` file as a template. This script is useful when you need to update the `.env` file with new values from the `.env.example` file or when you checkout or clean the repo and need to restore the `.env` file.
 - `yarn refresh:directory`: This will remove any deprecated package folders that are no longer in use. The goal is to make migrating to a new project architecture easier for the most number of users.
-- `yarn lint:components`: Provides helpful updates and warnings for a component's package.json file. This helps keep all components in alignment.
+- `yarn lint`: Provides helpful updates and warnings for a component's package.json file. This helps keep all components in alignment.
+  - Use `yarn lint --fix` to automatically fix any issues that are found.
+  - To run on a single component, use `yarn linter accordion` (where `accordion` is the name of the component or components you want to lint).
```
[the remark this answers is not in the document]
castastrophe: If possible, I'd like to add this as a backlog story as well so we can focus on the build rework and return to this for documentation.

castastrophe:
> stylelint-no-missing-var - is this still needed?
> I already get a regular syntax error from the lint command when testing out the missing bracket examples: "Unclosed bracket CssSyntaxError". Is this now covered by the linter by default? I wasn't able to create an example where I could see a warning from this plugin because of the syntax error being raised first.

This isn't verifying missing brackets, but rather, missing var functions. Did you test adding a `color: --test-variable` to see if it would throw the error?

jawinn:
> > stylelint-no-missing-var - is this still needed?
> > I already get a regular syntax error from the lint command when testing out the missing bracket examples: "Unclosed bracket CssSyntaxError". Is this now covered by the linter by default? I wasn't able to create an example where I could see a warning from this plugin because of the syntax error being raised first.
> 
> This isn't verifying missing brackets, but rather, missing var functions. Did you test adding a `color: --test-variable` to see if it would throw the error?

Sorry, that comment was actually meant for the "stylelint-no-missing-parenthesis".

plugins/stylelint-no-unknown-custom-properties/index.js
castastrophe: (Part of my hesitation to do it now is that I didn't write the plugin, it was written by Rajdeep for the build rework so to add a solid JS-doc, I'd need to read through the whole thing again and re-acquaint myself with it's inner workings)

castastrophe:
@jawinn Should be ready for a re-review

castastrophe:
@jawinn Ah yes good call-out. This latest update includes removing `stylelint-no-missing-parenthesis`. I've also addressed the need to add verbose all the time by adding it to the yarn command and then we can go about documenting the different ways of controlling console output during the build rework.

jawinn:
Looking good! I am seeing `spectrum-tools/no-unused-custom-properties` working now. And the added warning output with the `lint` and `linter` commands.

One last thing I saw when running `yarn linter accordion`:
"_Invalid Option: Invalid option value "false" for rule "spectrum-tools/no-unknown-custom-properties". Are you trying to disable this rule? If so use "null" instead_"

Do we also want to disable the `--fix` running on commit in this PR? ([main](https://github.com/adobe/spectrum-css/blob/b3d69af5b4077522c52fc8cab8a6d2b83ca5cfb7/package.json#L110))

castastrophe:
> Looking good! I am seeing `spectrum-tools/no-unused-custom-properties` working now. And the added warning output with the `lint` and `linter` commands.
> 
> One last thing I saw when running `yarn linter accordion`: "_Invalid Option: Invalid option value "false" for rule "spectrum-tools/no-unknown-custom-properties". Are you trying to disable this rule? If so use "null" instead_" After that, the PR should be good to go.
> 
> Do we also want to disable the `--fix` running on commit in this PR? ([main](https://github.com/adobe/spectrum-css/blob/b3d69af5b4077522c52fc8cab8a6d2b83ca5cfb7/package.json#L110))

Genuinely no idea what's that referring to. If you look at the config, it's using null.  I don't see any signs of false being used.

castastrophe:
@jawinn Oh and I pushed up a removal of stylelint from the lint-staged task list and just replaced it with a vanilla prettier.

jawinn (approved):

PR #2407, opened 2024-01-05
build: allow empty theme assets during build step

castastrophe:
## Description

Currently, whether a package uses the themes/*.css assets for styling or not, those placeholder files must exist in order for the build to work. If you remove empty theme files, the build for that component will fail even though the files are not needed.

This fix removes that requirement so that we can delete theme assets when they are not needed and clean-up technical cruft.

### Setup

- Remove the themes folder in `components/accordion` (**note** do not commit this change to the branch)

### Validation steps

- [x] Run `yarn builder accordion`; expect the build to complete successfully and not output any warnings or errors regarding missing theme files. No dist/themes assets should be generated.

## To-do list

### DIFF
--- a/tools/component-builder/css/vars.js
+++ b/tools/component-builder/css/vars.js
@@ -12,17 +12,11 @@ governing permissions and limitations under the License.
 
 const gulp = require("gulp");
 const through = require("through2");
-const postcss = require("postcss");
 const logger = require("gulplog");
 const fsp = require("fs").promises;
 const path = require("path");
 
 const varUtils = require("./lib/varUtils");
-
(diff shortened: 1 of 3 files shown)


tools/component-builder/css/vars.js
```diff
@@ -12,18 +12,12 @@ governing permissions and limitations under the License.
 
 const gulp = require("gulp");
 const through = require("through2");
-const postcss = require("postcss");
```
castastrophe, line 15: These variables and imports were unused so I removed them.

mdt2 (approved):

PR #2424, opened 2024-01-12
docs(tabs): improvements to storybook for tabs

jenndiaz:
## Description

This PR adds the [tabs overflow example](https://opensource.adobe.com/spectrum-css/tabs.html#basictabswithoverflow) to storybook and decreasing the number of stories from 26 to 3 using our "kitchen sink" approach. 

[Design Spec for Tab overflow](https://spectrum.adobe.com/page/tabs/#Tab-overflow)

### Validation steps

Within the storybook for tabs: @mdt2 

- [x] toggling controls can replicate all of the stories for tabs that current exist on main 
- [x] toggling controls works as expected 
- [x] [Chromatic is displaying](https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=1212) the "kitchen sink" tabs groups as expected 

### Regression testing

Validate:

## To-do list


pfulton (changes requested):
One thing that I noticed here is that the t-shirt sizing control doesn't seem to affect the overflow variants. Can we address that?

jenndiaz:
> One thing that I noticed here is that the t-shirt sizing control doesn't seem to affect the overflow variants. Can we address that?

I implemented t-shirt sizing but this introduced complexity with positioning off popover. Since we're more concerned about the tabs implementation and not verifying the popover position, I also removed the popover from the overflow stories.

components/tabs/stories/tabs-horizontal-compact.stories.js
```diff
@@ -29,29 +28,60 @@ export default {
...
-CompactEmphasized.args = {
-  isEmphasized: true,
-  ...horizontalProps
+const TabsGroup = ({
+	...args
+}) => {
+	return html`
+		<div style="display: flex; flex-direction: column; gap: 2rem;">
+			${Template({
+          selectorStyle: {"width": "35px"},
```
mdt2, line 37: I'm seeing some weirdness when toggling controls that I think is caused by this static selectorStyle. For example, if I turn on icons, the selector indicator looks too short:

Screen Shot 2024-02-07 at 2 40 36 PM

jenndiaz: Addressed this!

components/tabs/stories/tabs-horizontal-compact.stories.js
```diff
@@ -29,29 +27,54 @@ export default {
...
+          labelWithIcons: true
+        })
+      : null }
+      ${window.isChromatic() ?
+        Template({
+          ...args,
+          isEmphasized: true,
+          iconOnly: true
+        })
+      : null }
```
mdt2, line 69: This is just style preference, aka optional, but you could move these into a const and use lit's `when` here. Something like [Combobox](https://github.com/adobe/spectrum-css/blob/main/components/combobox/stories/combobox.stories.js), but that aligns with the new `window.isChromatic()` setup.

mdt2 (approved):
This is looking good to me! One optional suggestion.

castastrophe:
@jenndiaz This is looking better but I still feel like there are way too many separate pages for the tabs. When we eventually expand test coverage, we'd be running against 16+ different snapshots. Are there ways we can collate these down further? Maybe just horizontal variants + vertical variants?

components/tabs/stories/index.js
```diff
@@ -1,6 +1,8 @@
 export const argTypes = {
   items: { table: { disable: true }},
   selectorStyle: { table: { disable: true }},
+  style: { table: { disable: true }},
```
castastrophe, line 4: Instead of style, can we use `customStyles` which is baked into the global storybook settings.

components/tabs/stories/tabs-horizontal.stories.js
```diff
@@ -30,11 +28,110 @@ export default {
   }
 };
 
-export const Default = Template.bind({});
-Default.args = horizontalProps;
+const chromaticKitchenSink = ({
```
castastrophe, line 31:
```suggestion
const HorizontalTabsGroup = ({
```

I've noticed a move toward `chromaticKitchenSink` but this name is awfully generic. If we want to export the set to use elsewhere, we'd have to rename it on entry. I also love when variable names are as specific as possible to make it clear without documentation.

components/tabs/stories/tabs-horizontal.stories.js
```diff
@@ -30,11 +28,110 @@ export default {
...
+	...args
+}) => {
+	return html`
+		<div style="display: flex; flex-direction: column; gap: 2rem;">
+    <h2>Horizontal Default</h2>
```
castastrophe, line 36: If we're going to label items in the snapshots, can we run them through typography? That protects us from unexpected environment changes (font-size, typography, etc) from influencing all our snapshots. See button for some examples.

components/tabs/stories/tabs-horizontal.stories.js
```diff
@@ -30,11 +28,110 @@ export default {
...
+        ...args
+			})}
+			${Template({
+        ...args,
+        labelWithIcons: true
+       })}
+      ${Template({
+        ...args,
+        iconOnly: true
+        })}
```
castastrophe, line 47: Since these are repeated in the 3 contexts below, can we pull that out into its own function? Maybe those are named `TabsGroup`? Also, since this same set appears in the vertical set, maybe move that definition to index.js and export it for use in both files?

components/tabs/stories/template.js
```diff
@@ -16,11 +17,115 @@ export const Template = ({
...
+				label: "Tab 2",
+			},
+			{
+				id: "tab-3",
+				label: "Tab 3",
+			}
+		]
+	}
+
+	const overflowHtml = html`
```
castastrophe, line 97: I couldn't tell if overflowHtml is used elsewhere or if it's just returned below?  If just returned below, can we skip adding it to a variable at all and just wrap this in the if statement and return the HTML directly?

components/tabs/stories/tabs-horizontal.stories.js
```diff
@@ -30,11 +30,57 @@ export default {
...
-export const Default = Template.bind({});
-Default.args = horizontalProps;
+const horiztonalTabsGroup = ({
+	...args
+  }) => {
+    return html `
+      ${Typography({
+        semantics: "heading",
+        size: "s",
+        content: ["Horizontal Default"],
```
castastrophe, line 40:
```suggestion
        content: ["Horizontal default"],
```

Would you update headings to use sentence case? Thank you!

castastrophe (approved):

castastrophe:
@jenndiaz Just looking through the baselines and they're looking good but the spacing is a little tight with the visuals and typography, do you think you could increase the gap? https://www.chromatic.com/test?appId=64762974a45b8bc5ca1705a2&id=65cf8cc9273596f853df88ce

components/tabs/stories/tabs-horizontal.stories.js
```diff
@@ -30,11 +30,59 @@ export default {
...
+        })}
+        ${TabsGroup({...args, isQuiet: true, isCompact: true})}
+      </div>
+    `
+};
+
+const overflowTabsGroup = ({
+  ...args
+}) => {
+  return html `
```
castastrophe, line 69: I can't really tell if this is working in the snapshots or not - can you add some more contextual data to the snapshot so we can tell at a glance if it's handling overflow correctly?

pfulton (approved):

PR #2437, opened 2024-01-16
fix: deprecate transform logical plugin for vanilla values

castastrophe:
## Description

Transform logical is currently converting values to rtl versions when a special keyword is provided. This is useful however, because it involves adding a non-(browser-)specified keyword to transforms, causes more confusion when onboarding than it helps.

This PR proposes removing the transform-logical utility in favor of hardcoding the values for better clarity for contributors.

Should result in no rtl regressions for the following components:

- [x] [accordion](https://pr-2437--spectrum-css.netlify.app/accordion) - indicator chevron direction swaps in rtl
- [x] [actionbutton](https://pr-2437--spectrum-css.netlify.app/actionbutton) - hold icon appears on opposite side in rtl
- [x] [assetlist](https://pr-2437--spectrum-css.netlify.app/assetlist) - item indicator appears on opposite side in rtl
- [x] [breadcrumb](https://pr-2437--spectrum-css.netlify.app/breadcrumb) - separator faces opposite direction in rtl
- [x] [calendar](https://pr-2437--spectrum-css.netlify.app/calendar) - prev & next month chevrons swap in rtl
- [x] [pagination](https://pr-2437--spectrum-css.netlify.app/pagination) - prev & next month chevrons swap in rtl
- [x] [slider](https://pr-2437--spectrum-css.netlify.app/slider) - ramp increases in opposite direction for rtl
- [x] [table](https://pr-2437--spectrum-css.netlify.app/table#collapsiblerows) - table disclosure rotates in opposite direction for rtl
- [x] [treeview](https://pr-2437--spectrum-css.netlify.app/treeview) - open item indicator faces opposite direction for rtl

### Regression testing

Validate:

## To-do list


components/accordion/index.css
```diff
@@ -621,19 +621,23 @@ governing permissions and limitations under the License.
...
-			> .spectrum-Accordion-itemIndicator {
-			/* stylelint-disable declaration-property-value-no-unknown */
-			transform: logical rotate(90deg);
-			/* stylelint-enable declaration-property-value-no-unknown */
+		> .spectrum-Accordion-itemHeading > .spectrum-Accordion-itemIconContainer > .spectrum-Accordion-itemIndicator {
+			transform: rotate(90deg);
+
+			[dir="rtl"] &,
+			&:dir(rtl) {
+				transform: matrix(-1, 0, 0, 1, 0, 0) rotate(90deg);
```
pfulton, line 629: We had initially [discussed](https://adobedesign.slack.com/archives/CS22GNRU2/p1702332156131419) removing this plugin and @jawinn mentioned that we might be able to get creative with "global" custom properties to help us get the value(s) we want out of these `transform` properties. Should we revisit that discussion?

Alternatively, if we don't go fully toward what was discussed there, perhaps we could at least pull the `matrix` part out of these places and house that in a custom property.

pfulton (commented):
Just the one comment for discussion.

components/accordion/index.css
castastrophe: I took a stab at a lower impact version of this approach.  I added a logical rotation definition to the root class for each impacted component, then leveraged that variable in the uses for the file.

```
[dir="rtl"] &,
&:dir(rtl) {
  --spectrum-logical-rotation: matrix(-1, 0, 0, 1, 0, 0);
}
```

pfulton (changes requested):
I went through the validation links, and the ones that I've left unchecked seem to have chevrons that are either pointing the incorrect direction in the LTR state, or in the RTL state, or both.

I think maybe we're missing the `rotate()` portion of the `transform` on some of those?

components/calendar/index.css
```diff
@@ -86,7 +89,7 @@ governing permissions and limitations under the License.
 
 .spectrum-Calendar-prevMonth,
 .spectrum-Calendar-nextMonth {
-  transform: logical rotate(0deg); /* stylelint-disable-line declaration-property-value-no-unknown */
+  transform: var(--spectrum-logical-rotation,);
```
jawinn, line 92: Should this be either `transform: var(--spectrum-logical-rotation,) rotate(0deg);` or `transform: var(--spectrum-logical-rotation);` without the comma? With the comma and nothing after the var() it seems like it would be creating `transform:;`.

castastrophe: I just pushed up a change without the comma, let’s see if that works

pfulton (approved):

PR #2439, opened 2024-01-16
build: remove postcss-logical plugin

jawinn:
## Description

This plugin is no longer needed. It previously was adding support for logical properties when they were new to the CSS spec, but our dist CSS files are already making use of logical properties directly and not converting properties back to their non-logical alternatives.

This plugin affected the dist output of all un-migrated components still using `component-builder` rather than the newer `component-builder-simple`. It had replaced the use of logical properties with non-logical equivalents, for example creating additional style rules with `[dir="ltr"]` and `[dir="rtl"]` prepended, and replacing `min-inline-size` with `min-width`.

_Note: the difference in dist output can be compared locally be running `yarn compare` and clicking on each component._

### Validation steps

The following components have not changed visually in LTR or RTL, in the docs and in Storybook.
- [x] Asset [@castastrophe]
- [x] CycleButton [@castastrophe]
- [x] Icon [@castastrophe]
- [x] QuickAction [@castastrophe]
- [x] SearchWithin [@castastrophe]
- [x] Site [@castastrophe]
- [x] SplitButton [@castastrophe]

### Regression testing

Validate:

- [x] The pages render correctly, are accessible, and are responsive.  [@castastrophe]

## To-do list

### DIFF
--- a/components/splitbutton/index.css
+++ b/components/splitbutton/index.css
@@ -131,6 +131,10 @@ governing permissions and limitations under the License.
     }
   }
 
+  & .spectrum-SplitButton-action .spectrum-Button-label + .spectrum-Icon {
+    margin-inline-start: var(--spectrum-splitbutton-icon-gap);
+  }
+
   &.spectrum-SplitButton--left {
     & .spectrum-SplitButton-action {
       border-start-start-radius: 0;
@@ -184,9 +188,5 @@ governing permissions and limitations under the License.
         border-inline-end-width: var(--spectrum-button-m-primary-outline-texticon-border-size);
       }
     }
-  }
-
-  & .spectrum-SplitButton-action .spectrum-Button-label + .spectrum-Icon {
-    margin-inline-start: var(--spectrum-splitbutton-icon-gap);
   }
 }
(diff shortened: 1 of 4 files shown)


castastrophe (approved):

castastrophe:
@jawinn I'm thinking the split-button is a fluke. I just rebased and am curious if it shows up again in the VRTs. If there are no changes to the compiled output, no reason the VRTs should diff.

jawinn:
After comparing the dist output locally with `yarn compare`, this does affect the dist CSS of non-migrated components still using `component-builder`. I've updated the description and validation instructions, and am doing some more testing.

It looks like SplitButton will need some slight CSS adjustments, since the loss of some specificity by the plugin added `[dir="ltr"]` and `[dir="rtl"]` created the issue seen in the VRTs. It only manifested itself in Storybook, not the docs, because it is coming from the class `spectrum-Button--iconOnly` that is only included in Storybook. And that more specific class is setting the border-radius to 50%.

jawinn:
The last VRT issue has been addressed. This needs some review of the affected components listed in the validation steps.

castastrophe (approved):

pfulton (approved):

PR #2417, opened 2024-01-10
fix(card): rename misnamed instance of background color mod

jawinn:
## Description

The custom mod property for the background color was named incorrectly as `--mod-spectrum-card-background-color` in one of the declarations. The correct mod name `--mod-card-background-color` already exists and is used by the rest of the file, so this is more of a bug fix than an overall renaming. A fallback and deprecation notice were left in case of existing usage of the old mod name.

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps @jenndiaz 

- [x] Set `--mod-card-background-color` to a color and it should change the background color (does not in prod). Setting the deprecated fallback `--mod-spectrum-card-background-color` should also change it.
- [x] Existing variants of Card look the same on docs and in Storybook. Double-check quiet, gallery, horizontal, and drop target

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [x] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [x] VRTs have been run and looked at.
- [x] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

## To-do list

- [x] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [ ] I have tested these changes in Windows High Contrast mode.
- [x] If my change impacts **other components**, I have tested to make sure they don't break.
- [x] If my change impacts **documentation**, I have updated the documentation accordingly.
- [x] ✨ This pull request is ready to merge. ✨

PR #2444, opened 2024-01-17
refactor(card): rename mod and refactor some usage of mods and vars

jawinn:
## Description

fix(card): rename misnamed instance of background color mod

The custom mod property for the background color was named incorrectly in one of the declarations.
This mod renaming was made in the reverted PR #2417 , which had some issues with the Horizontal variant. This does an additional bit of refactoring cleanup to address this:

refactor(card): cleanup mods and vars around use of preview bg

The custom properties and mods were not differentiating between the card background and the preview div background used in the quiet, gallery, and horizontal variations. This did not work well with the horizontal variant which has both a card background and a preview background, when some of the mis-named mods were fixed.

A few unset custom properties with spectrum prefixed naming also appeared to be intended as mods, and now have mod equivalents. These did not exist in the repo or in the tokens package, and have been removed/replaced:
--spectrum-card-background-color-quiet
--spectrum-card-background-color-hover-quiet
--spectrum-card-content-margin-top-quiet
--spectrum-card-minimum-width-quiet

This PR creates new custom properties for the preview background colors, that were previously being set through changing a mod value. It also renames `--spectrum-card-preview-border-width` to `--spectrum-card-preview-border-width-selected` to match with its usage.

A fallback was kept to maintain the use of `--mod-card-background-color` for the quiet variants. The horizontal variant now uses `--mod-card-preview-background-color`, as it has both a card background color and a "preview" element background color.

### Validation steps

- [x] No visual changes to Card variations in docs site, including focus and hover
- [x] No visual changes to Card in Storybook, including focus, hover, selected
- [x] Set `--mod-card-background-color` to a color and it should change the background color (does not in prod). Setting the deprecated fallback `--mod-spectrum-card-background-color` should also change it.
- [x] Setting `--mod-card-preview-background-color` should change the preview area bg color for quiet, and horizontal variants
- [x] Setting `--mod-card-preview-background-color-hover` should change the preview area bg color for quiet variants

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/card/index.css
+++ b/components/card/index.css
@@ -52,6 +52,7 @@ governing permissions and limitations under the License.
 	--spectrum-card-actions-spacing: var(--spectrum-spacing-300);
 	--spectrum-card-actions-size: var(--spectrum-card-selection-background-size);
 	--spectrum-card-actions-border-radius: var(--spectrum-corner-radius-100);
+
 	/* TODO update to --spectrum-card-selection-background-color token once an RGB stripped value is available */
 	--spectrum-card-actions-background-color-rgb: var(--spectrum-gray-100-rgb);
 	--spectrum-card-actions-background-color-opacity: var(
@@ -70,14 +71,17 @@ governing permissions and limitations under the License.
 
 	/* Selected */
 	--spectrum-card-selected-background-opacity: 0.1; /* table-selected-row-background-opacity does not exist in tokens yet */
+
 	/* TODO: These are placeholder until recursive RGB is available */
 	.spectrum--light &,
 	.spectrum--lightest & {
 		--spectrum-card-selected-background-color-rgb: var(--spectrum-blue-900-rgb);
 	}
+
 	.spectrum--dark & {
 		--spectrum-card-selected-background-color-rgb: var(--spectrum-blue-500-rgb);
 	}
+
 	.spectrum--darkest & {
 		--spectrum-card-selected-background-color-rgb: var(--spectrum-blue-600-rgb);
 	}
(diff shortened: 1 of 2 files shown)


components/card/index.css
```diff
@@ -745,8 +760,11 @@ governing permissions and limitations under the License.
 		.spectrum-Card-preview {
 			transition: none;
 			background-color: var(
-				--mod-card-background-color,
-				var(--spectrum-card-background-color)
+				--mod-card-preview-background-color, 
+				var(
+					--mod-card-background-color,
```
jawinn, line 765: Should I mark `--mod-card-background-color` as being deprecated here? It was setting the background on the preview elements, while the main card background is set to transparent directly.

components/card/index.css
```diff
@@ -824,21 +842,20 @@ governing permissions and limitations under the License.
 			var(--spectrum-card-horizontal-preview-padding)
 		);
 		background-color: var(
-			--mod-card-background-color,
-			var(--spectrum-card-background-color)
+			--mod-card-preview-background-color,
```
mdt2, line 845: Is there a reason we wouldn't include both `--mod-card-preview-background-color` and `--mod-card-background-color` here?

components/card/index.css
mdt2: I kinda like this hierarchy, reminds me of a font stack. Most specific is a preview-namespaced mod, then the general card mod, then the variable. So I don't mind it as is, but there's also not really a precedent for it in our components as far as I know. Not really an answer, just thoughts 😁

mdt2 (commented):
Looks great. Just a couple questions/thoughts.

components/card/index.css
jawinn: Yeah, this is tied to the general issue around the mod renaming. Here's a visual to better demonstrate:

Screenshot 2024-02-07 at 11 10 24 AM

```
--mod-card-preview-background-color: red;
--mod-card-background-color: lightblue;
```
These are two separate background colors, and the left side defaults to a different grey that I wouldn't expect to also change if changing the main bg color (that changes the right side).

components/card/index.css
jawinn: What I was thinking is that ideally it seems like the gallery and quiet variations should be setting `--mod-card-background-color: transparent` in the CSS instead of `background-color: transparent;`, and then the `--mod-card-preview-background-color` would be used to set the background color. That seems to fit with our existing conventions and keeps the general bg color mod available for gallery. That would break any existing usage of `--mod-card-background-color` though.

castastrophe (approved):
Nice updates and I appreciate the linting fixes introduced here too!

jawinn:
It looks like the docs build is not including the Checkbox CSS on the page. It _is_ including the Quick Actions CSS, that wraps the Checkbox:
Screenshot 2024-03-26 at 11 22 26 AM

castastrophe:
> It looks like the docs build is not including the Checkbox CSS on the page. It _is_ including the Quick Actions CSS, that wraps the Checkbox: Screenshot 2024-03-26 at 11 22 26 AM

This appears to be a bug on main so it doesn't necessarily have to block this PR. I can work on a patch for main that fixes this issue.

castastrophe:
@jawinn I've got a fix going in here - https://github.com/adobe/spectrum-css/pull/2611

PR #2445, opened 2024-01-18
feat(button): pending state

mdt2:
## Description

Add pending state variant to the existing button

#### New/additional validation steps
- [ ] The "iconOnly" version of the button (aka the circular one) remains a perfect circle when the pending state is applied. Check that the height and width match and that the iconOnly pending state button has the same height as other variants of the button. This is best checked in the docs site but can be tested in Storybook as well.
- [ ] In Storybook, Pending has it's own story that shows pending state default and static white with labels (Cassondra recommended this and is implementing similar things in her [decorator PR](https://github.com/adobe/spectrum-css/pull/2420/files#diff-b965736461e30accd067d0f9fe09aa0e60ac58b4910b6dfb39c194bbb5200f5a))
- [ ] WithForcedColors story includes Pending state
- [ ] Test WHCM. Buttons should look like this per chats with the accessibility team:
Screen Shot 2024-01-23 at 2 25 48 PM

#### Original validation steps
- [x] Design approval @mdt2 
  - Outstanding design questions [and answers](https://adobedesign.slack.com/archives/G019JTYMT6H/p1705609053178079?thread_ts=1705006780.231489&cid=G019JTYMT6H):
    - [x] Should the pending button be disabled to prevent user interaction and hover/focus states? **yes**
    - [x] Assuming it is disabled, the background color and border should be the same no matter what background color variant the button originally was, correct? For example even if it is a negative variant button, when it is pending it should look like the pending spec with the disabled gray background/border and the default blue for the progress circle? **yes, it can use disabled-background-color and disabled-border-color**
    - [x] What size Progress Circle (PC) should I use for each size of the button? **small for all size buttons, including desktop and mobile**
    - [x] What should the pending state look like against static white or black backgrounds? Should we use a different version of the Progress Circle in either case? **Static white uses the static white progress circle. Design is continuing to not support static black for progress circle, which means buttons on static black aren't eligible for pending state.**

- [x] Pending state has storybook representation @jenndiaz 
- [x] Pending state has docs site representation @jenndiaz 
- [x] Pending state matches [the spec](https://spectrum.adobe.com/page/button/#Pending) @jenndiaz 
- [x] ProgressCircle has increased chromatic coverage for indeterminate and static white variants @jenndiaz 

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/button/index.css
+++ b/components/button/index.css
@@ -56,6 +56,11 @@ governing permissions and limitations under the License.
   --spectrum-button-padding-label-to-icon: var(--spectrum-text-to-visual-100);
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-medium);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-medium);
+
+  &[pending],
+  &.is-pending {
+    --mod-button-edge-to-visual-only: calc(1px + var(--spectrum-button-edge-to-visual-only));
+  }
 }
 
 .spectrum-Button--sizeL {
@@ -72,6 +77,12 @@ governing permissions and limitations under the License.
   --spectrum-button-padding-label-to-icon: var(--spectrum-text-to-visual-200);
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-large);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-large);
+
+
+  &[pending],
+  &.is-pending {
+    --mod-button-edge-to-visual-only: calc(2px + var(--spectrum-button-edge-to-visual-only));
+  }
 }
 
 .spectrum-Button--sizeXL {
@@ -89,6 +100,15 @@ governing permissions and limitations under the License.
   --spectrum-button-top-to-text: var(--spectrum-button-top-to-text-extra-large);
   --spectrum-button-bottom-to-text: var(--spectrum-button-bottom-to-text-extra-large);
 
+  .spectrum--medium &[pending],
+  .spectrum--medium &.is-pending {
+    --mod-button-edge-to-visual-only: calc(3px + var(--spectrum-button-edge-to-visual-only));
+  }
+
+  .spectrum--large &[pending],
+  .spectrum--large &.is-pending {
+    --mod-button-edge-to-visual-only: calc(4px + var(--spectrum-button-edge-to-visual-only));
+  }
 }
 
 .spectrum-Button {
@@ -229,10 +249,22 @@ a.spectrum-Button {
   }
 
   &:disabled,
-  &.is-disabled {
+  &.is-disabled,
+  &[pending],
+  &.is-pending {
     background-color: var(--highcontrast-button-background-color-disabled, var(--mod-button-background-color-disabled, var(--spectrum-button-background-color-disabled)));
     border-color: var(--highcontrast-button-border-color-disabled, var(--mod-button-border-color-disabled, var(--spectrum-button-border-color-disabled)));
     color: var(--highcontrast-button-content-color-disabled, var(--mod-button-content-color-disabled, var(--spectrum-button-content-color-disabled)));
+  }
+
+  &[pending],
+  &.is-pending {
+    cursor: default;
+
+    .spectrum-Icon,
+    .spectrum-Button-label {
+      display: none;
+    }
   }
 }
 
@@ -241,6 +273,9 @@ a.spectrum-Button {
   .spectrum-Button {
     --highcontrast-button-content-color-disabled: GrayText;
     --highcontrast-button-border-color-disabled: GrayText;
+    --mod-progress-circle-track-border-color: ButtonText;
+    --mod-progress-circle-track-border-color-over-background: ButtonText;
+    --mod-progress-circle-thickness: var(--spectrum-progress-circle-thickness-medium);
 
     &:focus-visible {
       &::after {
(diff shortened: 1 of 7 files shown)


components/button/metadata/button-accent.yml
```diff
@@ -258,3 +258,90 @@ examples:
           </button>
         </div>
       </div>
+  - id: button-pending
```
mdt2, line 261: A bit torn on how to display this in the docs site. Should it be it's own page? If not, should I show the same pending state examples for all button types since it will look the same for all of them (except maybe static color, waiting to hear back from design on this)?

components/button/index.css
```diff
@@ -229,11 +229,16 @@ a.spectrum-Button {
   }
 
   &:disabled,
-  &.is-disabled {
+  &.is-disabled,
+  &[pending] {
```
pfulton, line 233: Interesting! I know that Spectrum Web Components is using a `pending` attribute there, but I don't think this is a valid HTML attribute for a `button`. Perhaps we could leave this selector, but also add an `.is-pending` selector, as well?

pfulton (commented):
Just one thing here, @mdt2

components/button/index.css
mdt2: Yeah that's correct, it's not an HTML attribute. I'll push that update!

pfulton (approved):

components/button/stories/button.stories.js
```diff
@@ -136,6 +149,23 @@ const CustomButton = ({
 				hideLabel: true,
 				iconName: iconName ?? "Edit",
 			})}
+			${when(isPendingStateSupported, () =>
```
castastrophe, line 152: Should this queue off the isPending variable? I'm not seeing that included here at all.

castastrophe (approved):
One question about the use of isPending in the logic for storybook.

components/button/stories/button.stories.js
mdt2: These two examples always show the pending state, so I've done this to hide them on the Static Black story since that one doesn't support pending state. The isPending variable still works, it just shows what the other buttons look like when pending state is applied. I liked the visibility of this option for users

jenndiaz (approved):
Looks good to me

castastrophe:
Looking through the new baselines, I’m realizing that pending is a state where the others depicted are design variants, I’m not sure this fits in this set long-term. cc 
@pfulton thoughts on the vrt organization here?

castastrophe:
@mdt2 can you verify the high contrast snapshots? Does that look right to you? I’m not seeing a status on it.

pfulton:
> Looking through the new baselines, I’m realizing that pending is a state where the others depicted are design variants, I’m not sure this fits in this set long-term. cc @pfulton thoughts on the vrt organization here?

I could go either way on it, but maybe it makes the most sense to have it as a separate story. @mdt2 wanna make that change?

mdt2:
I've made some updates since the previous approvals that need re-reviewed (I've also added more specific validation steps for these changes in the description):

- High contrast mode has been adjusted and approved by the accessibility team
- Identified and fixed a bug (when pending state was active, the button was no longer a perfect circle) for the iconOnly button
- Pending state has it's own story in Storybook now
- Pending state is included in the existing With Forced Colors story

components/button/index.css
```diff
@@ -89,6 +100,15 @@ governing permissions and limitations under the License.
...
 
+  .spectrum--medium &[pending],
+  .spectrum--medium &.is-pending {
+    --mod-button-edge-to-visual-only: calc(3px + var(--spectrum-button-edge-to-visual-only));
+  }
+
+  .spectrum--large &[pending],
+  .spectrum--large &.is-pending {
+    --mod-button-edge-to-visual-only: calc(4px + var(--spectrum-button-edge-to-visual-only));
+  }
```
mdt2, line 111: I couldn't find a token value that properly reflected the distance from button edge to the progress circle since it has different dimensions than the icons, so this is pretty custom. Should these be custom tokens? Or maybe there is a token value somewhere that I missed (I was [looking here](https://github.com/adobe/spectrum-tokens/blob/29663e130dc9f221462de87ae7df4847e325cb4b/packages/tokens/src/layout.json#L596))?

components/button/index.css
```diff
@@ -229,18 +249,28 @@ a.spectrum-Button {
...
+    cursor: default;
+  }
 }
 
 /* windows high contrast mode over-writes for accent variant */
 @media (forced-colors: active) {
   .spectrum-Button {
     --highcontrast-button-content-color-disabled: GrayText;
     --highcontrast-button-border-color-disabled: GrayText;
+    --mod-progress-circle-track-border-color: CanvasText;
```
jawinn, line 271: Should this double border around the progress circle be set to `ButtonText` to ensure contrast when on top of `ButtonFace`?

components/button/metadata/button-pending.yml
```diff
@@ -0,0 +1,328 @@
+name: Button - pending
+status: Verified
+description: The pending button is for indicating that a quick progress action is taking place. In this case, the label and optional icon disappear and a progress circle appears. The progress circle always shows an indeterminate progress. We recommend the use of the `.is-pending` class on the component's parent container, but there is also an option to use an attribute of `pending` instead. Buttons should have the `disabled` attribute when the pending state is applied.
+SpectrumSiteSlug: https://spectrum.adobe.com/page/button/#Pending
+examples:
+  - id: button-pending
+    name: Pending for accent, negative, primary, and secondary buttons
```
jawinn, line 7: Would it be worth showing examples for pending on buttons that are accent, negative, primary, or secondary? It looks like these are not any of those variants and are showing on a default button without any variant class (that is looking like 'spectrum-Button--outline'---is button even supposed to be used without any variant set?). 

In Storybook, I can see that with accent, it will have a grey disabled background that I don't see on this page. You could show both outline and accent, since there only appears to be two different visual styles?

jawinn (changes requested):
This is looking great! Nice work. I have a few questions around what we're showing in the docs and one of the colors used in forced-colors mode.

For buttons that have label text and an optional icon, should they still be allowed to exist in the markup, and then be set to `display: none` in the styles when the pending class is present? I was testing things out in Storybook and I think if there is a label and text in there now like on the Default accent button story, they will still appear:
Screenshot 2024-01-25 at 4 57 03 PM

mdt2:
@jawinn thank you for catching these, great thoughts! I've just pushed up those changes.

jawinn (approved):

PR #2453, opened 2024-01-22
feat(actiongroup)!: migrate to S2

mdt2:
## Description

Migrate actiongroup to [S2 design](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=9602-9335&mode=design&t=zVWXXxGYASrB1tXr-0).

### Validation steps

- [ ] In Storybook, actiongroup gap spacing matches the [S2 spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=9602-9335&mode=design&t=zVWXXxGYASrB1tXr-0)
- [ ] In Storybook, actiongroup has representation of all variants
- [ ] In the docs site, actiongroup gap spacing matches the [S2 spec]
- [ ] actiongroup no longer has theme-specific styles (when I removed the files completely, Storybook was throwing errors, so the empty files remain for now)
- [ ] Design validation @mdt2 
  - [ ] Specifically check if changes at XS and S t-shirt sizes are still good

### Regression testing

Validate:

## To-do list


components/actiongroup/index.css
```diff
@@ -14,6 +14,11 @@ governing permissions and limitations under the License.
   --spectrum-actiongroup-button-spacing-reset: 0;
   --spectrum-actiongroup-border-radius-reset: 0;
   --spectrum-actiongroup-border-radius: var(--spectrum-corner-radius-100);
+  --spectrum-actiongroup-gap-size-compact: var(--spectrum-spacing-50);
+
+  /* account for button border */
+    --spectrum-actiongroup-horizontal-spacing-compact: calc(-1px * var(--spectrum-spacing-50));
+    --spectrum-actiongroup-vertical-spacing-compact: calc(-1px * var(--spectrum-spacing-50));
```
jenndiaz, line 21: instead of `-1px` do we want to use the button border token?

jenndiaz (commented):
Looked good to me, just had a question.

components/actiongroup/index.css
jenndiaz: Could you also explain what these are doing? it looks like previously the compact buttons have negative margin but I am not seeing this after your changes.

mdt2: This is a really good question. Originally I moved this code here from the [Express theme file](https://github.com/adobe/spectrum-css/pull/2453/files#diff-b652814554c91654b7f8a9e6caf1f3ba0fd869d7af1b03f371bd870fdfd32ae2L17-L20). But when I looked at the Express actiongroup in prod, there's also not any negative margins on those buttons despite using this same formula. Turns out that this calculation, which simplifies into `calc(-1px * 2px) is an invalid property value:

Screen Shot 2024-01-29 at 10 29 00 AM

This is because when using multiplication with the `calc()` function, it's required that [one of the numbers be unitless](https://css-tricks.com/a-complete-guide-to-calc-in-css/#aa-multiplication-requires-one-of-the-numbers-to-be-unitless). So we functionally have margins set to zero for the compact variant in Express. Given that [the S2 spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=8476-850&mode=design&t=5L7YOtD8BkZpNlm7-0) defines space between buttons in both the default and compact variant, we should be good to remove all these margin values. ✨ I'll push up that change.

In case it helps you test, I tested this change by checking the values in the PR deploy with dev tools but also by screenshotting the variants and pulling them into Figma to be able to see that the pixel values match the variants (2px spacing for compact and 8px for default).

components/actiongroup/metadata/mods.md
```diff
@@ -2,9 +2,6 @@
 | ---------------------------------------------- |
 | `--mod-actiongroup-border-radius`              |
 | `--mod-actiongroup-border-radius-reset`        |
-| `--mod-actiongroup-button-spacing-reset`       |
 | `--mod-actiongroup-gap-size-compact`           |
-| `--mod-actiongroup-horizontal-spacing-compact` |
 | `--mod-actiongroup-horizontal-spacing-regular` |
-| `--mod-actiongroup-vertical-spacing-compact`   |
```
mdt2, line 9: Is it ok to remove these mods since this is a breaking change, or is there a different way we need to do this?

pfulton:
✅ approved by design via [Slack](https://adobedesign.slack.com/archives/GQ09RCBA5/p1706302198861699).

components/actiongroup/metadata/mods.md
pfulton: Yes, I think this is an ok change to make!

PR #2477, opened 2024-01-31
docs(textarea): example text is more relevant to component

mdt2:
## Description

Adjusting the example text in the TextArea docs page to be more realistic for what the component is used for.

### Validation steps
- [ ] Open the [TextArea docs site page](https://pr-2477--spectrum-css.netlify.app/textarea) and confirm that the example labels and help text make more sense in regards to the type of info a TextArea might be used to gather. (open to suggestions, you'd think I'd never used a textarea before with how difficult I found it to think of applicable labels 😂)

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/textfield/metadata/textarea.yml
+++ b/components/textfield/metadata/textarea.yml
@@ -1,6 +1,6 @@
 name: Text area
 status: Verified
-description: A multi-line text field (aka textarea)
+description: A multi-line text field using the `<textarea>` element.
 SpectrumSiteSlug: https://spectrum.adobe.com/page/text-field/
 sections:
   - name: Migration Guide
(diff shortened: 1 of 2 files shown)


components/textfield/metadata/textarea.yml
```diff
@@ -1,6 +1,6 @@
 name: Text area
 status: Verified
-description: A multi-line text field (aka textarea)
+description: A multi-line text field.
```
jenndiaz, line 3: I think it could still be important to mention that it is in fact a text area since the html is using `<textarea>` and not `<input>` but perhaps in a more polished way. 

"A multi-line text field, using the text field element" 

And if we're updating it here do we also want to update the description in textfield?

mdt2: Ohh I didn't realize the descriptions were the same for both Text Field and Text Area. Good catch, I'll update 👍

jenndiaz (approved):

components/textfield/metadata/textfield.yml
```diff
@@ -1,6 +1,6 @@
 name: Text field
 status: Verified
-description: A single-line text field (aka textfield)
+description: A single-line text field using the `<input>` element with a `type="text"`.
```
jawinn, line 3:
```suggestion
description: A single-line text field using the `<input>` element`.
```
Is this only limited to `type="text"`? I think it can also be `url`,`email`, or `password`.

jawinn (changes requested):
Nice work. The examples make a lot more sense now.

jawinn (approved):

castastrophe (approved):

components/textfield/metadata/textfield.yml
```diff
@@ -1,6 +1,6 @@
 name: Text field
 status: Verified
-description: A single-line text field (aka textfield)
+description: A single-line text field using the `<input>` element`.
```
castastrophe, line 3:
```suggestion
description: A single-line text field using the `<input>` element.
```

I noticed on the site there's a stray back-tick.

mdt2: Ope good catch!

PR #2500, opened 2024-02-07
feat(tokens): support latest style-dictionary build

castastrophe:
## Description

When adding an experimental windows-latest run, I found that the style-dictionary compilation was failing pretty terrifically. This PR aims to fix that by wrapping pathing in path.join so that it can be rendered in a windows-friendly format.

**Update**: Our windows compatibility issues are much more complex than anticipated. Those updates, while excellent for hardening the tooling here, did not solve the problem(s). The scope of this pull request is no longer to support Windows builds for CI, but instead of handle the breaking change migration for style-dictionary.

### Hardening steps

- Update the .gitattributes to force windows to use lf instead of crlf so that the generated mods files don't show a diff after the build.

- Switch the style-dictionary build to use the vanilla command instead of the nx executor.

- [x] `yarn builder tokens`. Expect the token build to create the following assets:
  - tokens/dist/css/index.css    202.00 KB
  - tokens/dist/css/dark-vars.css  48.04 KB
  - tokens/dist/css/global-vars.css  44.24 KB
  - tokens/dist/css/large-vars.css  32.08 KB
  - tokens/dist/css/light-vars.css  48.00 KB
  - tokens/dist/css/medium-vars.css  32.24 KB

### Validation steps

This can be validated via CI which will try to run the tokens build in windows. If the task fails, this PR did not work.

## To-do list


castastrophe:
Something to evaluate: https://github.com/storybookjs/storybook/pull/29676/files

.github/workflows/build.yml
```diff
@@ -150,10 +150,15 @@ jobs:
                     echo "Changes detected"
                     git status
                     git add .
-                    git diff > changes.diff
                     exit 1
                 fi
 
+            # If there are changes, capture the changes and upload them as an artifact
+            - name: Capture changes
```
castastrophe, line 157: Doing this as a separate task helps us identify issues more easily in CI

tasks/utilities.js
```diff
@@ -237,10 +275,7 @@ async function copy(from, to, { cwd, isDeprecated = true } = {}) {
 	if (!fs.existsSync(path.dirname(to))) {
 		await fsp.mkdir(path.dirname(to), { recursive: true }).catch((err) => {
 			if (!err) return;
-			console.log(
-				`${"✗".red}  problem making the ${relativePrint(path.dirname(to), { cwd }).yellow} directory`,
-			);
-			return Promise.reject(err);
+			return `${"✗".red}  problem making the ${relativePrint(path.dirname(to), { cwd }).yellow} directory`;
```
castastrophe, line 278:
```suggestion
			return Promise.resolve(`${"✗".red}  problem making the ${relativePrint(path.dirname(to), { cwd }).yellow} directory`);
```

tasks/utilities.js
```diff
@@ -251,15 +286,15 @@ async function copy(from, to, { cwd, isDeprecated = true } = {}) {
...
+				});
 				return `${"✓".green}  ${relativePrint(from, { cwd }).yellow} -> ${relativePrint(to, { cwd }).padEnd(20, " ").yellow} ${`copied ${stats.length >= 0 ? stats.length : "0"} files (${bytesToSize(stats.reduce((acc, details) => acc + details.stats.size, 0))})`.gray}`;
 			})
 			.catch((err) => {
 				if (!err) return;
-				console.log(
-					`${"✗".red}  ${relativePrint(from, { cwd }).yellow} could not be copied to ${relativePrint(to, { cwd }).yellow}`,
-				);
-				return Promise.reject(err);
+				return `${"✗".red}  ${relativePrint(from, { cwd }).yellow} could not be copied to ${relativePrint(to, { cwd }).yellow}`;
```
castastrophe, line 297:
```suggestion
				return Promise.resolve(`${"✗".red}  ${relativePrint(from, { cwd }).yellow} could not be copied to ${relativePrint(to, { cwd }).yellow}`);
```

.gitattributes
```diff
@@ -1,3 +1,4 @@
+* text=auto eol=lf
```
jawinn, line 1: It'd be worth noting this line ending config change for all files in the repo.

postcss.config.js
```diff
@@ -51,7 +51,7 @@ module.exports = ({
 						const filePath = packageParts.length > 2 ? packageParts.slice(2).join("/") : "index.css";
 
 						if (packageParts[1] === "tokens") {
-							return join(__dirname, packageParts[1], "dist", "css", filePath);
+							return join(__dirname, packageParts[1], filePath);
```
jawinn, line 54: Are we no longer using the dist version of this file? Was this a fix?

tasks/component-builder.js
```diff
@@ -345,3 +359,9 @@ async function main({
 exports.processCSS = processCSS;
 exports.fetchContent = fetchContent;
 exports.default = main;
+
+let {
+	_: components,
+} = yargs(hideBin(process.argv)).argv;
+
+Promise.all(components.map((componentName) => main({ componentName })));
```
jawinn, line 367: I'm curious about what this is doing with yargs and the change to run main() on all components at the end of this script. And the same question for the similar change in `tasks/component-reporter.js`.

tasks/component-compare.js
```diff
@@ -150,10 +138,68 @@ async function generateDiff({
 			...renderData,
 			html,
 		},
-		join(outputPath, `index.html`)
+		join(outputPath, "index.html")
 	);
 }
 
+async function fetchPublishedComponent(packageName, {
```
jawinn, line 145: Thanks for making this into a function.

tasks/component-compare.js
jawinn, line 1: What are the expected changes to the component compare? I see there are a lot of changes here and want to make sure I can validate them.

jawinn (commented):
I didn't notice any specific issues here and mostly just had questions! Some additional validation instructions and PR description of the various changes would be helpful, as it was a little hard to review all the pieces in a single commit.

I wasn't able to check off the one validation instruction that is there now, as the size of the files have changed slightly, probably due to changes since it was originally written; they were only a few KB different.

tasks/component-compare.js
```diff
@@ -438,10 +458,10 @@ async function main(
 		});
 
 		// This is our report header to indicate the start of a new component's data
-		log.write(`\n${_.pad(` ${component} `, 20, "-").cyan}\n`);
+		cliOutput.push(`\n${_.pad(` ${component} `, 20, "-").cyan}\n`);
 
 		if (warnings.length > 0) {
-			warnings.forEach((warning) => log.write(`${warning}\n`));
+			warnings.forEach((warning) => cliOutput.push(`${warning}\n`));
```
marissahuysentruyt, line 464: I take it some of these changes are just old fashioned refactors? I don't think I should expect to see anything different in my dev console, correct?

tokens/utilities/index.js
```diff
@@ -0,0 +1,6 @@
+export { default as AttributeSetsTransform } from "./attribute-sets-transform.js";
+export { default as CSSOpenTypeTransform } from "./css-font-open-type-transform.js";
+export { default as CSSSetsFormatter } from "./css-sets-formatter.js";
+export { default as DataJsonFormatter } from "./data-json-formatter.js";
+export { default as JsonSetsFormatter } from "./json-sets-formatter.js";
```
marissahuysentruyt, line 5: Do we need both `JsonSetsFormatter` and `DataJsonFormatter`? It looked like we removed `JsonSetsFormatter` from `style-dictionary.config.js`. Was it replaced in favor of `DataJsonFormatter` in the config, or are we missing it in the config?

marissahuysentruyt (commented):
I think this looks good. As far as I can tell, things are working as they were before. When I checked the file sizes of the `dist/css`, mine didn't exactly line up to your validation notes- is that an issue? They were really close, but not exact - that's all.

Screenshot 2025-04-04 at 12 43 26 PM


My main question is just how else would you like me to validate this, or is there anything else you want me to check? Is there anything that's going to change once this is merged that we should be on the lookout for?

.github/workflows/build.yml
castastrophe: Yeah for sure, the --staged flags ensures we're only diffing the changes detected, meaning those we staged in the previous step. By separating it into its own step instead of including it above, the logs are now easier to read. The more command line steps you run together, the harder it is to distinguish the results in the log output.

.github/workflows/development.yml
```diff
@@ -52,9 +52,15 @@ jobs:
...
-                    # - windows-latest # todo: debug token style-dictionary failures on windows
                 node-version:
                     - 20
+                # experimental:
+                #     - false
+                # include:
+                #     - system: windows-latest
+                #       experimental: true
+                #     - system: windows-latest
+                #       experimental: true
```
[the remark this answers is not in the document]
castastrophe: Good point but I did already add a backlog ticket to Jira to cover Windows support so it might not be needed as a todo anymore.

package.json
```diff
@@ -106,9 +105,9 @@
 		"postcss-hover-media-feature": "^1.0.2",
 		"postcss-import": "^16.1.0",
 		"postcss-licensing": "^3.0.0",
-		"postcss-load-config": "^6.0.1",
+		"postcss-load-config": "^5.1.0",
 		"postcss-modules": "^6.0.1",
-		"postcss-preset-env": "^10.1.5",
+		"postcss-preset-env": "^9.5.13",
```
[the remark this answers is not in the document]
castastrophe: No I think it just needs a rebase - it's out of sync with the updates on main now.

postcss.config.js
castastrophe: This was a fix, the filepath includes the dist/css filepath already.

tasks/component-builder.js
castastrophe: It's a way to help support running these scripts as node entry points vs. calling the functions in the modules. A difference between saying: `node ./tasks/component-builder.js' or having to extract the default method and run it: `node -e 'import('./tasks/component-builder.js').default()`

tasks/component-compare.js
castastrophe: Great point, yes this is pulling identical utilities in from the shared utilities resource instead of redefining them. The previous approach rewrote a lot of the same tasks like reporting. It also re-ran the build again manually but the workflow calling this function runs the build first so it's really unnecessary.

tokens/project.json
```diff
@@ -40,46 +40,45 @@
...
-			"executor": "@nxkit/style-dictionary:build",
-			"inputs": ["scripts"],
+			"executor": "nx:run-commands",
+			"inputs": [
+				"{projectRoot}/postcss.config.js",
+				"{projectRoot}/style-dictionary.config.js",
+				"{projectRoot}/utilities/style-dictionary.utils.js"
+			],
 			"options": {
-				"deleteOutputPath": false,
```
castastrophe, line 65: The @nxkit tools don't seem to work in Windows _at all_ so moving to the style-dictionary CLI directly really helped improve how far our Windows workflow got. Still not enough to pass but much, much closer.

tokens/style-dictionary.config.js
```diff
@@ -11,53 +11,135 @@
  * governing permissions and limitations under the License.
  */
 
-const path = require("path");
+import { dirname, join, sep } from "path";
```
castastrophe, line 14: This was rewritten as ESM instead of CJS to support Style Dictionary's breaking change to v4

tokens/utilities/index.js
castastrophe: Yes great point, we aren't using it.

castastrophe (commented):
Made a few updates based on this feedback and added more context. Let me know if you want to sync soon and I can talk through any remaining questions as well.

tasks/component-compare.js
castastrophe: No output difference, yes. This method of pushing the output to an array and printing it at the end means we can run in parallel in the script without mixing up the resulting data. Before it was writing immediately to the log so we either had to run them one-by-one or risk mixing up the outputs.

castastrophe:
> My main question is just how else would you like me to validate this, or is there anything else you want me to check? Is there anything that's going to change once this is merged that we should be on the lookout for?

@marissahuysentruyt It's a tough one to validate yeah because the goal is that everything works the same but now instead of CJS, we're leveraging ESM syntax and the latest release of style-dictionary. It's maybe a subtle performance gain in how fast it runs but overall, hopefully, no significant change.

As for the numbers being off, we probably had a token change since I wrote up the test cases so the file sizes may have updated.

tokens/project.json
```diff
@@ -40,46 +40,45 @@
...
-		"lint": {},
-		"report": {
+		"diff": {
 			"cache": true,
 			"executor": "nx:run-commands",
 			"inputs": [{ "externalDependencies": ["@adobe/token-diff-generator", "@adobe/spectrum-tokens"] }],
 			"options": {
 				"commands": [
-					"tdiff report --old-token-version @adobe/spectrum-tokens@12.25.0 --new-token-version @adobe/spectrum-tokens@0.0.0-s2-foundations-20241121221506 --format markdown --output ./dist/tokens-diff.md"
+					"tdiff report --old-token-version @adobe/spectrum-tokens@{args.v1} --new-token-version @adobe/spectrum-tokens@{args.v2} --format markdown --output {projectRoot}/dist/tokens-diff.md"
```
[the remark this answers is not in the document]
marissahuysentruyt: Thanks for this command. The tokens diff is awesome!

Screenshot 2025-04-16 at 12 40 31 PM

marissahuysentruyt (approved):
As you noted, this branch seems to be running the same as before. 👍 I don't think I see any glaring issues, so I'm good to approve. Thanks also for the extra context and the few extra commands to help us validate!

package.json
```diff
@@ -117,10 +116,10 @@
 		"prettier": "^3.5.3",
 		"rimraf": "^6.0.1",
 		"semver": "^7.7.1",
-		"stylelint": "^16.15.0",
+		"stylelint": "^16.16.0",
 		"stylelint-config-standard": "^36.0.1",
-		"stylelint-header": "^2.1.0",
-		"stylelint-high-performance-animation": "^1.11.0",
+		"stylelint-header": "^2.0.1",
+		"stylelint-high-performance-animation": "^1.10.0",
```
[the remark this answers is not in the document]
castastrophe: This has been corrected now, thanks!

.storybook/package.json
```diff
@@ -76,7 +76,7 @@
 		"react-syntax-highlighter": "^15.6.1",
 		"remark-gfm": "^4.0.0",
 		"rollup-plugin-postcss-lit": "^2.1.0",
-		"storybook": "^8.4.7",
+		"storybook": "^8.6.12",
```
jawinn, line 79: Looks like a Storybook upgrade got added in here. This newer version still causes all Docs page descriptions to disappear.

jawinn (changes requested):

.storybook/package.json
castastrophe: Ah thank you, yes!

castastrophe: Pushed up that roll-back

jawinn (dismissed):

yarn.lock
```diff
@@ -5811,177 +5868,176 @@ __metadata:
...
-  checksum: 10c0/3e8267a5bb6aa8cd3fcbf81a5fead96e6e0416568518a966e90b2409b998e7c49d87cd97e928be43cc9e978784f5474d19cc71c707937adf712db390784f4f05
+  checksum: 10c0/2f0427afb97cd445e7dde5cde9022ae65ef4a9b2c79e2d6f51757d7bd53fb844b4167a85d21d3904ea5f6b95f46df4ca34fca0ead0ae6e992884123ebabc4af0
   languageName: node
   linkType: hard
 
-"@storybook/core@npm:8.4.7":
-  version: 8.4.7
-  resolution: "@storybook/core@npm:8.4.7"
+"@storybook/core@npm:8.6.12":
+  version: 8.6.12
```
jawinn, line 6034: It looks like the Storybook updates are still in the yarn.lock file, so the Docs descriptions are still missing on the PR build. Do we need to limit the Storybook version in `main` like we did for `spectrum-two` in #3640, to prevent updated dependencies from getting a version that breaks our docs?

jawinn (changes requested):

yarn.lock
castastrophe: Good point. I've updated to remove the caret in the package.json and regenerated the lockfile.

PR #2380, opened 2023-12-19
fix(textfield): icon positioning with wrapped fieldlabel

jenndiaz:
## Description

The position of the textfield validation icons current does not match design spec when the associated feildlable wraps or when a side label is applied. 

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [ ] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [ ] VRTs have been run and looked at.
- [ ] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

### Before 

Screenshot 2023-12-18 at 1 57 12 PM
Screenshot 2023-12-18 at 1 57 25 PM
Screenshot 2023-12-18 at 1 58 52 PM
Screenshot 2023-12-18 at 1 59 29 PM

### After 
Screenshot 2023-12-18 at 2 48 24 PM
Screenshot 2023-12-18 at 2 49 04 PM

## To-do list

- [ ] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [ ] I have updated relevant storybook stories and templates.
- [ ] I have tested these changes in Windows High Contrast mode.
- [ ] If my change impacts **other components**, I have tested to make sure they don't break.
- [ ] If my change impacts **documentation**, I have updated the documentation accordingly.
- [ ] ✨ This pull request is ready to merge. ✨

PR #2506, opened 2024-02-09
fix(textarea): drag and grow behaviors

jenndiaz:
## Description

Updates text area to display grow and drag behaviors. 

[Design Guidance](https://spectrum.adobe.com/page/text-area/#Drag-icon):
- TextArea should display drag by default 
- When grow is enabled this drag behavior and icon should go away 

Changes: 

- Update CSS for all text areas to be draggable by default 
- update docs site to include grow description

### Validation steps

On the docs page for TextArea 
- [x] All variants, expect for grows, should be draggable
- [ ] Validation icons stay aligned on top of the textarea after resize.

In storybook:

- [x] TextArea is draggable by default
- [x] drag behavior goes away with Grows control toggled to true
- [ ] Validation icons stay aligned on top of the textarea after resize. Include testing in Chromatic-only view of Storybook. 

### Regression testing

Validate:

### Before
Screenshot 2024-02-09 at 2 36 32 PM

### After 
Screenshot 2024-02-09 at 2 39 23 PM

## To-do list

### DIFF
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -1267,6 +1267,7 @@ governing permissions and limitations under the License.
 
 	.spectrum-Textfield-input {
 		resize: both;
+		block-size: unset;
 		min-inline-size: var(
 			--mod-text-area-min-inline-size,
 			var(--spectrum-text-area-min-inline-size)
@@ -1277,6 +1278,7 @@ governing permissions and limitations under the License.
 		);
 	}
 
+	/* CSS to support text area grow behavior */
 	&.spectrum-Textfield--grows {
 		.spectrum-Textfield-input {
 			grid-row: 2 / auto;
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -553,26 +553,15 @@ governing permissions and limitations under the License.
 
 	/****** Validation Icon - Invalid ⚠️ ******/
 	.spectrum-Textfield.is-invalid & {
+		justify-self: end;
+
 		block-size: var(
 			--mod-textfield-icon-size-invalid,
 			var(--spectrum-textfield-icon-size-invalid)
 		);
-		inline-size: var(
-			--mod-textfield-icon-size-invalid,
-			var(--spectrum-textfield-icon-size-invalid)
-		);
-		inset-block-start: var(
-			--mod-textfield-icon-spacing-block-invalid,
-			var(--spectrum-textfield-icon-spacing-block-invalid)
-		);
-		inset-block-end: var(
-			--mod-textfield-icon-spacing-block-invalid,
-			var(--spectrum-textfield-icon-spacing-block-invalid)
-		);
-		inset-inline-end: var(
-			--mod-textfield-icon-spacing-inline-end-invalid,
-			var(--spectrum-textfield-icon-spacing-inline-end-invalid)
-		);
+		padding-block-start: var(--spectrum-textfield-icon-spacing-block-invalid);
+		padding-inline-end: var(--spectrum-textfield-icon-spacing-inline-end-invalid);
+
 		color: var(
 			--highcontrast-textfield-icon-color-invalid,
 			var(
@@ -1264,6 +1253,11 @@ governing permissions and limitations under the License.
 /*** Text Area ***/
 .spectrum-Textfield--multiline {
   --spectrum-textfield-input-line-height: normal;
+
+	&.spectrum-Textfield {
+		inline-size: unset;
+		min-inline-size: var(--mod-textfield-width, var(--spectrum-textfield-width));
+	}
 
 	.spectrum-Textfield-input {
 		resize: both;
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -1265,7 +1265,8 @@ governing permissions and limitations under the License.
 .spectrum-Textfield--multiline {
   --spectrum-textfield-input-line-height: normal;
 
-  .spectrum-Textfield-input {
+	.spectrum-Textfield-input {
+		resize: both;
 		min-inline-size: var(
 			--mod-text-area-min-inline-size,
 			var(--spectrum-text-area-min-inline-size)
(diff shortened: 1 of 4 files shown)


components/textfield/metadata/textarea.yml
```diff
@@ -174,8 +174,18 @@ examples:
...
+    description: Use `spectrum-Textfield--grows` to disable the drag behavior along with appropriate JavaScript to enable a text area that grows with input.
+    markup: |
+      <div class="spectrum-Textfield spectrum-Textfield--multiline spectrum-Textfield--grows">
+        <label for="textfield-helptext" class="spectrum-FieldLabel spectrum-FieldLabel--sizeM">Password</label>
+        <textarea id="textfield-helptext" name="field" class="spectrum-Textfield-input"></textarea>
+      </div>
+
   - id: textfield-quiet
     name: Quiet
+    description: Text areas should rarely be presented in the quiet style. The height of the field should use grow logic so the height will adjust with the height of the entered value.
```
mdt2, line 188: > The height of the field should use grow logic so the height will adjust with the height of the entered value.

I'm not clear on what this means. Do we expect this to require JS implementation?

mdt2 (commented):
Thanks for working on this! It's looking good. One thing I'm seeing in Storybook with Grows turned on, if I turn Invalid on, then the icon gets misaligned. Is this expected to be fixed in #2380 and therefore not a concern for this PR?

Screen Shot 2024-02-15 at 5 16 10 PM

components/textfield/index.css
```diff
@@ -1265,7 +1265,9 @@ governing permissions and limitations under the License.
 .spectrum-Textfield--multiline {
   --spectrum-textfield-input-line-height: normal;
 
-  .spectrum-Textfield-input {
+	.spectrum-Textfield-input {
+		resize: both;
+		block-size: unset;
```
jawinn, line 1270: 👍 I'm glad you got this update in there. In addition to the XL height bug fix you mentioned in the PR description, it also lets the textarea height size appropriately if the rows attribute it used, e.g. `rows="5"`.

components/textfield/metadata/textarea.yml
```diff
@@ -174,8 +174,18 @@ examples:
...
+    description: Use `spectrum-Textfield--grows` to disable the drag behavior along with appropriate JavaScript to enable a text area that grows with input.
+    markup: |
+      <div class="spectrum-Textfield spectrum-Textfield--multiline spectrum-Textfield--grows">
+        <label for="textfield-helptext" class="spectrum-FieldLabel spectrum-FieldLabel--sizeM">Password</label>
+        <textarea id="textfield-helptext" name="field" class="spectrum-Textfield-input"></textarea>
+      </div>
+
   - id: textfield-quiet
     name: Quiet
+    description: Text areas should rarely be presented in the quiet style. When used, quiet text areas should use Javascript to enable a text area that grows with input.
```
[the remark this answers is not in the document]
jenndiaz: this has been added.

pfulton (changes requested):
In Chrome in Storybook, if I set `invalid` to true, and then drag the textarea to make it larger, the invalid icon's position stays in its initial spot

Screenshot 2024-02-23 at 1 02 49 PM

components/textfield/index.css
```diff
@@ -553,26 +553,15 @@ governing permissions and limitations under the License.
 
 	/****** Validation Icon - Invalid ⚠️ ******/
 	.spectrum-Textfield.is-invalid & {
+		justify-self: end;
```
jenndiaz, line 556: This is attempt to address the invalid and valid icon position when drag behavior is enabled. 
`justify-self: end;` positioning the icon at the end of the grid and using padding instead of inset address the spacing around the icon. 

However there is still an issue with the min-size for the text-area, where you can drag the text area to be smaller than the original text area and the icon appears outside of the text area. I think there may be some browser differences that would need looking into here as well. 

Screenshot 2024-02-27 at 12 06 42 PM

components/textfield/index.css
jawinn: While I still think that using `block-size: unset;` is a good idea, I've removed it from this PR as it causes a few issues downstream with SWC and with the quiet variants. 

Text areas default to two rows in most browsers, resulting in fields like quiet that previously displayed as one line, displaying as two by default. It also would change the heights on some fields. This is significant enough of a change to be its own PR and can be revisited later, likely as part of the Spectrum 2 migration. It would likely require work on the SWC end.

The benefits of keeping the change would be the ability to use the `rows` attribute on the textarea. SWC lists rows in their API for Text area but it does not seem to work currently. The design specs do not specify a fixed height, only  a minimum height.

jawinn:
I got _really_ close with addressing some of the alignment issues and improving the resize/drag behavior. There are still some issues with it that I'll explain. I also discovered some areas that need more Storybook coverage.

### Improvements and changes

- Improved alignment of validation icon in the majority of cases after resize and drag.
- The Text area parent and input + validation icon no longer get out of alignment and keep the same width when you horizontally resize the field down really small
- When you horizontally resize larger, this no longer can grow too big and escape the container
- I decided to remove the `block-size: unset` declaration, possibly to be revisited later, per [the notes in the discussion](https://github.com/adobe/spectrum-css/pull/2506#discussion_r1526625609)

https://github.com/adobe/spectrum-css/assets/965114/6737247e-b633-4d95-806c-fc5f18f5ab6a

### Remaining issues

In the Chromatic testing view of Text area, I noticed that the wrapping example was no longer wrapping. This surfaced an alignment issue as well. On the parent, `inline-size: unset;` is used so that it is no longer a fixed width, and the width of the resizable input also affects the width of the parent. The problem with this is that the input will no longer start with the fixed width defined in the design spec, when the text of the field label is too long. And that means it would not wrap as it did previously. The grid layout will stretch out, and leave the input behind. In the following screenshot, "grid" highlighting is turned on in the Chrome inspector: 

Screenshot 2024-03-15 at 12 51 00 PM

https://github.com/adobe/spectrum-css/assets/965114/04310182-383d-43c1-bc76-99175a5f050e


I'm not sure if there is a way around this---I'm open to ideas! 
This makes me wonder if we should instead of using `inline-size: unset;`, rely on the implementation to use JS to update the parent width when the resize is used. Note that when the native resize functionality is used in the browser, the browser sets fixed width and height values on the element being resized.

### Increased Storybook coverage needed

I will do this as a separate PR, but in testing this I discovered a few areas that could used increased Storybook coverage and controls for Text area, that would be useful to fully test this. The character counter and help text are demoed in the docs but not represented in Storybook. The character counter especially is important to test as it is displayed within the second column of the grid layout being used for alignment.

PR #2519, opened 2024-02-13
fix(textfield): remove extra padding on nested label and help

jawinn:
## Description

Previously some styles were adding a small amount of extra inline-start padding to the label and help text, when they are nested within Text field (as shown on the docs page for Text field). This was particularly noticeable on the quiet variation—see screenshot.
    
It looks like this previously was adding padding that matched the amount of corner rounding, so the label text would start where the corner started its curve.
    
On the design, there is no additional left padding. This update removes that padding to adhere to the design.

This also includes a small Storybook fix for the alignment of the valid checkmark. Due to recent Icon fixes, it was necessary to specify the icon set as "ui" in some of the args.

CSS-674

### Validation steps @jenndiaz 

- [x] There is no additional inline-start padding on the label and help text within the docs for Textfield **non-quiet variants**
- [x] There is no additional inline-start padding on the label and help text within the docs for Textfield **quiet variants**

### Regression testing

Validate:

Note: Variations of Text field, including with a Field label, are not currently well represented in Chromatic/VRT, but there are already improvements to this coming in PR #2380 .

## Screenshots

Design:
design-reference

Previous behavior:
previous-behavior

Screenshot 2024-02-13 at 3 08 36 PM

## To-do list

### DIFF
--- a/components/textfield/index.css
+++ b/components/textfield/index.css
@@ -616,12 +616,6 @@ governing permissions and limitations under the License.
 		);
 		grid-row: 1;
 		grid-column: 1 / span 1;
-		padding-inline-start: calc(
-			var(
-					--mod-textfield-corner-radius,
-					var(--spectrum-textfield-corner-radius)
-				) / 2
-		);
 	}
 
 	.spectrum-Textfield--quiet & {
@@ -645,12 +639,6 @@ governing permissions and limitations under the License.
 		);
 		grid-row: 3;
 		grid-column: 1 / span 2;
-		padding-inline-start: calc(
-			var(
-					--mod-textfield-corner-radius,
-					var(--spectrum-textfield-corner-radius)
-				) / 2
-		);
 	}
 }
 
--- a/components/textfield/stories/template.js
+++ b/components/textfield/stories/template.js
@@ -33,6 +33,7 @@ export const Template = ({
 	labelPosition = "top",
 	labelText,
 	iconName,
+	iconSet,
 	pattern,
 	placeholder,
 	name,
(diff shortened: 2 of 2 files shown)


jenndiaz (approved):
Looks good to me, nice catch 🎣

components/textfield/index.css
```diff
@@ -617,12 +617,6 @@ governing permissions and limitations under the License.
 		);
 		grid-row: 1;
 		grid-column: 1 / span 1;
-		padding-inline-start: calc(
```
castastrophe, line 620: @jawinn @pfulton Is there a risk at all that this could break a downstream override?  Should we keep the property with the mod + calc and just remove the fallback?

components/textfield/stories/template.js
```diff
@@ -50,8 +51,13 @@ export const Template = ({
...
 	}
 
-	if (isInvalid) iconName = "Alert";
-	else if (isValid) iconName = "Checkmark";
+	if (isInvalid) {
+		iconName = "Alert";
+		iconSet = "workflow";
+	} else if (isValid) {
+		iconName = "Checkmark";
+		iconSet = "ui";
```
castastrophe, line 59: Why pass in iconName and iconSet as variables? It seems like they're being hardcoded here? I'm also curious why this has an isInvalid and isValid flag? Should we just use one or the other and use negation to toggle between them?

jawinn: It looks like the icon is an arg because the Search component passes in a third option, `iconName: "Magnify"`. If we wanted to refactor to remove those args that allow choosing the icon for this story, I think we would need to add a different arg like`isSearch`?

isInvalid and isValid are for toggling those classes for the alert or the checkmark, and the third state is not set, which doesn't display either. It looks like Datepicker is also doing it this way. We could use a select list of options with undefined, is-valid, or is-invalid, but I think it's probably good to be able to test when both classes are present to make sure that the error always overrides valid.

castastrophe: If search passes in Magnify and it's listed as isInvalid - does the icon get overridden?

castastrophe: I suppose search can't be invalid - just trying to think what states things could get in to

castastrophe (approved):

PR #2520, opened 2024-02-13
feat(downstate): utility class/placeholder mixed implementation

mdt2:
## Description

I ran into some challenges with a purely class based implementation of down state. For components like Button where the whole component is interactable, class based works fine because you can apply the classes to the root/parent of the component. But with a more complex component like Checkbox, you have to apply the styles to the box itself, but the `active` state that we're targeting happens at the root/parent level.

So I experimented with this mix of utility class and placeholder approaches. Maybe we should drop the utility class and use the placeholder approach since we want these states to ship with our S2 components? Maybe users will need the utility classes? It feels a bit ick to keep both because it can lead to confusion for users. Looking for thoughts and opinions for now.

### Validation steps
- Take a look at the down state docs site, which shows the class based approach on button and should also link you to the checkbox page to see the placeholder approach working there
- Take a look at the down state page in Storybook. This contains both the button using the class approach and the checkbox using the placeholder approach

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/checkbox/index.css
+++ b/components/checkbox/index.css
@@ -16,6 +16,8 @@ governing permissions and limitations under the License.
  * .spectrum-Checkbox-box::after is the focus indicator
  */
 
+@import "@spectrum-css/downstate/index.css";
+
 /* Component tokens by t-shirt size */
 .spectrum-Checkbox {
   /* Color */
@@ -372,6 +374,11 @@ governing permissions and limitations under the License.
     }
   }
 }
+
+.spectrum-Checkbox:active:not(.is-readOnly, :disabled, .is-disabled) .spectrum-Checkbox-box {
+  @extend %spectrum-DownState--min-perspective;
+}
+
 /* stylelint-enable max-nesting-depth */
 
 .spectrum-Checkbox-label {
@@ -571,7 +578,7 @@ governing permissions and limitations under the License.
       outline-style: auto;
       outline-offset: var(--highcontrast-checkbox-focus-indicator-gap, var(--mod-checkbox-focus-indicator-gap, var(--spectrum-checkbox-focus-indicator-gap)));
       /* stylelint-disable-next-line declaration-block-no-redundant-longhand-properties */
-      outline-width: var(--mod-focus-indicator-thickness, var(--spectrum-focus-indicator-thickness)); 
+      outline-width: var(--mod-focus-indicator-thickness, var(--spectrum-focus-indicator-thickness));
 
       &::after {
         box-shadow:
--- a/components/downstate/README.md
+++ b/components/downstate/README.md
@@ -1,0 +1,13 @@
+# @spectrum-css/downstate
+
+> > The Spectrum CSS, down state is a utility class used to apply active states to Spectrum 2 components. Consult design documentation for further usage and infomation.
+
+`.spectrum-DownState`
+
+To use this classes:
+- add @spectrum-css/downstate as a dependency
+- Use the class on the component's parent element (for example, it would be on the same element as the `.spectrum-Checkbox` class, or the `.spectrum-Button` class)
+
+This package is part of the [Spectrum CSS project](https://github.com/adobe/spectrum-css).
+
+See the [Spectrum CSS documentation](https://opensource.adobe.com/spectrum-css/downstate).
--- a/components/downstate/metadata/downstate.yml
+++ b/components/downstate/metadata/downstate.yml
@@ -1,35 +1,23 @@
 name: Down state
 SpectrumSiteSlug: https://spectrum.adobe.com/page/down-state/
 description: |
-  Down state is a Spectrum 2 feature that creates the illusion of components being pressed-in when active. To implement down state:
+  Down state is a Spectrum 2 (S2) feature that creates the illusion of components being pressed-in when active. There are two different ways to implement down state depending on the component.
+
+  If the component is interactable at it's root level:
     - Add the `.spectrum-DownState` class to the root of the element that receives interactions.
+      - Components with a width of 24px or less require the addition of the `.spectrum-DownState--min-perpsective` class next to the `.spectrum-DownState` class.
     - Develop an implementation to determine the width and height of the component. Assign this to the `--spectrum-downstate-width` and `--spectrum-downstate-height` custom properties in a `style` attribute on the HTML element.
+
+  If the component has an interactable element nested within it, this will require CSS changes:
+    - Use the `%spectrum-DownState` or `%spectrum-DownState--min-perspective` placeholder on the interactable element when the root of the component is in an `:active` state.
+      - Components with a width of 24px or less require the `%spectrum-DownState--min-perspective` placeholder instead.
+    - Note that this functionality is included in S2 components that require down state in this project. This can be seen in the [checkbox component](/checkbox.html).
 examples:
   - id: downstate
-    name: Down state - default
+    name: Down state - utility class
     markup: |
       <div class="spectrum-Examples">
         <button class="spectrum-DownState spectrum-Button spectrum-Button--fill spectrum-Button--accent spectrum-Button--sizeM" style="--spectrum-downstate-width: 72px; --spectrum-downstate-height: 32px;">
           <span class="spectrum-Button-label">Edit</span>
         </button>
-      </div>
-  - id: downstate-min
-    name: Down state - minimum perspective
-    description: |
-      Components with a width of 24px or less require the addition of the `.spectrum-DownState--min-perpsective` class next to the `.spectrum-DownState` class.
-      * Note that in this example, the element that will be interacted with is nested in a component. In this case, the "root" is the element that is interacted with, not the parent component.
-    markup: |
-      <div class="spectrum-Examples">
-        <label class="spectrum-Checkbox spectrum-Checkbox--sizeM">
-          <input type="checkbox" class="spectrum-Checkbox-input" id="checkbox-0">
-          <span class="spectrum-Checkbox-box spectrum-DownState spectrum-DownState--min-perspective" style="--spectrum-downstate-width: 14px; --spectrum-downstate-height: 14px;">
-            <svg class="spectrum-Icon spectrum-UIIcon-Checkmark100 spectrum-Checkbox-checkmark" focusable="false" aria-hidden="true">
-              <use xlink:href="#spectrum-css-icon-Checkmark100" />
-            </svg>
-            <svg class="spectrum-Icon spectrum-UIIcon-Dash100 spectrum-Checkbox-partialCheckmark" focusable="false" aria-hidden="true">
-              <use xlink:href="#spectrum-css-icon-Dash100" />
-            </svg>
-          </span>
-          <span class="spectrum-Checkbox-label">Checkbox</span>
-        </label>
       </div>
--- a/components/downstate/metadata/downstate.yml
+++ b/components/downstate/metadata/downstate.yml
@@ -1,0 +1,35 @@
+name: Down state
+SpectrumSiteSlug: https://spectrum.adobe.com/page/down-state/
+description: |
+  Down state is a Spectrum 2 feature that creates the illusion of components being pressed-in when active. To implement down state:
+    - Add the `.spectrum-DownState` class to the root of the element that receives interactions.
+    - Develop an implementation to determine the width and height of the component. Assign this to the `--spectrum-downstate-width` and `--spectrum-downstate-height` custom properties in a `style` attribute on the HTML element.
+examples:
+  - id: downstate
+    name: Down state - default
+    markup: |
+      <div class="spectrum-Examples">
+        <button class="spectrum-DownState spectrum-Button spectrum-Button--fill spectrum-Button--accent spectrum-Button--sizeM" style="--spectrum-downstate-width: 72px; --spectrum-downstate-height: 32px;">
+          <span class="spectrum-Button-label">Edit</span>
+        </button>
+      </div>
+  - id: downstate-min
+    name: Down state - minimum perspective
+    description: |
+      Components with a width of 24px or less require the addition of the `.spectrum-DownState--min-perpsective` class next to the `.spectrum-DownState` class.
+      * Note that in this example, the element that will be interacted with is nested in a component. In this case, the "root" is the element that is interacted with, not the parent component.
+    markup: |
+      <div class="spectrum-Examples">
+        <label class="spectrum-Checkbox spectrum-Checkbox--sizeM">
+          <input type="checkbox" class="spectrum-Checkbox-input" id="checkbox-0">
+          <span class="spectrum-Checkbox-box spectrum-DownState spectrum-DownState--min-perspective" style="--spectrum-downstate-width: 14px; --spectrum-downstate-height: 14px;">
+            <svg class="spectrum-Icon spectrum-UIIcon-Checkmark100 spectrum-Checkbox-checkmark" focusable="false" aria-hidden="true">
+              <use xlink:href="#spectrum-css-icon-Checkmark100" />
+            </svg>
+            <svg class="spectrum-Icon spectrum-UIIcon-Dash100 spectrum-Checkbox-partialCheckmark" focusable="false" aria-hidden="true">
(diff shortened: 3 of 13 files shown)


components/downstate/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+  - If the interactable element has a width of 24px or less, we use the minimum perspective token: `transform: perspective(var(--spectrum-component-size-minimum-perspective-down)) translateZ(var(--spectrum-component-size-difference-down));`
+  - If the interactable element has a width that is greater than 24px, we use the max formula: `transform: perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down))) translateZ(var(--spectrum-component-size-difference-down));`
+    - Note that in this case, users are required to develop an implementation to determine the width and height of the component. Assign these values to the `--spectrum-downstate-width` and `--spectrum-downstate-height` custom properties in a `style` attribute on the HTML element to expose them for use in the CSS.
+
+This package is part of the [Spectrum CSS project](https://github.com/adobe/spectrum-css).
+
+See the [Spectrum CSS documentation](https://opensource.adobe.com/spectrum-css/downstate).
```
jenndiaz, line 11: this url does not work, is this s2 documentation still VPN access only, how do we want to handle s2 doc links in the meantime?

jenndiaz:
I don't necessarily thing this is something that should be addressed but I did notice on Safari and Windows(via assistive labs) that the checkbox active state squares out the checkbox borders while other browsers do not. 

### Safari Active
Screenshot 2024-02-15 at 10 13 59 AM

components/downstate/metadata/downstate.yml
```diff
@@ -0,0 +1,82 @@
+name: Down state
```
mdt2, line 1: We'll need to add the max-perspective token setup here once it's available, but design hasn't defined the token yet and therefore it's not in the token system at this time.

components/downstate/README.md
mdt2: Ah good point. This URL is wrong, my bad, but it is still blocked by VPN. I'll remove it for now.

components/button/index.css
```diff
@@ -147,6 +147,11 @@ governing permissions and limitations under the License.
     box-shadow: none;
   }
 
+  &:active {
+    /* stylelint-disable-next-line spectrum-tools/no-unknown-custom-properties */
+    transform: perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down))) translateZ(var(--spectrum-component-size-difference-down));
```
mdt2, line 152: If we're expecting users to dynamically fetch this width/height data, will we need to namespace these?

Rajdeepc: Any use cases you have in mind where a user will do it from their end?

components/downstate/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "@spectrum-css/downstate",
+  "version": "1.0.0-alpha.0",
+  "description": "The Spectrum CSS downstate utility class",
+  "license": "Apache-2.0",
+  "author": "Adobe",
+  "homepage": "https://opensource.adobe.com/spectrum-css/",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/adobe/spectrum-css.git",
+    "directory": "components/downstate"
```
Rajdeepc, line 11: Is there a reason to keep this down state "concept" as a part of the component directory?

Rajdeepc:
Should we also think and test this concept in mobile devices too with different themes?

components/button/index.css
mdt2: Originally when we were exploring a class-based approach, users were be expected to set these width and height variables in their implementation. But if we go with the component-based approach shown in this PR, we may be able to set these and ship them as part of the CSS instead.

components/downstate/package.json
mdt2: Nope! After this PR was opened we got some input from design, so this will change to look more like the [corner rounding PR](https://github.com/adobe/spectrum-css/pull/2559) (which, disclaimer, is also a work in progress like this one)

mdt2:
@Rajdeepc Good questions!

1. Yes, `:active` is the way we can simulate a "down state" in CSS. From [the design docs](https://s2.spectrum.corp.adobe.com/page/states/#down-state): down state is the concept to make a component look "pressed in" when activated by a user. We're using that phrase "down state" to align with design.
2. We don't have a `:pressed` pseudo class available in CSS, so we're using the `:active` [CSS pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:active). It should be equivalent to the idea of a pressed state, but let me know if that's not the case with web components

I agree, we should do some testing on mobile as well once the team agrees on the approach.

Let me know if this answers your questions 🙂

mdt2:
> > should the down state have a `prefers-reduced-motion` alternative?
> 
> There isn't anything for this in the design docs so I reached out to PJ on Slack to ask 👍

After talking with Damian and PJ [in Slack](https://adobedesign.slack.com/archives/CSNDSSA5D/p1709241964489619), this down state motion isn't jarring or problematic from an accessibility perspective, so we don't need a reduced motion alternative.

components/button/stories/button.stories.js
```diff
@@ -349,4 +349,13 @@ Wrapping.args = {
...
+};
+
+export const DownStateDocs = CustomButton.bind({});
+DownStateDocs.args = {
+	variant: "accent",
+	customStyles: {
+		"--spectrum-downstate-width": "72px",
+		"--spectrum-downstate-height": "32px"
+	}
+}
```
mdt2, line 361: This is a WIP and I don't love the direction it's going so far...if the team agrees on this approach (to setup down state in each component's index.css), then we should do this in the template.js, and make it compatible with changing t-shirt sizing and platform scale. But this was the first step to just see it working on the new Foundations/Down state page.

Checkbox doesn't need this because it uses the `component-size-minimum-perspective-down` token, which is already defined. But for components that are larger than 24px, we have to [calculate the perspective](https://github.com/adobe/spectrum-css/pull/2520/files#diff-97b5a7e747486aca74b3076354c28a00ef69dab3c91dc0806d926078198571a2R152) based on width and height. We've thought from the beginning that consumers would have to create their own implementation for setting the values of those width and height variables, and in button's case I think that's still true, because we don't statically set a width and height in the CSS...

components/downstate/package.json
mdt2: I've pushed up a WIP commit that starts making the changes needed for this.

.storybook/foundations/down-state.mdx
```diff
@@ -0,0 +1,24 @@
...
+  - If the interactable element has a width that is greater than 24px, we use the max formula:
+    ```
+    transform:
+      perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down)))
+      translateZ(var(--spectrum-component-size-difference-down));
+    ```
+    - Note that in this case, users are required to develop an implementation to determine the width and height of the component. Assign these values to the `--spectrum-downstate-width` and `--spectrum-downstate-height` custom properties in a `style` attribute on the HTML element to expose them for use in the CSS.
+
+<Story of={CheckboxStories.Default} />
+<Story of={ButtonStories.DownStateDocs} />
```
mdt2, line 24: There's an outstanding issue with these where the global styles don't load unless you first go to a component page, so the buttons look wonky. Not sure yet if that will be a fix in this PR or a separate, more global one

components/button/index.css
jawinn: Namespacing might be necessary if there are cases where a component with a down state is nested within another component that has a down state? I'm not sure.

Maybe we could use --mod naming on these? `--mod-button-downstate-height`?

.storybook/foundations/down-state.mdx
```diff
@@ -0,0 +1,24 @@
+import { Meta, Story } from '@storybook/blocks';
+import * as ButtonStories from '../../components/button/stories/button.stories.js';
+import * as CheckboxStories from '../../components/checkbox/stories/checkbox.stories.js';
+
+<Meta title="Foundations/Down state" />
+
+<h1>Down state</h1>
+
+Down state is a Spectrum 2 (S2) feature that creates the illusion of components being pressed-in when active. This functionality is already included in S2 components that require down state in this project. It is implemented with a CSS transform. The implementation depends on the size of the interactable element:
```
jawinn, line 9: Should we spell out "Spectrum 2" each time rather than writing "S2"?

.storybook/foundations/down-state.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+  - If the interactable element has a width of 24px or less, we use the minimum perspective token:
```
jawinn, line 11: Formatting-wise for this section, I'd drop the bullet points, and add an "Examples" heading above the stories. 

Another idea would be to have two separate headings for the minimum perspective and the max formula, with each relevant story moved under each. 

It might also be worth noting the values of the two custom variables being set on the button stories in text above it.

jawinn (commented):
I don't see a problem with defining these in each component.

For the max formula, you mentioned that maybe we could set these for some components. 
> But if we go with the component-based approach shown in this PR, we may be able to set these and ship them as part of the CSS instead.
What if we had fallback default values for the components, for when the width and height variables aren't set? Button for example might use the defined minimum component height as another fallback on downstate height. This could vary per component. Some might be able to be set directly if the component has a fixed size. Just a thought. It may not be necessary if we are expecting the consumer to always set these (per your discussion here).

Rajdeepc:
> @Rajdeepc Good questions!
> 
> 1. Yes, `:active` is the way we can simulate a "down state" in CSS. From [the design docs](https://s2.spectrum.corp.adobe.com/page/states/#down-state): down state is the concept to make a component look "pressed in" when activated by a user. We're using that phrase "down state" to align with design.
> 2. We don't have a `:pressed` pseudo class available in CSS, so we're using the `:active` [CSS pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:active). It should be equivalent to the idea of a pressed state, but let me know if that's not the case with web components
> 
> I agree, we should do some testing on mobile as well once the team agrees on the approach.
> 
> Let me know if this answers your questions 🙂

Sure thanks. This make sense. Also curious to know how are you shipping javascript? We haven't had a chance to discuss this but if you had any discussion do let me know. Since we are building a logic to tackle this from SWC side we would love to see an example.

.storybook/foundations/down-state.mdx
rise-erpelding: Was this a known issue before or something you discovered for the first time adding this story? I noticed this when creating a new story for the Checkboxes and Action Buttons in Foundations for corner-rounding as well.

mdt2: I think it has to do with the existing setup and where we're importing global tokens (probably in preview.js). It can probably be its own Jira card to work on.

.storybook/foundations/down-state.mdx
mdt2: > It might also be worth noting the values of the two custom variables being set on the button stories in text above it, since those are only visible if you use the inspector.

I'm not clear on this, do you mean the width and height tokens?

jawinn: Yeah, I was thinking that it could say something like:

In this example, the width and height tokens are set to the following with an inline style: `--spectrum-downstate-width:72px; --spectrum-downstate-height: 32px;`

components/button/stories/button.stories.js
```diff
@@ -12,6 +13,7 @@ export default {
 	description:
 		"Buttons allow users to perform an action or to navigate to another page. They have multiple styles for various needs, and are ideal for calling attention to where a user needs to do something in order to move forward in a flow.",
 	component: "Button",
+	decorators: [withDownStateDimensionCapture('.spectrum-Button:not(:disabled)')],
```
mdt2, line 16: This will need to be added to all stories of components that require the capture of width/height as they migrate to S2. I'll try to note that in the appropriate Jira tickets.

mdt2:
Question about components in which button and checkbox are a dependency (alert banner, card). I'm assuming we don't want down state to be included in those components until they're migrated to S2, yes? Either way, this feature is probably a breaking change, so for the components that use button and checkbox, we just won't update their dependencies to this feature version until migration to S2, is that correct?

rise-erpelding (commented):
Excellent job handling all of the complexities related to capturing the dimensions and applying the formulas! The down states appear to be working in Storybook for both checkboxes and buttons, and I can see that the decorator is working to assign the correct values for the width and height custom properties.

I have minor documentation suggestions for the tokens in Storybook; otherwise, everything looks fantastic!

jawinn (changes requested):
Nice work! This addresses my previous comments about formatting. And I am seeing the down state working as intended on the Button and Checkbox stories. With Rise's suggestions, I think the Foundations docs page is good to go.

One thing I noticed is that the down state is happening when clicking on the disabled Checkbox, in the Checkbox->Default story. I'm assuming that the down state animation should not appear when it is disabled.

.storybook/foundations/down-state/down-state.mdx
```diff
@@ -0,0 +1,39 @@
...
+import * as Checkbox from './checkbox-down-state.stories.js';
+import * as Button from './button-down-state.stories.js';
+
+<Meta title="Foundations/Down state" />
+
+<h1>Down state</h1>
+
+Down state is a Spectrum 2 feature that creates the illusion of components being pressed-in when active. This functionality is already included in Spectrum 2 components that require down state in this project. It is implemented with a CSS transform. The implementation depends on the size of the interactable element, as shown in the examples below.
+
+<h2>Examples</h2>
```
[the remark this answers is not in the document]
mdt2: Definitely prefer consistency, good catch. I'll update to use markdown, it looks like that's what we're using in the Storybook guides files as well 👍

.storybook/foundations/down-state/down-state.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+<h3>Minimum perspective</h3>
+
+The checkbox component is an example of an element that has a width of 24px or less by default.
```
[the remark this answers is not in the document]
mdt2: Sure thing! Thanks for including a wording suggestion, that helps.

pfulton:
> Question about components in which button and checkbox are a dependency (alert banner, card). I'm assuming we don't want down state to be included in those components until they're migrated to S2, yes? Either way, this feature is probably a breaking change, so for the components that use button and checkbox, we just won't update their dependencies to this feature version until migration to S2, is that correct?

Great question about the semantics of how we'd release this! I'm thinking about it like this:
1. Anything migrated to S2 will be a breaking change for that initial migration bit. (meaning, the component points to the S2 tokens rather than the S1 tokens)
2. This PR is adding a feature, so normally, I'd say it's a minor bump (at least), BUT, I think the S2 migration bit from my first point overrides that; it'll be a major.

As for components that use button and checkbox as dependencies: yeah, let's just wait for those for now.

Can you update the `package.json` for Button and Checkbox to [14.0.0-next.2](https://www.npmjs.com/package/@spectrum-css/tokens/v/14.0.0-next.2)?

jawinn (approved):

pfulton (approved):

components/checkbox/index.css
```diff
@@ -150,6 +150,12 @@ governing permissions and limitations under the License.
     }
   }
 
+  &:not(.is-readOnly):active {
```
Rajdeepc, line 153: This approach will not work in case the SWC wants to disable downstate for checkbox. This is tightly coupled with the css transform. We need to cascade the `--spectrum-downstate-width` and `--spectrum-downstate-height` to this css.

mdt2: Hey Rajdeep! Can you help me understand what needs to change? One piece of context that might be helpful is for components like Checkbox which are smaller than 24px, the transform perspective is not based on the width and height of the component, it is based on a set value [per the design spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9069-310&t=CQ2ikZHR2mb5Ku6g-0).

PR #2523, opened 2024-02-15
docs: move release instructions to Storybook

castastrophe:
## Description

In an effort to move developer-focused and longer-form documentation into the Storybook Guides section, this PR converts the release steps from the root README.md into a stand-alone guide.

## To-do list


pfulton (approved):
This is great, and is now making me wonder if we should also find a way to talk about the fact that we don't really have a release schedule—we just release on an as-needed basis, basically.

I think we can find the right words, the right time, and the right place for that info, but it's something for us to consider as we go forward with providing docs like this!

PR #2530, opened 2024-02-15
chore(docs): increase dark mode compatibility

mdt2:
## Description

Updating the pug templates to:

- make the deprecationNotice part of the deprecation alert dark mode compatible
- make links in the deprecationNotice and description fields dark mode compatible

### Validation steps

- [x] In the docs site, go to Quick Actions and turn on dark mode. All text in the deprecation notice should be white other than the link (but the link should also be using dark mode colors)
- [x] Go to Accordion, scroll down to the Custom Properties API section. With dark mode on, the link should be dark mode compatible

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.storybook/deprecated/quickaction/quickaction.yml
+++ b/.storybook/deprecated/quickaction/quickaction.yml
@@ -1,7 +1,7 @@
 id: quickactions
 name: Quick actions
 status: Deprecated
-deprecationNotice: Use an <a href="actionbar.html">action bar</a> to allow users to perform actions on either a single or multiple items at the same time, instead.
+deprecationNotice: Use an <a class="spectrum-Link" href="actionbar.html">action bar</a> to allow users to perform actions on either a single or multiple items at the same time, instead.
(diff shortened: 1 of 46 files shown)


site/includes/exampleContent.pug
```diff
@@ -1,6 +1,9 @@
+-
+  const withThemeCompatibleLinks = (markdown) => markdown.replace(/<a href/g, '<a class="spectrum-Link" href');
```
castastrophe, line 2: Doesn't this end up being a general problem for any markdown links rendered? I think instead of trying to inject the change here, maybe we just update the links in the docs to use the right class? We'll be fixing this programmatically with some of the markdown updates 11ty will bring with it.

mdt2: Yeah it's a problem for any that don't manually include the class. My thought was that doing it programmatically like this avoids human error, but if it'll be fixed in 11ty anyway then yeah I can adjust the links in the docs instead

mdt2: Ohhh it looks like Pug is handling this already when we use markdown instead of HTML (Accordion usage notes as an example). Should I spend the time now to update all of those mod links to use markdown?

castastrophe (approved):

PR #2559, opened 2024-02-26
feat(cornerrounding): add custom token and foundations documentation

jenndiaz:
## Description

S2 migration work for corner rounding tokens. Includes:

- Foundations page for documenting corner rounding as a concept for consumers (according to direction from this Slack thread https://adobedesign.slack.com/archives/G019JTYMT6H/p1708721169366349)
- Checkbox, action button component changes to partially migrate to only include new corner rounding
- Close button update to use the new custom token from #2597 

### Validation steps
- [x] The new and updated corner rounding tokens from [the design spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=8078-675&mode=design&t=DLupOmAesiqeoMyr-0) are set to the appropriate values
- [x] In Storybook, Corner Rounding has a Foundations docs page with a doc that explains the tokens in a way that makes sense 
- [x] Checkbox component includes new corner rounding tokens
- [x] Action button includes new corner rounding tokens
- [ ] Close button uses the new custom token `--spectrum-corner-radius-1000` for border-radius (
[Jenn's code pen](https://codepen.io/forthewinn/pen/zYbLGmp) that shows why this is needed for posterity)
- [x] Design team has reviewed the docs page @mdt2 ([approved by Matt Knorr in Slack](https://adobedesign.slack.com/archives/GQ09RCBA5/p1711402667435969?thread_ts=1708116176.602529&cid=GQ09RCBA5))

## To-do list

- [X] I have updated relevant storybook stories and templates.
- [X] If my change impacts **documentation**, I have updated the documentation accordingly.
- [x] Dependencies have a tokens version set that includes the custom-vars additions

### DIFF
--- a/.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
+++ b/.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
@@ -1,0 +1,110 @@
+import { html } from "lit";
+import { Template } from "../../../components/actionbutton/stories/template";
+
+export default {
+	title: "Foundations/Corner rounding",
+	description:
+		"The action button component represents an action a user can take.",
+	component: "ActionButton",
+	args: {
+		rootClass: "spectrum-ActionButton",
+	},
+	parameters: {
+		actions: {
+			handles: ["click .spectrum-ActionButton:not([disabled])"],
+		},
+		status: {
+			type: process.env.MIGRATED_PACKAGES.includes("actionbutton")
+				? "migrated"
+				: undefined,
+		},
+	},
+	tags: ['foundation'],
+};
+
+const ActionButton = ({
+	...args
+}) => {
+	return html`
+		<div style="padding: 1rem 0;">
+			${Template({
+				...args,
+				iconName: undefined,
+			})}
+		</div>
+	`;
+};
+
+const ActionButtonTable = ({ ...args }) => {
+	return html`
+		<table class="spectrum-Foundations-Example-CornerRounding-table">
+			<thead>
+				<tr>
+					<th scope="col">Token</th>
+					<th scope="col" style="padding: 0 2rem;">Value</th>
+					<th scope="col">Medium example</th>
+				</tr>
+			</thead>
+			<tbody>
+				<tr>
+					<td>--spectrum-corner-radius-medium-size-extra-small</td>
+					<td style="padding: 0 2rem;">6px</td>
+					<td>
+						${ActionButton({
+							...args,
+							label: "Extra Small",
+							size: "xs"
+						})}
+					</td>
+				</tr>
+				<tr>
+					<td>--spectrum-corner-radius-medium-size-small</td>
+					<td style="padding: 0 2rem;">7px</td>
+					<td>
+						${ActionButton({
+							...args,
+							label: "Small",
+							size: "s"
+						})}
+					</td>
+				</tr>
+				<tr>
+					<td>--spectrum-corner-radius-medium-size-medium</td>
+					<td style="padding: 0 2rem;">8px</td>
+					<td>
+						${ActionButton({
+							...args,
+							label: "Medium",
+							size: "m"
+						})}
+					</td>
+				</tr>
+				<tr>
+					<td>--spectrum-corner-radius-medium-size-large</td>
+					<td style="padding: 0 2rem;">9px</td>
+					<td>
+						${ActionButton({
+							...args,
+							label: "Large",
+							size: "l"
+						})}
+					</td>
+				</tr>
+				<tr>
+					<td>--spectrum-corner-radius-medium-size-extra-large</td>
+					<td style="padding: 0 2rem;">10px</td>
+					<td>
+						${ActionButton({
+							...args,
+							label: "Extra Large",
+							size: "xl"
+						})}
+					</td>
+				</tr>
+			</tbody>
+		</table>
+	`;
+};
+
+export const ActionButtonExamples = ActionButtonTable.bind({});
+ActionButtonExamples.args = {};
--- a/.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
+++ b/.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
@@ -13,11 +13,6 @@ export default {
 		actions: {
 			handles: ["click .spectrum-ActionButton:not([disabled])"],
 		},
-		status: {
-			type: process.env.MIGRATED_PACKAGES.includes("actionbutton")
-				? "migrated"
-				: undefined,
-		},
 	},
 	tags: ['foundation'],
 };
--- a/.storybook/foundations/corner-rounding/corner-rounding.mdx
+++ b/.storybook/foundations/corner-rounding/corner-rounding.mdx
@@ -1,0 +1,26 @@
+import { Meta, Canvas, Story, Title, Unstyled } from '@storybook/blocks';
+import * as CornerRoundingStories from './corner-rounding.stories.js';
+import * as CheckboxStories from './checkbox-corner-rounding.stories.js';
+import * as ActionButtonStories from './action-button-corner-rounding.stories.js';
+
+<Meta of={CornerRoundingStories} />
+
+<Title />
+
+## Default values
+
+<Story of={CornerRoundingStories.CornerRounding} />
+
+## Component examples
+
+### Small tokens
+
+Checkbox includes small component size tokens to scale corner rounding.
+
+<Story of={CheckboxStories.CheckboxExamples} />
+
+### Medium tokens
+
+Action button includes medium component size tokens to scale corner rounding.
+
+<Story of={ActionButtonStories.ActionButtonExamples} />
--- a/.storybook/foundations/corner-rounding/corner-rounding.stories.js
+++ b/.storybook/foundations/corner-rounding/corner-rounding.stories.js
@@ -1,0 +1,75 @@
+// Import the component markup template
+import { html } from "lit";
+import { Template } from './template';
+
+export default {
+	title: "Foundations/Corner rounding",
+	description:
+		"Corner rounding is a foundation that shows the different border-radius tokens that can be applied to a component.",
+	component: "Corner rounding",
+	args: {
+		rootClass: "spectrum-Foundations-Example-CornerRounding",
+	},
+  tags: ['foundation'],
+};
+
+const CornerRadiusGroup = ({
+	customStyles = {},
+	...args
+}) => {
+	return html`
+		<div>
+      <table class="spectrum-Foundations-Example-CornerRounding-table">
+        <thead>
+          <tr>
+            <th scope="col">Token</th>
+            <th scope="col" style="padding: 0 2rem;">Value</th>
+            <th scope="col">Example<br>(No Border)</th>
+            <th scope="col">Example<br>(Border)</th>
+          </tr>
+        </thead>
+        <tbody>
+			${Template({
+				...args,
+        label: '--spectrum-corner-radius-none',
+        size: 'none',
+        value: '0px',
+			})}
+      ${Template({
+				...args,
+        label: '--spectrum-corner-radius-small-default',
+        size: 's',
+        value: '4px',
+			})}
+      ${Template({
+				...args,
+        label: '--spectrum-corner-radius-medium-default',
+        size: 'm',
+        value: '8px',
+			})}
+      ${Template({
+				...args,
+        label: '--spectrum-corner-radius-large-default',
+        size: 'l',
+        value: '10px',
+			})}
+      ${Template({
+				...args,
+        label: '--spectrum-corner-radius-extra-large-default',
+        size: 'xl',
+        value: '16px',
+			})}
+      ${Template({
+				...args,
+        label: '--spectrum-corner-radius-full',
+        size: 'full',
+        value: '9999px',
+			})}
+        </tbody>
+      </table>
+		</div>
+	`;
+};
+
+export const CornerRounding = CornerRadiusGroup.bind({});
+CornerRounding.args = {};
--- a/components/closebutton/index.css
+++ b/components/closebutton/index.css
@@ -20,8 +20,7 @@ governing permissions and limitations under the License.
 @import "@spectrum-css/commons/basebutton.css";
 
 .spectrum-CloseButton {
-  /* TO DO: replace with custom var during corner rounding work */
-  --spectrum-closebutton-border-radius: 9999px;
+  --spectrum-closebutton-border-radius: var(--spectrum-corner-radius-full);
 
   /* Cross icon */
   --spectrum-closebutton-icon-color-default: var(--spectrum-neutral-content-color-default);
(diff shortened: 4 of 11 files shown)


.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
```diff
@@ -0,0 +1,111 @@
+import { html } from "lit";
```
rise-erpelding, line 1: I really don't like this approach of adding separate stories for Checkbox and ActionButton but as the [guidance on this in the Slack thread](https://adobedesign.slack.com/archives/GQ09RCBA5/p1708130655322169?thread_ts=1708116176.602529&cid=GQ09RCBA5) stated:

> Component size tokens:
small – four checkboxes in small, medium, large, extra-large using these tokens
medium – five action buttons in extra-small, small, medium, large, extra-large using these tokens

This does technically fulfill that requirement, but I'm going to keep looking to see if there's a less awkward approach to get there.

.storybook/foundations/corner-rounding/corner-rounding.mdx
```diff
@@ -9,3 +11,18 @@ import * as CornerRoundingStories from './corner-rounding.stories.js';
 
 <Story of={CornerRoundingStories.CornerRounding} />
 
+
+## Component side tokens
```
rise-erpelding, line 15: Open to suggestions on how we might show or explain this better! I'm going to keep thinking on this one as well.

tokens/custom-spectrum/custom-vars.css
```diff
@@ -36,9 +36,9 @@ governing permissions and limitations under the License.
   --spectrum-animation-ease-out: cubic-bezier(0, 0, 0.40, 1);
   --spectrum-animation-ease-linear: cubic-bezier(0, 0, 1, 1);
 
-   /* TODO: refactor to use a translation layer that would adjust current `corner-radius-1000` to be 9999px instead of 0.5 */
-  /* should be used in place of `--spectrum-corner-radius-1000` and -`-spectrum-corner-radius-full`*/
   --spectrum-corner-radius-1000-full: 9999px;
+  --spectrum-corner-radius-full: var(--spectrum-corner-radius-1000-full);
+  --spectrum-corner-radius-1000: var(--spectrum-corner-radius-1000-full);
```
rise-erpelding, line 41: This overrides the current `--spectrum-corner-radius-1000` and `--spectrum-corner-radius-full` with the value we want, if there's a different way to do this that I overlooked, let me know!

mdt2: I think this is fine for now... Do you remember if Jenn said anything about where she left this? Had she started talking to Garth about getting this added to the tokens system?

jawinn (changes requested):
Nice work on creating the MDX page to demo the different rounding. This looks just like what we were trying to achieve.

One small thing, could we put some more padding so that 16px is not so close to the token name? It might also be worth adding a table header to describe the columns too.
Screenshot 2024-03-11 at 12 59 54 PM

.storybook/foundations/corner-rounding/component-size-tokens-medium.mdx
```diff
@@ -0,0 +1,42 @@
+<table>
```
rise-erpelding, line 1: It sounds like [there might be a plugin out there](https://github.com/storybookjs/storybook/issues/20091) that would enable us to write the tables in markdown and not HTML, but out of the box our implementation doesn't work with markdown tables.

.storybook/foundations/corner-rounding/corner-rounding.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+<ComponentSizeTokensMedium />
+
+These component size tokens are used to scale corner rounding in different-sized tokens. For instance, the corner-rounding for a checkbox will be `4px` for a medium-sized checkbox, `3px` for small, `5px` for large, and `6px` for an extra-large checkbox, corresponding to the small component size tokens table above.
+
+### Checkbox corner rounding
+
+Checkbox will use small-sized component size tokens to scale corner rounding.
+
+Note that the story below has not yet been updated to show corner rounding that scales with component size.
```
rise-erpelding, line 34: [Design wants to show the scaling for corner-rounding on different components](https://adobedesign.slack.com/archives/GQ09RCBA5/p1708116176602529) by showing components in different sizes with differently-sized corner-rounding, but our components don't have this baked into them yet. I've written down as a to do for myself to make sure this part of the documentation gets updated once those components are migrated and displaying as design intended.

.storybook/foundations/corner-rounding/corner-rounding.stories.js
```diff
@@ -0,0 +1,75 @@
...
+
+export default {
+	title: "Foundations/Corner Rounding",
+	description:
+		"Corner rounding is a foundation that shows the different border-radius tokens that can be applied to a component.",
+	component: "Corner Rounding",
+	args: {
+		rootClass: "spectrum-Foundations-Example-CornerRounding",
+	},
+  tags: ['foundation'],
```
rise-erpelding, line 13: The `foundation` tag to hide the story from the sidebar so the primary function of the story is just to exist within the Corner Rounding Foundations doc.

I think it might be possible to make an .mdx file with the table similar to the component size tokens, but incorporating the CSS appropriately was a bit of a challenge when I tried this approach.

.storybook/manager.js
```diff
@@ -53,5 +53,10 @@ addons.setConfig({
 	}),
 	sidebar: {
 		showRoots: false,
+		filters: {
+			patterns: (item) => {
+				return !item.tags.includes('foundation');
```
rise-erpelding, line 58: Hides stories with this tag from the sidebar

.storybook/foundations/corner-rounding/action-button-corner-rounding.stories.js
mdt2: We chatted about this in dev huddle, decided to go with [this API solution](https://github.com/storybookjs/storybook/issues/9209#issuecomment-1866309500) to hide the documentation stories from the sidebar. We should probably still check if Chromatic sees these...

tokens/custom-spectrum/custom-vars.css
mdt2: We chatted about this in a Slack DM with Patrick Fulton and determined that the current setup in this PR is ok for now 👍

mdt2 (commented):
Thanks for your work on this Rise!! A couple comments for you born from my getting a little caught on documentation semantics 😅

.storybook/foundations/corner-rounding/corner-rounding.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+<Title />
+
+## Alias tokens
+
+<Story of={CornerRoundingStories.CornerRounding} />
+
+## Component size tokens
+
+In addition to the alias tokens listed above, Spectrum 2 also provides component size alias tokens that align with global size tokens:
```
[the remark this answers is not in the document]
jawinn: I have some similar thoughts on simplifying the table displayed here as well. I think the table should only have two columns, "Component Size Token" and "Value". I'm a little unclear about the purpose of the "Alias Token (if available)" column here.

.storybook/foundations/corner-rounding/corner-rounding.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+### Small component size tokens
+
+<ComponentSizeTokensSmall />
+
+### Medium component size tokens
+
+<ComponentSizeTokensMedium />
+
+These component size tokens are used to scale corner rounding in different-sized tokens. For instance, the corner-rounding for a checkbox will be `4px` for a medium-sized checkbox, `3px` for small, `5px` for large, and `6px` for an extra-large checkbox, corresponding to the small component size tokens table above.
```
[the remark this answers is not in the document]
jawinn: While we use the phrase "t-shirt size" internally, is this common for the public facing docs? I know in the components docs and Storybook it is just called "Sizes" or "Sizing", and I see some component Guidelines pages for Spectrum 1 that just say "Sizes" as well. I'm not sure if Spectrum 2 is making more use of that terminology or not.

Also, I'd try to keep "extra large" instead of "x-large" where possible, to be extra clear when considering screen readers or readers less fluent in english.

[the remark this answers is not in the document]
mdt2: Great points, thank you!

.storybook/foundations/corner-rounding/corner-rounding.stories.js
```diff
@@ -0,0 +1,75 @@
+// Import the component markup template
+import { html } from "lit";
+import { Template } from './template';
+
+export default {
+	title: "Foundations/Corner Rounding",
```
jawinn, line 6:
```suggestion
	title: "Foundations/Corner rounding",
```
Just a minor suggestion here to keep title case on the story name. This would be consistent with the capitalization on the Down state PR as well.

jawinn:
I just noticed that when I scrolled to the bottom of the Storybook sidebar, there is a section appearing for corner rounding. Is there a way to hide these?
Screenshot 2024-03-15 at 10 45 12 AM

.storybook/foundations/corner-rounding/corner-rounding.mdx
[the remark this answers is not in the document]
rise-erpelding: My thought in having the tables was that if I were building (or I guess even migrating) a component that required corner-rounding, for instance a checkbox, I'd want to know that there were all these nice component size tokens available to use on each size variant of the checkbox.

Since we're moving toward medium size being the default, I'd want to know that I probably should use the alias token `--spectrum-corner-radius-small-default` for corner-rounding on the default checkbox, but then use the component size tokens like `--spectrum-corner-radius-small-size-large` for the large size checkbox, and other component size tokens for other sizes.

Given that, maybe some sort of note on using the alias token for default sizes would be nice here?

We didn't get a specific ask from design to show all the non-alias tokens, but the design spec does line up all the different tokens (alias, component size, global) with the values so you can see where they fall, and that seemed like it might be a nice thing to try to capture here, but I can also see how that might be unnecessary--it doesn't seem like we typically include that level of detail in Storybook.

Does that help either of you lean one way or another about whether to include the tables/what to include on them?

components/closebutton/index.css
```diff
@@ -18,8 +18,7 @@ governing permissions and limitations under the License.
 @import "@spectrum-css/commons/basebutton.css";
 
 .spectrum-CloseButton {
-  /* TO DO: replace with custom var during corner rounding work */
-  --spectrum-closebutton-border-radius: 9999px;
+  --spectrum-closebutton-border-radius: var(--spectrum-corner-radius-1000-full);
```
rise-erpelding, line 21: Stumbled upon this TODO comment that probably needs to be addressed in this PR?

Using the token seems to work fine in Storybook, but the linter doesn't seem very happy about it--if I should be using a different token, I'd be happy to change this.

.storybook/foundations/corner-rounding/corner-rounding.mdx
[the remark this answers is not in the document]
mdt2: Thanks for this excellent explanation. I was confused about the use of non-alias tokens, but since they'll be used in some cases we definitely want to surface them here as you've done 👏 

Like you suggested, I'm going to add a note to use alias tokens when possible.

components/closebutton/index.css
jawinn: Should we use `--spectrum-corner-radius-full` in this component and future components, since we're now overriding its value?:
`--spectrum-closebutton-border-radius: var(--spectrum-corner-radius-full);`

tokens/custom-spectrum/custom-vars.css
```diff
@@ -36,6 +36,9 @@ governing permissions and limitations under the License.
   --spectrum-animation-ease-out: cubic-bezier(0, 0, 0.40, 1);
   --spectrum-animation-ease-linear: cubic-bezier(0, 0, 1, 1);
 
+  --spectrum-corner-radius-1000-full: 9999px;
+  --spectrum-corner-radius-full: var(--spectrum-corner-radius-1000-full);
+  --spectrum-corner-radius-1000: var(--spectrum-corner-radius-1000-full);
```
jawinn, line 41:
```suggestion
--spectrum-corner-radius-1000: 9999px;
```
Thoughts on changing these three lines of changes into this single line? I noticed that the `--spectrum-corner-radius-full` alias is already set to the value of `--spectrum-corner-radius-1000` in the s2 tokens. And then we would shouldn't need a separate `--spectrum-corner-radius-1000-full` if we're using `--spectrum-corner-radius-full` in our components' CSS.

pfulton: I think that makes sense. The cascade is going to be important here to ensure that `--spectrum-corner-radius-1000` is actually resolving to our `9999px` instead of the Spectrum Tokens value (that is, until we get a translation up and running to transform it for our needs).

components/closebutton/index.css
mdt2: Good callout, due to [this change](https://github.com/adobe/spectrum-css/pull/2559#discussion_r1530724751), I'll update this to `--sectrum-corner-radius-1000`.

tokens/custom-spectrum/custom-vars.css
mdt2: Note for posterity: Patrick handled this in a tokens PR (#2597).

components/closebutton/index.css
```diff
@@ -18,8 +18,7 @@ governing permissions and limitations under the License.
 @import "@spectrum-css/commons/basebutton.css";
 
 .spectrum-CloseButton {
-  /* TO DO: replace with custom var during corner rounding work */
-  --spectrum-closebutton-border-radius: 9999px;
+  --spectrum-closebutton-border-radius: var(--spectrum-corner-radius-1000);
```
jawinn, line 21:
```suggestion
  --spectrum-closebutton-border-radius: var(--spectrum-corner-radius-full);
```

jawinn (commented):
This looks great! I think we're ready to move forward with this. 

The only thing I see left to do is to update the minimum required version number for the `@spectrum-css/tokens` peer dependencies in Action button, Checkbox, Close button, and Storybook itself. Once the next beta of the tokens releases tomorrow, we should be able to set that as the minimum version.

components/closebutton/index.css
mdt2: Ah yes, thank you for catching this 🙌

components/actionbutton/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@spectrum-css/actionbutton",
-  "version": "5.2.7-next.1",
+  "version": "14.0.0-next.2",
```
mdt2, line 3: Am I misunderstanding how we want to version packages? Should this be:

```suggestion
  "version": "5.2.7-next.2",
```

jawinn (approved):

PR #2564, opened 2024-02-27
feat(closebutton)!: migrate to S2

mdt2:
## Description

- Migrate close button to S2 per [design spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=9503-23185&mode=design&t=MUQYhyd5cziDsZHR-0).

- Includes a refactor that removes all static-specific mods since they aren't needed:
    ```
    | `--mod-closebutton-static-background-color-default` |
    | `--mod-closebutton-static-background-color-down`    |
    | `--mod-closebutton-static-background-color-focus`   |
    | `--mod-closebutton-static-background-color-hover`   |
    ```

    Consumers can use these existing background-color mods instead:
    ```
    | `--mod-closebutton-background-color-default` |
    | `--mod-closebutton-background-color-down`    |
    | `--mod-closebutton-background-color-focus`   |
    | `--mod-closebutton-background-color-hover`   |
    ```

- Also removes `--mod-closebutton-size` since we'll be setting rounded border-radius with a single rounding token, to be formally implemented as part of [S2 Corner Rounding](https://github.com/adobe/spectrum-css/pull/2559/files#diff-d6e9caa794bd8e1c659425d223e23ab53a93a254f00fb11a626e1b7563bd0c61R39-R41).

### Validation steps

- [ ] Design validation @mdt2 
- [ ] In storybook:
    - [ ] close button is the correct size at all t-shirt sizes
    - [ ] close button matches the [design spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=9503-23185&mode=design&t=MUQYhyd5cziDsZHR-0)

### Regression testing

Validate:

## To-do list


components/closebutton/metadata/closebutton.yml
```diff
@@ -28,66 +22,6 @@ examples:
...
-            <button class="spectrum-CloseButton spectrum-CloseButton--sizeXL">
-              <svg class="spectrum-CloseButton-UIIcon spectrum-Icon spectrum-UIIcon-Cross300" focusable="false" aria-hidden="true">
-                <use xlink:href="#spectrum-css-icon-Cross300" />
-              </svg>
-            </button>
-          </div>
-        </div>
-      </div>
-  - id: closebutton-largeicon
-    name: Icon Size - Large
```
mdt2, line 76: Per[ the changelog](https://www.figma.com/file/M9P2s2IDhGWZuNyvJtXrx3/S2-foundations-and-components-changelog?type=design&node-id=740-1679&mode=design&t=4cIhoO8rTWmFDwRy-0), icon size goes away in S2

components/closebutton/metadata/closebutton.yml
```diff
@@ -1,25 +1,19 @@
...
-      | `.spectrum-CloseButton--sizeS`  | `.spectrum-UIIcon-Cross75`  |
-      | `.spectrum-CloseButton--sizeM`  | `.spectrum-UIIcon-Cross100` |
-      | `.spectrum-CloseButton--sizeL`  | `.spectrum-UIIcon-Cross200` |
-      | `.spectrum-CloseButton--sizeXL` | `.spectrum-UIIcon-Cross300` |
-
-      ### Remove focus-ring class
-      We've migrated away from the focus-ring class in favor of the native `:focus-visible` pseudo-class due to changes in browser support.
 examples:
   - id: close
     name: Icon Size - Regular
```
rise-erpelding, line 19: Since we're removing the "Icon Size - Large" heading, I'm wondering if there are guidelines for changing this text as well? Looking at other components, it seems like "Sizing" could be a more appropriate heading here.

jawinn (changes requested):
This is looking pretty good so far. 

One thing I noticed is that it looks like the "Cross icon (static white)" and "Cross icon (static black)" tokens on the Figma may still need to be set within the `.spectrum-CloseButton--staticWhite` and `.spectrum-CloseButton--staticBlack` classes. I didn't see any CSS that had the `transparent-white-800` and `transparent-white-900` for example.

The other issue may be build related, but I'm not seeing the hover background color on the docs page. It looks fine in Storybook.

**Cleanup of "Hardcoded" tokens in CSS**
The other suggestion I had when looking at the previously existing CSS, is that we can probably remove all of these variables:

--spectrum-closebutton-size-300
--spectrum-closebutton-size-400
--spectrum-closebutton-size-500
--spectrum-closebutton-size-600
--spectrum-closebutton-size

All they seem to be used for is setting the border-radius to rounded. We can handle that with a single `9999px` rounding token, like the one being proposed in PR #2559 . Maybe for now use a temporary token to be replaced with the `--spectrum-corner-radius-1000-full` when it is available?

components/closebutton/metadata/closebutton.yml
```diff
@@ -1,25 +1,19 @@
...
-  A button used to close or dismiss components.
+  A button used to close or dismiss components. Close button supports different sized icons using these class couplings:
+
+  | Component size  | Close button classname          | UI icon classname           |
+  | --------------  | ------------------------------- | --------------------------- |
+  | Small           | `.spectrum-CloseButton--sizeS`  | `.spectrum-UIIcon-Cross400` |
+  | Medium (default)| No size class for default       | `.spectrum-UIIcon-Cross500` |
+  | Large           | `.spectrum-CloseButton--sizeL`  | `.spectrum-UIIcon-Cross400` |
+  | X-Large         | `.spectrum-CloseButton--sizeXL` | `.spectrum-UIIcon-Cross500` |
+
```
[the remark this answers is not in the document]
mdt2: 😂 that's what I get for using replace all. Thanks for catching this

components/closebutton/index.css
```diff
@@ -231,55 +245,6 @@ a.spectrum-CloseButton {
   }
 }
 
-/* Modifier Classes */
-.spectrum-CloseButton--staticBlack:not(:disabled),
-.spectrum-CloseButton--staticWhite:not(:disabled) {
```
mdt2, line 236: I don't think we need these? We should be able to redefine the token values [like I've done here](https://github.com/adobe/spectrum-css/pull/2564/files#diff-a4409bdab1797d61664b594472edb087eeb1c2c2ce52c0c14b68dc4b5fe574c7R75-R109) and just rely on that for the static background updates.

components/closebutton/metadata/mods.md
```diff
@@ -1,33 +1,28 @@
...
-| `--mod-closebutton-icon-color-down`                 |
-| `--mod-closebutton-icon-color-focus`                |
-| `--mod-closebutton-icon-color-hover`                |
-| `--mod-closebutton-margin-inline`                   |
-| `--mod-closebutton-margin-top`                      |
-| `--mod-closebutton-size`                            |
-| `--mod-closebutton-static-background-color-default` |
-| `--mod-closebutton-static-background-color-down`    |
-| `--mod-closebutton-static-background-color-focus`   |
-| `--mod-closebutton-static-background-color-hover`   |
```
mdt2, line 30: Some of these are going away. I figured it would be ok to just remove instead of following a deprecation flow for these since this is a breaking change already, but I'm open to thoughts!

components/closebutton/index.css
jawinn: Nice refactor that makes better use of CSS variables 👍

components/closebutton/metadata/mods.md
jawinn: I think these are fine to be going away; I saw the bg color ones would just use the existing mods and don't need a static specific mod. And the size was not changing anything. It'd probably be a good idea to document these removals in the PR at least, especially the bg colors and the alternative mod name that takes their place.

mdt2: Good call, I've added those to the PR description 👍

jawinn (changes requested):
Those updates look good! That resolves my previous comments about the icon color and the corner rounding.

Two last questions / suggestions:
- Should the `express.css` and `spectrum.css` files be deleted completely, now that they're no longer required in the build process?
- Could your notes about the mod change in the PR be added to the migration guide? Since consumers could be relying on those old mod properties. Here's [a general idea for the formatting of the S2 migration guide note](https://github.com/adobe/spectrum-css/pull/2569/files#diff-72afc31bc9ab2a614ffa71d6510367743b25930798808ef19e146e95ba98f50bR9-R11), if that's helpful.

jawinn (approved):

Issue #2566, opened 2024-02-28
[Picker][RTL] Spinner margins are wrong


Rocss:
## Description

If you're using the Picker in RTL direction, the spinner has a margin added between the spinner and the menu icon.
This margin should actually be between the spinner and the text.

## Steps to reproduce

1. Go to https://opensource.adobe.com/spectrum-css/preview/?path=/story/components-picker--default&args=isLoading:!true&globals=textDirection:rtl
3. Observe the space between the menu icon and the spinner is bigger than it should be.

## Expected behavior

The `margin-inline-start` property should be applied for the space between the spinner and the text. However, because of the `direction: ltr` property set on the spinner, this margin is applied wrong.

## Screenshots

![Screenshot 2024-02-28 at 12 10 31](https://github.com/adobe/spectrum-css/assets/13311865/d4d63af3-1f2c-4d70-b618-7f8f5dd2e1c2)

Invalid icon placement, for reference:
![Screenshot 2024-02-28 at 12 14 26](https://github.com/adobe/spectrum-css/assets/13311865/600f4e0e-94bf-40d3-85bd-db3111179e42)

## Environment

-   **Version of the impacted component(s):** 
-   **Browser(s) and OS(s):** 

## Additional context

Tracked in https://jira.corp.adobe.com/browse/CSS-701

PR #2567, opened 2024-02-28
fix(picker): spinner position in RTL direction

Rocss:
## Description

Aims to fix https://github.com/adobe/spectrum-css/issues/2566

### Validation steps

  1. Open the [docs](https://pr-2567--spectrum-css.netlify.app/picker) for the picker component (where is storybook deployed? 🙈 ):
- [x] `is-loading` LTR - margin is applied between the spinner and the label
- [x] `is-loading` RTL - margin is applied between the spinner and the label
- [x]  `is-invalid` LTR - margin is applied between the spinner and the icon
- [x] `is-invalid` RTL - margin is applied between the spinner and the icon
- [x] `not invalid` - label has no margin end
- [x] `not loading` - label has no margin end

### Regression testing

Validate:

## To-do list


components/picker/index.css
```diff
@@ -135,7 +135,7 @@ governing permissions and limitations under the License.
   --spectrum-picker-spacing-bottom-to-text: var(--spectrum-component-bottom-to-text-300);
   --spectrum-picker-spacing-edge-to-text: var(--spectrum-component-edge-to-text-300);
   --spectrum-picker-spacing-text-to-icon: var(--spectrum-text-to-visual-300);
-  --spectrum-picker-spacing-text-to-alert-icon-inline-start: var(--spectrum-field-text-to-alert-icon-extra-large);
+  --spectrum-picker-spacing-text-to-alert-icon-inline-end: var(--spectrum-field-text-to-alert-icon-extra-large);
```
Rocss, line 138: I am not sure if I should actually do this change 😅 , made sense since the property will be `margin-inline-end`, but is it OK to change mod tokens like that?

mdt2 (commented):
Visually this change is looking good! Just a couple of comments related to deprecating mods.

To answer your question about finding storybook, in the docs site side navigation, you can scroll down to the bottom and click on "Component preview" and it should pull up storybook.

components/picker/index.css
castastrophe: @mdt2 Generally you're right but I don't see any --mods in use in this case so I don't think we need to do any steps for deprecation. We don't recommend users overwrite --spectrum prefixed variables anyway since those are expected to change.

castastrophe:
Thanks @mdt2, great feedback. I have no additional feedback, just +1 to her suggestions.

Rocss:
One thing I observed, unrelated to this change because it was happening before too, is the size of the warning icon in XL Picker, these seem like a suspicious size:

Screenshot 2024-02-29 at 12 03 05

mdt2:
> One thing I observed, unrelated to this change because it was happening before too, is the size of the warning icon in XL Picker, this seems like a suspicious size:
> 
> Screenshot 2024-02-29 at 12 03 05

This does look strange. I took a quick peek and it looks like the token definition matches the [S1 design spec](https://www.figma.com/file/nchxla8gD4CxWYwt9eiRrz/S1-token-specs?type=design&node-id=496-7958&mode=design&t=rHYiN5Fm8ZQJB7kA-0), but we have a margin-inline-start value that seems to be limiting it at the XL size:
picker-invalid

Rocss:
Guys, I have no more things to add to this PR, happy to implement any other feedback, or else we can merge it.
cc: @mdt2 @castastrophe

mdt2 (approved):
Looks good to me, thank you! @castastrophe, @pfulton this should be ready for a final look/merge 👍

PR #2600, opened 2024-03-20
feat(button): migrate to Spectrum 2

jawinn:
## Description

Migrates the Button component to Spectrum 2. Button now uses the tokens defined on the Spectrum 2 design spec. This includes a lot of color changes for the different variants. It removes the express.css and spectrum.css theme files, and combines their contents into the main index.css.

#### Medium as default, and removing "fill" class
Medium size is now the default. The class `.spectrum-Button--sizeM` is now unnecessary for this size, and has been removed. This follows the same type of change being progressively made to all components.

All references to `.spectrum-Button--fill` have been removed, as it is now unnecessary. It was only used in the CSS for one high contrast mode selector, which has been changed. The fill style has already been the default without the need for this class.

#### Cleanup work
This also includes a bit of cleanup and refactoring work, that is detailed in the commit messages and noted where necessary in the migration guide. This includes:
- Re-organizes some CSS and selectors, and removes some repeated and unnecessary declarations. Of note, there were some older skin.css styles separated from the main styles, and several spread out styles related to the focus indicator---some of which were overriding each other.
- A a good chunk of the custom properties from the spectrum.css theme file are consolidated and slimmed down.
- Removes some leftover classes that do not apply to button: `.is-selected`, `emphasized`, and `quiet`.

#### Some other changes of note:
- Renames `--mod-bold-font-weight` to `--mod-button-font-weight`
- Updates the Migration guide to include versions and dates

CSS-616

### Validation steps

WIP

### Regression testing

Validate:

## To-do list


components/button/index.css
```diff
@@ -106,6 +104,258 @@ governing permissions and limitations under the License.
...
+  }
+
+  /* ---- Accent ---- */
+  &.spectrum-Button--accent {
+    --spectrum-button-background-color-default: var(--spectrum-accent-background-color-default);
+    --spectrum-button-background-color-hover: var(--spectrum-accent-background-color-hover);
+    --spectrum-button-background-color-down: var(--spectrum-accent-background-color-down);
+    --spectrum-button-background-color-focus: var(--spectrum-accent-background-color-key-focus);
+
+    &.spectrum-Button--outline {
```
mdt2, line 130: I'm having trouble finding the S2 spec for Accent Outline, could you point me to that?

components/button/index.css
```diff
@@ -106,6 +104,258 @@ governing permissions and limitations under the License.
...
+      --spectrum-button-content-color-hover: var(--spectrum-negative-content-color-hover);
+      --spectrum-button-content-color-down: var(--spectrum-negative-content-color-down);
+      --spectrum-button-content-color-focus: var(--spectrum-negative-content-color-key-focus);
+    }
+  }
+
+  /* ---- Primary ---- */
+  /* Also shows as the default when a variant is not set. */
+  &.spectrum-Button--primary,
+  & {
```
mdt2, line 176: Do we need to continue supporting the `.spectrum-Button--primary` class if these styles are the default on `.spectrum-Button`?

components/button/index.css
```diff
@@ -106,6 +104,258 @@ governing permissions and limitations under the License.
...
+      --spectrum-button-background-color-hover: var(--spectrum-transparent-white-200);
+      --spectrum-button-background-color-down: var(--spectrum-transparent-white-200);
+      --spectrum-button-background-color-focus: var(--spectrum-transparent-white-200);
+  
+      --spectrum-button-content-color-default: var(--spectrum-transparent-white-800);
+      --spectrum-button-content-color-hover: var(--spectrum-transparent-white-900);
+      --spectrum-button-content-color-down: var(--spectrum-transparent-white-900);
+      --spectrum-button-content-color-focus: var(--spectrum-transparent-white-900);
+  
+      &.spectrum-Button--outline {
```
mdt2, line 285: I think we might be missing the background changes for this secondary static white with outline

mdt2 (commented):
Sooo much good work on this one, Josh 🏆  A few comments for you. Also +1 on needing more design direction for wrapping buttons, I see you mentioned that in Slack for the design review. I'm not sure we should merge this one until we have that squared away. It's nice to have that wrapping story now to account for that.

components/button/metadata/button-accent.yml
```diff
@@ -10,27 +10,26 @@ sections:
     description: |
       ### x/x/2024 - Version 13.0.0
```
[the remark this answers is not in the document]
jawinn: I like the idea of using versions only. The date doesn't seem necessary for the upgrade process. I'll update it here.

components/button/metadata/button-staticcolor.yml
```diff
@@ -43,29 +39,38 @@ sections:
...
+      ### 2/22/2021 - Version 3.0.0
+      #### T-shirt sizing
+      Button now supports t-shirt sizing and requires that you specify the size of button by adding a `.spectrum-Button--size*` class.
+
 examples:
   - id: button-staticcolor
     name: Static White
     markup: |
-      <div style="background-color: rgb(15, 121, 125); color: rgb(15, 121, 125); --spectrum-alias-heading-text-color: white; padding: 15px 20px;">
+      <div style="background-color: rgb(5, 108, 92); --mod-heading-font-color: white; padding: 15px 20px;">
```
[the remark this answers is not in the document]
jawinn: I haven't heard anything in regards to the gradients. I used the solid colors that are used on this Figma. I can follow up on this.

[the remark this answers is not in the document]
mdt2: Sounds good, thanks. I kinda wonder if we shouldn't change it until we get formal word from them for consistency's sake

components/buttongroup/package.json
```diff
@@ -16,7 +16,7 @@
   "main": "dist/index-vars.css",
   "peerDependencies": {
     "@spectrum-css/button": ">=11",
-    "@spectrum-css/tokens": "14.0.0-next.0"
+    "@spectrum-css/tokens": "^14.0.0-next.3"
```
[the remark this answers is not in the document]
jawinn: For Button, the next 3 version is the minimum with the rounding token in it.

The updates in the tools folder, button group, and commons were in [the commit that Patrick added](https://github.com/adobe/spectrum-css/pull/2600/commits/42bb3d20a764cae44dba4f78568fc4dbc513dfa1), when we were seeing the wrong tokens version loading in the docs (resulting in square buttons, etc). That update seems to have fixed it, along with the linter warnings.

components/button/index.css
jawinn: You're right that it isn't necessary to have both, if one is the default. 

I wasn't sure about whether this change was the best course of action or which variant would be the default. Or how that might affect SWC. I was hesitant to remove the primary class without knowing. I'll start a Slack discussion about this to get some more opinions.

Previously, if you didn't have one of the variant classes on the button, it does not display with any of the designed colors. So basically one of the variant classes is required. It seems like it would be a good idea to have the default display nicely if there is no variant class.

components/button/index.css
jawinn: Good callout on this one! The token values are left as they were previously, but do not appear to be on the specs Figma. I'll ask the design team and see if that needs to be added to the Figma and if any of those token values have changed.

jawinn: Per design, the outline style has been deprecated for the accent and negative variants. I've removed those styles. I will reflect this better in the controls once PR #2578 gets merged in, as that makes big changes to the structure of the stories.

components/button/index.css
jawinn: After our discussion today, I've made "accent" the default, as accent shows as the default on the guidelines. And I  have kept the class around that is used for the default, to avoid a drastic change here but also leaving the option to remove it in the future.

components/button/index.css
```diff
@@ -12,18 +12,40 @@ governing permissions and limitations under the License.
...
-  --spectrum-button-focus-ring-border-radius: calc(var(--spectrum-button-border-radius) + var(--spectrum-button-focus-ring-gap));
   --spectrum-button-focus-ring-thickness: var(--spectrum-focus-indicator-thickness);
   --spectrum-button-focus-indicator-color: var(--spectrum-focus-indicator-color);
-  --spectrum-button-intended-icon-size: var(--spectrum-workflow-icon-size-50);
+
+  --spectrum-button-min-width: calc(var(--spectrum-component-height-100) * var(--spectrum-button-minimum-width-multiplier));
+  --spectrum-button-height: var(--spectrum-component-height-100);
+
+  /* @todo set line-height value per size, using font size specific line-height tokens when they are finalized. */
+  --spectrum-button-line-height: 1.2;
```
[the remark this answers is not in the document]
jawinn: The icon only button being too large seems to be an issue with the values of the `--spectrum-component-pill-edge-to-visual-only-*` tokens now being too large with the change in workflow icon sizes. I'm checking with the design team about this.

components/button/index.css
jawinn: Good catch 👍 I've pushed up a commit that fixes this.

pfulton (commented):
This looks awesome, @jawinn! I really like the Docs page.

One question: can `accent` or `negative` have an `outline` treatment, or do they always receive `fill` and only `fill`? If they can't receive the `outline` treatment, can we [conditionally disable](https://storybook.js.org/docs/essentials/controls#conditional-controls) the `fill` and `outline` control?

jawinn:
> This looks awesome, @jawinn! I really like the Docs page.
> 
> One question: can `accent` or `negative` have an `outline` treatment, or do they always receive `fill` and only `fill`? If they can't receive the `outline` treatment, can we [conditionally disable](https://storybook.js.org/docs/essentials/controls#conditional-controls) the `fill` and `outline` control?

I'd love to disable that control when the variant control has either of those two values. I looked into that and so far Storybook doesn't have a way to have multiple conditionals, a custom conditional function, or a conditional that checks against two values, though there is some progress on it: https://github.com/storybookjs/storybook/issues/21281

pfulton (approved):

PR #2605, opened 2024-03-22
feat: add css minification output

castastrophe:
## Description

This PR would add a minified and gzipped output for all existing CSS exports. These assets are only created on production builds.

[Jira CSS-880](https://jira.corp.adobe.com/browse/CSS-880)

### Validation steps

- [x] Expect only whitespace changes to existing CSS exports
- [x] Expect additional *.min.css and *.min.css.gz outputs in component dist directories
- [x] Expect minified assets to include no whitespace and have a smaller size than the unminified version
- [x] Expect the g-zipped asset to be significantly smaller than the unminified version; file should be unreadable without decompression

## To-do list


5t3ph (approved):

jawinn (approved):
Nice addition. I was able to test locally by running `yarn ci` and confirm the validation steps. 🎉 

Of note: I tried to unzip the `.gz` in MacOS via Finder and also with `gunzip` via command line but it didn't understand the format. It seems this is because the headers use zlib format (file info is "zlib compressed data"). TIL. I think there's an option for gzip style headers or maybe the file could be named .zlib; I'm not sure if that's needed though.

PR #2616, opened 2024-03-28
feat: cssnano optimizations

castastrophe:
## Description

Reducing whitespace in our CSS files (especially var functions) optimizes them for minification. This PR implements CSSNano for optimization and minification however CSS minifiers will not remove whitespace from var functions (as whitespace could be valid input) but prettier _will_. To this end, I've added prettier to the build process so that it will process the raw CSS coming in and then pass the pre-prettified version over to postcss which will handle minification.

This PR updates to the latest version of prettier. A significant CSS bug existed at 2.x where it would remove commas from the end of var functions: i.e., `var(--spectrum-logical-transform,)`. That format is critical to defaulting the value correctly and cannot be removed from the resulting code. This bug does not occur at 3.x version of prettier.

### Regression testing

Expect to see no VRT changes.

Validate:

## To-do list


pfulton (approved):
This looks good, and I think the `dist` output is looking even more tidy (to me, at least!)

PR #2617, opened 2024-03-29
fix(storybook): body classes for story and docs pages

mdt2:
## Description

With this fix, Foundation documentation pages will receive the global classes required for styles to show correctly. The fix was to exclude foundation pages from the contextsWrapper. This allows the existing code in `manager.js` to add the `spectrum spectrum--light spectrum--medium` class stack to the `<body>` tag inside the foundations page iframe instead of adding it to only the first `<div id="root-inner>` (it works to add it here for our Story pages because we only have one Story instance on those pages, but in the docs pages we have 1+ story instances so we need the classes to be higher up on the iframe body).

### Validation steps

- [ ] VRTs are passing
- [ ] When you navigate to storybook, navigate to a Foundations page first. All styles should look correct
- [ ] Click around to a component or two and then back to a Foundations page, the styles should look correct on all pages

## To-do list

### DIFF
--- a/.storybook/decorators/contextsWrapper.js
+++ b/.storybook/decorators/contextsWrapper.js
@@ -8,7 +8,7 @@ export const withContextWrapper = makeDecorator({
 	name: "withContextWrapper",
 	parameterName: "context",
 	wrapper: (StoryFn, context) => {
-		const { args, argTypes, viewMode } = context;
+		const { args, argTypes, id, viewMode } = context;
 
 		const getDefaultValue = (type) => {
 			if (!type) return null;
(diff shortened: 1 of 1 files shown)


jawinn (commented):
The component styles do load now when visiting the two foundation pages first. I see the button and checkbox styles now on those pages. 👍

**Issue with background colors also on spectrum-two**
I am seeing one difference with the background colors. I tested on the spectrum-two branch though and am seeing the same thing. So this could be a problem for a separate PR, unless you're aware of a fix related to the changes you made. The problem seems somewhat connected.

On first load of a foundations page, I see these two sections with a different grey background:
Screenshot 2024-03-29 at 2 47 34 PM
Screenshot 2024-03-29 at 2 47 41 PM

If I click around to another component story and then back to a foundation page, the extra background colors go away and it's back to normal.

mdt2:
> On first load of a foundations page, I see these two sections with a different grey background

@jawinn weird. I was also seeing this initially while working on this PR, but after making the changes to the contextWrapper, I'm not seeing the backgrounds anymore 😵 Locally I:

1. Run `git clean -dfX && yarn install`
2. Run `yarn start`
3. Let it load to the default guide page
4. Click on one of the Foundations pages and I don't see the grey background

Did you take different steps when you saw the weird backgrounds?

rise-erpelding (approved):
This looks good to me! The components displayed in Foundations now have the correct styles applied before clicking on a component page, and I can confirm that I am not seeing that gray background on this branch, but do on the `spectrum-two` branch:

image
image

castastrophe (approved):

castastrophe:
@mdt2 Should we be adding this change to main and then bringing it back into spectrum-two or is it a spectrum-two specific infrastructure change?

jawinn (approved):
I'm no longer seeing the issue I was seeing before. LGTM! 👍

PR #2657, opened 2024-04-11
refactor(ui-icons)!: s2 migration of assets

castastrophe:
## Description

This PR pulls in the S2 icons and commits the dist output. This is a breaking change with entirely new icons.

### Regression testing

Validate:

## To-do list


jawinn (changes requested):
The latest version of these icons is quite a big change! I have a few questions and noticed a few issues.

### Missing Workflow icons
Some workflow icons (Chevron, Arrow) are no longer displaying. I believe I see what was causing this, as noted in the code comments.

### Order of UI icons displayed
It would be good if the arrows showed up next to each other here. They are spread out in different areas of the list. This also happens on the select menu for the icon in the Storybook control. I would expect all of the icon names starting with "Arrow" to be next to each other:
Screenshot 2024-04-16 at 10 35 39 AM

### Needed CSS changes and refactoring: classes with sizing for new UI icons, and removal of Medium/Large related code
Each UI icon is paired with a class and token that determines its size. With the addition of new icons and removal of some icons, there is now CSS that needs to be both added and removed. This can be seen currently in the first set of five "Add" icons all appearing as the same size, since there is no `.spectrum-UIIcon-Add50` etc:
Screenshot 2024-04-16 at 10 43 21 AM
It doesn't look like some of these new icons have tokens that exist yet.

The removal of the separate Large and Medium UI icons also means that some code can be refactored and removed.

This could be follow-up work. Should we create another Jira issue for this?

### Communication of deprecations?
This update removes a lot of existing sizes and icons. For example, Arrow200, Arrow300, Asterisk75, etc. Should we communicate these changes somehow? Those sizes are likely in use for some users of the library. Some are also in use in our own library, though I am assuming once everything is migrated to Spectrum 2 then they won't be anymore.

## Storybook Chromatic-only template reference
**On PR**
Screenshot 2024-04-16 at 10 04 19 AM

**Production**
Screenshot 2024-04-16 at 10 05 13 AM

castastrophe:
@jawinn I tried to adjust for the issues you raised but there were some significant changes to the format and naming of the icons so I tried to approach it in a new way. You can see the new baseline proposed here - let me know what you think! https://pr-2657--spectrum-css.netlify.app/preview/?path=/story/components-icon--default&globals=testingPreview:!true

jawinn:
> @jawinn I tried to adjust for the issues you raised but there were some significant changes to the format and naming of the icons so I tried to approach it in a new way. You can see the new baseline proposed here - let me know what you think! https://pr-2657--spectrum-css.netlify.app/preview/?path=/story/components-icon--default&globals=testingPreview:!true

The new grid layout is great. It's much more effective at showing which size is which. Could you also include an arrow and a chevron in the list of workflow icons? That way we can continue to test for any conflicts with those icons that can have the same name as UI icons (which had rendering issues previously). 

The ordering of the icon names in the control looks like it still needs an update so they are alphabetical (e.g. Arrow* names grouped with each other).

What do you think about the questions raised in "_Needed CSS changes and refactoring: classes with sizing for new UI icons, and removal of Medium/Large related code_"? As well as communicating the deprecations?

PR #2550, opened 2024-02-22
refactor(menu): adjust styling for disabled drill-in chevron

jawinn:
## Description

Some difficulty was encountered with the SWC implementation making sure the chevron on a disabled drill-in menu item displayed as disabled, without overriding CSS. On the SWC end, the way a generic icon is styled is different from how chevrons or checkmarks are styled, so the styles added didn't apply to the chevron.

Discussion here https://github.com/adobe/spectrum-css/pull/2199#pullrequestreview-1881393032

Docs migration guide: also adds versions and dates to the existing migration guide content to conform with new recommended format. 

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [ ] Confirm that solution will work for SWC.
- [ ] Chevron on drill-in menu items continues to look the same in Spectrum CSS, including for a disabled menu item.

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [ ] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [ ] VRTs have been run and looked at.
- [ ] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

## To-do list

- [x] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [x] I have updated relevant storybook stories and templates.
- [ ] I have tested these changes in Windows High Contrast mode.
- [x] If my change impacts **other components**, I have tested to make sure they don't break.
- [x] If my change impacts **documentation**, I have updated the documentation accordingly.
- [ ] ✨ This pull request is ready to merge. ✨

PR #2689, opened 2024-04-22
fix: resolve missing build artifacts

pfulton:

### Regression testing

Validate:

## To-do list

### DIFF
--- a/site/tasks/builder.js
+++ b/site/tasks/builder.js
@@ -158,6 +158,10 @@ async function copy_Resources() {
 			cwd: path.join(dirs.root, "assets"),
 			outputDir: "img"
 		}),
+		copy_Assets(["**"], {
+			cwd: path.join(dirs.root, "components"),
+			outputDir: "components"
+		}),
 	]).then((results = []) => {
 		const end = process.hrtime(start);
 		return [


castastrophe (approved):

pfulton:
@castastrophe I think this might still be looking a little funky, but I need to dig into it more. It seems like the underlay is present when you first hit the docs site...not sure why.

rise-erpelding (commented):
I did some digging on this--one of the first things I noticed was that running the docs site on `main` wasn't breaking for me locally, until I changed the version of Node I was using (I was previously on 18.19.1, and it looks like that's what our [GitHub Actions were using on #2550](https://github.com/adobe/spectrum-css/actions/runs/8054142387/job/21998045555), which has a [Netlify preview](https://pr-2550--spectrum-css.netlify.app/) that works as we expect it to).

The GitHub Action picks its node version based on which is the LTS, so it's been using 18.20.2--I didn't look into why exactly, but this appears to break the docs site on Netlify and locally.

I set up a draft PR #2694 to see what would happen using Node 20, and as far as I can tell the [preview of the docs site](https://pr-2694--spectrum-css.netlify.app/accordion) looks like it did before, so that could be something we could look into as an alternative to this.

PR #2718, opened 2024-04-30
fix: remove custom storybook keyboard shortcuts for text direction

jawinn:
## Description

This PR removes the Storybook "r" and "n" key shortcuts for the text direction toolbar control.

When pressing the "r" or "n" key, the document direction would change due to this custom keydown code. This was causing problems such as typing into a text input and the document direction would flip if you typed anything with the letter "r". This also conflicted with other shortcuts, such as Cmd+Shift+R to reload the page.

CSS-647

### Validation steps

- [ ] Storybook - pressing the "r" or "n" keys no longer changes text direction (after clicking into story content area)
- [ ] Storybook - text direction toolbar control continues to function

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.storybook/decorators/index.js
+++ b/.storybook/decorators/index.js
@@ -6,27 +6,15 @@ export { withTestingPreviewWrapper } from "./withTestingPreviewWrapper.js";
 
 /**
  * @type import('@storybook/csf').DecoratorFunction<import('@storybook/web-components').WebComponentsFramework>
- * @description Rendered as controls; these properties are assigned to the document root element
+ * @description Sets the text direction of the document, using the global set with a toolbar control. These properties are assigned to the document root element.
  **/
 export const withTextDirectionWrapper = makeDecorator({
 	name: "withTextDirectionWrapper",
 	parameterName: "textDecoration",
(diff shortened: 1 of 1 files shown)


jawinn:
Per some discussion, if we'd like to bring back a different keyboard shortcut to do the same thing (for example `Opt + T` to toggle between RTL and LTR), it would be better if this was done through Storybook's API method `api.setAddonShortcut`. The text direction selector would need to change into an addon for that to happen.

Making it into an addon and using the API could improve some issues that it had:
- The value of the text direction toolbar item did not update when the old shortcut was used. Using an addon, these could be made to be in sync.
- The keyboard shortcut did not activate until after clicking into the story content area. This is not the case with Storybook's built in keyboard shortcuts.
- The keyboard shortcut did not respect the config option "enableShortcuts"
- The keyboard shortcut was not shown to the user on the list of shortcuts (viewable in Storybook through the cog icon)

I've added a backlog item CSS-647 if we want to add back a keyboard shortcut in the future.

pfulton (approved):

Issue #2707, opened 2024-04-26
[Tooltip] Left/right tooltips in RTL have incorrect tip placements


Westbrook:
## Steps to reproduce

1. Go to [here](https://pr-2698--spectrum-css.netlify.app/tooltip)
2. Switch to RTL
3. Scroll down to "Directions"
4. See the `left/right` tooltips
5. They don't touch the container and get larger as they move away from the container

## Expected behavior
Tips should touch the container on their wide edge and get smaller

## Screenshots
<img width="333" alt="image" src="https://github.com/adobe/spectrum-css/assets/1156657/62e5b128-fdb4-440e-90d0-9f9beca90513">

pfulton:
Tracking this in Jira via [CSS-748](https://jira.corp.adobe.com/browse/CSS-749)

pfulton:
@Westbrook this fix is available in `@spectrum-css/tooltip@6.1.1`: https://github.com/adobe/spectrum-css/commit/f487f473cf39cd6ec8a8646e6671391985e955f1

PR #2736, opened 2024-05-06
fix(tooltip): rtl places tip in correct position

mdt2:
## Description

This PR aims to address issue #2707 by adjusting the left and right positions of the Tooltip for RTL. Also included is an updated Chromatic kitchen sink story to increase VRT coverage.

### Validation steps
- [x] Head over to Storybook and turn on the [testing preview for Tooltip](https://64762974a45b8bc5ca1705a2-qgypyyjwyz.chromatic.com/?path=/story/components-tooltip--default&globals=testingPreview:!true). You should see all the options for tooltip placement there
- [x] When you switch to RTL, all the tip placements should be touching the body of the tooltip
- [x] When you switch to RTL, the left and right tooltips are facing the correct way (which I believe should be opposite in RTL, so if the heading says "left" the tooltip should be to the right of the imaginary source when RTL is on)
- [x] Check [the docs site](https://pr-2736--spectrum-css.netlify.app/tooltip) in LTR and RTL. Confirm no regressions, specifically for "show on hover" section

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/kind-mangos-jam.md
+++ b/.changeset/kind-mangos-jam.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/tooltip": patch
+---
+
+fix(tooltip): Left and right placements look correct for RTL
--- a/components/tooltip/index.css
+++ b/components/tooltip/index.css
@@ -154,7 +154,7 @@ governing permissions and limitations under the License.
 	inline-size: var(--mod-tooltip-tip-square-size, var(--spectrum-tooltip-tip-square-size));
 
 	inset-block-start: 100%;
-	/* stylelint-disable-next-line csstools/use-logical */
+	/* stylelint-disable-next-line csstools/use-logical -- intentional use of non-logical property */
 	left: 50%;
 	transform: translateX(-50%);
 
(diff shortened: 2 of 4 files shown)


rise-erpelding (approved):
This looks great! I'm approving because this fixes the bug and the Chromatic coverage is a big improvement! 🎉 

I have a question about something that existed before and may not necessarily need to be fixed:
I find these Bottom Left/Bottom Right/Top Left/Top Right tooltips very confusing when they're presented in rtl (see screenshots below). It does make sense if they're supposed to flip directions in rtl so that they're not actually left or right, but thought it would be good to confirm that's how we want them while we're here.
image
image

mdt2:
Great question @rise-erpelding! When RTL is on, the assumption is that the component would show a right-to-left language like Arabic, so the flipping of the UI aligns with that. I agree that calling it Left/Right is confusing in this case. My first thought was to try to name these the way logical properties are handled, but I'm not sure that really applies 1:1. Does that help? I'm definitely open to suggestions 😄

jawinn (approved):
This looks good and fixes the issue with the tip placement.

I did notice one thing that seems to have regressed, that may be a slightly different bug. It might be worth trying to tackle here as well? Or a followup PR/issue.

In the YML docs, under "**Show on hover**", these don't appear to behave the same way anymore in RTL. These are using the CSS within a `.u-tooltip-showOnHover` class (which seems unusual for our library):

[Older version built on prod](https://opensource.adobe.com/spectrum-css/tooltip.html):
Screenshot 2024-05-07 at 10 02 49 AM

Current local build of main and PR:
Screenshot 2024-05-07 at 10 08 07 AM

mdt2:
@jawinn I can't seem to replicate this, maybe I'm misunderstanding 🤔 Are you looking at "Show on hover" for the [PR deploy of this PR](https://pr-2736--spectrum-css.netlify.app/tooltip)? This is what I see:

Screen Shot 2024-05-07 at 11 09 34 AM

jawinn:
> I can't seem to replicate this, maybe I'm misunderstanding 🤔 Are you looking at "Show on hover" for the [PR deploy of this PR](https://pr-2736--spectrum-css.netlify.app/tooltip)? This is what I see:

Yes, on that PR link. Viewed in Direction RTL.

jawinn (approved):
Looks great! Thanks for fixing that "Show on hover" issue and for adding documentation.

PR #2737, opened 2024-05-07
fix(picker): correct disabled state colors

jawinn:
## Description

**Picker**: The following states were showing a change of colors when they shouldn't:
- Disabled + hover
- Disabled + invalid + hover
- Disabled + open
- Disabled + open + hover

This was caused by a few different issues. One is that that the `.spectrum-Picker:hover` styles are being moved farther down in the built dist file, as part of the hover media query from the postcss-hover-media-feature plugin. The hover styles with the same specificity as the disabled styles were then later in the order than intended and overriding the disabled styles. The other is that the `.is-open` styles had a higher specificity than some of the disabled styles.

This also removes an unused variable flagged by the linter, `--spectrum-picker-spacing-top-to-text-side-label-quiet`. This was used nowhere in the file or repo.

CSS-754

### Validation steps

- [x] Disabled + hover does not show color changes in Storybook
- [x] Disabled + invalid + hover does not show color changes in Storybook
- [x] Disabled + open does not show color changes in Storybook
- [x] Disabled + open + hover does not show color changes in Storybook

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/picker/index.css
+++ b/components/picker/index.css
@@ -277,6 +277,26 @@ governing permissions and limitations under the License.
 		}
 	}
 
+	&.is-open:not(.spectrum-Picker--quiet, :disabled, .is-disabled) {
+		color: var(--highcontrast-picker-content-color-default, var(--mod-picker-font-color-default-open, var(--spectrum-picker-font-color-default-open)));
+		background-color: var(--highcontrast-picker-background-color, var(--mod-picker-background-color-default-open, var(--spectrum-picker-background-color-default-open)));
+		border-color: var(--highcontrast-picker-border-color-default, var(--mod-picker-border-default-open, var(--spectrum-picker-border-color-default-open)));
+
+		&:hover {
+			color: var(--highcontrast-picker-content-color-default, var(--mod-picker-font-color-hover-open, var(--spectrum-picker-font-color-hover-open)));
+			background-color: var(--highcontrast-picker-background-color, var(--mod-picker-background-color-hover-open, var(--spectrum-picker-background-color-hover-open)));
+			border-color: var(--highcontrast-picker-border-color-hover, var(--mod-picker-border-color-hover-open, var(--spectrum-picker-border-color-hover-open)));
+
+			.spectrum-Picker-menuIcon {
+				color: var(--highcontrast-picker-content-color-default, var(--mod-picker-icon-color-hover-open, var(--spectrum-picker-icon-color-hover-open)));
+			}
+		}
+
+		.spectrum-Picker-menuIcon {
+			color: var(--highcontrast-picker-content-color-default, var(--mod-picker-icon-color-default-open, var(--spectrum-picker-icon-color-default-open)));
+		}
+	}
+
 	&.is-invalid:not(:disabled, .is-disabled) {
 		border-color: var(--highcontrast-picker-border-color-default, var(--mod-picker-border-color-error-default, var(--spectrum-picker-border-color-error-default)));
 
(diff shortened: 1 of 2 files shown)


mdt2 (approved):
Looks good to me!

components/picker/index.css
```diff
@@ -14,488 +14,481 @@ governing permissions and limitations under the License.
...
+		&.is-placeholder {
+			color: var(--highcontrast-picker-content-color-default, var(--mod-picker-font-color-key-focus, var(--spectrum-picker-font-color-key-focus)));
+		}
+
+		.spectrum-Picker-menuIcon {
+			color: var(--highcontrast-picker-content-color-default, var(--mod-picker-icon-color-key-focus, var(--spectrum-picker-icon-color-key-focus)));
+		}
+	}
+
+	&.is-open:not(.spectrum-Picker--quiet, :disabled, .is-disabled) {
```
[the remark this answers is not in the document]
jawinn: This section just moved from below, but `:disabled, .is-disabled` was added. Looks like all the whitespace changed.

components/picker/index.css
```diff
@@ -321,47 +337,28 @@ governing permissions and limitations under the License.
...
-		.spectrum-Picker-menuIcon {
-			color: var(--highcontrast-picker-content-color-default, var(--mod-picker-icon-color-hover-open, var(--spectrum-picker-icon-color-hover-open)));
-		}
+/* Disabled state. */
+.spectrum-Picker.spectrum-Picker:disabled,
+.spectrum-Picker.spectrum-Picker.is-disabled {
+	cursor: default;
+	background-color: var(--highcontrast-picker-background-color, var(--mod-picker-background-color-disabled, var(--spectrum-picker-background-color-disabled)));
+	border-color: var(--highcontrast-picker-border-color-disabled, transparent);
+	color: var(--highcontrast-picker-content-color-disabled, var(--mod-picker-font-color-disabled, var(--spectrum-picker-font-color-disabled)));
```
castastrophe, line 352:
```suggestion
.spectrum-Picker:disabled,
.spectrum-Picker.is-disabled {
	cursor: default !important;
	background-color: var(--highcontrast-picker-background-color, var(--mod-picker-background-color-disabled, var(--spectrum-picker-background-color-disabled))) !important;
	border-color: var(--highcontrast-picker-border-color-disabled, transparent) !important;
	color: var(--highcontrast-picker-content-color-disabled, var(--mod-picker-font-color-disabled, var(--spectrum-picker-font-color-disabled))) !important;
```

I think last time we spoke about doubling up on classes to increase specificity, there was concern this would not translate well to downstream consumers like web components who convert it to `:host`. I have read a lot lately in CSS that supports the use of !important when setting disabled styles to ensure they are always preferred and it best solves the web components use-case. Is there a reason why this might not work?

castastrophe (changes requested):

components/picker/index.css
jawinn: I'm not keen on us having to start using `!important` for this for the typical reasons to avoid it (including it overriding inline styles and needing another !important to override it), though it may be more okay when the CSS is on the web components side.

I've used an alternative approach in the latest commit, that moves the disabled styles to the very end, to make sure it's after the hover media query that's causing the order issue. What do you think of that? Relying on the order isn't great either.

I wish this could be solved at the tooling level, so the CSS is not changed in unintended ways, but currently I do not see any configuration options available for `postcss-hover-media-feature`. I tried creating a separate `@media (hover: hover)` in the source CSS but it still bundled them all into one at the end.

Doubling up on the class felt like a lighter touch that just adds that little bit more specificity, without resulting to using`!important`. Another idea would be to double up the second part, `.spectrum-Picker:disabled:disabled`, if that's better with how web components converts it. Open to other ideas as well.

components/picker/index.css
```diff
@@ -32,7 +32,6 @@ governing permissions and limitations under the License.
 	--spectrum-picker-spacing-bottom-to-text: var(--spectrum-component-bottom-to-text-100);
 	--spectrum-picker-spacing-edge-to-text: var(--spectrum-component-edge-to-text-100);
 	--spectrum-picker-spacing-edge-to-text-quiet: var(--spectrum-field-edge-to-text-quiet);
-	--spectrum-picker-spacing-top-to-text-side-label-quiet: var(--spectrum-component-top-to-text-100);
```
castastrophe, line 35: Can you outline why these variables are okay to be removed in the PR description when you get a chance? I see that it's fine in the baselines but just want to have it documented for the future.

castastrophe (approved):

PR #2743, opened 2024-05-09
feat(tooltip): S2 migration

mdt2:
## Description

This migration includes updated colors, rounding, a bigger tip, and the removal of variants (only neutral is available in Spectrum 2). As a result of the deprecation of variants, icons have also been removed.

The redesign of the tip, specifically the rounding, required a reworking of how we use clip-path and transform.

Some custom property mods have been removed:

- `--mod-tooltip-background-color-informative`
- `--mod-tooltip-background-color-negative`
- `--mod-tooltip-background-color-positive`
- `--mod-tooltip-icon-spacing-block-start`
- `--mod-tooltip-icon-spacing-inline-end`
- `--mod-tooltip-icon-spacing-inline-start`
- `--mod-tooltip-icon-width`

And one mod has been added:

- `--mod-tooltip-tip-corner-radius`

### Validation steps

- [ ] Design review (@mdt2)
  - Specifically check on the new distance of tooltip from edge. This needed adjusted because of the new corner rounding. It looks funky when the tooltip text is only one line.
- [x] Check out Tooltip in Storybook, specifically the [testing preview](https://pr-2743--spectrum-css.netlify.app/preview/?path=/story/components-tooltip--default&globals=testingPreview:!true) will be helpful. The variants should align with the token changes in the [design spec](https://www.figma.com/file/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?type=design&node-id=9602-6409&mode=design&t=TLl0bmGBfKk31oG6-0) (@rise-erpelding)
- [x] Now turn on RTL. All the tips should be in the correct place. (@rise-erpelding)
- [x] Any controls for the deprecated variants (informative, positive, negative) have been removed from storybook (@rise-erpelding)
- [x] Now let's spin up the docs site to check that the "show on hover" functionality is still looking good. You'll have to do this locally because of the difference in tokens package:
  - First update the `site/package.json` tokens version to be `>=14.0.0-next.7`
  - Then you can run a `yarn install`
  - And finally a `yarn dev`
  - Check on that hover functionality and make sure it's looking good
- [x] And still in the docs site, turn on RTL and make sure the hover functionality looks good there too

### Regression testing

Validate:

## Screenshots

Screen Shot 2024-05-09 at 3 29 34 PM

## To-do list


components/tooltip/index.css
```diff
@@ -14,367 +14,271 @@ governing permissions and limitations under the License.
...
-  align-self: flex-start;
-  flex-shrink: 0;
+	position: absolute;
+
+	block-size: var(--mod-tooltip-tip-square-size, var(--spectrum-tooltip-tip-square-size));
+	inline-size: var(--mod-tooltip-tip-square-size, var(--spectrum-tooltip-tip-square-size));
+
+	inset-block-start: calc(100% - (0.5 * var(--mod-tooltip-tip-square-size, var(--spectrum-tooltip-tip-square-size))) - 0.5px);
+	/* stylelint-disable-next-line csstools/use-logical -- intentional use of non-logical property */
+	left: calc(50% - (0.5 * var(--mod-tooltip-tip-square-size, var(--spectrum-tooltip-tip-square-size))));
```
mdt2, line 116: I carried this over from PR #2736 until we can get that work into the spectrum-two branch

components/tooltip/index.css
```diff
@@ -14,367 +14,271 @@ governing permissions and limitations under the License.
...
+	transform: rotate(-45deg);
+
+	background-color: var(--highcontrast-tooltip-background-color-default, var(--mod-tooltip-background-color-default, var(--spectrum-tooltip-background-color-default)));
+
+	/* Default: Pointing down ▽ */
+	clip-path: polygon(0 calc(0% - var(--mod-tooltip-tip-antialiasing-inset, var(--spectrum-tooltip-tip-antialiasing-inset))), 100% var(--mod-tooltip-tip-height-percentage, var(--spectrum-tooltip-tip-height-percentage)), calc(0% - var(--mod-tooltip-tip-antialiasing-inset, var(--spectrum-tooltip-tip-antialiasing-inset))) 100%);
```
mdt2, line 122: The problem with the existing clip-path was that I couldn't apply a border-radius to it when there was a `50%` point. So I adjusted the sizing of the square ([see above](https://github.com/adobe/spectrum-css/pull/2743/files?diff=unified&w=1#diff-14d4b487bc2de8da9dccaef67bce6cccc92cbce28e2ac16912884edbce436cbcL58-R49)) and the clip-path points to be able to apply the new border-radius to the tip.

mdt2, line 122: The problem with the [previous clip-path](https://github.com/adobe/spectrum-css/pull/2743/files#diff-14d4b487bc2de8da9dccaef67bce6cccc92cbce28e2ac16912884edbce436cbcL164-L168) was that I couldn't apply a border-radius to it when there was a `50%` point. So I adjusted the sizing of the square ([see above](https://github.com/adobe/spectrum-css/pull/2743/files?diff=unified&w=1#diff-14d4b487bc2de8da9dccaef67bce6cccc92cbce28e2ac16912884edbce436cbcL58-R49)) and the clip-path points to be able to apply the new border-radius to the tip.

This change also has a slightly added bonus that we don't have to redefine the clip-path for the different placements, we just have to adjust the `transform: rotate()` value ✨

components/tooltip/index.css
```diff
@@ -38,50 +35,24 @@ governing permissions and limitations under the License.
...
+	--spectrum-tooltip-background-color-default: var(--spectrum-neutral-background-color-default);
 
 	/* tip */
-	--spectrum-tooltip-tip-inline-size: var(--spectrum-tooltip-tip-width);
 	--spectrum-tooltip-tip-block-size: var(--spectrum-tooltip-tip-height);
 
 	/* Width and height of square element used to render the tip triangle. */
-	--spectrum-tooltip-tip-square-size: var(--spectrum-tooltip-tip-inline-size);
+	/* To get a square that measure 10px on the diagonal, the sides have to be 7px. */
+	--spectrum-tooltip-tip-square-size: 7px;
```
rise-erpelding, line 46: Yay math and Pythagorean theorem! I wonder if it's helpful to explain further where that number comes from, but I think that's really up to you, in my opinion this comment does make it possible to figure out why it's 7.

For the large platform scale, I think this ends up being `1px` off (because `tooltip-tip-height` is `6px` rather than `5px`, and the square root of 6 squared + 6 squared is about 8.48... I'm not sure this is significant enough to address though, and visually there aren't any issues being caused by that 1px difference. It's nice that that's the only size change that we'd need to think about!

rise-erpelding (commented):
This looks great and I really appreciated the clear validation instructions! All of those things look good and I'm happy to approve, but I have some outstanding questions, one above about the tip size being just slightly off the design for large platform scale, but I also noticed if I look closely at the tip (I changed the color to make it easy to see--and this is Chrome which is more obvious than Firefox) it almost looks like the rotation is slightly off (I think maybe by 1 degree?), which is weird because why would you need to rotate it `-46deg` instead of `-45deg`? 🤷‍♀️ Totally something we could explore more or talk about or reason through in case there's some weird logical reason for it!
image

jawinn (commented):
One thing I noticed was that the tip looks a little overlapped in high contrast mode. Is that something that is possible to adjust?
Screenshot 2024-05-14 at 12 38 06 PM

components/tooltip/index.css
```diff
@@ -647,3 +570,14 @@ governing permissions and limitations under the License.
...
 	}
 }
+
+@media (forced-colors: active) {
+	.spectrum-Tooltip {
+		border: 1px solid transparent;
+	}
+
+	.spectrum-Tooltip-tip {
+		forced-color-adjust: none;
```
jawinn, line 580: I don't think the `forced-color-adjust` is needed here?

components/tooltip/index.css
```diff
@@ -647,3 +570,14 @@ governing permissions and limitations under the License.
...
+		--highcontrast-tooltip-background-color-default: CanvasText;
```
jawinn, line 581: This could probably be moved underneath `@media (forced-colors: active)` directly.

components/tooltip/index.css
```diff
@@ -647,3 +570,14 @@ governing permissions and limitations under the License.
 		}
 	}
 }
+
+@media (forced-colors: active) {
+	.spectrum-Tooltip {
+		border: 1px solid transparent;
```
jawinn, line 576:
```suggestion
		border: 1px solid CanvasText;
```

jawinn (changes requested):

components/tooltip/index.css
mdt2: Great callout on the large platform scale. I'll add a custom token for this.

mdt2:
@rise-erpelding great callout about the strange rotation. This was caused by trying to account for antialiasing in the clip-path. When I moved that to the positioning properties instead, this resolved 🎉

components/tooltip/index.css
jawinn: 👍 Gotcha. I had thought the main background and tip background were using different custom properties.

rise-erpelding (approved):
I just took a quick look at this one again and it looks like the tip issues previously noted are sorted out now! Nice work! If there are any further adjustments based on design feedback, I'm happy to take another look, but for now, I'm giving it my approval!

components/tooltip/stories/tooltip.stories.js
```diff
@@ -162,10 +162,12 @@ const PlacementVariants = (args) => html`
 									content: [`${optionDescription()}`],
 								})}
 							`)}
+							<div>
 							${Template({
 								...args,
 								placement: option,
 							})}
+							</div>
```
rise-erpelding, line 170: Wrapping the Template in a <div> solves the issue that was happening here where the Tooltip starts expanding to fill the container:
image (1)

components/tooltip/index.css
```diff
@@ -15,6 +15,7 @@ governing permissions and limitations under the License.
 
 .spectrum-Tooltip {
 	--spectrum-tooltip-animation-duration: var(--spectrum-animation-duration-100);
+	--spectrum-tooltip-animation-distance: var(--spectrum-spacing-75);
```
rise-erpelding, line 18: Design spec says `spacing-75`, I'll double-check with design and revert if needed but it doesn't sound like it differs for the large platform size.

tokens/dist/index.css
```diff
@@ -3688,9 +3688,8 @@
 
   --spectrum-tooltip-animation-distance:var(--spectrum-spacing-75);
 
-  /* To get a square that measure 10px on the diagonal, the sides have to be 7px, but this has to
-		 be 8px because we have to account for antialiasing so there isn't a gap between tooltip and tip */
-	--spectrum-tooltip-tip-square-size:8px;
+  /* To get a square that measures 10px on the diagonal, the sides have to be 7px */
+	--spectrum-tooltip-tip-square-size:7px;
```
rise-erpelding, line 3688: This small discrepancy in size was called out in the design review--clearly there was a reason `8px` was used instead of `7px`, and `10px` rather than `9px` for large, but testing it again I'm not seeing it, so I'm having a hard time justifying it, I'd love a second opinion here.

marissahuysentruyt: I don't think I understand enough about anti-aliasing. The only way I can see that hairline is if I change `--spectrum-tooltip-tip-antialiasing-inset` to 0px. 

I do not see any gap between the tip and the rest of the tooltip with this changed to 7px, or to the 9px (as long as we keep that property I mentioned at 0.5px~)

tokens/dist/css/spectrum/custom-large-vars.css
```diff
@@ -101,7 +101,8 @@
   --spectrum-assetcard-header-content-font-size:var(--spectrum-heading-size-xxs);
   --spectrum-assetcard-content-font-size:var(--spectrum-body-size-xs);
 
-  --spectrum-tooltip-animation-distance:5px;
+  /* To get a square that measures 12px on the diagonal, the sides have to be 10px */
+	--spectrum-tooltip-tip-square-size:10px;
```
rise-erpelding, line 105: To clarify, we're back to the same values as we had at the beginning for this, design and dev both agreed that these were the most accurate 🙃

components/tooltip/index.css
```diff
@@ -25,7 +25,9 @@ governing permissions and limitations under the License.
 	--spectrum-tooltip-border-radius: var(--spectrum-corner-radius-400);
 
 	--spectrum-tooltip-font-size: var(--spectrum-font-size-75);
-	--spectrum-tooltip-line-height: var(--spectrum-line-height-100);
+
+	/* @todo set line-height using font size specific line-height tokens when they are finalized along with the new variable font. */
+	--spectrum-tooltip-line-height: 1.2;
```
rise-erpelding, line 30: This is [similar to button](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/button/index.css#L29) where the line-height token causes the overall height of the tooltip to be larger than the height token, so we need to adjust the line-height manually to make it work.

marissahuysentruyt (commented):
@rise-erpelding Thanks for taking this on! Left a few questions for you, mainly getting myself up to speed on a few of the changes. 

The line-height stuff: I left a question about it, but there's a spot in the Figma file that looks like maayyybe this is what design was seeing and we're trying to fix with the hardcoded 1.2. There's a little comparison chart. Is that correct?

components/tooltip/index.css
```diff
@@ -38,50 +38,22 @@ governing permissions and limitations under the License.
...
-	--spectrum-tooltip-background-color-informative: var(--spectrum-informative-background-color-default);
-	--spectrum-tooltip-background-color-positive: var(--spectrum-positive-background-color-default);
-	--spectrum-tooltip-background-color-negative: var(--spectrum-negative-background-color-default);
-
-	--spectrum-tooltip-content-color: var(--spectrum-white);
+	--spectrum-tooltip-content-color: var(--spectrum-gray-25);
+	--spectrum-tooltip-background-color-default: var(--spectrum-neutral-background-color-default);
 
 	/* tip */
-	--spectrum-tooltip-tip-inline-size: var(--spectrum-tooltip-tip-width);
```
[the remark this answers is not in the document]
rise-erpelding: Yes, that's correct, we need to set the `tooltip-tip-square-size` using a custom token now because of the design of the new tip!

components/tooltip/index.css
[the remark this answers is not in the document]
rise-erpelding: 🤔 That might be a good thing to follow up on--if anything it's useful to have it documented somewhere.

components/tooltip/index.css
```diff
@@ -15,19 +15,19 @@ governing permissions and limitations under the License.
...
 
 	--spectrum-tooltip-height: var(--spectrum-component-height-75);
 	--spectrum-tooltip-max-inline-size: var(--spectrum-tooltip-maximum-width);
-	--spectrum-tooltip-border-radius: var(--spectrum-corner-radius-100);
-
-	--spectrum-tooltip-icon-width: var(--spectrum-workflow-icon-size-50);
-	--spectrum-tooltip-icon-height: var(--spectrum-workflow-icon-size-50);
+	--spectrum-tooltip-border-radius: var(--spectrum-corner-radius-400);
 
 	--spectrum-tooltip-font-size: var(--spectrum-font-size-75);
```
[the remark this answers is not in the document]
rise-erpelding: I think it would take care of _most_ of them, and honestly it's an interesting idea that I played around with for a bit after you mentioned it, but it doesn't cover the line-height issues--[line-height for the body resolves to line-height-200](https://github.com/adobe/spectrum-css/blob/main/tokens/dist/index.css#L1413C37-L1413C61), which is too big, and we need a custom one anyway. You're right that it's supposed to grow, if there's more than what fits on 1 line, if we have cjk line heights we'd want to accommodate those, etc., but I think at a minimum there needs to be a circumstance where it does fit into that 30px space similar to how the line-height needs to be adjusted to make button fit a minimum height.

components/tooltip/index.css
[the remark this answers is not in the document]
rise-erpelding: Actually now that I can get on the VPN, I can see [it's already been documented in the old version](https://spectrum.corp.adobe.com/page/tooltip/#Container-padding), that might be sufficient

jawinn (approved):

components/tooltip/index.css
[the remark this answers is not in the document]
marissahuysentruyt: Oh yeah, great find!

Issue #2703, opened 2024-04-26
[Radio] ::after:dir(...) considered incorrect in some parsers 


Westbrook:
Some strict parsing contexts will consider placing `:dir` after pseudo element incorrect and throw.

https://github.com/adobe/spectrum-css/blob/3c7de5e07d34ca6ca1ae873ff32f4e7a02d1f0ad/components/radio/index.css#L486-L488

See errors thrown in https://lightningcss.dev/playground/index.html#%7B%22minify%22%3Atrue%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22div%3A%3Aafter%3Adir(ltr)%20%7B%5Cn%20%20color%3A%20red%3B%5Cn%7D%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D

Not sure if this means it just gets "CSS error" (e.g. dropped) in some browsers or whether browser parsers are more lenient or something else, but it would be great if we could normalize this so that down stream tooling can more easily handle the Spectrum CSS published content.

PR #2744, opened 2024-05-10
fix(slider,radio): dir before pseudo element

mdt2:
## Description

Per issue #2703 and #2704, some parsers see `:pseudo:dir` as invalid, so we've changed it so that the pseudo element comes last `:dir :pseudo`. 

Before, the dist output was:

```
.spectrum-Slider-handle:before:dir(rtl),
[dir="rtl"] .spectrum-Slider-handle:before

.spectrum-Radio-button:after:dir(rtl),
[dir="rtl"] .spectrum-Radio-button:after
```

After, with this improvement, the dist output is:

```
.spectrum-Slider:dir(rtl) .spectrum-Slider-handle:before,
[dir="rtl"] .spectrum-Slider .spectrum-Slider-handle:before {
  transform: translate(50%, -50%);
}

.spectrum-Radio:dir(rtl) .spectrum-Radio-button:after,
[dir="rtl"] .spectrum-Radio .spectrum-Radio-button:after {
  transform: translateX(50%) translateY(-50%);
}
```

### Validation steps

- [x] VRTs pass (@mdt2)

Slider

- [x] Validate that the dist output looks correct after building locally
- [x] Validate that the slider handle location looks correct in [Storybook when RTL](https://pr-2744--spectrum-css.netlify.app/preview/?path=/story/components-slider--default&globals=textDirection:rtl) mode is on

Radio

- [x] Validate that the dist output looks correct after building locally
- [x] Validate that the radio focus looks correct in [Storybook when RTL](https://pr-2744--spectrum-css.netlify.app/preview/?path=/story/components-radio--default&globals=textDirection:rtl) mode is on

## To-do list

### DIFF
--- a/components/radio/index.css
+++ b/components/radio/index.css
@@ -421,7 +421,7 @@ governing permissions and limitations under the License.
 			opacity var(--mod-radio-animation-duration, var(--spectrum-radio-animation-duration)) ease-out,
 			margin var(--mod-radio-animation-duration, var(--spectrum-radio-animation-duration)) ease-out;
 
-		&:dir(rtl) {
+		.spectrum-Radio:dir(rtl) & {
 			transform: translateX(50%) translateY(-50%);
 		}
 	}
--- a/components/slider/index.css
+++ b/components/slider/index.css
@@ -348,7 +348,7 @@ governing permissions and limitations under the License.
 
 		transform: translate(-50%, -50%);
 
-		&:dir(rtl) {
+		.spectrum-Slider:dir(rtl) & {
 			transform: translate(50%, -50%);
 		}
 	}
--- a/.changeset/neat-windows-yawn.md
+++ b/.changeset/neat-windows-yawn.md
@@ -1,0 +1,6 @@
+---
+"@spectrum-css/slider": patch
+"@spectrum-css/radio": patch
+---
+
+Includes similar fixes for both Slider and Radio. Some parsers see `:pseudo:dir` as invalid, so we've changed it so that the pseudo element comes last `:dir :pseudo`.


components/slider/index.css
```diff
@@ -13,775 +13,756 @@ governing permissions and limitations under the License.
...
-  &:focus {
-    outline: none;
-  }
+		inline-size: var(--mod-slider-handle-size, var(--spectrum-slider-handle-size));
+		block-size: var(--mod-slider-handle-size, var(--spectrum-slider-handle-size));
+		border-radius: 100%;
+
+		transform: translate(-50%, -50%);
+
+		:dir(rtl) & {
```
mdt2, line 350: This was my change, all the rest are auto linting changes.

jawinn (approved):
Looks good! I validated the changes in the dist output with `yarn compare` and looked at those two components in Storybook in RTL/LTR.

components/radio/index.css
```diff
@@ -20,471 +20,412 @@ governing permissions and limitations under the License.
...
+		inset-block-start: 50%;
+		transform: translateX(-50%) translateY(-50%);
+
+		transition:
+			opacity var(--mod-radio-animation-duration, var(--spectrum-radio-animation-duration)) ease-out,
+			margin var(--mod-radio-animation-duration, var(--spectrum-radio-animation-duration)) ease-out;
+
+		:dir(rtl) & {
+			transform: translateX(50%) translateY(-50%);
+		}
```
castastrophe, line 429: I'm wondering if this should be moved to query for the direction of the root class selector rather than looking too far up the DOM tree; in this case it's looking up to root implicitly. Something like:

```suggestion
		.spectrum-Radio:dir(rtl) &::after  {
			transform: translateX(50%) translateY(-50%);
		}
```

castastrophe (changes requested):
Can you take a look at the impact of changing the dir selector from querying locally to the component vs. looking to the root of the document? My thought is that if a product has a region of the page localized for example in a language requiring rtl, looking to the root element in the DOM tree will give the wrong result.

components/radio/index.css
mdt2: Great callout, thank you!!

PR #2753, opened 2024-05-13
fix(popover): correct left/right tips for RTL, increase VRT coverage

mdt2:
## Description

Fixes an issue in Popover where Left and Right tips were mirrored incorrectly for RTL. This is simliar to some [work done on Tooltip](https://github.com/adobe/spectrum-css/pull/2736). Note that the Left and Right placement for the tips are always on the LTR left and right, even if RTL is turned on. Start and End, however, are adjusted based on the text direction.

Also included Storybook enhancements including:
- Show all placement options (previously Start and End placements weren't included in the Storybook controls)
- Increase Chromatic VRT coverage for the tip variant

### Validation steps 

- [x] VRTs passing @mdt2 
- [x] Check Popover With Tip's new [testing preview](https://pr-2753--spectrum-css.netlify.app/preview/?path=/story/components-popover--with-tip&args=isOpen:!true&globals=testingPreview:!true) in Storybook. All the tips should look correctly positioned in LTR @jawinn
- [x] Now check RTL, tips should be correctly positioned
- [x] All placement options are shown in the Storybook controls

### Regression testing

Validate:

## Screenshots

Before
before-popover

After
after-popover

## To-do list

### DIFF
--- a/.changeset/three-geese-relax.md
+++ b/.changeset/three-geese-relax.md
@@ -1,0 +1,7 @@
+---
+"@spectrum-css/popover": patch
+---
+
+fix(popover): correct left/right tips for RTL
+
+Includes exposing Start and End tip placement variants in Storybook, and increased VRT coverage for Chromatic.
--- a/components/popover/index.css
+++ b/components/popover/index.css
@@ -150,7 +150,8 @@ governing permissions and limitations under the License.
 	/* spacing to include tip in calculation of offset from source */
 	&.spectrum-Popover--withTip {
 		/* tip size minus where it overlaps with popover border */
-		margin-inline-start: calc(var(--mod-popover-pointer-width, var(--spectrum-popover-pointer-width)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+		/* stylelint-disable-next-line csstools/use-logical -- intentional, right stays on the same side even for RTL */
+		margin-left: calc(var(--mod-popover-pointer-width, var(--spectrum-popover-pointer-width)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
 	}
 
 	/* popover animates towards right ⮕ */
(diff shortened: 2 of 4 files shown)


components/popover/stories/popover.stories.js
```diff
@@ -113,21 +127,71 @@ export default {
...
 };
 
+
+const ChromaticTipPlacementVariants = (args) => html`
+	${placementOptions.map(option => {
+		const optionDescription = () => {
+			if (option.startsWith("start") || option.startsWith("end"))
+				return "Changes side with text direction (like a logical property)";
+			if (option.startsWith("left") || option.startsWith("right"))
+				return "Text direction does not effect the position";
```
jawinn, line 150:
```suggestion
				return "Text direction does not affect the position";
```

jawinn (approved):
Nice work fixing this and adding the Storybook improvements. The Chromatic-only view with descriptions makes it much easier to understand which ones should be flipping.

I only have one minor grammar suggestion.

.changeset/three-geese-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/popover": patch
+---
+
+fix(popover): correct left/right tips for RTL
```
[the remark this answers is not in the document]
mdt2:
```suggestion
fix(popover): correct left/right tips for RTL

Includes exposing Start and End tip placement variants in Storybook, and increased VRT coverage for Chromatic.
```

components/popover/index.css
```diff
@@ -150,7 +150,8 @@ governing permissions and limitations under the License.
 	/* spacing to include tip in calculation of offset from source */
 	&.spectrum-Popover--withTip {
 		/* tip size minus where it overlaps with popover border */
-		margin-inline-start: calc(var(--mod-popover-pointer-width, var(--spectrum-popover-pointer-width)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+		/* stylelint-disable-next-line csstools/use-logical -- intentional, right stays on the same side even for RTL */
```
castastrophe, line 153: Thanks for adding this detailed comment!  Very helpful and ensures we don't accidentally "correct" it later.

components/popover/stories/template.js
```diff
@@ -159,3 +171,39 @@ export const Template = ({
 		</div>
 	`;
 };
+
+export const SourcelessTemplate = ({
```
castastrophe, line 175: Can you add some more context in a comment here about why this additional template is needed?

castastrophe (changes requested):
This is looking good, just a question about the goals of the sourceless template. I'm wondering if we have to update the class list or add new features, is it going to be hard to remember to update that in both templates? Is there a way to have 1 be a "source of truth" without needing the duplication?

components/popover/stories/template.js
mdt2: Great callout on this! Turns out all I had to do was pass a null trigger where I didn't want a source button to show up (for Chromatic), so I've removed this code 👍

castastrophe (approved):

PR #2761, opened 2024-05-15
fix(textfield): no pointer events for focus pseudo element

mdt2:
## Description

Fixes an issue specifically for the Quiet variant where hovering over the focus pseudo element triggers the hover event, which causes the hover event to flicker if the user hovers over the focus, then the space between the focus (which doesn't have pointer events and turns off the hover effect), then onto the textfield (which does have pointer events and the hover effect).

See Josh's helpful screen capture for an example of the issue this fixes: 

https://github.com/adobe/spectrum-css/assets/23404383/148fee9f-1077-49c5-93c2-7acaeac23d8d

### Validation steps

- [x] Pointer events arent triggered by the focus pseudo element: @jawinn 
  - Open the Quiet, keyboard-focused textfield in [Storybook](https://pr-2761--spectrum-css.netlify.app/preview/?path=/story/components-text-field--default&args=isKeyboardFocused:!true;isQuiet:!true)
  - To be able to see what you're doing better, zoom in on the component and in dev tools, set `.spectrum-Textfield:hover { background-color: red }`
  - Hover the mouse over the focus line. The background shouldn't turn red.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/lazy-rivers-leave.md
+++ b/.changeset/lazy-rivers-leave.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/textfield": patch
+---
+
+Fixes an issue in the Quiet variant where hovering over the focus pseudo element was firing pointer events that weren't meant to be seen unless the user hovered over the input element itself.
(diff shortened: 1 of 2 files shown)


jawinn (approved):
Looks good! This should help address the issue that was reported to us in Slack.

rise-erpelding (approved):
This looks good! I can see the difference between how the hover state was being triggered before and see that it does not now.

PR #2652, opened 2024-04-10
feat(alertbanner): migrate to spectrum 2

jawinn:
## Description

Alert banner now uses Spectrum 2 tokens and specifications. In this new design, the divider has been removed. Some spacing has been updated to match the specs.

The following changes have been made to the mod custom properties:
  - `--mod-alert-banner-size` was renamed to `--mod-alert-banner-inline-size`
  - `--mod-alert-banner-neutral-background` was previously misspelled. The mod `--mod-alert-banner-background` was also added, which will take precendence over the variant background mods.
  - `--mod-alert-banner-top-text` was renamed to `--mod-alert-banner-top-to-text`
  - `--mod-alert-banner-top-icon` was renamed to `--mod-alert-banner-top-to-icon`
  - `--mod-alert-banner-bottom-text` was renamed to `--mod-alert-banner-bottom-to-text`
  - `--mod-alert-banner-start-edge` was renamed to `--mod-alert-banner-inline-start-to-content`
  - `--mod-alert-banner-edge-to-button` was renamed to `--mod-alert-banner-block-edge-to-button`
  - The spacing on either side of the close button is now controlled by two separate mods; `--mod-alert-banner-close-button-to-inline-end` and `--mod-alert-banner-close-button-to-content`. The previous `--mod-alert-banner-close-button-spacing` has been removed.
  - A new mod `--mod-alert-banner-inline-end-to-content` was added, which handles the inline end spacing when the alert banner does not have a close button. Or when there is neither a close button or an action button.

To-do after merge into `spectrum-two`:
- Use updated S2 icons and design review

CSS-704

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [x] The divider and its relevant code no longer exists
- [x] Test the Default story as well as the Chromatic testing preview. This looks like the design spec.
- [ ] Updated spacing looks correct. There should not be extra vertical spacing between text and button when the button wraps to the next line (as compared to current version in S1 which has this issue).
- [ ] Proofread migration guide updates
- [ ] Correct spacing (especially at inline end) exists when there is no close button
- [ ] Correct spacing (especially at inline end) exists when there is no close button and no action button

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [x] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [ ] VRTs have been run and looked at.
[...]

PR #2762, opened 2024-05-16
feat(alertbanner): use width token for max-width instead of width

### LINKED ISSUE #2715: [Alert banner] Fixed width causes truncation on small devices
## Description

The alert banner has a fixed width of 832px (M)/ 680px (L). This is set by `--spectrum-alert-banner-width` variable in [spectrum-css tokens](https://github.com/adobe/spectrum-css/blob/main/tokens/dist/index.css#L1859). This introduces an issue for small screens because the alert banner gets truncated on mobile devices.

Since the fixed sizes were introduced to prevent having alert banners that are too wide and hard to read, we should use the fixed sizes as `max-width` instead of width. This way the alert banner would occupy all the available space until it reaches its maximum allowed width of `832px/680px`, depending on the theme size.

## Steps to reproduce

1. Go to <https://opensource.adobe.com/spectrum-css/alertbanner.html/>
2. Open dev tools and toggle device toolbar
3. Choose a mobile device from the list of available options
4. Observe that alert banner content gets truncated

## Expected behavior

The alert banner should be fully visible even on small devices.

## Screenshots

Screenshot 2024-04-30 at 13 56 54

jawinn:
## Description

The alert banner now occupies all available space until it reaches the maximum allowed width. By default it will no longer cut off in viewports smaller than the width token.

The token `--spectrum-alert-banner-width` was previously used as the width, as it was shown in the design specs. Per a discussion with design, this had been intended as a max width, to prevent having alert banners that are too wide and hard to read. That token is now used for `max-inline-size` and the `inline-size` is set to `auto`.

Includes addition of a new mod custom property: `--mod-alert-banner-max-inline-size`

Resolves #2715 
CSS-748

### Validation steps

- [x] Alert banner shrinks down for viewport sizes smaller than the defined max width
- [x] Alert banner does not grow larger than `--spectrum-alert-banner-width`

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/brave-seas-retire.md
+++ b/.changeset/brave-seas-retire.md
@@ -1,0 +1,7 @@
+---
+"@spectrum-css/alertbanner": minor
+---
+
+The alert banner component now occupies all available space until it reaches the maximum allowed width.
+The token previously used for the fixed width, `--spectrum-alert-banner-width`, is now used for the maximum width.
+This includes the additional of a new mod custom property: `--mod-alert-banner-max-inline-size`
--- a/components/alertbanner/index.css
+++ b/components/alertbanner/index.css
@@ -13,7 +13,8 @@ governing permissions and limitations under the License.
 
 .spectrum-AlertBanner {
 	--spectrum-alert-banner-min-height: var(--spectrum-alert-banner-minimum-height);
-	--spectrum-alert-banner-size: var(--spectrum-alert-banner-width);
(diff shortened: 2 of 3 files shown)


jawinn:
I'm going to include Storybook coverage of less than the max size in the S2 migration of Alert banner #2652 , where the stories have already been refactored a bit.

mdt2 (approved):
Looks good, thanks for fixing this!

rise-erpelding (approved):
This is addressing the issue! Nice work.
image

PR #2771, opened 2024-05-21
feat: cleanup legacy vendor prefixes and normalization

jawinn:
## Description

In this work I searched for all uses of `-webkit-*`, `-moz-*`, and `-ms-*` prefixed properties in the repo, to find any cases where we no longer need them. I discovered a few that could be removed, and some tied to normalization styles that needed some cleanup and clarification.

### ===== accordion =====
The `appearance` property no longer needs vendor prefixes. Any necessary vendor prefixes for supported browsers would be added in the build process (none are added for our current browser support list).

### ===== statuslight =====
Removed the legacy `-ms-high-contrast-adjust` property. This CSS property was removed from Microsoft Edge starting with version 79 (when Edge started using Chromium). Our current browser support list does not support Edge versions before that.

This also moves the high contrast mode properties to the bottom of the file, and moves one high contrast property for the dot to within the forced-colors media query.

#### Vendor prefixes removed
`-webkit-user-select`: the `user-select` property is already set and no longer needs manual prefixing.

#### note on `appearance` normalization prefixes
The use of `-webkit-appearance: button` is for normalization; this is still included in modern-normalize and csstools/normalize.css (with the same code comment). I've left these as-is for now because of some lingering questions, and created a followup Jira issue.

#### Simplified ::-moz-focus-inner normalization
These normalization styles only needed two declarations. They now match with what's done in modern-normalize and normalize.css.

#### Legacy vendor prefixes removed
These both are no longer needed in our library given our current browser support list:
  * `::-ms-clear` -- removed in Microsoft Edge 79
  * `::-ms-input-placeholder` -- removed in Microsoft Edge 79
  * `-webkit-text-fill-color` -- only needed for setting the text color on disabled inputs in Safari versions before version 12 (2018).

#### Cleanup and selector update for hide spinner styles
There were styles and comments for removing the native spin buttons for number fields. These now are now only applied to `[type="number"]`. I confirmed that `appearance: none` does not hide them in Firefox and the `textfield` value is still needed.

#### Removed some properties that weren't doing anything
There was a duplicative `appearance: textfield` on `.spectrum-Textfield` which didn't make sense there, as that isn't the input element; `.spectrum-Textfield-input` already has it set.

Two instances of `text-indent: 0` were removed. There is nowhere where text-indent is being set, and this already defaults to zero in the browser, and is not present in any of the normalization libraries.

Note: I also confirmed that the existing `:-moz-ui-invalid` in Text field is part of modern normalization libraries, so it was kept.

CSS-663

### Validation steps

_Note for code review: you'll likely want to hide whitespace changes in files changed for BaseButton, as that does not appear to have gotten the tab/space whitespace change yet._

- [ ] Confirm no visual regressions in accordion, statuslight, textfield, and components that inherit from BaseButton
- [ ] -- Textfield input with `type="number"` should continue to show no spinners in both Firefox and Chrome
- [ ] -- Focus indicator still looks the same in Firefox
- [ ] -- Status light dot looks the same in high contrast mode

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/tidy-balloons-cover.md
+++ b/.changeset/tidy-balloons-cover.md
@@ -1,0 +1,8 @@
+---
+"@spectrum-css/statuslight": minor
+"@spectrum-css/accordion": minor
+"@spectrum-css/textfield": minor
+"@spectrum-css/commons": minor
+---
+
+Removes some legacy vendor prefixes that are no longer necessary, including some that were for older browsers that are no longer in browser support list for Spectrum CSS. Such as Microsoft Edge before version 79. Includes some cleanup around some of the related normalization styles in the Text field component and the Commons `%spectrum-BaseButton` (used for multiple button components).
--- a/components/accordion/index.css
+++ b/components/accordion/index.css
@@ -242,10 +242,6 @@ governing permissions and limitations under the License.
 	font-family: var(--mod-accordion-item-header-font, var(--spectrum-accordion-item-header-font));
 
 	/* reset styling if button element is used */
-	/* stylelint-disable-next-line property-no-vendor-prefix -- todo may no longer be needed */
-	-webkit-appearance: none;
-	/* stylelint-disable-next-line property-no-vendor-prefix -- todo may no longer be needed */
-	-moz-appearance: none;
 	appearance: none;
 	border: 0;
 	text-align: start;
--- a/components/commons/basebutton.css
+++ b/components/commons/basebutton.css
@@ -12,9 +12,6 @@ governing permissions and limitations under the License.
 
 %spectrum-BaseButton {
 	cursor: pointer;
-
-	/* stylelint-disable-next-line property-no-vendor-prefix -- need to validate if this can be removed */
-	-webkit-user-select: none;
 	user-select: none;
 
 	/* Contain halo */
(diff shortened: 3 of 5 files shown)


castastrophe:
Interesting that our polyfilling tool added the webkit prefix inside the `:is(button, [type="button"],...)`. Do we know if this will cause an issue for the web components since they're assigning the styles for our root classes to the host selector which will never match `button` (but could match the type selectors if those are being applied correctly).

Screenshot 2024-05-22 at 1 44 46 PM

jawinn:
> Interesting that our polyfilling tool added the webkit prefix inside the `:is(button, [type="button"],...)`. Do we know if this will cause an issue for the web components since they're assigning the styles for our root classes to the host selector which will never match `button` (but could match the type selectors if those are being applied correctly).
> 
That one is a normalization that only needs the `-webkit` prefix (which is prefixed in the source CSS and ignored via a comment). The style should only be necessary if it does match AFAIK Here is the related normalization in [modern-normalize](https://github.com/sindresorhus/modern-normalize/blob/c997a1e1d37aadbf9b6bc46ed396e7d9ebc8632f/modern-normalize.css#L174) and [csstools/normalize.css](https://github.com/csstools/normalize.css/blob/a4a0c202e8b34426e0cd8e0411005906cd94e407/normalize.css#L123). 

I do see [this CSS override in SWC](https://github.com/adobe/spectrum-web-components/blob/5d189c2adf91545eb64a0136fa02d7c8caee8f48/packages/button/src/button-base.css#L23), which this selector change might help with it being more targeted to the types/element it's needed on, but I'm unclear about why they are only selecting `[dir]` there.

castastrophe:
I think I'd like to do a beta release for this one and run a quick test in SWC

jawinn:
@castastrophe I've removed the questionable `-webkit-appearance` change from`%spectrum-BaseButton`, and have created a followup Jira issue CSS-774 if we want to look into that later.

castastrophe (approved):

PR #2802, opened 2024-05-30
feat(menu): s2 migration

rise-erpelding:
## Description

This migration was a big one--lots of features were added to the design spec for S2. To make way for these, menu was previously refactored on `main` to make the Storybook implementation easier to add to. There are two temporary commits at the beginning of this PR ("keep selector list in dist output..." and "improve chromatic coverage") containing, respectively, a fix that addresses menu, and recent menu refactor changes made on `main` that this work depends on--I'll remove this once `spectrum-two` has been rebased to include these changes.

In addition to small token and minor style changes to match the [design spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-571&t=KyqswJ2j7M7ykpdv-0), there were several new features that were added to this version of menu, including:
* A thumbnail can now be added to the menu item label
* A section description can now be added below the menu section heading
* The actions area previously held value text, action switches for multi-select, and the drill-in for a submenu--in this version, an external link action can now be added
* Menu items now receive corner rounding, which is visible on hover/focus/active states
* Menu items now have the S2 down state transform
* The focus indicator has been updated--instead of a blue line on the left (start side) of the menu item, the whole element now receives an outline
* Because the focus outline has been added, the tray and popover components that contain menu will require padding. Padding for popover is in the [S2 design spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-696&t=0ODBxYdGbSwvwLmP-0), and [there has been discussion about including it for tray](https://adobedesign.slack.com/archives/GQ09RCBA5/p1717512040878679?thread_ts=1717459282.852639&cid=GQ09RCBA5) as well.

Since `@adobe/spectrum-tokens` has been updated to include many tokens relating to the menu component, including some that replace custom tokens that had previously been added. As such, these custom menu item color tokens that are now available from `@adobe/spectrum-tokens` have been removed.

The following mods have been added:
- `--mod-menu-item-corner-rounding`
- `--mod-menu-item-description-font-weight`
- `--mod-menu-item-label-font-weight`
- `--mod-menu-item-thumbnail-height`
- `--mod-menu-item-thumbnail-opacity-disabled` 
- `--mod-menu-item-thumbnail-to-label`
- `--mod-menu-item-thumbnail-width`
- `--mod-menu-item-top-to-drillin`
- `--mod-menu-item-top-to-icon`
- `--mod-menu-item-top-to-thumbnail`
- `--mod-menu-linkout-icon-height`
- `--mod-menu-linkout-icon-width`
- `--mod-menu-section-description-color`
- `--mod-menu-section-description-font-size`
- `--mod-menu-section-description-font-weight`
- `--mod-menu-section-description-line-height`
- `--mod-menu-section-description-line-height-cjk`
- `--mod-menu-section-header-to-description`
- `--mod-popover-content-area-spacing-horizontal`
- `--mod-tray-edge-to-content-area`

Note that there is currently one outstanding item that needs to be addressed before merging:
- [ ] Spectrum 2 Icons need to be added to this once available - therefore the external link icon _does not_ match the spec

### Validation steps

- New features added are visible in Storybook and are covered by a Chromatic testing view for one or more stories on the menu component. @marissahuysentruyt 
    - [x] Thumbnail
    - [x] Section description
    - [x] External link
    - [x] Corner-rounding
    - [x] Focus indicator
    - [x] Down state
- Menu item elements (icon, thumbnail, checkmarks, checkboxes, switches, external linkout, etc.) all play nicely with each other and nothing appears out of place when combining various elements (**Note:** Storybook will throw hooks error messages on the screen when switching select modes, it goes away if you reload) @marissahuysentruyt 
    - [x] No selection mode
    - [x] Single select mode
    - [x] Multi-select mode with switches
    - [x] Muti-select mode with checkboxes
- Menu items and other components within menu behave as expected on hover/active/focus states (for instance, nothing disappears that shouldn't, disabled items don't get hover/active states or pointer cursors, contrast seems appropriate)
    - [ ] In regular light/dark contexts
    - [ ] In Windows High Contrast Mode
- New tokens listed in [Figma](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-571&t=t4QTWi13rbKIw1ce-0) are used @marissahuysentruyt 
    - [x] `menu-item-label-to-description` size tokens
    - [x] `section-header-to-description` size tokens
    - [x] `menu-top-to-thumbnail` size tokens
    - [x] `menu-item-background-color` and `menu-item-background-opacity` tokens
    - [x] `text-to-visual-400` (used for thumbnail to label inline spacing)
    - [x] `link-out-icon` size tokens
- Menu matches the [design spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-571&t=t4QTWi13rbKIw1ce-0) @marissahuysentruyt 
    - [x] Layout - heights and spacings appear to match spec, and corner rounding tokens are in use (**Note:** Sub-menu alignment with menu item is outside of the scope of Spectrum CSS, _I think_)
    - [x] Assets - icons, checkboxes, and thumbnails are appropriately sized (**Note:** External link icon needs to be updated to S2 icon when these icons are available to us)
    - [x] Colors - background and content colors use the appropriate color tokens
    - [x] Typography - menu headers and items are using the appropriate typography tokens
- Popover and Tray components have 8px of horizontal margin added per discussion [about adding spacing to accommodate focus outline](https://adobedesign.slack.com/archives/G019JTYMT6H/p1717529135522869) @marissahuysentruyt 

Links that may help in validation:
- [S2 Token Specs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-571&t=t4QTWi13rbKIw1ce-0)
- [S2 Foundations and Components Changelog](https://www.figma.com/design/M9P2s2IDhGWZuNyvJtXrx3/S2-foundations-and-components-changelog?node-id=2635-5785&t=PvOYvCTqTr5duOa7-0)
- [S2 Desktop](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?node-id=37252-553&t=JjqDcmTcp5v3b9Zc-0)
- [Internal Menu Docs](https://spectrum.corp.adobe.com/page/menu/)

### Regression testing

Validate:

## Screenshots
image

## To-do list


components/menu/index.css
```diff
@@ -12,8 +12,10 @@ governing permissions and limitations under the License.
 
 .spectrum-Menu {
 	--spectrum-menu-item-background-color-default-rgba: rgba(var(--spectrum-menu-item-background-color-default), var(--spectrum-menu-item-background-opacity));
+	--spectrum-menu-item-background-color-key-focus: var(--spectrum-menu-item-background-color-keyboard-focus);
```
rise-erpelding, line 15: It seems a little odd that the token is named with `keyboard-focus` instead of `key-focus` like most other tokens I've seen, so I'm using it to fit with the custom property we're already using for these styles

jawinn (changes requested):
Just an initial review here after comparing against the changes noted on the design and poking around some things in Storybook.

**Thumbnail + description spacing**
Screenshot 2024-06-06 at 1 49 08 PM
The way things wrap here seems odd. I don't think design has this combination included on the Figma for reference? Should the description text be right below the label text?

**"Link out" UI icon and UI icon sizing docs**
It looks like the "Link out" icon as defined on the Figma is a UI icon with specific sizing using tokens that would come from the tokens component itself. I took a peak at the WIP UI icons package update and saw the "Link out" icons in there. So we may have to wait for that until we can access them?

I'm not sure if we need to be defining CSS custom properties in this component for the sizes. It would be similar to what's already done in the Storybook template for the `scales` of the arrow and checkmark.

It may be good idea to have in the docs / usage notes, which UI icon size number is used for each t-shirt size, for all three of those icons (checkmark, chevron, link out). Similar to what was done for the Close Button YML docs  where there is a table. Right now this info just kind of lives in the Storybook template code and on the Figma.

**Down state and focus on Collapsible**
The down state and focus for collapsible is a little odd, and I'm wondering what design intends for those. Down state affects all items within the same level, rather than just itself. And the focus ring for the top level item goes around the whole thing including children. I know the nested structure as it stands might make this difficult to change?

**Storybook thoughts from a user perspective**

Menu Item Only -- These sets of two controls could be combined into a single control for the text, and if the text is empty, it does not display:
- Menu item description for single menu item & Has menu item description
- Menu item value for single menu item & Has value text

Has Icon -- would be nice to default to a workflow icon value if possible. It seemed like the control wasn't doing anything until I saw the new control that appeared.

Menu with Variants -- any reason why there isn't a divider appearing before the drill-in section, when "has dividers" is enabled?

Truncate menu item label -- I was wondering what this was doing and had to dig into the code. It could use a description. It appears to add the "spectrum-Menu-itemLabel--truncate" class which changes the overflow behavior. I also see there is a "maxInlineSize" that could use a description that its for testing and perhaps that should have a different control category.

I'm also feeling like the primary story should not the collapsible variant, and that the Standard with divider should be shown to users as a more bare bones example?

components/menu/stories/menu.stories.js
```diff
@@ -53,7 +63,7 @@ export default {
 		selectionMode: "none",
 		size: "m",
 		shouldTruncate: false,
-		maxInlineSize: "150px"
+		maxInlineSize: "150px",
 	},
 	parameters: {
 		actions: {
```
marissahuysentruyt, line 69: This is separate from the work you did, but did we want to continue tracking every time one of the `.spectrum-Menu-item` elements is clicked? I think these results are just verifying that it's listening to the event appropriately (https://storybook.js.org/docs/essentials/actions#action-event-handlers)

Screenshot 2024-06-06 at 4 24 04 PM

components/menu/index.css
marissahuysentruyt: Agreed. It looks like there might just be some inconsistency though, too. In combobox, it's opposite 😆 Although these examples are specifically the `mod` properties, so maybe it's purposely different.

```
--mod-textfield-text-color-keyboard-focus: var(--mod-combobox-font-color-key-focus);
--mod-textfield-border-color-invalid-keyboard-focus: var(--mod-combobox-border-color-invalid-key-focus, var(--spectrum-combobox-border-color-invalid-key-focus));
```

components/popover/metadata/mods.md
```diff
@@ -1,18 +1,19 @@
...
-| `--mod-popover-shadow-vertical`               |
+| Modifiable custom properties                    |
+| ----------------------------------------------- |
+| `--mod-overlay-animation-duration`              |
+| `--mod-overlay-animation-duration-opened`       |
+| `--mod-popover-animation-distance`              |
+| `--mod-popover-background-color`                |
+| `--mod-popover-border-color`                    |
+| `--mod-popover-border-width`                    |
+| `--mod-popover-content-area-spacing-horizontal` |
```
marissahuysentruyt, line 9: Does this mod need a callout in the `popover.yml` at all since it was added?

components/tray/metadata/mods.md
```diff
@@ -4,6 +4,7 @@
 | `--mod-tray-bottom-to-content-area`         |
 | `--mod-tray-corner-radius`                  |
 | `--mod-tray-corner-radius-portrait`         |
+| `--mod-tray-edge-to-content-area`           |
```
marissahuysentruyt, line 7: Same question like popover- does this need to go in the `tray.yml`? Are we supporting that file still- isn't just for the docs site?

components/tray/index.css
```diff
@@ -11,97 +11,87 @@ governing permissions and limitations under the License.
...
+		visibility var(--mod-tray-exit-animation-duration, var(--spectrum-tray-exit-animation-duration)) linear calc(var(--mod-tray-exit-animation-delay, var(--spectrum-tray-exit-animation-delay)) + var(--mod-tray-exit-animation-duration, var(--spectrum-tray-exit-animation-duration))),
+		transform var(--mod-tray-exit-animation-duration, var(--spectrum-tray-exit-animation-duration)) cubic-bezier(0.5, 0, 1, 1) var(--mod-tray-exit-animation-delay, var(--spectrum-tray-exit-animation-delay));
+
+	background-color: var(--highcontrast-tray-background-color, var(--mod-tray-background-color, var(--spectrum-tray-background-color)));
+
+	&.is-open {
+		/* Entry animations */
+		transition:
+			transform var(--mod-tray-entry-animation-duration, var(--spectrum-tray-entry-animation-duration)) cubic-bezier(0, 0, 0.4, 1) var(--mod-tray-entry-animation-delay, var(--spectrum-tray-entry-animation-delay)),
+			opacity var(--spectrum-tray-entry-animation-duration, var(--mod-tray-entry-animation-duration)) cubic-bezier(0, 0, 0.4, 1) var(--mod-tray-entry-animation-delay, var(--spectrum-tray-entry-animation-delay));
```
marissahuysentruyt, line 74: I'm struggling a little with this `opacity` block, so I might be totally wrong. But should the mod property come before the `--spectrum` property? 

from this:
`opacity var(--spectrum-tray-entry-animation-duration, var(--mod-tray-entry-animation-duration))`

to this: 
`opacity var(--mod-tray-entry-animation-duration, var(--spectrum-tray-entry-animation-duration))`

I noticed it was the only pairing in this section that's got the `spectrum` var first.

tokens/dist/css/spectrum/custom-light-vars.css
```diff
@@ -1,12 +1,5 @@
 .spectrum--light,
 .spectrum--lightest{
-  --spectrum-menu-item-background-color-default-rgb:0, 0, 0;
-  --spectrum-menu-item-background-color-default-opacity:0;
-  --spectrum-menu-item-background-color-default:rgba(var(--spectrum-menu-item-background-color-default-rgb), var(--spectrum-menu-item-background-color-default-opacity)); /* --spectrum-gray-900 */
-  --spectrum-menu-item-background-color-hover:var(--spectrum-transparent-black-200);
-  --spectrum-menu-item-background-color-down:var(--spectrum-transparent-black-200);
-  --spectrum-menu-item-background-color-key-focus:var(--spectrum-transparent-black-200);
-
```
marissahuysentruyt, line 9: Were all of these `tokens` files affected from a particular version or something? Did you have to use a specific package or something that no longer has these variables?

tokens/custom-spectrum/custom-dark-vars.css
```diff
@@ -12,11 +12,6 @@ governing permissions and limitations under the License.
 /* This file contains overrides and additions to core tokens */
 
 .spectrum--dark {
-  --spectrum-menu-item-background-color-default: rgba(255, 255, 255, 0); /* --spectrum-gray-900 */
-  --spectrum-menu-item-background-color-hover: var(--spectrum-transparent-white-200);
-  --spectrum-menu-item-background-color-down: var(--spectrum-transparent-white-200);
-  --spectrum-menu-item-background-color-key-focus: var(--spectrum-transparent-white-200);
```
marissahuysentruyt, line 18: Did these just move into the menu's `index.css` instead then?

components/menu/index.css
```diff
@@ -1,5 +1,5 @@
 /*!
-Copyright 2023 Adobe. All rights reserved.
+Copyright 2024 Adobe. All rights reserved.
```
marissahuysentruyt, line 2: oooooh great catch here. I better do this for status light!

components/menu/index.css
marissahuysentruyt, line 1: Is the `background-color` [here](https://github.com/adobe/spectrum-css/blob/7a77ce5f9d54dc3d0698f1583d445b6c28c431ee/components/menu/index.css#L811) supposed to point to the new `menu-item-bakground-color-disabled` token? I don't think I can find that token in the css...I might be wrong!

components/menu/index.css
```diff
@@ -11,849 +11,945 @@ governing permissions and limitations under the License.
...
-  --spectrum-menu-item-top-edge-to-text: var(--spectrum-component-top-to-text-200);
-  --spectrum-menu-item-bottom-edge-to-text: var(--spectrum-component-bottom-to-text-200);
+	--spectrum-menu-item-description-font-size: var(--spectrum-font-size-100);
 
-  --spectrum-menu-item-text-to-control: var(--spectrum-text-to-control-200);
+	--spectrum-menu-section-header-font-size: var(--spectrum-font-size-200);
+	--spectrum-menu-section-header-min-width: var(--spectrum-component-height-200);
+	--spectrum-menu-section-description-font-size: var(--spectrum-font-size-100);
+	--spectrum-menu-section-header-to-description: var(--spectrum-menu-section-header-to-description-large);
+	--spectrum-menu-item-label-to-description-spacing: var(--spectrum-menu-item-label-to-description-large);
```
marissahuysentruyt, line 196: What should we do with the word "spacing" in these few properties? Should `--spectrum-menu-section-header-to-description` be `--spectrum-menu-section-header-to-description-spacing` (adding `-spacing`)
OR
Should --`spectrum-menu-item-label-to-description-spacing` be `--spectrum-menu-item-label-to-description` (without `-spacing`)? 

I like the idea of having spacing in the property name, but none of the other properties have that so I'm cool with removing it from `label-to-description`. What do you think?

components/menu/index.css
marissahuysentruyt: How are we handling the `menu-item-edge-to-content-not-selected-*` tokens? Are we using `--spectrum-menu-item-selectable-edge-to-text-not-selected-*` sizes instead? I think there's some discrepancy on whether it's 24px or 34px in `spectrum--large`/mobile, at least in the `tokens/dist/index.css` when comparing those properties.

components/menu/index.css
```diff
@@ -11,849 +11,945 @@ governing permissions and limitations under the License.
...
+		color: var(--highcontrast-menu-item-color-focus, var(--mod-menu-checkmark-icon-color-hover, var(--spectrum-menu-checkmark-icon-color-hover)));
+	}
 }
 
 .spectrum-Menu-item:active {
-  background-color: var(--highcontrast-menu-item-background-color-focus, var(--mod-menu-item-background-color-down, var(--spectrum-menu-item-background-color-down)));
+	background-color: var(--highcontrast-menu-item-background-color-focus, var(--mod-menu-item-background-color-down, var(--spectrum-menu-item-background-color-down)));
+
+	/* stylelint-disable-next-line spectrum-tools/no-unknown-custom-properties -- height and width are set by implementations */
+	transform: perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down))) translateZ(var(--spectrum-component-size-difference-down));
```
marissahuysentruyt, line 499: Would the `component-size-minimum-perspective-down` token come into play here at all? It's called out in the design, but I'm not super clear on how we would use it here.

components/menu/index.css
```diff
@@ -11,849 +11,945 @@ governing permissions and limitations under the License.
...
+
+	--spectrum-menu-section-header-font-size: var(--spectrum-font-size-100);
+	--spectrum-menu-section-header-min-width: var(--spectrum-component-height-100);
+	--spectrum-menu-section-description-font-size: var(--spectrum-font-size-75);
+	--spectrum-menu-section-header-to-description: var(--spectrum-menu-section-header-to-description-medium);
+
+	--spectrum-menu-item-selectable-edge-to-text-not-selected: var(--spectrum-menu-item-selectable-edge-to-text-not-selected-medium);
+
+	--spectrum-menu-item-checkmark-height: var(--spectrum-checkmark-icon-size-100);
+	--spectrum-menu-item-checkmark-width: var(--spectrum-checkmark-icon-size-100);
```
marissahuysentruyt, line 45: These are pointing to the incorrect values, right? It looks like 75 & 100 are both 10px right now, but that's not how it should be according to Figma, correct?

components/menu/index.css
```diff
@@ -11,849 +11,945 @@ governing permissions and limitations under the License.
...
+	--spectrum-menu-section-description-line-height: var(--spectrum-line-height-100);
+	--spectrum-menu-section-description-line-height-cjk: var(--spectrum-cjk-line-height-100);
+	--spectrum-menu-section-description-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-menu-section-description-color: var(--spectrum-neutral-subdued-content-color-default);
+	--spectrum-menu-collapsible-icon-color: var(--spectrum-gray-900);
+
+	--spectrum-menu-checkmark-icon-color-default: var(--spectrum-accent-content-color-default);
+	--spectrum-menu-checkmark-icon-color-hover: var(--spectrum-accent-content-color-hover);
+	--spectrum-menu-checkmark-icon-color-down: var(--spectrum-accent-content-color-down);
+	--spectrum-menu-checkmark-icon-color-focus: var(--spectrum-accent-content-color-key-focus);
```
marissahuysentruyt, line 97: Should our property be `--spectrum-menu-checkmark-icon-color-key-focus`? They weren't like that before, I'm just thinking of making it all consistent.

marissahuysentruyt: Oh and does this "checkmark" chunk need a "disabled" property? To use the `disabled-content-color` token? I don't have a good grasp on how `.is-disabled` is affecting menu items like a checkmark though...

marissahuysentruyt (commented):
@rise-erpelding holy cow this was a big one! I left lots of questions for you, and only found a few spots where maybe we were using a token. That could also be a purposeful decision based on conversations I just don't have context on right now too, so ignore those if you need! 

I agree with Josh that the layout of the menu item, when it has a thumbnail, label and description is weird. It feels like we need to some how wrap the spans in an extra div maybe? Or maybe we just need to edit the grid? I don't quite know the solution to that yet...hopefully nobody decides to put switches and values in a single item too!!! 😆

Screenshot 2024-06-07 at 3 08 37 PM

components/popover/metadata/mods.md
rise-erpelding: Probably wouldn't be a bad idea (and for tray)! Thanks for the suggestion.

tokens/dist/css/spectrum/custom-light-vars.css
rise-erpelding: The short answer: These were custom variables that I think were written in because there weren't tokens, but now there are, so I'm (mostly) using those instead.

The longer answer: [It looks like `spectrum-tokens@13.0.0-beta.33`](https://github.com/adobe/spectrum-tokens/releases/tag/%40adobe%2Fspectrum-tokens%4013.0.0-beta.33) adds `--spectrum-menu-item-background-color-default` plus tokens for hover, down, keyboard focus, and disabled states, and they're in the `dist/index.css` from [this PR](https://github.com/adobe/spectrum-css/pull/2797/files#diff-6e8dc6638f2705c29fc830de60cfde41bca79f26502c4bb0607b7e0055786005). Since they're now coming from tokens, I'd rather use that value than the older custom value that was being used before.

Does this (either of these answers 😆 ) answer your questions?

tokens/custom-spectrum/custom-dark-vars.css
rise-erpelding: These are coming from the tokens now! See https://github.com/adobe/spectrum-css/pull/2802#discussion_r1633830351

components/menu/index.css
rise-erpelding: Good catch on both of these! Disabled menu items should be using that `menu-item-background-color-disabled` token along with the `menu-item-background-opacity` token. And we should be using `--spectrum-menu-item-edge-to-content-not-selected-*` tokens, I think those might have the values that match Figma, I wonder if that was a new addition that wasn't marked?

components/menu/index.css
rise-erpelding: Good point, it'd be good if they matched. Since we already had `--spectrum-menu-item-label-to-description-spacing` and I'm hesitant to change it, I'll update `--spectrum-menu-section-header-to-description` to be `--spectrum-menu-section-header-to-description-spacing` instead.

components/menu/index.css
rise-erpelding: Based on the [Figma spec for down state](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=8079-857&t=CarrjiAwalhIRslb-0) and the [Storybook Docs for down state](https://pr-2802--spectrum-css.netlify.app/preview/?path=/docs/foundations-down-state--docs), that `minimum-perspective-down` token is used if the component is less than 24px wide, which menu item can never be by default because of the padding, so I think this is ok to be left out here.

marissahuysentruyt: ohohohoh. I see- thanks!

components/menu/index.css
rise-erpelding: Correct, it looks like the tokens are not pointing to the same values in Figma, but they're pointing to the right tokens--something to call out when this gets design reviewed. Thanks!

components/menu/index.css
rise-erpelding: 🤔 We didn't use `key-focus` for any of the property definitions before, so I'm hesitant to change it in the migration, but I spy one `key-focus` property definition that I added in that branch, I'm wondering if maybe I should remove that one.

I didn't add a disabled icon color, it looks like the disabled color is coming through from the `.spectrum-Menu-itemIcon` class being set on it and all the icons are set to `--spectrum-menu-item-label-icon-color-disabled`.

5t3ph:
Closing per updated work on this migration in #3686

PR #2793, opened 2024-05-28
docs(breadcrumb, checkbox, clearbutton, coachindicator): Site Docs to Storybook

marissahuysentruyt:
## Description

This PR continues migrating documentation from the docs site into Storybook. The focus is on `breadcrumb`, `checkbox`, `clearbutton` and `coachindicator`. All variants of the component should now be displayed on the Storybook `Docs` page.

- `breadcrumb`: ~~new MDX docs page and~~ 3 `docs-only` stories
- `checkbox`: ~~new MDX docs page~~, usage notes and a `docs-only` story
- `clearbutton`: ~~new MDX docs page and~~ additional docs (Usage Notes) regarding proper icon size
- `coachindicator`: ~~new MDX docs page~~

**This PR doesn't need a changeset since it's docs-only.**

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps
@jawinn 
- [x] The [breadcrumb storybook docs](https://pr-2793--spectrum-css.netlify.app/preview/?path=/docs/components-breadcrumbs--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/breadcrumb.html)
- [x] The [checkbox storybook docs](https://pr-2793--spectrum-css.netlify.app/preview/?path=/docs/components-checkbox--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/checkbox.html)
- [x] The [clear button storybook docs](https://pr-2793--spectrum-css.netlify.app/preview/?path=/docs/components-clear-button--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/clearbutton.html)
- [x] The [coach indicator storybook docs](https://pr-2793--spectrum-css.netlify.app/preview/?path=/docs/components-coach-indicator--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/coachindicator.html)
- [ ] Ensure the Chromatic testing view captures all new variants.

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [x] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [x] VRTs have been run and looked at.
- [x] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

## To-do list

[...]

PR #2833, opened 2024-06-10
docs(dial,dialog,divider,dropindicator): Migrate docs to storybook pt7

marissahuysentruyt:
## Description

This PR continues migrating documentation from the static docs site into Storybook. The focus is on `dial`, `dialog` (standard dialog), `divider` and `dropindicator`. All variants of the components should now be displayed on the Storybook `Docs` page.

- `dial`: ~~new MDX docs page~~ new stories
- `dialog`: !!new MDX docs page~~ new stories and controls (plus an additional note regarding alert dialog)
- `divider`: ~~new MDX docs page~~ new stories
- `dropindicator`: ~~new MDX docs page~~ 

**This PR doesn't need a changeset since it's docs-only.**

### Validation steps

- [x] The [dial storybook docs](https://pr-2833--spectrum-css.netlify.app/preview/?path=/docs/components-dial--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/dial.html)
- [x] The [dialog storybook docs](https://pr-2833--spectrum-css.netlify.app/preview/?path=/docs/components-dialog--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/dialog.html)
- [x] The [divider button storybook docs](https://pr-2833--spectrum-css.netlify.app/preview/?path=/docs/components-divider--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/divider.html)
- [x] The [drop indicator storybook docs](https://pr-2833--spectrum-css.netlify.app/preview/?path=/docs/components-drop-indicator--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/dropindicator.html)
- [ ] Ensure the Chromatic testing view for each component captures all new variants.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/dial/stories/dial.stories.js
+++ b/components/dial/stories/dial.stories.js
@@ -1,4 +1,5 @@
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
+import { isFocused } from "@spectrum-css/preview/types";
 import { version } from "../package.json";
 import { DialGroup } from "./dial.test.js";
 import { Template } from "./template.js";
@@ -28,15 +29,7 @@ export default {
 			},
 			control: "text",
 		},
-		isFocusVisible: {
-			name: "Focused",
-			type: { name: "boolean" },
-			table: {
-				type: { summary: "boolean" },
-				category: "State",
-			},
-			control: "boolean",
-		},
+		isFocusVisible: isFocused,
 		isDragged: {
 			name: "Dragged",
 			type: { name: "boolean" },
--- a/components/dial/stories/dial.stories.js
+++ b/components/dial/stories/dial.stories.js
@@ -1,7 +1,7 @@
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
-import { isFocused } from "@spectrum-css/preview/types";
 import { version } from "../package.json";
 import { DialGroup } from "./dial.test.js";
+import { Template } from "./template.js";
 
 /**
  * A dial is an input control used for selecting a value within a range, similar to a slider. It's often used in audio and video mixing and editing applications, where horizontal space is limited.
@@ -28,7 +28,15 @@ export default {
 			},
 			control: "text",
 		},
-		isFocused,
+		isFocusVisible: {
+			name: "Focused",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "State",
+			},
+			control: "boolean",
+		},
 		isDragged: {
 			name: "Dragged",
 			type: { name: "boolean" },
@@ -51,7 +59,7 @@ export default {
 	args: {
 		rootClass: "spectrum-Dial",
 		size: "m",
-		isFocused: false,
+		isFocusVisible: false,
 		isDragged: false,
 		isDisabled: false,
 	},
@@ -62,6 +70,33 @@ export default {
 
 export const Default = DialGroup.bind();
 Default.args = {};
+
+export const Small = Template.bind();
+Small.args = {
+	size: "s",
+};
+Small.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Small.tags = ["!dev"];
+
+export const WithLabel = Template.bind();
+WithLabel.args = {
+	label: "Volume",
+};
+WithLabel.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+WithLabel.tags = ["!dev"];
+
+export const Disabled = Template.bind();
+Disabled.args = {
+	isDisabled: true,
+};
+Disabled.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Disabled.tags = ["!dev"];
 
 // ********* VRT ONLY ********* //
 export const WithForcedColors = DialGroup.bind({});
--- a/components/dial/stories/template.js
+++ b/components/dial/stories/template.js
@@ -10,7 +10,7 @@ export const Template = ({
 	rootClass = "spectrum-Dial",
 	size = "m",
 	label,
-	isFocused = false,
+	isFocusVisible = false,
 	isDragged = false,
 	isDisabled = false,
 	min = 0,
@@ -38,58 +38,58 @@ export const Template = ({
         if (isDisabled) return;
         if (!document.body.classList.contains("u-isGrabbing")) return;
 
-        const dial = e.target.closest(".spectrum-Dial");
-        const handle = dial.querySelector(".spectrum-Dial-handle");
-        const input = dial.querySelector("input");
-        const min = -45;
-        const max = 225;
-        const dialOffsetLeft = dial.offsetLeft + dial.offsetParent.offsetLeft;
-        var x = Math.max(Math.min(e.x - dialOffsetLeft, dial.offsetWidth), 0);
-        var percent = (x / dial.offsetWidth) * 100;
-        var deg = percent * 0.01 * (max - min) + min;
-        handle.style.transform = "rotate(" + deg + "deg" + ")";
-        input.value = Math.round(
-          percent * 0.01 * (input.max - input.min) + input.min
-        );
-      }}
-    >
-      ${when(
-        label,
-        () => html`<div class="${rootClass}-labelContainer">
-          <label id="dialLabel" class="${rootClass}-label" for="labeledDial"
-            >${label}</label
-          >
-          <div
-            class="${rootClass}-value"
-            role="textbox"
-            aria-readonly="true"
-            aria-labelledby="dialLabel"
-          >
-            ${min}
-          </div>
-        </div>`
-      )}
-      <div class="${rootClass}-controls">
-        <div
-          class="${rootClass}-handle ${isDragged ? "is-dragged" : ""} ${isFocused
-            ? "is-focused"
-            : ""}"
-          tabindex="0"
-        >
-          <input
-            type="range"
-            class="${rootClass}-input"
-            min="${min}"
-            max="${max}"
-            value="${min}"
-            @change=${(e) => {
-              const value = e.target.value;
-              const label = document.getElementById("dialLabel");
-              label.nextSibling.textContent = value;
-            }}
-          />
-        </div>
-      </div>
-    </div>
-  `;
+				const dial = e.target.closest(".spectrum-Dial");
+				const handle = dial.querySelector(".spectrum-Dial-handle");
+				const input = dial.querySelector("input");
+				const min = -45;
+				const max = 225;
+				const dialOffsetLeft = dial.offsetLeft + dial.offsetParent.offsetLeft;
+				var x = Math.max(Math.min(e.x - dialOffsetLeft, dial.offsetWidth), 0);
+				var percent = (x / dial.offsetWidth) * 100;
+				var deg = percent * 0.01 * (max - min) + min;
+				handle.style.transform = "rotate(" + deg + "deg" + ")";
+				input.value = Math.round(
+					percent * 0.01 * (input.max - input.min) + input.min
+				);
+			}}
+		>
+			${when(
+				label,
+				() => html`<div class="${rootClass}-labelContainer">
+					<label id="dialLabel" class="${rootClass}-label" for="labeledDial"
+						>${label}</label
+					>
+					<div
+						class="${rootClass}-value"
+						role="textbox"
+						aria-readonly="true"
+						aria-labelledby="dialLabel"
+					>
+						${min}
+					</div>
+				</div>`
+			)}
+			<div class="${rootClass}-controls">
+				<div
+					class="${rootClass}-handle
+					${isDragged ? "is-dragged": ""}
+					${isFocusVisible ? "is-focus-visible": ""}"
+					tabindex="0"
+				>
+					<input
+						type="range"
+						class="${rootClass}-input"
+						min="${min}"
+						max="${max}"
+						value="${min}"
+						@change=${(e) => {
+							const value = e.target.value;
+							const label = document.getElementById("dialLabel");
+							label.nextSibling.textContent = value;
+						}}
+					/>
+				</div>
+			</div>
+		</div>
+	`;
 };
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -3,47 +3,71 @@ import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isOpen } from "@spectrum-css/preview/types";
 import { Template as Typography } from "@spectrum-css/typography/stories/template.js";
 import { version } from "../package.json";
+// import { DialogGroup } from "./dialog.test.js";
 import { DialogFullscreen, DialogFullscreenTakeover, DialogGroup } from "./dialog.test.js";
+import { Template } from "./template";
 
 /**
  * A dialog displays important information that users need to acknowledge. They appear over the interface and block further interactions.
+ * 
+ * The alert variants that were previously a part of Dialog were moved to their own component, [Alert Dialog](/docs/components-alert-dialog--docs).
  */
 export default {
 	title: "Dialog",
 	component: "Dialog",
 	argTypes: {
-		heading: { table: { disable: true } },
+		dialogHeading: {
+			name: "Heading",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "text" },
+		},
 		content: { table: { disable: true } },
-		footer: { table: { disable: true } },
 		hasFooter: {
 			name: "Has footer",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
-				category: "Component",
+				category: "Content",
 			},
 			control: "boolean",
 		},
-		isDismissable: {
-			name: "Dismissable",
+		footer: { table: { disable: true } },
+		size: {
+			name: "Size",
+			type: { name: "string", required: true },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
+			},
+			options: ["small", "medium", "large"],
+			control: "select",
+		},
+		layout: {
+			name: "Layout",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
+			},
+			options: ["default", "fullscreen", "fullscreenTakeover"],
+			control: "select",
+		},
+		isDismissible: {
+			name: "Dismissible",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
 				category: "Component",
 			},
 			control: "boolean",
+			if: { arg: "layout", eq: "default" },
 		},
-		isFullScreen: {
-			name: "Fullscreen",
-			type: { name: "boolean" },
-			table: {
-				type: { summary: "boolean" },
-				category: "Component",
-			},
-			control: "boolean",
-		},
-		isFullScreenTakeover: {
-			name: "Fullscreen takeover",
+		hasDivider: {
+			name: "Divider",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
@@ -61,15 +85,37 @@ export default {
 			control: "boolean",
 		},
 		isOpen,
+		hasHeroImage: {
+			name: "Has hero image",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Content",
+			},
+			control: "boolean",
+			if: { arg: "layout", eq: "default" },
+		},
+		heroImageUrl: {
+			name: "Hero Image",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "file", accept: ".svg,.png,.jpg,.jpeg,.webc" },
+			if: { arg: "hasHeroImage", truthy: true },
+		},
 	},
 	args: {
 		rootClass: "spectrum-Dialog",
 		hasFooter: true,
-		isDismissable: true,
-		isFullScreen: false,
-		isFullScreenTakeover: false,
+		isDismissible: false,
+		hasDivider: true,
 		isOpen: true,
 		showModal: true,
+		size: "medium",
+		hasHeroImage: false,
+		layout: "default",
 	},
 	parameters: {
 		actions: {
@@ -88,30 +134,172 @@ export default {
 	],
 };
 
+// the "TallerViewport" modes are accommodating the underlay, which is position: fixed,
+// and Chromatic's treatment of position:fixed elements. By increasing the viewport height,
+// it doesn't look like the background color just stops without wrapping the
+// entire container of templates.
+// const defaultModesWithTallerViewport = Object.keys(modes).reduce((acc, key) => {
+// 	acc[key] = { 
+// 		...modes[key],
+// 		viewport: {
+// 			height: "5300px",
+// 		}
+// 	};
+// 	return acc;
+// }, {});
+
+// const mobileModeWithTallerViewport = Object.keys(mobile).reduce((acc, key) => {
+// 	acc[key] = { 
+// 		...mobile[key],
+// 		viewport: {
+// 			height: "5600px",
+// 		}
+// 	};
+// 	return acc;
+// }, {});
+
+const ExampleContent = Typography({
+	semantics: "body",
+	size: "m",
+	content: [
+		"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Auctor augue mauris augue neque gravida. Libero volutpat sed ornare arcu. Quisque egestas diam in arcu cursus euismod quis viverra. Posuere ac ut consequat semper viverra nam libero justo laoreet. Enim ut tellus elementum sagittis vitae et leo duis ut. Neque laoreet suspendisse interdum consectetur libero id faucibus nisl. Diam volutpat commodo sed egestas egestas. Dolor magna eget est lorem ipsum dolor. Vitae suscipit tellus mauris a diam maecenas sed. Turpis in eu mi bibendum neque egestas congue. Rhoncus est pellentesque elit ullamcorper dignissim cras lobortis."
+	]
+});
+
+/**
+ * The default size for dialog is medium.
+ */
 export const Default = DialogGroup.bind({});
-Default.tags = ["dev"];
 Default.args = {
-	heading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+	dialogHeading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
 	content: [
 		(passthroughs, context) => Typography({
 			semantics: "body",
 			size: "m",
-			content: [
-				"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Auctor augue mauris augue neque gravida. Libero volutpat sed ornare arcu. Quisque egestas diam in arcu cursus euismod quis viverra. Posuere ac ut consequat semper viverra nam libero justo laoreet. Enim ut tellus elementum sagittis vitae et leo duis ut. Neque laoreet suspendisse interdum consectetur libero id faucibus nisl. Diam volutpat commodo sed egestas egestas. Dolor magna eget est lorem ipsum dolor. Vitae suscipit tellus mauris a diam maecenas sed. Turpis in eu mi bibendum neque egestas congue. Rhoncus est pellentesque elit ullamcorper dignissim cras lobortis."
-			],
+			content: ExampleContent,
 			...passthroughs,
 		}, context),
 	],
 };
+// Default.parameters = {
+// 	chromatic: {
+// 		modes: {
+// 			...defaultModesWithTallerViewport,
+// 			...mobileModeWithTallerViewport }
+// 	},
+// };
 
+// ********* DOCS ONLY ********* //
+export const DefaultSmall = Template.bind({});
+DefaultSmall.tags = ["!dev"];
+DefaultSmall.storyName = "Dialog - small",
+DefaultSmall.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+DefaultSmall.args = {
+	...Default.args,
+	size: "small",
+};
+
+export const DefaultLarge = Template.bind({});
+DefaultLarge.tags = ["!dev"];
+DefaultLarge.storyName = "Dialog - large",
+DefaultLarge.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+DefaultLarge.args = {
+	...Default.args,
+	size: "large",
+};
+
+/**
+ * A dialog that can be dismissed without taking an action. Dismissible dialogs should never have buttons.
+ */
+export const Dismissible = Template.bind({});
+Dismissible.tags = ["!dev"];
+Dismissible.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Dismissible.args = {
+	...Default.args,
+	isDismissible: true,
+};
+
+/**
+ * Dialogs can forgo the divider if they have content that spans the entire width of the dialog.
+ */
+export const NoDivider = Template.bind({});
+NoDivider.tags = ["!dev"];
+NoDivider.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+NoDivider.args = {
+	...Default.args,
+	isDismissible: true,
+	hasDivider: false,
+};
+
+/**
+ * Dialogs can have a hero or cover image header.
+ */
+export const WithHero = Template.bind({});
+WithHero.tags = ["!dev"];
+WithHero.parameters = {
+	docs: {
+		story: {
+			height: "650px",
+		},
+	},
+	chromatic: { disableSnapshot: true },
+};
+WithHero.args = {
+	...Default.args,
+	isDismissible: true,
+	hasHeroImage: true,
+	heroImageUrl: "example-card-portrait.png",
+};
+
+/**
+ * The content inside the body area can scroll when necessary.
+ */
+export const WithScroll = Template.bind({});
+WithScroll.args = {
+	...Default.args,
+	customStyles: {
+		"max-block-size": "400px",
+	}
+};
+WithScroll.tags = ["!dev"];
+WithScroll.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * A fullscreen dialog will automatically fill almost all of the available screen space. A margin is included around the outside of the dialog.
+ */
 export const Fullscreen = DialogFullscreen.bind({});
-Fullscreen.tags = ["!autodocs", "dev"];
-Fullscreen.args = Default.args;
+// export const Fullscreen = Template.bind({});
+Fullscreen.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Fullscreen.args = {
+	...Default.args,
+	layout: "fullscreen",
+};
 
+/**
+ * A fullscreen takeover dialog will fill all of the available screen space.
+ */
 export const FullscreenTakeover = DialogFullscreenTakeover.bind({});
-FullscreenTakeover.tags = ["!autodocs", "dev"];
-FullscreenTakeover.args = Default.args;
-
+// export const FullscreenTakeover = Template.bind({});
+FullscreenTakeover.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+FullscreenTakeover.args = {
+	...Default.args,
+	layout: "fullscreenTakeover",
+	content: [ExampleContent, ExampleContent, ExampleContent, ExampleContent],
+};
 
 // ********* VRT ONLY ********* //
 export const WithForcedColors = DialogGroup.bind({});
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -86,25 +86,16 @@ export default {
 			control: "boolean",
 		},
 		isOpen,
-		hasHeroImage: {
-			name: "Has hero image",
-			type: { name: "boolean" },
-			table: {
-				type: { summary: "boolean" },
-				category: "Content",
-			},
-			control: "boolean",
-			if: { arg: "layout", eq: "default" },
-		},
 		heroImageUrl: {
 			name: "Hero Image",
 			type: { name: "string" },
+			description: "Adds a cover image to the header of a dialog.",
 			table: {
 				type: { summary: "string" },
 				category: "Content",
 			},
 			control: { type: "file", accept: ".svg,.png,.jpg,.jpeg,.webc" },
-			if: { arg: "hasHeroImage", truthy: true },
+			if: { arg: "layout", eq: "default" },
 		},
 	},
 	args: {
@@ -115,7 +106,6 @@ export default {
 		isOpen: true,
 		showModal: true,
 		size: "medium",
-		hasHeroImage: false,
 		layout: "default",
 	},
 	parameters: {
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -48,9 +48,11 @@ export default {
 		layout: {
 			name: "Layout",
 			type: { name: "string" },
+			defaultValue: "Default",
 			table: {
 				type: { summary: "string" },
 				category: "Component",
+				defaultValue: { summary: "Default" },
 			},
 			options: ["default", "fullscreen", "fullscreenTakeover"],
 			control: "select",
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -246,7 +246,6 @@ WithScroll.parameters = {
  * A fullscreen dialog will automatically fill almost all of the available screen space. A margin is included around the outside of the dialog.
  */
 export const Fullscreen = DialogFullscreen.bind({});
-// export const Fullscreen = Template.bind({});
 Fullscreen.parameters = {
 	chromatic: { disableSnapshot: true },
 };
@@ -259,7 +258,6 @@ Fullscreen.args = {
  * A fullscreen takeover dialog will fill all of the available screen space.
  */
 export const FullscreenTakeover = DialogFullscreenTakeover.bind({});
-// export const FullscreenTakeover = Template.bind({});
 FullscreenTakeover.parameters = {
 	chromatic: { disableSnapshot: true },
 };
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -16,7 +16,7 @@ export default {
 	title: "Dialog",
 	component: "Dialog",
 	argTypes: {
-		dialogHeading: {
+		heading: {
 			name: "Heading",
 			type: { name: "string" },
 			table: {
@@ -171,7 +171,7 @@ const ExampleContent = Typography({
  */
 export const Default = DialogGroup.bind({});
 Default.args = {
-	dialogHeading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+	heading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
 	content: [
 		(passthroughs, context) => Typography({
 			semantics: "body",
--- a/components/dialog/stories/dialog.stories.js
+++ b/components/dialog/stories/dialog.stories.js
@@ -3,7 +3,6 @@ import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isOpen } from "@spectrum-css/preview/types";
 import { Template as Typography } from "@spectrum-css/typography/stories/template.js";
 import { version } from "../package.json";
-// import { DialogGroup } from "./dialog.test.js";
 import { DialogFullscreen, DialogFullscreenTakeover, DialogGroup } from "./dialog.test.js";
 import { Template } from "./template";
 
@@ -134,30 +133,6 @@ export default {
 	],
 };
 
-// the "TallerViewport" modes are accommodating the underlay, which is position: fixed,
-// and Chromatic's treatment of position:fixed elements. By increasing the viewport height,
-// it doesn't look like the background color just stops without wrapping the
-// entire container of templates.
-// const defaultModesWithTallerViewport = Object.keys(modes).reduce((acc, key) => {
-// 	acc[key] = { 
-// 		...modes[key],
-// 		viewport: {
-// 			height: "5300px",
-// 		}
-// 	};
-// 	return acc;
-// }, {});
-
-// const mobileModeWithTallerViewport = Object.keys(mobile).reduce((acc, key) => {
-// 	acc[key] = { 
-// 		...mobile[key],
-// 		viewport: {
-// 			height: "5600px",
-// 		}
-// 	};
-// 	return acc;
-// }, {});
-
 const ExampleContent = Typography({
 	semantics: "body",
 	size: "m",
@@ -181,13 +156,6 @@ Default.args = {
 		}, context),
 	],
 };
-// Default.parameters = {
-// 	chromatic: {
-// 		modes: {
-// 			...defaultModesWithTallerViewport,
-// 			...mobileModeWithTallerViewport }
-// 	},
-// };
 
 // ********* DOCS ONLY ********* //
 export const DefaultSmall = Template.bind({});
--- a/components/dialog/stories/dialog.test.js
+++ b/components/dialog/stories/dialog.test.js
@@ -30,6 +30,7 @@ export const DialogGroup = Variants({
 		{
 			testHeading: "Dismissible",
 			isDismissible: true,
+			hasFooter: false,
 			wrapperStyles: {
 				"background-color": "var(--spectrum-gray-50)"
 			},
@@ -42,9 +43,44 @@ export const DialogGroup = Variants({
 			},
 		},
 		{
-			testHeading: "No divider",
+			testHeading: "With hero/cover image, dismissible",
+			hasHeroImage: true,
 			isDismissible: true,
+			hasFooter: false,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "No divider",
 			hasDivider: false,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "No divider, dismissible",
+			hasDivider: false,
+			isDismissible: true,
+			hasFooter: false,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "With hero/cover image, no divider",
+			hasDivider: false,
+			hasHeroImage: true,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "With hero/cover image, no divider, dismissible",
+			hasDivider: false,
+			isDismissible: true,
+			hasFooter: false,
+			hasHeroImage: true,
 			wrapperStyles: {
 				"background-color": "var(--spectrum-gray-50)"
 			},
--- a/components/dialog/stories/dialog.test.js
+++ b/components/dialog/stories/dialog.test.js
@@ -2,18 +2,46 @@ import { Variants } from "@spectrum-css/preview/decorators";
 import { Template } from "./template.js";
 
 export const DialogGroup = Variants({
-	Template,
+	Template: (args, context) => (
+		/*
+		 * This template forces each test case to showModal: false, and give a background 
+		 * color to the dialog grid container when isChromatic() is true.
+		 * This should ensure that the Sizes within the Variants() template to display correctly.
+		 */
+		window.isChromatic() ?
+			Template({
+				...args,
+				showModal: false,
+				customStyles: {
+					"background-color": "var(--spectrum-gray-100)",
+				},
+			}, context)
+			: Template(args, context)
+	),
 	testData: [
 		{
-			showModal: false,
 			wrapperStyles: {
 				"background-color": "var(--spectrum-gray-50)"
 			},
 		},
 		{
-			testHeading: "Not dismissable",
-			isDismissable: false,
-			showModal: false,
+			testHeading: "Dismissible",
+			isDismissible: true,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "With hero/cover image",
+			hasHeroImage: true,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "No divider",
+			isDismissible: true,
+			hasDivider: false,
 			wrapperStyles: {
 				"background-color": "var(--spectrum-gray-50)"
 			},
@@ -23,12 +51,13 @@ export const DialogGroup = Variants({
 
 export const DialogFullscreen = Variants({
 	Template,
+	withSizes: false,
 	testData: [
 		{
 			showModal: false,
-			isFullScreen: true,
+			layout: "fullscreen",
 			wrapperStyles: {
-				"background-color": "var(--spectrum-gray-100)"
+				"background-color": "var(--spectrum-gray-50)"
 			},
 		},
 	],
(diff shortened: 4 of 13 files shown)


components/dial/stories/dial.stories.js
```diff
@@ -1,5 +1,5 @@
 import { Template } from "./template";
-
+import { html } from "lit";
 /**
  * A dial is an input control used for selecting a value within a range, similar to a slider. It's often used in audio and video mixing and editing applications, where horizontal space is limited.
  */
```
marissahuysentruyt, line 5: I didn't migrate the "usage notes" for dial (it was a funny quip about turning it up to 11). I'm happy to though 😆

marissahuysentruyt: I don't think the `.is-focused` class is doing anything when you toggle the control in storybook. There's nothing in the `index.css` file for dial regarding that class and any styles.

components/dialog/stories/dialog.stories.js
```diff
@@ -4,6 +4,8 @@ import { Template as Typography } from "@spectrum-css/typography/stories/templat
 
 /**
  * A dialog displays important information that users need to acknowledge. They appear over the interface and block further interactions.
+ * 
+ * Alert variants have been removed from Dialog to their own component, [Alert Dialog](/docs/components-alert-dialog--docs).
```
marissahuysentruyt, line 8: I didn't migrate [the usage notes about blurry dialogs and wrapping](https://opensource.adobe.com/spectrum-css/dialog.html#usage) these dialogs in `.spectrum-Dialog-wrapper`. That was added in the yml like 5 years ago, and the link didn't work either. Let me know if you want me to add it, though.

components/dialog/stories/dialog.stories.js
```diff
@@ -19,8 +21,27 @@ export default {
...
-		isDismissable: {
-			name: "Dismissable",
+		size: {
+			name: "Size",
+			type: { name: "string", required: true },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
+			},
+			options: ["small", "medium", "large", "fullscreen", "fullscreenTakeover"],
```
marissahuysentruyt, line 31: I wasn't sure how to handle the sizing and the fullscreen variants. The components I've worked on so far are using that more standardized t-shirt classes (like `--sizeL`), but dialog/index.css is using classes like `--small`. 

When we migrate dialog to S2, is that the time we could refactor those class names?

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +110,159 @@ export default {
...
+}, {
+	treatment: "fill",
+	label: "Enable",
+	variant: "accent"
+}];
+
 export const Default = Template.bind({});
 Default.args = {
-	heading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+	heading: "Disclaimer",
```
marissahuysentruyt, line 133: This very long heading was messing with the grid for dialogs, and really undoing the layout we were trying to create I think. I figured changing this heading was easier than editing the grid or something. And I also thought a dialog in the wild probably would have a heading that's 8 words long. 🤞 

Screenshot 2024-06-11 at 1 31 40 PM

Screenshot 2024-06-11 at 1 30 59 PM

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +110,159 @@ export default {
...
+	content: [
+		() => ExampleContent, ExampleContent, ExampleContent, ExampleContent
+	],
+};
+WithScroll.tags = ["docs-only"];
+WithScroll.parameters = {
+	chromatic: {disabledSnapshot: true},
+};
+
+export const Fullscreen = Template.bind({});
```
marissahuysentruyt, line 240: So the fullscreen and fullscreenTakeover don't really behave appropriately in storybook. Is that a storybook limitation, or is it something I can fix with CSS? Either way, that's potentially a different PR.

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +110,159 @@ export default {
...
+
+const ExampleButtons = [{
+	variant: "secondary",
+	treatment: "outline",
+	label: "Remind me later"
+}, {
+	treatment: "fill",
+	label: "Enable",
+	variant: "accent"
+}];
```
marissahuysentruyt, line 129: None of the dialogs on the docs site currently have buttons, so I added them in any of the examples that _weren't_ dismissible.

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +110,159 @@ export default {
...
+	buttons: ExampleButtons,
+	content: [
+		() => ExampleContent
+	],
+};
+
+/*
+ * Stories for the MDX "docs" only.
+ */
+export const Small = Template.bind({});
```
marissahuysentruyt, line 144: The `--small` variant styles currently don't display properly. I think if we moved the `--small` selector block in the CSS below the `--medium`, it would work with the cascade better (similar to how we have it set up with some of the other migrated components). 

Additionally, there might be some refactoring that could be possible in the future. If I inspect any of the dialogs, I saw both modifier classes are being added. That seemed odd, but I couldn't quite figure out where things are buggy. 

For instance, on the large variant, I see these classes: `spectrum-Dialog spectrum-Dialog--medium spectrum-Dialog--large ` even though I should probably only see `spectrum-Dialog spectrum-Dialog--large `

components/divider/stories/divider.stories.js
```diff
@@ -1,3 +1,4 @@
+import { html } from "lit";
 import { Template } from "./template";
 
 /**
```
marissahuysentruyt, line 4: I didn't [migrate anything about the icon sizes](https://opensource.adobe.com/spectrum-css/divider.html#migrationguide). Let me know if I should add that.

components/dial/stories/dial.stories.js
```diff
@@ -62,25 +63,71 @@ export default {
 	},
 };
 
+// the !auto-docs tag removes the component from the auto-template
 export const Default = Template.bind();
 Default.args = {};
+Default.tags = ['!autodocs'];
```
marissahuysentruyt, line 69: I found `!autodocs` here: https://storybook.js.org/docs/writing-docs/autodocs and I'm using it to remove this Default dial story from our auto-template, so that I can just display the dial groups like on the [docs site](https://opensource.adobe.com/spectrum-css/dial.html)

components/dial/stories/dial.stories.js
```diff
@@ -62,25 +63,71 @@ export default {
...
+	})}
+	${Template({
+		...args,
+		size: "s",
+		isDisabled: true,
+	})}
+	</div>
+`;
+DialGroup.tags = ["docs-only"];
+DialGroup.storyName = "Standard";
```
marissahuysentruyt, line 120: I also found `storyName` here: https://storybook.js.org/docs/writing-stories/naming-components-and-hierarchy#single-story-hoisting so I could override the component name  when it gets rendered.

components/dialog/stories/dialog.stories.js
jawinn: The `fullscreen` and `fullscreenTakeover` values are not for `size`. So it looks like Storybook was not representing either of the classes `spectrum-Dialog--fullscreen` and `spectrum-Dialog--fullscreenTakeover` yet, yikes! Which is why it's a good thing we're going through the YML docs.

Dialog needs another control for displaying these classes. I would call it `layout` and reference the similar control in Modal. Also make sure that the similar class is applied to the modal when they are set (reference the example HTML on the YML docs).

components/dialog/stories/dialog.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+### Scrolling
+
+A dialog without `.spectrum-Dialog--alert` can expand to a larger size to house larger contents.
+
+<Canvas of={DialogStories.WithScroll} />
+
+### Fullscreen
+
+A fullscreen dialog will automatically fill almost all of the available screen space.
```
jawinn, line 50:
```suggestion
A fullscreen dialog will automatically fill almost all of the available screen space. A margin is included around the outside of the dialog.
```
Some clarification I just discovered and documented while working on Modal.

components/dialog/stories/dialog.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+<Canvas of={DialogStories.Default} />
+
+### Large
+
+<Canvas of={DialogStories.Large} />
+
+### Dismissible
+
+A dialog that can be dismissed without taking an actions. Dismissible dialogs should never have buttons.
```
jawinn, line 26: Should all/most of the story descriptions / notes in this MDX file be added to the stories file descriptions instead?

components/dialog/stories/dialog.stories.js
jawinn: Setting a `size` arg here for the small, medium, and large makes sense. Since this differs from the standard size naming and options.

I'm assuming the naming of those hasn't been changed yet to avoid breaking things downstream. You could ahead and create a backlog Jira issue about renaming these and getting them in sync in the future. And you're right that things may be changing with S2.

_(Also see related discussion on "_So the fullscreen and fullscreenTakeover don't really behave appropriately in storybook_". Fullscreen and fullscreenTakeover are their own classes that are separate from sizing, so wouldn't be options here.)_

components/dialog/stories/dialog.stories.js
jawinn: Definitely leave the longer heading. That's there to test wrapping text and catch these sort of things. If this is causing an issue with the grid than we don't want to hide it. Oddly [I'm not seeing the text wrapping issue on the older version of what's built on prod currently](https://opensource.adobe.com/spectrum-css/preview/index.html?path=/story/components-dialog--default), so I think there may have been a regression recently? Something that went into main?

components/dialog/stories/dialog.stories.js
jawinn: It looks like this existing template code may be wrong. It's applying the platform scale instead of the size classes:
`[`${rootClass}--${scale}`]: true`. That should be removed now that you have `size`. The platform scale changes things (variables) from the root via `spectrum--medium` and `spectrum--large`.

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +47,107 @@ export default {
...
+		story: {
+			height: "auto",
+		}
+	},
+};
+
+const AllDividerSizes = (args) => html`
+	<div>
+		<div style="display: flex; flex-direction: column; padding: 16px;">
+			<h3 style="margin: 0;">Small</h4>
```
jawinn, line 66: Could you set the color to white on the headings when Static white is set, like on the docs page? I noticed they were a little hard to read on top of the background.

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +47,107 @@ export default {
...
+Small.parameters = {
+	chromatic: { disableSnapshot: true },
+	docs: {
+		story: {
+			height: "auto",
+		}
+	},
+};
+
+/**
```
jawinn, line 121: Small, medium, and large could be displayed within the same story, like they are for static black and static white. I don't think we need to separate them out.

jawinn (changes requested):
Nice work so far on this. Dial, divider, and drop indicator are looking good as compared to the docs site; just a few suggestions on those.

Dialog was a beast and it looks like you uncovered some things that were missing or wrong in Dialog's Storybook.

components/dial/stories/dial.stories.js
```diff
@@ -62,25 +63,71 @@ export default {
...
 Disabled.args = {
 	isDisabled: true,
 };
+Disabled.tags = ["!autodocs"];
+
+/*
+ * Stories for the MDX "docs" only.
+ * "storyName" refers to the display name/heading for a component
+ */
+export const DialGroup = (args) => html `
```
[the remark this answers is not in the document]
marissahuysentruyt: Sounds good. I cleaned up that file here: https://github.com/adobe/spectrum-css/pull/2833/commits/17261b54d92c54814a1d040af70e5956ce93c8ec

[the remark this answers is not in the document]
marissahuysentruyt: It also takes care of the `isFocusVisible` stuff. Thanks!!

components/dialog/stories/dialog.stories.js
marissahuysentruyt: `is-open` is getting applied to `.spectrum-Modal` instead. It sort of seems like that was outdated info.

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Was I getting the scale (--medium and --large) confused with the dialog classes (which were also --medium and --large)? I got it now though. thank you.

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Is this sort of what you were thinking? I'm adding the `--fullscreen` and `--fullscreenTakeover` to `.spectrum-Modal`, and I think it's working as expected (I don't really have anything to go off of except for comments in the CSS that describe what it should do/look).

https://github.com/adobe/spectrum-css/assets/69602589/022c9d64-af0e-4fb1-8a4c-fa0e382edb6c

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Made a note to make a card.

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Ok that makes sense- I didn't really think about the wrapping being used purposely like that. Reverted back to the longer heading and removed the buttons for now. I made another note to look into the buttons.

https://github.com/adobe/spectrum-css/pull/2833/commits/96464ed463b8c34341f7614d82e4eafdc07b2df6

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Here's the commit for that 👆: https://github.com/adobe/spectrum-css/pull/2833/commits/ca66e378cd233db53ec266b4b308a576d6970475

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Removing buttons in favor of that work being done in a different PR.

components/dialog/stories/dialog.stories.js
```diff
@@ -19,8 +21,27 @@ export default {
...
+			name: "Size",
+			type: { name: "string", required: true },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
+			},
+			options: ["small", "medium", "large", "fullscreen", "fullscreenTakeover"],
+			control: "select",
+		},
 		isDismissable: {
```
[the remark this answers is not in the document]
marissahuysentruyt: 🎉 https://github.com/adobe/spectrum-css/pull/2833/commits/2cb35f7a453f8622983957b522207d984300acba

components/dialog/stories/dialog.mdx
marissahuysentruyt: Brilliant! I fixed that with the rest of the mdx removal: https://github.com/adobe/spectrum-css/pull/2833/commits/2cb52cbd0a6373d5963c13d96b8ac856b5cda1fa

components/dialog/stories/dialog.mdx
marissahuysentruyt: I can get them over there. They're in a weird order (at least to me), but that might be ok for now. https://github.com/adobe/spectrum-css/pull/2833/commits/2cb52cbd0a6373d5963c13d96b8ac856b5cda1fa

components/dialog/stories/dialog.stories.js
```diff
@@ -46,12 +67,37 @@ export default {
...
+				category: "Component",
+			},
+			control: "boolean",
+		},
+		heroImageUrl: {
+			name: "Hero Image",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
```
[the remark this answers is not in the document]
marissahuysentruyt: Oh yeah that's a good idea. https://github.com/adobe/spectrum-css/pull/2833/commits/f6b5ca80499260c8c998f6df364144df201f84b9

components/divider/stories/divider.stories.js
marissahuysentruyt: I took care of the staticWhite color and then this sizing stuff all in one commit: https://github.com/adobe/spectrum-css/pull/2833/commits/b0028c9aebbf080502de7e3e78b84880083d2325

jawinn (changes requested):

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +111,159 @@ export default {
...
+	hasDivider: false,
+	content: [
+		() => ExampleContent
+	],
+};
+
+export const WithHero = Template.bind({});
+WithHero.tags = ["docs-only"];
+WithHero.parameters = {
+	chromatic: {disabledSnapshot: true},
```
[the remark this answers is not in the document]
marissahuysentruyt: This time for real 😊 https://github.com/adobe/spectrum-css/pull/2833/commits/2dc7680608feae6209acc96596ae40caa91bc576

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +48,105 @@ export default {
...
+			">
+				Large
+			</h3>
+			${Template({
+				...args,
+				size: "l"
+			})}
+		</div>	
+	</div>
+`;
```
rise-erpelding: I really like this as a way to display all the variants in one story! I think it's appropriate for divider because it's such a simple component but I wonder if it's a strategy we could use elsewhere 🤔

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +48,105 @@ export default {
...
+ * "storyName" refers to the display name/heading for a component
+ */
+const AllDividerSizes = (args) => html`
+	<div>
+		<div style="display: flex; flex-direction: column; padding: 16px;">
+			<h3 style="
+				margin: 0;
+				${styleMap({
+					"color": args.staticColor === "white" ? "rgb(255, 255, 255)": "inherit",
+				})}
```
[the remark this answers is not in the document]
marissahuysentruyt: https://github.com/adobe/spectrum-css/pull/2833/commits/53b2f6bb96e427a7953a72ae54deaca3ff93433f

components/dialog/stories/template.js
```diff
@@ -88,6 +106,7 @@ export const Template = ({
 				...globals,
 				isOpen,
 				content: Dialog,
+				customClasses: [`spectrum-Modal--${layout}`],
```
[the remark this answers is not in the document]
marissahuysentruyt: 👍: https://github.com/adobe/spectrum-css/pull/2833/commits/8b55970907fb1a12babe4bbaacb2ea926857f8fb

components/dialog/stories/template.js
```diff
@@ -13,42 +13,60 @@ import "../index.css";
...
 
 	const Dialog = html`
 		<div
 			class=${classMap({
 				[rootClass]: true,
-				[`${rootClass}--${scale}`]: true,
-				[`${rootClass}--dismissable`]: isDismissable,
+				[`${rootClass}--${size}`]: typeof size !== "undefined", 
+				[`${rootClass}--dismissable`]: isDismissible,
+				[`${rootClass}--${layout}`]: layout,
```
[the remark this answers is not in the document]
marissahuysentruyt: Oh I see. sort of like `size`. https://github.com/adobe/spectrum-css/pull/2833/commits/8b55970907fb1a12babe4bbaacb2ea926857f8fb

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +120,150 @@ export default {
...
+	showModal: true,
+	hasHeroImage: true,
+	heroImageUrl: "example-card-portrait.png",
+	content: [
+		() => ExampleContent
+	],
+};
+
+/**
+ * A dialog without `.spectrum-Dialog--alert` can expand to a larger size to house larger contents.
```
rise-erpelding, line 224: I think all of the dialogs might now scroll if their contents are longer than their available space, and I didn't see `.spectrum-Dialog--alert` in the CSS anymore, I wonder if you could double-check if this comment needs updating?

components/dialog/stories/dialog.stories.js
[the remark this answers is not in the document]
marissahuysentruyt: Your's is much more like if you were to say this to someone. I like it. https://github.com/adobe/spectrum-css/pull/2833/commits/8b8b3fb1ed0504742921ec5f22d92c5bf84511b9

components/dialog/stories/template.js
```diff
@@ -13,42 +13,60 @@ import "../index.css";
...
+				[`${rootClass}--noDivider`]: !hasDivider,
 				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
 			})}
 			id=${ifDefined(id)}
 			role="dialog"
 			tabindex="-1"
 			aria-modal="true"
+			style="max-height: ${ifDefined(maxHeight)}px;"
```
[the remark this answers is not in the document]
marissahuysentruyt: That makes sense now, good catch. https://github.com/adobe/spectrum-css/pull/2833/commits/8e90a6efba6247683660b9337a6b864779e112b9

rise-erpelding (commented):
Hi! I really like what you've done here and the callouts I have are pretty small. I have a few other comments on Dialog as well that didn't fit in the code:

* Kind of a nitpick, but I wish these stories didn't show up out of order, especially with the size variants, but I can't figure out how to reorder the stories on the component level and I'm not entirely convinced it's possible. I'd thought it might be nice to rename the Small and Large variants to something like "Default - Small" and "Default - Large" so that they might show up next to the Default (medium) but that's not such a great idea, maybe something like "Dialog - Default" for medium, "Dialog - Small", and "Dialog - Large"? If there's an obvious reason not to do it, we could definitely do without it though.
* I did confirm that the Hero image works if I uploaded an image, I wish there was a default image in there already. Could we throw a default image in there as well as giving the option to upload an image, like the `example-card-portrait.png`? I think it's the same image in the Card component if that is helpful to reference.
* The fullscreen variants of Dialog had buttons that look like they're supposed to allow you to close the dialog, I'm wondering if you have thoughts on whether these should be preserved in the Storybook version? It would be nice for these variants in particular to be able to have some way to close it, either through a button group or a close button.
* This Docs Site usage note on Dialog is no longer needed, right? I didn't see anything wrapped in `.spectrum-Dialog-wrapper` and nothing looked blurry, but want to confirm that wasn't overlooked.
image

components/divider/stories/divider.stories.js
marissahuysentruyt: Yeah definitely! I have it in drop indicator, and then in my other storybook docs branch, checkbox is using something similar (coach indicator too, I think). I had one for dial, but we decided not to use it since dial already had a bunch of stories represented.

components/dialog/stories/dialog.stories.js
```diff
@@ -65,17 +120,163 @@ export default {
...
+	heading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+	showModal: true,
+	size: "large",
+	content: [
+		() => ExampleContent
+	],
+};
+
+/**
+ * A dialog that can be dismissed without taking an actions. Dismissible dialogs should never have buttons.
```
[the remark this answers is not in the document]
marissahuysentruyt: Good catch: https://github.com/adobe/spectrum-css/pull/2833/commits/e2e5bc9e832c200687a68bddbeaa0a19fa476277

components/dial/stories/dial.stories.js
```diff
@@ -52,23 +52,26 @@ export default {
...
 		rootClass: "spectrum-Dial",
 		size: "m",
-		isFocused: false,
+		isFocusVisible: false,
 		isDragged: false,
 		isDisabled: false,
 	},
 };
 
+// the !auto-docs tag removes the component from the auto-template
```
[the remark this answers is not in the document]
marissahuysentruyt: YES: https://github.com/adobe/spectrum-css/pull/2833/commits/cdc5be2dc8fc96af0cd9945f7f892ed4d46eafb4

components/dialog/stories/dialog.stories.js
marissahuysentruyt: @rise-erpelding I wasn't sure how to handle this comment. I assume that was the class name before alert dialog was its own component. Is that a question for design? (that all dialogs scroll if their content is longer than the available space?) I know I'm mixing the two branches, but scrolling isn't represented in the S2 designs, so it might be a good question anyways.

marissahuysentruyt:
@rise-erpelding thanks for these! 

> I can't figure out how to reorder the stories on the component level and I'm not entirely convinced it's possible.

I couldn't either, unless I add an MDX file. The MDX isn't really needed, except to reorganize them, since I could get the documentation onto the page without it. 

The "Dialog - Small" format is how they have it on the docs site now I think. I initially didn't use that format since a few of the other components weren't like that, but I agree that it sort of bothers me that the sizing variants are scattered on the page. Here's the change: https://github.com/adobe/spectrum-css/pull/2833/commits/5daca40a70768913cf65abcf79153c63c029f742

I didn't change the `storyName` for "Default" yet since it would also change the story name in the sidebar.

> I wish there was a default image in there already. Could we throw a default image in there as well as giving the option to upload an image

I meant to do that, great catch!! How's this? https://github.com/adobe/spectrum-css/pull/2833/commits/81cef32a0b0e3f6f81cb140363b972097e04f89a

> * The fullscreen variants of Dialog had buttons that look like they're supposed to allow you to close the dialog, I'm wondering if you have thoughts on whether these should be preserved in the Storybook version? It would be nice for these variants in particular to be able to have some way to close it, either through a button group or a close button.

The fullscreen variants are a little funky. They're pretty broken on the docs site now. I had adding buttons earlier, but after reviewing it with Josh ([here](https://github.com/adobe/spectrum-css/pull/2833#discussion_r1635274470), [here](https://github.com/adobe/spectrum-css/pull/2833#discussion_r1635255615) and [here](https://github.com/adobe/spectrum-css/pull/2833#discussion_r1635270121)), we decided separating out any button work might be best, since we both think there's more going on.

> This Docs Site usage note on Dialog is no longer needed, right? I didn't see anything wrapped in `.spectrum-Dialog-wrapper` and nothing looked blurry, but want to confirm that wasn't overlooked.

I purposely didn't include it for the same reasons. The `-wrapper` I think may have gotten moved onto the `spectrum-Modal` component instead. That's also where `.is-open` is being applied.

components/dialog/stories/dialog.stories.js
rise-erpelding: I think checking on the scrolling behavior whenever we get to it in S2 would be ok, I think it's probably out of scope for this work. And it seems like we might want it to scroll: 

> [The content inside the body area can scroll when necessary](https://spectrum-contributions.corp.adobe.com/page/standard-dialog-beta/#Body-area).

Quickly scanning the repo history it seems like this line in the documentation has been around for a [number of years](https://github.com/adobe/spectrum-css/blob/%40adobe/spectrum-css%402.14.0-alpha.10/components/dialog/docs.yml#L256) and it looks like that `.spectrum-Dialog--alert` class was used at one point but isn't anymore. I would say we could probably drop this comment/story description as it no longer seems needed, or if you still wanted to keep some text there, you could write "The content inside the body area can scroll when necessary."

marissahuysentruyt: Ok, I added scrolling behavior to my list of questions for design for dialog S2. 👍

I used that sentence you suggested. https://github.com/adobe/spectrum-css/pull/2833/commits/e5abf1453cf4e803d67b88e4f0952338bbfd923b

components/dropindicator/stories/dropindicator.stories.js
```diff
@@ -36,5 +38,33 @@ export default {
...
+		${Template({
+			...args,
+			direction: "horizontal",
+		})}
+	</div>
+`;
+
 export const Default = Template.bind({});
 Default.args = {};
+// the !auto-docs tag removes the component from the auto-template
```
rise-erpelding, line 56:
```suggestion
// the !autodocs tag removes the component from the auto-template
```

rise-erpelding (approved):
One super tiny typo in the comments, but otherwise I think this is looking good! Nice work!

components/dial/stories/dial.stories.js
```diff
@@ -52,23 +52,26 @@ export default {
 			},
 			control: "boolean",
 		},
+		reducedMotion: { table: { disable: true } },
```
castastrophe, line 55:
```suggestion
```

This is no longer used because reducedMotion is a global setting now.

components/dialog/stories/dialog.stories.js
```diff
@@ -66,17 +120,154 @@ export default {
...
-		}),
+		() => ExampleContent
+	],
+};
+
+/*
+ * Stories for "docs only."
+ */
+export const DefaultSmall = Template.bind({});
+DefaultSmall.tags = ["docs-only"];
```
castastrophe, line 147:
```suggestion
DefaultSmall.tags = ["autodocs", "!dev"];
```

Can you update these tags with our new tagging knowledge when you get a chance?

castastrophe: We should also exclude these from chromatic snapshots and create a combined snapshot on Default.

components/dialog/stories/template.js
```diff
@@ -6,49 +6,68 @@ import { Template as Underlay } from "@spectrum-css/underlay/stories/template.js
...
+				${when(hasDivider, () =>
+				Divider({
 						horizontal: true,
 						customClasses: [`${rootClass}-divider`],
 						...globals,
-					}, context),
-				])}
-				<section class="${rootClass}-content">${content.map((c) => (typeof c === "function" ? c({}, context) : c))}</section>
-				${when(isDismissable, () =>
+					}),
```
castastrophe, line 65:
```suggestion
					}, context),
```

Missing the context passthrough in this update.

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +48,105 @@ export default {
...
+			height: "auto",
+		}
+	},
+};
+
+/*
+ * Stories for the MDX "Docs" only.
+ * "storyName" refers to the display name/heading for a component
+ */
+const AllDividerSizes = (args) => html`
```
castastrophe, line 68: I'd like to move the Variant templates into the template.js as an export instead of keeping markup in the stories.js files.

components/dial/stories/dial.stories.js
castastrophe: (applies to other stories too)

components/divider/stories/divider.stories.js
```diff
@@ -45,5 +48,105 @@ export default {
...
+			${Template({
+				...args,
+				size: "s"
+			})}
+		</div>
+		<div style="display: flex; flex-direction: column; padding: 16px;">
+			<h3 style="
+				${styleMap({
+					"margin": 0,
+					"color": args.staticColor === "white" ? "rgb(255, 255, 255)": "inherit",
```
castastrophe, line 88: This is not necessary if you push them through Typography instead

castastrophe (changes requested):
Love the additions to the dialog story! Lots of good stuff but here but needs a little moving around and updated tagging to get it pulling through as expected.

components/dial/stories/dial.stories.js
marissahuysentruyt: Ah yes, thank you!

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Corrected tags can be found here: https://github.com/adobe/spectrum-css/pull/2833/commits/8305a2d4a376316b28340ecc6756684715bd4451
^ that should take care of the tagging in all 4 components

Excluded Chromatic snapshots: 
- dial: https://github.com/adobe/spectrum-css/pull/2833/commits/d547a48e60ae1537daaa42d6547719f91a5dfdfa
- dialog: https://github.com/adobe/spectrum-css/pull/2833/commits/4cf1b8aed2b20f8eb9463958c70f06d866627bb2

components/dialog/stories/template.js
marissahuysentruyt: Good catch! That should be fixed here: https://github.com/adobe/spectrum-css/pull/2833/commits/95dfddcfd56af9959011fea2b855702e96ccadbc

components/dialog/stories/dialog.stories.js
marissahuysentruyt: > We should also exclude these from chromatic snapshots and create a combined snapshot on Default.

@castastrophe I created the chromatic group for dialog in the s2 branch ([PR](https://github.com/adobe/spectrum-css/pull/2860) or [preview](https://pr-2860--spectrum-css.netlify.app/preview/?path=/docs/components-dialog--docs&globals=testingPreview:!true)). We had decided that adding chromatic coverage in these docs migration branches was making the PRs quite large, so removed that criteria from [the ticket](https://jira.corp.adobe.com/browse/CSS-759). 

I can either take this feedback and apply it over on that S2 branch (with regards to making sure the chromatic group is in the template.js file), or do something like we did for menu, where we had a separate PR completely for the chromatic coverage (so that the chromatic baselines are in `main` sooner. Do you have a preference?

components/dialog/stories/template.js
```diff
@@ -75,11 +95,29 @@ export const DialogGroup = Variants({
...
+		{
+			testHeading: "Fullscreen",
+			layout: "fullscreen",
+			showModal: false,
+		},
+		{
+			testHeading: "Takeover",
+			layout: "fullscreenTakeover",
+			showModal: false,
+		}
```
jawinn, line 121: I noticed that the "Sizing" template sizes are not rendering in the Chromatic only template. This looks like it's because `showModal: false` needs to also be set where `Sizes()` is called in the `Variants()` template (in utilities.js).
Screenshot 2024-07-18 at 5 03 10 PM

I'm not sure about the best way to pass that through. @castastrophe Could we add a `sizingArgs = {}` parameter to `Variants()`, that's spread after args in Sizes(), like:
`${Sizes({ Template, direction: sizeDirection, ...args, ...sizingArgs }, context)}`

Or is there another way to pass this through for just Chromatic? The only other way I could think of feels a little less clean:
```
Default.decorators = [
	// Make sure Sizes template used within Variants template changes this arg.
	(story, context) => window.isChromatic() ? story({ ...{ args: { showModal: false }}, ...context }) : story(context),
];
```

rise-erpelding (commented):
I did, as expected, see this error running the branch locally on Chrome, but confirmed that it doesn't happen in Firefox/Safari or in the preview link
image.

I'd also had comments to point out that Dial had extra stories and that Dialog is getting the small and medium size classes, but it looks like Josh mentioned those already!

The only other thing is that Divider had a static black and static white variant and I wasn't sure if there was a reason we were leaving those out.

marissahuysentruyt:
@rise-erpelding -
> The only other thing is that Divider had a static black and static white variant and I wasn't sure if there was a reason we were leaving those out.

I sort of approached the Divider static white/black stories like the clear button, where its "overbackground" story moved  to vrt-only, even though I guess technically it's a static white story. @castastrophe do you have a preference on where the static white/black stories go? Should they all be moved to vrt-only?

components/dial/stories/dial.stories.js
```diff
@@ -63,6 +73,31 @@ export default {
...
+};
+
+export const Disabled = Template.bind();
+Disabled.args = {
+	isDisabled: true,
+};
+Disabled.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
```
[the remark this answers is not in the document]
marissahuysentruyt: Oh thank you. https://github.com/adobe/spectrum-css/pull/2833/commits/dc8065c3297e9e3cc760a4c31e327c167cfa1552

components/dialog/stories/template.js
```diff
@@ -28,26 +35,38 @@ export const Template = ({
 			class=${classMap({
 				[rootClass]: true,
 				[`${rootClass}--${scale}`]: true,
```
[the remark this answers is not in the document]
marissahuysentruyt: There was a [change with `scale` in the modes PR](https://github.com/adobe/spectrum-css/pull/2844/files#diff-dd11973c8d7472852e8c9bb250b44fe9ecdc299b0abbd0c9a2fae8b303840662R23), so I left it back in (mainly out of concern that I would be overriding something). But yes, small doesn't work with it in.

Removed now. https://github.com/adobe/spectrum-css/pull/2833/commits/575b40b8d30480402432796aafa1be6f71957207

components/dialog/stories/template.js
marissahuysentruyt: How's this looking? Thanks for figuring it out!

https://github.com/adobe/spectrum-css/pull/2833/commits/4b535e9ff35ee5e5972c27d5e968f615016e8b0c

jawinn (approved):
This looks good to me! Thanks for rebasing and updating those last things.

One last suggestion I have is to show the horizontal drop indicator on the docs page as well as the vertical. Previously they were both shown on the docs site.

marissahuysentruyt:
> One last suggestion I have is to show the horizontal drop indicator on the docs page as well as the vertical. Previously they were both shown on the docs site.

I just noticed that too- I had it called `DropIndicatorGroup` before, and when I pulled over the variants template, which was also called `DropIndicatorGroup`, I never renamed that story/template.

components/dial/stories/template.js
```diff
@@ -88,157 +87,30 @@ export const Template = ({
...
+		},
+	],
+	stateData: [
+		{
+			heading: "Disabled",
+			isDisabled: true,
+		},
+		{
+			heading: "Focused",
+			isFocused: true,
```
marissahuysentruyt, line 109: Ack- this should be `isFocusVisible` I believe

components/dial/stories/dial.stories.js
```diff
@@ -28,7 +28,15 @@ export default {
...
-		isFocused,
+		isFocusVisible: {
+			name: "Focused",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "State",
+			},
+			control: "boolean",
+		},
```
castastrophe, line 39: What would you think of doing this to extend the isFocused definition but still allow the FocusVisible variable name:

```suggestion
		isFocusVisible: isFocused,
```

components/dialog/stories/dialog.stories.js
```diff
@@ -3,47 +3,71 @@ import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isOpen } from "@spectrum-css/preview/types";
 import { Template as Typography } from "@spectrum-css/typography/stories/template.js";
 import { version } from "../package.json";
+// import { DialogGroup } from "./dialog.test.js";
```
castastrophe, line 6: You can remove this comment if you don't need this anymore.

components/dialog/stories/dialog.test.js
```diff
@@ -2,18 +2,49 @@ import { Variants } from "@spectrum-css/preview/decorators";
...
+			testHeading: "With hero/cover image",
+			hasHeroImage: true,
+			wrapperStyles: {
+				"background-color": "var(--spectrum-gray-50)"
+			},
+		},
+		{
+			testHeading: "No divider",
+			isDismissible: true,
+			hasDivider: false,
```
marissahuysentruyt, line 47: Should I add cases for: 
- With hero, isDismissible: true
- with hero, no divider, isDismissible: true
- with hero, no divider, isDismissible: false
- No divider, isDismissible: false

What if I refactor "dismissible" as a state instead? That's not really a "state" though, is it? It feels like this view is really long, but I suppose we do have a lot of dialog variants.

Screenshot 2024-08-20 at 4 42 41 PM

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Yes, thank you! There are a few things in Dialog I kept immediately after the rebase just in case I needed them, but it looks like I don't! Removed: https://github.com/adobe/spectrum-css/pull/2833/commits/e9310f3a16d6a47215ef76f57b11e545c3da9c3b

components/dial/stories/dial.stories.js
marissahuysentruyt: 👏 ooooo that's fantastic! Yes! https://github.com/adobe/spectrum-css/pull/2833/commits/92eb2de7ff23cc4876803b6598561a2bd8b2d1d2

components/divider/stories/template.js
```diff
@@ -25,8 +25,10 @@ export const Template = ({
 				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
 			})}
 			style=${styleMap({
+				"min-block-size": vertical == true ? "20px" : undefined,
 				"block-size": vertical ? "auto" : undefined,
-				"inline-size": !vertical ? "100%" : undefined,
+				"min-inline-size": !vertical ? "200px" : "var(--spectrum-divider-thickness)",
+				"inline-size": !vertical ? "auto" : undefined,
```
marissahuysentruyt, line 31: I was really struggling with coordinating this styleMap and the customStyles that get passed after these. As it was, I noticed if I toggled the default story to vertical, there was nothing there. I think it was because `block-size: auto` wasn't resolving to anything (because there is no content in the container).

Screenshot 2024-08-21 at 10 50 43 AM
Screenshot 2024-08-21 at 10 50 34 AM

Then I noticed if I was looking at the chromatic preview, and toggled the vertical dividier, the sizing group wouldn't render at all and non-hr test case would get weird too, because there was a min-block-size, as well as the min-inline-size: 

Screenshot 2024-08-21 at 10 50 22 AM

Divider now functions more like what's on prod (I don't think the changes on main have been deployed yet where the vertical toggle doesn't display). However, I'm worried I introduced extra work where we'll have to pass more customStyles to the divider used in other components (like the changes in alert dialog and menu now). Can I refactor this at all?

components/dialog/stories/dialog.test.js
castastrophe: Though the `isDismissable` variable name is structured the same as our States generally are, it adds UI to the component and isn't a user-driven change but a developer choice about how the UI is structured. I think that makes it a "Variant".

components/dialog/stories/dialog.stories.js
```diff
@@ -4,46 +4,69 @@ import { isOpen } from "@spectrum-css/preview/types";
...
  * A dialog displays important information that users need to acknowledge. They appear over the interface and block further interactions.
+ * 
+ * The alert variants that were previously a part of Dialog were moved to their own component, [Alert Dialog](/docs/components-alert-dialog--docs).
  */
 export default {
 	title: "Dialog",
 	component: "Dialog",
 	argTypes: {
-		heading: { table: { disable: true } },
+		heading: {
```
castastrophe, line 18: Thanks for making this interactive!

components/dialog/stories/dialog.stories.js
```diff
@@ -61,15 +84,37 @@ export default {
...
+		hasHeroImage: {
+			name: "Has hero image",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Content",
+			},
+			control: "boolean",
+			if: { arg: "layout", eq: "default" },
+		},
```
castastrophe, line 96: Instead of adding a boolean to the UI, do we just want to assume that if `heroImageUrl` has content in the template that hasHeroImage is true?

components/dialog/stories/dialog.test.js
```diff
@@ -2,18 +2,49 @@ import { Variants } from "@spectrum-css/preview/decorators";
...
+				...args,
+				showModal: false,
+
+				// TODO: remove this heading arg when the "Sizing" heading is no longer overwritten by it.
+				heading: "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
+				customStyles: {
+					"background-color": "var(--spectrum-gray-100)",
+				},
+			}, context)
+			: Template(args, context)
```
castastrophe, line 22: It helps React's hot module reloader if we only call a template once and toggle the inputs based on settings:

```suggestion
		Template({
			...args,
			showModal: window.isChromatic() ? false : args.showModal,

			// TODO: remove this heading arg when the "Sizing" heading is no longer overwritten by it.
			heading: window.isChromatic() ? "Lorem ipsum dolor sit amet, consectetur adipiscing elit" : args.heading,
			customStyles: {
			    ...(args.customStyles ?? {}),
				"background-color": window.isChromatic() ? "var(--spectrum-gray-100)" : undefined,
			},
		}, context)
```

components/dialog/stories/dialog.test.js
castastrophe: Definitely feel free to expand the test coverage those, those all seem like great additions.

components/dialog/stories/template.js
```diff
@@ -24,65 +24,86 @@ export const Template = ({
...
 
 	const Dialog = html`
 		<div
 			class=${classMap({
 				[rootClass]: true,
-				[`${rootClass}--${scale}`]: true,
-				[`${rootClass}--dismissable`]: isDismissable && !isFullScreen && !isFullScreenTakeover,
-				[`${rootClass}--fullscreen`]: isFullScreen,
-				[`${rootClass}--fullscreenTakeover`]: isFullScreenTakeover,
+				[`${rootClass}--dismissable`]: isDismissible && (layout !== "fullscreen" || layout !== "fullscreenTakeover"),
```
castastrophe, line 40: Both ways are perfect, just wanted to share a nice shorthand I like to use sometimes leveraging the array and the `every` or `some` functions to make updates easier (just add new values to the array to update the logic).

```suggestion
				[`${rootClass}--dismissable`]: isDismissible && ["fullscreen", "fullscreenTakeover"].every(l => layout !== l),
```

components/dialog/stories/template.js
```diff
@@ -24,65 +24,86 @@ export const Template = ({
...
+				[`${rootClass}--${layout}`]: typeof layout !== "undefined" && layout !== "default",
```
castastrophe, line 41: Maybe we just remove default and the absence of layout sets the class since default doesn't use a class.

```suggestion
				[`${rootClass}--${layout}`]: typeof layout !== "undefined",
```

components/dialog/stories/dialog.stories.js
```diff
@@ -4,46 +4,69 @@ import { isOpen } from "@spectrum-css/preview/types";
...
+			control: "select",
+		},
+		layout: {
+			name: "Layout",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Component",
+			},
+			options: ["default", "fullscreen", "fullscreenTakeover"],
```
castastrophe, line 55:
```suggestion
			options: ["fullscreen", "fullscreenTakeover"],
```

Propose removing the default option from the list and just make the absence of a setting drive the class.

components/dialog/stories/template.js
```diff
@@ -13,9 +14,8 @@ import "../index.css";
 
 export const Template = ({
 	rootClass = "spectrum-Dialog",
-	isDismissable = true,
-	isFullScreen = false,
-	isFullScreenTakeover = false,
+	isDismissible = false,
```
castastrophe, line 17:
```suggestion
	isDismissible = false,
```
^ Was this a change in the design documentation? Is being dismissible no longer the default for a dialog?

components/dialog/stories/template.js
```diff
@@ -24,65 +24,86 @@ export const Template = ({
...
 			id=${ifDefined(id)}
 			role="dialog"
 			tabindex="-1"
 			aria-modal="true"
+			style=${ifDefined(styleMap(customStyles))}
 		>
 			<div class="${rootClass}-grid">
-				${when(heading, () => [
-					html`<h1 class="${rootClass}-heading">${heading}</h1>`,
+				${when(hasHeroImage, () =>
```
castastrophe, line 53:
```suggestion
				${when(typeof heroImageUrl !== "undefined", () =>
```

components/dialog/stories/template.js
```diff
@@ -24,65 +24,86 @@ export const Template = ({
...
+					html`
+						<div 
+							class="spectrum-Dialog-hero"
+							style="background-image:url(${heroImageUrl ?  heroImageUrl : "example-card-portrait.png"})">
```
castastrophe, line 57:
```suggestion
							style="background-image:url(${heroImageUrl})">
```

Can simplify this if we use the heroImageUrl's presence as the boolean

components/dialog/stories/template.js
```diff
@@ -94,13 +115,29 @@ export const Template = ({
...
+				treatment: "outline",
+				customClasses: [],
+				customStyles: {
+					position: "absolute",
+					insetInlineStart: "50%",
+					insetBlockStart: "50%",
+					transform: "translate(-50%, -50%)",
+				},
+				onclick: toggleOpen,
+			}, context)}
```
castastrophe, line 131:
```suggestion
```

I'm not sure we need this in this template - the `isOpen` boolean will be able to toggle the hide/show of the Dialog without adding a button to the Storybook UI.

components/divider/stories/divider.stories.js
```diff
@@ -44,18 +46,83 @@ export default {
...
+}, context);
+Sizing.tags = ["!dev"];
+Sizing.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * When a vertical divider is used inside of a flex container, use `align-self: stretch; block-size: auto;` on the divider.
+ */
+export const VerticalGroup = (args, context) => Sizes({
```
castastrophe, line 76:
```suggestion
export const VerticalSizing = (args, context) => Sizes({
```

^ Propose aligning the story's name with the story above since they're both pushing through the same Sizes function.

components/divider/stories/divider.stories.js
```diff
@@ -44,18 +46,83 @@ export default {
...
+VerticalGroup.storyName = "Vertical";
+VerticalGroup.tags = ["!dev"];
+VerticalGroup.args = {
+	vertical: true,
+};
+VerticalGroup.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+export const StaticWhiteGroup = (args, context) => Sizes({
```
castastrophe, line 91: There's no size-based changes that happen when the static white variant is used so it might not make sense for us to push this test through Sizes and perhaps use the same template as Default so that we're capturing any future states.

components/tray/stories/tray.test.js
```diff
@@ -11,7 +11,7 @@ export const TrayGroup = Variants({
 					...passthroughs,
 					heading: "You have new messages waiting in your inbox",
 					content: ["You have 5 new messages! This notification is extra long so it wraps to the next line"],
-					isDismissable: true,
+					isDismissible: true,
```
castastrophe, line 14: Nice spelling catch!

castastrophe (changes requested):
A couple minor questions about which templates we're using for rendering.

components/divider/stories/template.js
jawinn: Since `.spectrum-Divider` in the CSS has:
- horizontal: `inline-size: 100%;` and a `block-size` of the divider height
- vertical: `block-size: 100%;` and an `inline-size` of the divider height

We really only should need something like this, to make sure there is a width for horizontal and a height for vertical:
```
"min-block-size": vertical ? "20px" : undefined,
"min-inline-size": !vertical ? "200px" : undefined,
```
and no properties changing `inline-size` or `block-size`.

I don't think we should be using any of these `styleMap` properties (including `align-self`) in the default Template though (including as it exists in prod). Since any other components using divider are not testing with the shipped component CSS, which means VRTs can be showing the wrong behavior. The `min-block-size` and
`min-inline-size` are really only needed here because divider's stories need to be within a container with a width or height in order to be fully visible.

Perhaps we could remove all properties from this template `styleMap`, and then either add an boolean arg (`useMinDimensions`, default of false) to `Template` that determines whether the min styles are added. Or, create a separate template (e.g. `TemplateTest`) that uses `Template()` and sets the min styles (or creates a wrapper div around divider with the min-styles which might be more future-proof). And then stories would bind to that `TemplateTest`, while other components would use the regular `Template` that isn't injecting any CSS that doesn't exist in the CSS file. Stylistically, I'm not sure which is preferable. Just the boolean arg might be fine. Something like:

```
style=${styleMap({
	"min-block-size": useMinDimensions && vertical ? "20px" : undefined,
	"min-inline-size": useMinDimensions && !vertical ? "200px" : undefined,
	...customStyles,
})}
```

components/dialog/stories/template.js
marissahuysentruyt: @castastrophe Being totally honest, I don't know what the default is. I haven't found that super clear, but if that's the default, I'll change it!! Where would I find that info?

I do know, that after we [reimplemented the buttons on the docs site](https://github.com/adobe/spectrum-css/pull/2861), where those dialogs didn't have anything, (no close button _or_ button group), I had added them to most of dialogs (obviously not the previously dismissible ones). I was probably referencing that page with my own changes 🤦‍♀️ Let me know if I need to change this back- I'm absolutely open to it.

cdransf (approved):

components/dialog/stories/template.js
marissahuysentruyt: Reverted ✔️

components/divider/stories/divider.stories.js
marissahuysentruyt: Sure thing: https://github.com/adobe/spectrum-css/pull/2833/commits/0be603d2bb21b21308a838698a6380c92560581b

components/divider/stories/divider.stories.js
marissahuysentruyt: Oh ok, sure. The [docs site](https://opensource.adobe.com/spectrum-css/divider.html) displayed all three sizes, so I was trying to replicate that. But making sure these static colors represent everything on the testing grid is more valuable, for sure.

components/dialog/stories/template.js
marissahuysentruyt: yesssss

components/dialog/stories/dialog.test.js
marissahuysentruyt: ooooo awesome. Thank you!

components/dialog/stories/dialog.test.js
marissahuysentruyt: Here they are! https://github.com/adobe/spectrum-css/pull/2833/commits/38ab7ec73dd3034f62e11ba4bc6cbb6a4af5f550

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Based on our conversation in the breadcrumbs PR ([2793](https://github.com/adobe/spectrum-css/pull/2793#discussion_r1726039443)), I was going to keep this as is for now. It's a select drop down currently, and I added the `defaultValue` to the arg table. Does that all sound ok?

https://github.com/adobe/spectrum-css/pull/2833/commits/bb342743ca3ce29ad8e859859157e832c8356942

components/dialog/stories/dialog.stories.js
marissahuysentruyt: Yes! https://github.com/adobe/spectrum-css/pull/2833/commits/4c1cec586a4721b37931dacbfa78fa105e91d65b

components/dialog/stories/template.js
marissahuysentruyt: https://github.com/adobe/spectrum-css/pull/2833/commits/340ef9b9c38362f75fc46a01f92432f98d9523b8

components/dialog/stories/template.js
marissahuysentruyt: https://github.com/adobe/spectrum-css/pull/2833/commits/340ef9b9c38362f75fc46a01f92432f98d9523b8

components/divider/stories/divider.stories.js
marissahuysentruyt: Here's the commit with this update: https://github.com/adobe/spectrum-css/pull/2833/commits/b84e78e7e5533ac6b9f4d67b31583bc9c5718cc1

components/divider/stories/template.js
marissahuysentruyt: @jawinn THANK YOU! I just could not wrap my head around this! I went with the boolean approach, with a new `minDimensionValues` arg (but it's hidden from the args table) so let me know what you think! This is exactly what I was trying to do. 😮‍💨 😮‍💨 😮‍💨

https://github.com/adobe/spectrum-css/pull/2833/commits/b84e78e7e5533ac6b9f4d67b31583bc9c5718cc1

michael-scott-thank-you

aramos-adobe (approved):
Migration is there and looks great in the preview link 🎉 Everything looks on point! 

Just a few comments:

- Dialog Docs renders perfectly in the preview link, but has an error in localhost on my end. The preview.js templates doesn't render.
- The dragging on Dial feels different on the Storybook than the OG docs version


https://github.com/user-attachments/assets/9ef2c10d-da74-45cc-bf36-661a4fbccb92


https://github.com/user-attachments/assets/6ed9155b-631c-4aae-8448-0e1ed18e8bbf




https://github.com/user-attachments/assets/b3723d18-b29d-4c92-9c4d-195e4f0ca6ce

Screenshot 2024-08-22 at 10 34 27 PM

castastrophe (approved):

pfulton (approved):

PR #2840, opened 2024-06-12
feat(storybook): move args to global scope

castastrophe:
## Description

This PR moves a few of our local "arg" properties to the global scope:

- color
- scale
- reducedMotion
- express (migrated to "context" so it can be converted to a selection instead of a boolean)

This update supports backwards compatibility so that if an arg value exists at the component-level, that is pushed up to the global scope automatically.  This allows us to migrate without having to update our existing stories.

### Validation steps

Open [storybook](https://64762974a45b8bc5ca1705a2-ebiewpngbg.chromatic.com):

- [x] Navigate to [Accordion default](https://64762974a45b8bc5ca1705a2-ebiewpngbg.chromatic.com/?path=/story/components-accordion--default), toggle the color setting to dark and darkest. Expect the background color and theming to update quickly.
- [x] Navigate to [Dial express](https://64762974a45b8bc5ca1705a2-ebiewpngbg.chromatic.com/?path=/story/components-dial--default-express). Expect to see the top toolbar context button to automatically update to "Express". Toggle the context to Spectrum 1 and expect styles to update to S1 and remove Express styles.
- [x] Navigate to [Color area default](https://64762974a45b8bc5ca1705a2-ebiewpngbg.chromatic.com/?path=/story/components-color-area--default) and update the toolbar button "Left to right" to "Right to left". Expect to see the color area shift to the right of the screen.

### Regression testing

Validate:

## To-do list


pfulton (approved):

.storybook/decorators/context.js
```diff
@@ -8,81 +8,117 @@ import { html } from "lit";
...
+
+		if (args.scale && args.scale !== scale) {
+			updateGlobals({ scale: args.scale });
+			// prevents unnecessary re-renders
+			updateArgs({ scale: undefined });
+		}
+
+		window.__color = color;
+		window.__context = context;
+		window.__scale = scale;
```
jawinn, line 67: Could you explain why these globals are being stored in `window` now, and some components (e.g. combobox, datepicker, slider) that were previously `useGlobals()` are now accessing the window object?

jawinn (changes requested):
One issue I am seeing is that the Button "Docs" page doesn't load and then freezes up--I get the "page has become unresponsive" popup in Chrome.

A suggestion for the reduced motion toolbar item; the play button makes it look like it would do something like play an interaction. Could we use a different icon and/or label that differently? Perhaps the "Orbit" Workflow icon would work? Perhaps the "Invert Adj" could work for the color context option too.

I'm also wondering if a prefix of "Scale:" and "Colors:" (and "System:"?) might be useful to understand those toolbar items before hovering.

Would calling "Express" "Spectrum 1 - Express" make it more clear that Express uses the older S1 design system, when "Spectrum 2" is an option?

PR #2854, opened 2024-06-20
docs(modal,pagination,pickerbutton): site docs to storybook

jawinn:
## Description

This continues the migration work to move the documentation from the deprecated docs site to Storybook, covering the components Modal, Pagination, and Picker button. Variants shown on the docs site for these components have been added to Storybook as new stories if they did not already exist. No custom MDX files were needed for these components.

### Component-specific changes and notes:

- Modal
  - Updates docs text and sets Docs story to `inline: false` to avoid it being cut off horizontally
- Pagination
  - Makes sure pagination variants are included in the Storybook docs. Note: the "button" style variant is represented in VRTs but not included for users as it is using the deprecated Split button component. I wasn't sure about the best way to document this as deprecated. That variant has no unique CSS and is basically just a Split button with two classes added.
- Picker button
  - Creates docs only stories to cover what is on the docs site
  - Moves old migration notes to the changelog
  - Note on the "Custom icon only" variant from the docs site: this appears to be using a workflow icon but is using the class `spectrum-PickerButton--uiicononly` that gets rid of the padding. It's unclear whether this class should be named something else that applies to both icon sets, if `spectrum-PickerButton--icononly` should use the same CSS (this currently is applied in Storybook if a workflow icon is used), if the component really should only allow UI icons, or if the wrong class was used in the example. I've included both a UI icon and a workflow icon on the Docs story in order to show that these render differently (without `spectrum-PickerButton--uiicononly` there is extra padding on the left and right).

### Validation steps

@marissahuysentruyt ✅✅✅
- [Modal](https://pr-2854--spectrum-css.netlify.app/preview/?path=/docs/components-modal--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/modal.html)
- [Pagination](https://pr-2854--spectrum-css.netlify.app/preview/?path=/docs/components-pagination--docs) - includes all necessary information compared to the [docs site 1](https://opensource.adobe.com/spectrum-css/pagination-explicit.html) [docs site 2](https://opensource.adobe.com/spectrum-css/pagination-listing.html)
- [Picker button](https://pr-2854--spectrum-css.netlify.app/preview/?path=/docs/components-picker-button--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/pickerbutton.html)

### Regression testing

Expected VRT changes: label added to quiet Picker button.

Validate:

## To-do list

### DIFF
--- a/components/modal/stories/template.js
+++ b/components/modal/stories/template.js
@@ -6,6 +6,9 @@ import { when } from "lit/directives/when.js";
 
 import "../index.css";
 
+/**
+ * Just the modal markup.
+ */
 const Modal = ({
 	rootClass = "spectrum-Modal",
 	customClasses = [],
@@ -29,6 +32,9 @@ const Modal = ({
 	`;
 };
 
+/**
+ * The modal, optionally wrapped with .spectrum-Modal-wrapper.
+ */
 export const Template = ({
 	rootClass = "spectrum-Modal",
 	skipWrapper = false,
--- a/components/pagination/stories/pagination.stories.js
+++ b/components/pagination/stories/pagination.stories.js
@@ -2,9 +2,10 @@ import { default as ActionButton } from "@spectrum-css/actionbutton/stories/acti
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { version } from "../package.json";
 import { PaginationGroup } from "./pagination.test.js";
+import { Template } from "./template";
 
 /**
- * The pagination component displays numbered buttons or an input field to allow for navigation.
+ * The pagination component displays numbered buttons or an input field to allow for navigation. 
  */
 export default {
 	title: "Pagination",
@@ -21,7 +22,27 @@ export default {
 			control: "select",
 		},
 		variant: {
-			table: { disable: true },
+			name: "Variant",
+			type: { name: "string" },
+			table: {
+				category: "Component",
+				defaultValue: {
+					summary: "listing",
+				},
+			},
+			options: [
+				"listing",
+				"explicit",
+			],
+			control: "select",
+		},
+		items: {
+			name: "Items",
+			description: "In the \"listing\" variant, each item represents a page button and its label.",
+			table: {
+				category: "Content",
+			},
+			control: "object",
 		},
 	},
 	args: {
@@ -58,7 +79,11 @@ export default {
 	},
 };
 
+/**
+ * The default listing/page variant uses buttons for each page number. 
+ */
 export const Default = PaginationGroup.bind({});
+Default.storyName = "Default (listing)";
 Default.args = {};
 
 // ********* VRT ONLY ********* //
@@ -69,4 +94,19 @@ WithForcedColors.parameters = {
 		forcedColors: "active",
 		modes: disableDefaultModes
 	},
+};
+
+// ********* DOCS ONLY ********* //
+/**
+ * Pagination's explicit variant uses the text field component to represent the current page number,
+ * and action buttons for the previous and next navigation. It also displays text with the total
+ * number of pages.
+ */
+export const Explicit = Template.bind({});
+Explicit.tags = ["!dev"];
+Explicit.args = {
+	variant: "explicit",
+};
+Explicit.parameters = {
+	chromatic: { disableSnapshot: true },
 };
--- a/components/pagination/stories/template.js
+++ b/components/pagination/stories/template.js
@@ -13,52 +13,52 @@ export const Template = ({
 	size = "m",
 	customClasses = [],
 	variant,
-	items,
+	items
 } = {}, context = {}) => {
-	if (variant === "explicit") {
-		return html`
-			<nav
-				class=${classMap({
-					[rootClass]: true,
-					[`${rootClass}--explicit`]: true,
-					...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
-				})}
-			>
-				${ActionButton({
-					size,
-					isQuiet: true,
-					iconSet: "ui",
-					iconName: "ChevronLeft",
-					customClasses: [`${rootClass}-prevButton`],
-				}, context)}
-				${Textfield({
-					size,
-					value: "1",
-					customClasses: [`${rootClass}-textfield`],
-				}, context)}
-				<span class="${rootClass}-counter">of 89 pages</span>
-				${ActionButton({
-					size,
-					isQuiet: true,
-					iconSet: "ui",
-					iconName: "ChevronRight",
-					customClasses: [`${rootClass}-nextButton`],
-				}, context)}
-			</nav>
-		`;
-	}
-	else if (variant == "button") {
-		return SplitButton({
-			position: "left",
-			variant: "accent",
-			label: "Next",
-			iconName: "ChevronLeft100",
-			labelIconName: "ChevronRight100",
-			customFirstButtonClasses: ["spectrum-Pagination-prevButton"],
-			customLastButtonClasses: ["spectrum-Pagination-nextButton"]
-		}, context);
-	}
-	return html`
+
+	const explicitVariant = html`
+		<nav
+			class=${classMap({
+				[rootClass]: true,
+				[`${rootClass}--explicit`]: true,
+				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
+			})}
+		>
+			${ActionButton({
+				size,
+				isQuiet: true,
+				iconSet: "ui",
+				iconName: "ChevronLeft",
+				customClasses: [`${rootClass}-prevButton`],
+			}, context)}
+			${Textfield({
+				size,
+				value: "1",
+				customClasses: [`${rootClass}-textfield`],
+			}, context)}
+			<span class="${rootClass}-counter">of 89 pages</span>
+			${ActionButton({
+				size,
+				isQuiet: true,
+				iconSet: "ui",
+				iconName: "ChevronRight",
+				customClasses: [`${rootClass}-nextButton`],
+			}, context)}
+		</nav>
+	`;
+
+	// @todo This variant should be deprecated, as it uses the deprecated SplitButton component. 
+	const buttonVariant = SplitButton({
+		position: "left",
+		variant: "accent",
+		label: "Next",
+		iconName: "ChevronLeft100",
+		labelIconName: "ChevronRight100",
+		customFirstButtonClasses: ["spectrum-Pagination-prevButton"],
+		customLastButtonClasses: ["spectrum-Pagination-nextButton"]
+	}, context);
+
+	const listingVariant = html`
 		<nav
 			class=${classMap({
 				[rootClass]: true,
@@ -98,4 +98,12 @@ export const Template = ({
 			}, context)}
 		</nav>
 	`;
+
+	if (variant === "explicit") {
+		return explicitVariant;
+	}
+	else if (variant == "button") {
+		return buttonVariant;
+	}
+	return listingVariant;
 };
--- a/components/pickerbutton/CHANGELOG.md
+++ b/components/pickerbutton/CHANGELOG.md
@@ -325,70 +325,34 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 \*feat(pickerbutton)!: migrate to use spectrum tokens (#1940)([ad6051d](https://github.com/adobe/spectrum-css/commit/ad6051d)), closes[#1940](https://github.com/adobe/spectrum-css/issues/1940)
 
-    	###
-    	🛑 BREAKING CHANGES
+### 🛑 BREAKING CHANGES
 
-    		*
-    		migrates the Picker Button component to use `@adobe/spectrum-tokens`
+#### Migrates the Picker Button component to use `@adobe/spectrum-tokens`
 
 - feat(pickerbutton)!: update to use tokens
-
-- feat(pickerbutton): updating tokens
-
 - chore(pickerbutton): move css from generated to index
-
-- feat(pickerbutton): updating with tokens
-
 - docs(pickerbutton): update docs html to correct icon sizes
-
-- refactor(pickerbutton): fixing variable names and tokens
-
-- chore(pickerbutton): update story
-
 - chore(pickerbutton): update storybook controls and template
-
 - chore(pickerbutton): adding more storybook stories for express, quiet etc
-
 - fix(pickerbutton): fixing icon color
-
-- chore(pickerbutton): reset yarn file
-
-- chore: reset yarn file
-
 - refactor(pickerbutton): fix button fill padding calculations
-
 - chore(pickerbutton): remove invalid and isKeyboardFocused variants
-
 - chore(pickerbutton): prevent focused and open when disabled
-
 - fix(pickerbutton): fix disabled hover
-
-update mods
-
 - fix(searchwithin): pass through mod for picker button border color
-
-- chore(pickerbutton): bumping up tokens release
-
 - chore(pickerbutton): update token peer dependency
-
-- fix(searchwithin): update searchwithin story to match docs site
-
-use picker instead of pickerbutton
-remove extra border
-
-- chore(pickerbutton): manual version increase for beta release
-
-- chore(pickerbutton): remove pickerbutton-generated css
-
-- fix(pickerbutton): remove icononly class and remove padding from uiicononly
-
-removes padding from uiicononly class to allow for larger icons in slots
-
+- fix(searchwithin): update searchwithin story to match docs site  
+  use picker instead of pickerbutton and remove extra border
+- fix(pickerbutton): remove icononly class and remove padding from uiicononly  
+  removes padding from uiicononly class to allow for larger icons in slots
 - fix(pickerbutton): explicitly add box-sizing border-box
-
 - fix(pickerbutton): fix icon size
 
-- chore(pickerbutton): manual version increase for beta release
+### Migration guide
+
+#### .spectrum-PickerButton-UIIcon class removed
+
+The `.spectrum-PickerButton-UIIcon` class no longer matches our naming convention. Both types of icons now use the `spectrum-PickerButton-icon` class
 
 <a name="3.0.34"></a>
 
@@ -821,6 +785,17 @@ removes padding from uiicononly class to allow for larger icons in slots
 - remove high loudness selectors from pickerbutton, apply quiet background color from inputgroup, change loudness api to quiet for infieldbutton
 - replaces `medium` loudness with `quiet`
 - refactor spectrum-PickerButton--high to use base spectrum-PickerButton styles
+
+### Migration guide
+
+#### Picker button uses the Quiet variant instead of loudness levels.
+
+The Loudness level classes, `.spectrum-PickerButton--low`, `.spectrum-PickerButton--medium`, and `.spectrum-PickerButton--high`, have been removed.
+
+- Use the base class, `.spectrum-PickerButton`, to apply the default button styles. The default styles correspond to what was previously the Loudness - High variant, which used the class `.spectrum-InfieldButton--high`.
+- Use the modifier class, `spectrum-PickerButton--quiet`, to apply the quiet variant styles. Quiet corresponds to what was previously the Loudness - Low variant, which used the class `.spectrum-InfieldButton--low` class.
+
+The Loudness - Medium variant has been removed, so there is no equivalent.
 
 <a name="1.1.22"></a>
 
--- a/components/pickerbutton/stories/pickerbutton.stories.js
+++ b/components/pickerbutton/stories/pickerbutton.stories.js
@@ -1,12 +1,13 @@
 import { default as Icon } from "@spectrum-css/icon/stories/icon.stories.js";
+import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isFocused, isOpen } from "@spectrum-css/preview/types";
 import { version } from "../package.json";
 import { PickerGroup } from "./pickerbutton.test.js";
-import { Template } from "./template.js";
+import { CustomIconTemplate, Template } from "./template";
 
 /**
- * The picker button component is used as a dropdown trigger. See Combobox.
+ * The picker button component is used as a dropdown trigger within other components such as [combobox](?path=/docs/components-combobox--docs).
  */
 export default {
 	title: "Picker button",
@@ -51,6 +52,7 @@ export default {
 		},
 		isRounded: {
 			name: "Rounded",
+			description: "Increases the amount of rounding on the rounded corners.",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
@@ -82,7 +84,9 @@ export default {
 		},
 		position: {
 			name: "Position",
-			type: { name: "string" },
+			description:
+				"Denotes which side of a form field the button is displayed; this influences which corners are rounded.",
+			type: { name: "string", required: true },
 			table: {
 				type: { summary: "string" },
 				category: "Component",
@@ -100,10 +104,9 @@ export default {
 		isQuiet: false,
 		isDisabled: false,
 		isFocused: false,
-		isKeyboardFocused: false,
 		iconType: "ui",
 		iconName: "ChevronDown",
-		position: "right"
+		position: "right",
 	},
 	parameters: {
 		componentVersion: version,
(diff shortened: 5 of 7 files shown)


rise-erpelding (commented):
Hey, this looks great! Thanks for explaining all of your decision-making with this one in the PR description, that was really helpful in reviewing this.

I left a couple of minor suggestions for clarity but that was it!

components/pickerbutton/CHANGELOG.md
```diff
@@ -208,128 +254,110 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
...
 - fix(pickerbutton): explicitly add box-sizing border-box
-
 - fix(pickerbutton): fix icon size
 
-- chore(pickerbutton): manual version increase for beta release
+### Migration guide
+
+#### .spectrum-PickerButton-UIIcon class removed
+
+The `.spectrum-PickerButton-UIIcon` class no longer matches our naming convention. Both types of icons now use the `spectrum-PickerButton-icon` class
```
marissahuysentruyt, line 355: Is this bit of the migration guide here specifically because these changes were made in version 3.0.34? How did you know where to put this section, as well as the other section about Quiet variant?

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -108,35 +114,177 @@ export default {
...
+};
+
+export const Sizing = SizingTemplate.bind({});
+Sizing.tags = ["docs-only"];
+Sizing.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+export const Open = Template.bind({});
+Open.tags = ["docs-only"];
```
marissahuysentruyt, line 262: What's the distinction between using `vrt-only` and `docs-only` tags? They both remove the story from the sidebar, but our `docs-only` tagged stories have the chromatic snapshot disabled. How should I be deciding which one to use? We don't technically need the `vrt-only` tag if we want it to be included in the chromatic snapshot, right? 

I'll have to look this up!

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -108,35 +114,177 @@ export default {
...
+				size: "s",
+				content: ["UI icon"],
+				customStyles: {
+					"white-space": "nowrap",
+					"--mod-detail-font-color": "var(--spectrum-seafoam-900)",
+				},
+			})}
+			${Template({
+				...args,
+				iconName: "ArrowDown100",
```
marissahuysentruyt, line 194: Should we use the "ChevronDown" icon here? Or is that just a workflow icon?

marissahuysentruyt (changes requested):
@jawinn This looks good! I did leave a few questions for you. There were a few instances where I just wanted to make sure we didn't overlook things (like linking to a different component, or wording, etc). 

The biggest thing I wanted to bring up was the binding for pagination. I don't think I've had many components in these branches that have the chromatic coverage these 3 seem to have, so I might be doing it wrong or expecting the incorrect things. Let me know!

.storybook/decorators/context.js
```diff
@@ -134,6 +134,11 @@ export const withContextWrapper = makeDecorator({
 				if (hasStaticElement && staticKey && staticColorSettings[staticKey]) {
 					container.style.background = staticColorSettings[staticKey].background;
 				}
+
+				// Demonstrate the "Modal" component with a different darker background, so its bounds are visible.
+				if (rootClass == "spectrum-Modal") {
+					container.style.background = "var(--spectrum-transparent-black-400)";
+				}
```
castastrophe, line 141: I am really not sure about including a setting in the decorator that's specific to only one component and likely something we only want in test mode.

components/pagination/stories/pagination.stories.js
```diff
@@ -57,3 +74,15 @@ export default {
 
 export const Default = PaginationGroup.bind({});
 Default.args = {};
+
+/**
+ * Pagination's explicit variant uses the text field component to represent the current page, and text with the total number of pages. Along with action buttons for previous and next navigation.
+ */
+export const Explicit = Template.bind({});
+Explicit.tags = ["docs-only"];
```
castastrophe, line 82: Can you update the tags to our new standards when you get a chance?

components/pagination/stories/template.js
```diff
@@ -117,7 +119,10 @@ export const PaginationGroup = (args, context) => {
 			"flex-direction": "column",
 			"gap": "32px",
 		})}>
-			${Template(args, context)}
+			${Template({
+				...args,
+				variant: "listing",
+			}, context)}
```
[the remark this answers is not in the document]
castastrophe: Ideally this only shows when window.isChromatic() is true

castastrophe (changes requested):

components/pagination/stories/pagination.stories.js
```diff
@@ -57,3 +74,15 @@ export default {
 
 export const Default = PaginationGroup.bind({});
```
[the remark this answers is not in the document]
jawinn: Yeah, I was flip-flopping on how to do that as well, with still getting it to appear first in the autodocs order. I think I'll just rename as "Default (Listing)".

components/pickerbutton/CHANGELOG.md
jawinn: I looked at the git blame for the addition of the docs and found which release they were added in (also this is under version 4.0).

components/pagination/stories/pagination.stories.js
```diff
@@ -57,3 +74,15 @@ export default {
...
 Default.args = {};
+
+/**
+ * Pagination's explicit variant uses the text field component to represent the current page, and text with the total number of pages. Along with action buttons for previous and next navigation.
```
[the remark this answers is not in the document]
jawinn: Thanks. I've updated this wording.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: The tags are now changed to use the newer built-in Chromatic tags though (per Cassondra's suggestion).

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: This part of the template is to show a custom UI icon (other than the default Chevron).

components/pagination/stories/pagination.stories.js
jawinn: Sure. This is done for all of the tags in this PR.

.storybook/decorators/context.js
jawinn: This helps things for the user, not just for test mode. Before it was unclear where the boundaries were, especially when changing the variant control. Example:
- https://opensource.adobe.com/spectrum-css/preview/?path=/story/components-modal--default&args=variant:fullscreen vs 
-  https://pr-2854--spectrum-css.netlify.app/preview/?path=/story/components-modal--default&args=variant:fullscreen

I don't like the idea of a component specific exception either.
1. Should I remove it and just let it display as-is?
2. What about doing a `customStyles` with `--mod-modal-background-color: var(--spectrum-background-layer-2-color);` instead to give the modal a contrasting color?
3. Something else?

components/pagination/stories/template.js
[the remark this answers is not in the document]
jawinn: Some of the stories like this one (and Action button for example) display a different template for Chromatic that covers more options. We were calling this a kind of "kitchen sink" approach and it was partly to limit the number of snapshots from multiple stories. And also to cover testing specific things that we wanted to make sure are in the VRT.

You're right that it'd be a good not to show the chromatic testing (testing preview) view on the Docs / autodocs pages! I put up a fix on this PR that also checks if the `context.viewMode !== "docs"`. I also went ahead and made a similar but global change as part of PR #2885 that will prevent Chromatic templates from appearing on the Docs pages for _all_ components.

.storybook/decorators/context.js
marissahuysentruyt: I'll let Cassondra make the final decision here, but from the options you mentioned, maybe the `customStyles` is the way to go. Is that more in line with how we're using `customStyles` in other components?

components/pagination/stories/pagination.stories.js
```diff
@@ -56,4 +73,19 @@ export default {
...
+Default.storyName = "Default (Listing)";
 Default.args = {};
+
+/**
+ * Pagination's explicit variant uses the text field component to represent the current page number,
+ * and action buttons for the previous and next navigation. It also displays text with the total
+ * number of pages.
+ */
+export const Explicit = Template.bind({});
+Explicit.tags = ["!dev"];
```
marissahuysentruyt, line 85: Do we want to keep `Explicit` in the sidenav? The prod Storybook has docs, default and explicit all in the sidenav. I've been keeping whatever was already in the sidenav, but maybe I shouldn't be. 🤷‍♀️

components/pagination/stories/template.js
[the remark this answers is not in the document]
marissahuysentruyt: Oh great! Can't wait to check out that chromatic testing PR, then!

components/pagination/stories/pagination.stories.js
marissahuysentruyt: Same question in regards to the `!dev` stories for picker button. Do we or should we keep any of those in the sidenav since that's how it is on prod? If it's ok that some are removed, how should I be making that decision on my branches?

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -108,35 +114,178 @@ export default {
...
 		iconType: "ui",
 		iconName: "ChevronDown",
-		position: "right"
+		position: "right",
 	},
 };
 
+// Story specific templates
+const SizingTemplate = (args) => html`
+	<div
```
marissahuysentruyt, line 125: I think the preference is to get markup like this into the `template.js` file as an export (Cassondra left me some requests regarding that on #2793 & #2833 for things like the variant checkbox group template and the divider sizes templates.)

marissahuysentruyt (commented):
Just a few more thoughts and questions for you! The things I mentioned before are all taken care of- nice work, per usual!

.storybook/decorators/context.js
jawinn: I just wanted to note that the `customStyles` option would be changing the opposite of what's on the PR now (component foreground instead of story body background). Which may not be desirable.

components/pagination/stories/pagination.stories.js
jawinn: This was to go along with how we were condensing the stories previously, since the variants are already in the VRT for the default story in the Chromatic-only template.

I think it's worth us all having a conversation soon about the conventions, because I think there is the case to be made for keeping more stories in the sidebar, that may have their chromatic snapshots disabled, for user's sake. In this case I could see keeping them in the sidebar.

jawinn: For now I've added Pagination's "Explicit" back the sidebar, in addition to fixing the "Rendered more hooks than during the previous render" error that was happening when changing the variant control. This one I think is good to keep separate because they're significantly different sets of components (that used to have their own docs site pages), rather than a typical variant or state.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: I've moved these into the `template.js`.

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -108,35 +111,75 @@ export default {
...
 
 export const WithLabel = Template.bind({});
+WithLabel.storyName = "With label";
 WithLabel.args = {
 	label: "Select",
 };
 
 export const Disabled = Template.bind({});
-Disabled.tags = ["vrt-only"];
+Disabled.tags = ["!dev"];
```
marissahuysentruyt, line 130: Did we want to include the same sidebar items as on prod? Storybook prod has default, disabled, express, quiet (with label), and with label all listed.

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -108,35 +111,75 @@ export default {
...
 };
 
 export const Quiet = Template.bind({});
-Quiet.tags = ["vrt-only"];
+Quiet.storyName = "Quiet with label";
+Quiet.tags = ["!dev"];
 Quiet.args = {
-	isQuiet: true
+	isQuiet: true,
+	label: "All",
```
marissahuysentruyt, line 140: Adding the label here would change the chromatic snapshot, correct? It matches the docs site, so I like that update, but does that mean we should run chromatic on this branch? To update the baselines? And I guess if that's the case, modal might need to be checked as well (depending on what we decide to do with seeing the modal better).

marissahuysentruyt (approved):
Just a couple more questions for me. Nothing I left comments on would prevent this from being merged, so I'm approving 👍🎉 Nice work~

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: I don't think all of these need to be there anymore (especially Express). We've been moving away from having stories for every single state and combination of states, and condensing as much as possible into the Default. And handling the large number of possible states in VRTs and Chromatic-only templates. Unless we're changing our approach there. I think the team still needs to have that conversation about conventions for this reorganization.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: Correct. I have that noted as an expected VRT change, and VRT should be run to finalize the validation instructions. Thanks for calling that out! I've run the VRTs and will check on the differences.

components/pagination/stories/pagination.stories.js
```diff
@@ -1,9 +1,11 @@
 import { default as ActionButton } from "@spectrum-css/actionbutton/stories/actionbutton.stories";
 import { version } from "../package.json";
-import { PaginationGroup } from "./template";
+import { PaginationGroup, Template } from "./template";
 
 /**
  * The pagination component displays numbered buttons or an input field to allow for navigation.
+ * 
+ * The default listing/page variant uses buttons for each page number.
```
rise-erpelding, line 8: This is kind of a nitpick at this point and I don't have a strong opinion, but how would you feel about putting this as a comment on the Default story rather than the component overview?

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -1,9 +1,10 @@
+
 import { default as Icon } from "@spectrum-css/icon/stories/icon.stories.js";
 import { version } from "../package.json";
-import { Template } from "./template";
+import { CustomIconTemplate, SizingTemplate, Template } from "./template";
 
 /**
- * The picker button component is used as a dropdown trigger. See Combobox.
+ * The picker button component is used as a dropdown trigger within other components such as combobox.
```
rise-erpelding, line 7: Nitpick: we could link to combobox here.

rise-erpelding (approved):
I left a couple of small, nitpick-y things that you could optionally change if you wanted to, but otherwise I think this is looking pretty good!

components/pagination/stories/pagination.stories.js
jawinn: Thanks, I've moved that to the Default story.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: I've added that link.

.storybook/decorators/context.js
jawinn: I've dropped that commit and have addressed this by displaying the Underlay component beside the Modal, similar to Dialog. I've also added a `showUnderlay` control.

components/pagination/stories/template.js
```diff
@@ -103,21 +103,32 @@ export const Template = ({
...
 	return html`
 		<div style=${styleMap({
-			"display": window.isChromatic() ? "none" : undefined,
+			"display": window.isChromatic() && context.viewMode !== "docs" ? "none" : undefined,
 		})}>
 			${Template(args, context)}
 		</div>
 		<div style=${styleMap({
-			"display": window.isChromatic() ? "flex" : "none",
+			"display": window.isChromatic() && context.viewMode !== "docs" ? "flex" : "none",
```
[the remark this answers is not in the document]
castastrophe: The `.storybook/decorators/testing-preview.js` decorator does include a viewMode check already anyway:

```
window.isChromatic = typeof isChromatic === "function" && isChromatic() === true ? isChromatic : () => testingPreview && viewMode !== "docs";
```

components/modal/stories/template.js
```diff
@@ -27,6 +29,9 @@ export const Template = ({
 	`;
 
 	return html`
+		${when(showUnderlay,
+			() => Underlay({ isOpen })
+		)}
```
marissahuysentruyt, line 34: For dialog, I had to turn off the modal for the testing preview, so I'm really curious to see how we handle the testing preview for the modal itself!

components/pagination/stories/pagination.stories.js
```diff
@@ -70,3 +89,17 @@ WithForcedColors.parameters = {
...
 	},
 };
+
+// ********* DOCS ONLY ********* //
+/**
+ * Pagination's explicit variant uses the text field component to represent the current page number,
+ * and action buttons for the previous and next navigation. It also displays text with the total
+ * number of pages.
+ */
+export const Explicit = Template.bind({});
```
marissahuysentruyt, line 99: We should add the tags here, right (`!dev`)? To get this story out of the sidebar?

components/pagination/stories/template.js
```diff
@@ -99,21 +99,32 @@ export const Template = ({
...
+		return buttonVariant;
+	}
+	return listingVariant;
 };
 
 export const PaginationGroup = (args, context) => {
 	return html`
 		<div style=${styleMap({
-			"display": window.isChromatic() ? "none" : undefined,
+			"display": window.isChromatic() && context.viewMode !== "docs" ? "none" : undefined,
```
marissahuysentruyt, line 115: Depending on what gets merged first, and since we already have some of this Chromatic preview stuff in place, did we want to refactor this `PaginationGroup` to use `Variants()`? 

Looks like it's being addressed in[ #2912 ](https://github.com/adobe/spectrum-css/pull/2912/files#diff-ae92bc343a4abb638a6bfdc1168db6fcab504115107fb5f7891062510c4acce5), but we might want to keep that though in our back pocket if this PR looks like it'll get merged first. 🤷‍♀️ How would you handle it?

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -113,6 +115,7 @@ export const Default = Template.bind({});
 Default.args = {};
 
 export const WithLabel = Template.bind({});
+WithLabel.storyName = "With label";
```
marissahuysentruyt, line 118: Do we need the `!dev` tag here, too?

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -130,11 +133,50 @@ WithForcedColors.parameters = {
 export const Disabled = Template.bind({});
 Disabled.tags = ["!autodocs", "!dev", "test"];
 Disabled.args = {
-	isDisabled: true
+	isDisabled: true,
 };
 
 export const Quiet = Template.bind({});
 Quiet.tags = ["!autodocs", "!dev", "test"];
```
marissahuysentruyt, line 140: I think I asked this on the autodocs config PR, but why are these Quiet and Disabled stories removed from the docs page? They're on the docs site right now, but are we "saving" them for just VRT? Should we be doing that for other components as well?

marissahuysentruyt (commented):
This is looking good! Just looked at it with some fresh eyes and the modes stuff in mind, so I found a few more items we could address if you wanted!

components/pagination/stories/pagination.stories.js
jawinn: I'm purposefully including this one on the sidebar. It's a very different type of pagination that uses different elements/markup. I think it makes sense in this case, especially since these two types of pagination have their own separate docs site pages.

components/pagination/stories/template.js
jawinn: Thanks for pointing that out in that upcoming PR. I'll have to address the rebase conflicts when it merges.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: I had left it how it was, but I think that is a good idea. I've added that tag to remove it from the sidebar in the latest update.

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: I've added these to the docs page.

jawinn:
> Just a couple of really minor things that I don't feel strongly about and don't change the PR approval: In Modal and Pagination, we have some of these controls that are lowercased instead of capitalized, which doesn't really match the way the rest of the controls are named, how would you feel about adding a "name" property to these to adjust that? 

I've adjusted the `name` property on those so the capitalization is consistent. I also added a couple default value summaries and a missing control description.

components/pagination/stories/template.js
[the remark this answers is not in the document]
jawinn: I've removed those `viewMode` checks now that they're covered in the testing preview decorator changes that merged into main.

components/pagination/stories/pagination.stories.js
marissahuysentruyt: I didn't re-read your last response to me on this. sorry!

components/pagination/stories/pagination.stories.js
```diff
@@ -58,7 +79,11 @@ export default {
 	},
 };
 
+/**
+ * The default listing/page variant uses buttons for each page number. 
+ */
 export const Default = PaginationGroup.bind({});
+Default.storyName = "Default (Listing)";
```
castastrophe, line 86:
```suggestion
Default.storyName = "Default (listing)";
```

Is listing a name or a descriptor? I think it should probably be lowercase since it describes the variant.

components/pagination/stories/pagination.stories.js
```diff
@@ -70,3 +95,18 @@ WithForcedColors.parameters = {
 		modes: disableDefaultModes
 	},
 };
+
+// ********* DOCS ONLY ********* //
+/**
```
castastrophe, line 100: Great addition, thanks for adding this

components/pagination/stories/template.js
```diff
@@ -98,4 +98,12 @@ export const Template = ({
...
 		</nav>
 	`;
+
+	if (variant === "explicit") {
+		return explicitVariant;
+	}
+	else if (variant == "button") {
+		return buttonVariant;
+	}
+	return listingVariant;
```
castastrophe, line 108: Very succinct, nice restructuring

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -1,12 +1,14 @@
+
```
castastrophe, line 1: Can probably remove the empty whitespace at the top

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -1,12 +1,14 @@
...
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isFocused, isOpen } from "@spectrum-css/preview/types";
 import { version } from "../package.json";
 import { PickerGroup } from "./pickerbutton.test.js";
-import { Template } from "./template.js";
+import { CustomIconTemplate, Template } from "./template";
 
 /**
- * The picker button component is used as a dropdown trigger. See Combobox.
+ * The picker button component is used as a dropdown trigger within other components such as [combobox](?path=/docs/components-combobox--docs).
```
castastrophe, line 11: Nice internal linking here

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -82,7 +85,9 @@ export default {
 		},
 		position: {
 			name: "Position",
-			type: { name: "string" },
+			description:
+				"Which side the picker button is displayed on. This determines which corners are rounded.",
```
castastrophe, line 89:
```suggestion
				"Denotes which side of a form field the button is displayed; this influences which corners are rounded.",
```

castastrophe: Minor phrasing suggestion

components/pickerbutton/stories/pickerbutton.stories.js
```diff
@@ -142,6 +146,48 @@ Disabled.parameters = {
...
+CustomIcon.storyName = "With custom icon";
+CustomIcon.tags = ["!dev"];
+CustomIcon.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * The `spectrum-PickerButton--rounded` class increases the amount of rounding on the rounded corners.
+ */
+export const Rounded = Template.bind({});
```
castastrophe, line 182: Great additions here, thanks for taking care of this!

components/pickerbutton/stories/template.js
```diff
@@ -70,3 +71,66 @@ export const Template = ({
 		</button>
 	`;
 };
+
+/**
+ * Displays the component with a custom icon (instead of the chevron UI icon).
+ * Two examples are shown; with a custom UI icon and a custom Workflow icon.
+ */
+export const CustomIconTemplate = (args) => html`
```
castastrophe, line 79: Ideally I'd like to move away from hardcoding layouts and headings in the templates for the docs pages but I think this is an edge case where it makes sense for now. I'd love to brainstorm if there are ways we can support these custom demos without a lot of custom grids and headings.

castastrophe (approved):
A few minor suggestions but approved overall

components/pickerbutton/stories/pickerbutton.stories.js
jawinn: Thanks. This is updated.

PR #2855, opened 2024-06-21
chore(dialog): z-index for dialog docs

marissahuysentruyt:
## Description
Currently on the [static docs site](https://opensource.adobe.com/spectrum-css/dialog.html), the dialogs don't really function properly (the default dialog does however, work properly in [Storybook](https://opensource.adobe.com/spectrum-css/preview/?path=/docs/components-dialog--docs). If you close any open dialog variant, and try to reopen it, it is "stuck" underneath `spectrum-Underlay`. I believe this is because a `z-index: 1 !important` is overriding other `z-index: 2`s for `spectrum-Modal`. 

This PR adds a new `.spectrum-CSSExample-modal` class to mimic `.spectrum-Modal-wrapper` with a `z-index` that should position it above `.spectrum-Underlay`. It also adds/removes classes in `dialog.yml` to better implement the open/close behaviors. 

**Note:** The hero, no-divider, and dismissible dialogs are now closed on page load, as opposed to being open previously.

### Validation steps

@jawinn
- [x] Pull down the branch
- [x] Run `yarn run dev` and visit [the dialog docs page](http://localhost:3000/dialog.html)
- [x] Verify the no-divider, dismissible, and hero variants are closed on page load
- [x] Test all closed dialogs on the page: 
    - opening the dialog should add the `is-open` class to `spectrum-Modal` and `spectrum-Modal-wrapper`
    - once the dialog is open, any buttons (like the close button or a button group) should be clickable/focusable
    - the z-index for `spectrum-CSSExample-modal` should be `2 !important` to ensure it remains above the `z-index: 1 !important` set on `spectrum-Underlay`.
    - closing the dialog should remove `is-open` from those same modal elements 

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/dialog/metadata/dialog.yml
+++ b/components/dialog/metadata/dialog.yml
@@ -71,8 +71,8 @@ examples:
     markup: |
       <button class="spectrum-Button spectrum-Button--sizeM spectrum-Button--outline spectrum-Button--staticWhite spectrum-CSSExample-overlayShowButton" onclick="openDialog(this.nextElementSibling)"><span class="spectrum-Button-label">Open Dismissible Dialog</span></button>
 
-      <div class="spectrum-Modal-wrapper spectrum-CSSExample-dialog">
-        <div class="spectrum-Modal is-open" data-testid="modal">
+      <div class="spectrum-Modal-wrapper spectrum-CSSExample-modal">
+        <div class="spectrum-Modal" data-testid="modal">
           <section class="spectrum-Dialog spectrum-Dialog--medium spectrum-Dialog--dismissable" role="dialog" tabindex="-1" aria-modal="true">
             <div class="spectrum-Dialog-grid">
               <h1 class="spectrum-Dialog-heading">Disclaimer</h1>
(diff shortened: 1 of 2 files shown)


jawinn (approved):
I validated on the PR site and this fixes the issues on the docs page for Dialog 🎉  

There may be some changes we should make in the future to be more in line with SWC or to be clear about how source order affects both Underlay and Dialog being used together:

### Thoughts on the z-indexes:
I poked around a little bit and I see what is going on now, and am wondering if this should be looked into as a followup issue. The original display issue stems from both `.spectrum-Modal` and `.spectrum-Underlay` having `z-index: 1`, and whether one displays on top of the other depends on the source order. 

In the original prod issue, there is only a single underlay, later on in the HTML source (after the Dialog markup). In Storybook, the underlay markup appears before the Dialog, which is why it looks fine; you can test to see the bug behavior there by dragging the underlay to be after the Dialog in the DOM inspector.

I looked at Spectrum Web Components, and they have the underlay within what we call `.spectrum-Modal-wrapper` (their `sp-dialog-base` uses those styles), also before the dialog modal:
Screenshot 2024-06-25 at 9 22 20 AM

This brings up a few questions for me:
1. If we want to allow underlay to be later in the source, should `.spectrum-Modal` have `z-index: 2`?
2. If underlay should always be before the modal, should we have multiple underlay divs for each one of these examples, and the JS should be targeting those? And should that  be documented?
3. If so (# 2), should we be more in line with SWC and have the underlay nested within `.spectrum-ModalWrapper`?

rise-erpelding (approved):
This does indeed fix the issue! I’m honestly a little puzzled because it looks like `.spectrum-Modal` does have a `z-index: 2` in the code but every time I inspect (and also in the `dist`), it has a `z-index: 1`. Seems like something in the build is changing the z-index, maybe? 🤔 

image

image

If the modal is expected to always have `z-index: 1` then I would say let’s not throw `!important` in with `z-index: 2 !important` but since I'm seeing `z-index: 2` for `.spectrum-Modal` and `!important` is already pretty liberally sprinkled in with the CSSExample classes, this feels like the right choice.

+1 to Josh’s point that underlay isn’t causing issues elsewhere because of the stacking order (looks like from his comment he’s also seeing `.spectrum-Modal` having `z-index: 1`), this would be a good thing to clarify at some point.

castastrophe (approved):

PR #2844, opened 2024-06-13
feat(storybook): chromatic modes testing

castastrophe:
## Description

Enable chromatic modes testing to scale VRTs on all components.

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [ ] Expect to see VRTs run for all components against the following contexts: default, dark, rtl, express
- [ ] Expect to see the following components run an additional mode for mobile scale: coachmark, dialog, and icon

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [ ] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

- [ ] VRTs have been run and looked at.
- [ ] Any VRT changes have been accepted (by reviewer and/or PR author), or there are no changes.

## To-do list

- [ ] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [ ] I have updated relevant storybook stories and templates.
- [ ] I have tested these changes in Windows High Contrast mode.
- [ ] If my change impacts **other components**, I have tested to make sure they don't break.
- [ ] If my change impacts **documentation**, I have updated the documentation accordingly.
- [ ] ✨ This pull request is ready to merge. ✨

PR #2123, opened 2023-08-24
fix(textfield): add line-height to prevent missing diacritics

pfulton:
This adds a line-height definition to the textfield's input and sets that value equal to the height of the input. This prevents diacritics from being cut off in certain languages. Only applies to non-multiline (textarea) inputs.

Similar to https://github.com/adobe/react-spectrum/pull/4579 in React Spectrum.

[Jira ticket](https://jira.corp.adobe.com/browse/CSS-532)

There's no visual difference for English.

Local with changes vs. production
Screenshot 2023-08-24 at 2 23 28 PM

Screenshot 2023-08-24 at 2 23 58 PM

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Regression testing

Validate:

1. A legacy documentation page (such as [accordion](https://pr-###--spectrum-css.netlify.app/accordion.html)), including:

- [ ] The page renders correctly
- [ ] The page is accessible
- [ ] The page is responsive

2. A migrated documentation page (such as [action group](https://pr-###--spectrum-css.netlify.app/actiongroup.html)), including:

- [ ] The page renders correctly
- [ ] The page is accessible
- [ ] The page is responsive

## To-do list

- [ ] I have read the [contribution guidelines](/.github/CONTRIBUTING.md).
- [ ] I have updated relevant storybook stories and templates.
- [ ] I have tested these changes in Windows High Contrast mode.

- [ ] If my change impacts **other components**, I have tested to make sure they don't break.
- [ ] If my change impacts **documentation**, I have updated the documentation accordingly.

- [ ] ✨ This pull request is ready to merge. ✨

PR #2856, opened 2024-06-21
feat(textfield): migrate to spectrum 2

rise-erpelding:
## Description

CSS-767

* The textfield component was checked against the [tokens design spec ](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9602-1303&t=CJIk4zpSYeOHiT88-0) and any discrepancies were adjusted

* Note that there are some tokens that currently have different values than what is seen in Figma, this is (for the moment) intentional

### Validation steps

1. Confirm that code style matches S2:
    - [ ] Spectrum styles have been moved to index.css
    - [ ] Medium is default
1. Confirm that quiet variant is removed:
    - [ ] There is no remaining evidence of quiet variant in CSS or in implementation
1. Check that adjustments to design match [spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9602-1303&t=CJIk4zpSYeOHiT88-0):
    - [ ] Spacing start/end edge to value for XL component matches design (should be using `component-edge-to-text-300`)
    - [ ] Spacing for top edge to value text should change with size to match spec instead of remaining the same for every t-shirt size
    - [ ] Spacing for bottom edge to value text should change with size to match spec instead of remaining the same for every t-shirt size, note that the token values don't match the Figma value, and you can confirm the values [here](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9698-1839&t=CJIk4zpSYeOHiT88-0) if needed
    - [ ] Border thickness matches spec
    - [ ] Border colors match spec
    - [ ] Inline side label variant has `--spectrum-spacing-200` for every size
    - [ ] Disabled background color (text input color) and border color match design spec

### Regression testing

Validate:

3. If components have been modified, VRTs have been run on this branch:

## To-do list


components/textfield/index.css
```diff
@@ -10,414 +10,179 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-	);
-
-	--spectrum-textfield-width: 240px; /* override per api */
-	--spectrum-textfield-min-width: var(
-		--spectrum-text-field-minimum-width-multiplier
-	);
-	--spectrum-textfield-corner-radius: var(--spectrum-corner-radius-100);
+	/* set input line-height to the height of the textfield - prevents the cutting off of diacritics in some languages */
+	/* disallow mod for max compatibility */
+	--spectrum-textfield-input-line-height: var(--spectrum-textfield-height);
```
rise-erpelding, line 16: I did look into this, as this is _**not**_ the line height seen in Figma. Using the line height token from Figma would cause [the original bug that was the reason this line height was added](https://github.com/adobe/spectrum-css/pull/2123), so this has intentionally been left as is.

components/textfield/index.css
```diff
@@ -13,411 +13,199 @@ governing permissions and limitations under the License.
...
-  /* set input line-height to the height of the textfield - prevents the cutting off of diacritics in some languages */
-  /* disallow mod for max compatibility */
-  --spectrum-textfield-input-line-height: var(--spectrum-textfield-height);
-  --spectrum-texfield-animation-duration: var(
-		--spectrum-animation-duration-100
-	);
+	/* set input line-height to the height of the textfield - prevents the cutting off of diacritics in some languages */
+	/* disallow mod for max compatibility */
+	--spectrum-textfield-input-line-height: var(--spectrum-textfield-height);
+	--spectrum-texfield-animation-duration: var(--spectrum-animation-duration-100);
```
rise-erpelding, line 19: Wanting to keep this large diff out of other commits

components/textfield/stories/template.js
```diff
@@ -172,3 +173,137 @@ export const Template = ({
...
+			testHeading: "Read-only",
+			value: "username@reallylongemail.com",
+			isReadOnly: true,
+		}, 
+		{
+			testHeading: "Loading",
+			isLoading: true,
+		}
+	],
+});
```
rise-erpelding, line 309: The intention here is to make the test cases look pretty similar to what's happening in #2844 without bringing _everything_ over but adding test cases where it seems necessary.

components/textfield/stories/template.js
```diff
@@ -109,11 +125,11 @@ export const Template = ({
 		}))}
 		${when(multiline,
 			() => html`<textarea
-				placeholder=${ifDefined(placeholder)}
+				placeholder=${ifDefined(placeholder && placeholder.trim() ? placeholder : undefined)}
```
rise-erpelding, line 128: This change makes it so that if there is no string defined for `placeholder` that the attribute doesn't show up in the markup

components/textfield/stories/template.js
rise-erpelding: See also https://github.com/adobe/spectrum-css/blob/feat-chromatic-modes-testing/components/textfield/stories/template.js

components/textfield/index.css
```diff
@@ -410,7 +410,8 @@ governing permissions and limitations under the License.
 	}
 
 	/* keyboard focus */
-	.is-keyboardFocused & {
+	.is-keyboardFocused &,
+	&:focus-visible {
```
rise-erpelding, line 422: The way textfield was before, it didn't have that nice focus outline when you keyboard focus, and that didn't seem quite right.

components/textfield/index.css
```diff
@@ -501,6 +501,11 @@ governing permissions and limitations under the License.
 		&::placeholder {
 			color: var(--highcontrast-textfield-text-color-readonly, var(--mod-textfield-text-color-readonly, var(--spectrum-textfield-text-color-readonly)));
 			background-color: transparent;
+
+			.is-disabled &,
+			.is-disabled:hover & {
+				color: var(--highcontrast-textfield-text-color-disabled, var(--mod-textfield-text-color-disabled, var(--spectrum-textfield-text-color-disabled)));
```
rise-erpelding, line 507: Ensures we get a disabled color for placeholder text if both disabled and read-only are on

components/textfield/index.css
```diff
@@ -1265,17 +553,10 @@ governing permissions and limitations under the License.
 
 /*** Text Area ***/
 .spectrum-Textfield--multiline {
```
marissahuysentruyt, line 555: It looks like Text Area is its own component now (at least it's a separate page in Figma). Do you think that will change anything in this Text Field component? We do technically return a `textarea` with the multiline variant, so does that encroach at all? I have no idea, I'm just thinking out loud.

components/textfield/stories/template.js
```diff
@@ -109,11 +125,11 @@ export const Template = ({
...
 				name=${ifDefined(name)}
 				id=${ifDefined(id ? `${id}-input` : undefined)}
 				.value=${ifDefined(value)}
-				autocomplete=${autocomplete ? undefined : "off"}
+				autocomplete=${ifDefined(autocomplete ? undefined : "off")}
 				?required=${isRequired}
 				?disabled=${isDisabled}
```
marissahuysentruyt, line 134: For `required`, I see the attribute getting added, but do we need to represent that somehow in the UI? With the asterisk? I see the * in Figma, but wasn't sure if we needed it here somewhere.

components/textfield/stories/template.js
```diff
@@ -140,6 +156,15 @@ export const Template = ({
 				})}
 			/>`
 		)}
+		${when(helpText, () => HelpText({
+			...globals,
```
marissahuysentruyt, line 160: I was going to say, I don't think we need the `...globals`, but that's just on main now, right?! 😂🤦‍♀️

components/textfield/stories/textfield.stories.js
```diff
@@ -166,9 +185,25 @@ export default {
 				disable: true,
 			},
 		},
+		hasCharacterCount: {
+			name: "Has character count",
```
marissahuysentruyt, line 189: Should we add a `description` at all? It does seem pretty self-explanatory to me, so feel free not to.

components/textfield/stories/textfield.stories.js
```diff
@@ -180,10 +215,13 @@ export default {
...
 		labelPosition: "top",
 		labelText: "Username",
+		displayHelpText: false,
+		helpText: "Create a username with at least 5 characters",
+		placeholder: "Username",
 		size: "m",
 		multiline: false,
 		grows: false,
-		isQuiet: false,
+		pattern: "",
```
marissahuysentruyt, line 224: What would `pattern` be for? I see it as an attribute, but is that for, like, client-side validation or something? I'll have to look it up!

components/textfield/stories/template.js
marissahuysentruyt: Great idea! Want to do that for `pattern` too? There's a couple instances where that could be handy.

```
pattern=${ifDefined(pattern ? pattern : undefined)}
```

marissahuysentruyt (changes requested):
Phew, that was a big one!! I found some places where you're pointing to the correct token, but it seems like the value might be off. Maybe once another round of tokens gets released, some of those will take care of themselves.

I did find a weird custom property that I would want to see if you were getting too. Some `--system` prefixed properties are overriding the correct `--spectrum` properties, but I don't know where those are coming from. 

Other than that, I found a couple of small things! Excellent work!

components/textfield/stories/template.js
```diff
@@ -148,3 +173,137 @@ export const Template = ({
 	</div>
 	`;
 };
+
```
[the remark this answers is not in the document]
rise-erpelding: Yes exactly! Looking at what's on `main` now, I want to keep `TextFieldGroup` mostly as it is, except with a few additions to display the things that weren't being displayed before, so `renderDataWithHeading` and `Variants` are just there temporarily until we can use the newer functions in `utilities.js`. I will push some small adjustments to account for changes since that modes branch was merged to `main`.

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
-	--spectrum-textfield-min-width: var(
-		--spectrum-text-field-minimum-width-multiplier
-	);
-	--spectrum-textfield-corner-radius: var(--spectrum-corner-radius-100);
+	/* set input line-height to the height of the textfield - prevents the cutting off of diacritics in some languages */
+	/* disallow mod for max compatibility */
+	--spectrum-textfield-input-line-height: var(--spectrum-textfield-height);
+	--spectrum-texfield-animation-duration: var(--spectrum-animation-duration-100);
+
+	--spectrum-textfield-width: var(--spectrum-field-default-width-medium);
```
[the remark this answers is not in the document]
rise-erpelding: Probably not a bad idea to bring a little more organization, let me know what you think about the update.

components/textfield/index.css
```diff
@@ -532,323 +229,151 @@ governing permissions and limitations under the License.
...
-		var(--mod-textfield-corner-radius, var(--spectrum-textfield-corner-radius)) /
-			2
-	);
-
-	.spectrum-Textfield--quiet & {
-		margin-block-end: var(
-			--mod-textfield-character-count-spacing-block-quiet,
-			var(--spectrum-textfield-character-count-spacing-block-quiet)
-		);
+	padding-inline-end: calc(var(--mod-textfield-corner-radius, var(--spectrum-textfield-corner-radius)) / 2);
```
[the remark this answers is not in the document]
rise-erpelding: That's a good callout. I think I left this `padding-inline-end` here from before because there wasn't really any guidance one way or the other, but it doesn't look like there is any in the design.

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
-	);
-	--spectrum-textfield-spacing-block-end: var(
-		--spectrum-component-bottom-to-text-100
-	);
-	--spectrum-textfield-spacing-block-quiet: var(
-		--spectrum-field-edge-to-border-quiet
-	);
+	--spectrum-textfield-spacing-inline: var(--spectrum-component-edge-to-text-100);
+	--spectrum-textfield-spacing-block-start: var(--spectrum-component-top-to-text-100);
+	--spectrum-textfield-spacing-block-end: var(--spectrum-component-bottom-to-text-100);
```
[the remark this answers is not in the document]
rise-erpelding: Correct! Here's the link in Figma where they're talking about the token value vs. the Figma override value https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9698-1839&t=p7KpYlcfdrvsuIM9-0

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
 
 	/* focus indicator */
-	--spectrum-textfield-focus-indicator-width: var(
-		--spectrum-focus-indicator-thickness
-	);
+	--spectrum-textfield-focus-indicator-width: var(--spectrum-focus-indicator-thickness);
 	--spectrum-textfield-focus-indicator-gap: var(--spectrum-focus-indicator-gap);
 
+	/* border width */
+	--spectrum-textfield-border-width: var(--spectrum-border-width-200);
```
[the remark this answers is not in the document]
rise-erpelding: Yep. I didn't even notice this! I think what happened was, initially I'd pulled out the themes, then put them back, and didn't notice that this was causing problems... there are border definitions in the `components/textfield/themes/spectrum.css` file and it's throwing off all of the border custom properties that were set in `spectrum.css` already.

I asked, and we're going to keep doing S2 migrations with pulling out the themes like we were before, so I've pulled them out again and this problem should go away.

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
-		--spectrum-component-bottom-to-text-75
-	);
-	--spectrum-textfield-character-count-spacing-inline-side: var(
-		--spectrum-side-label-character-count-to-field
-	);
-	--spectrum-textfield-character-count-spacing-block-side: var(
-		--spectrum-side-label-character-count-top-margin-medium
-	);
+	--spectrum-textfield-character-count-color: var(--spectrum-neutral-subdued-content-color-default);
+	--spectrum-textfield-character-count-font-family: var(--spectrum-sans-font-family-stack);
```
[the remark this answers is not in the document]
rise-erpelding: Yes! You're right, we did talk about it.

In the tokens, we have this defined for `default-font-family`:
```
--spectrum-default-font-family:var(--spectrum-sans-serif-font-family);
--spectrum-sans-serif-font-family:Adobe Clean;
```

And then for `sans-font-family-stack` we have the following:
```
--spectrum-sans-font-family-stack:adobe-clean, var(--spectrum-sans-serif-font-family), 'Source Sans Pro', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Ubuntu, 'Trebuchet MS', 'Lucida Grande', sans-serif;
--spectrum-sans-serif-font-family:Adobe Clean;
```

What we'd talked about with @jawinn before is that they point to the same font but on the off-chance that for some reason the browser can't find that Adobe Clean font, we have a series of back-up fonts that can be used instead--it feels a lot more robust to have more backup options, and since it should render the same if the Adobe Clean font is found, it felt preferable to use the stack instead of the single font token.

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
+	--spectrum-textfield-icon-spacing-inline-start-valid: var(--spectrum-text-to-visual-100);
+	--spectrum-textfield-icon-spacing-inline-end-valid: var(--spectrum-field-edge-to-validation-icon-medium);
+	--spectrum-textfield-icon-spacing-block-valid: var(--spectrum-field-top-to-validation-icon-medium);
 
 	/* font styles */
 	--spectrum-textfield-font-family: var(--spectrum-sans-font-family-stack);
 	--spectrum-textfield-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-textfield-line-height: var(--spectrum-line-height-100);
+	--spectrum-textfield-line-height-cjk: var(--spectrum-cjk-line-height-100);
 	--spectrum-textfield-placeholder-font-size: var(--spectrum-font-size-100);
```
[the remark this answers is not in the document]
rise-erpelding: Definitely something to clarify with design since we do want to make sure we've got all the different pieces of text colored/sized etc. correctly!

components/textfield/index.css
```diff
@@ -13,503 +13,200 @@ governing permissions and limitations under the License.
...
 	--spectrum-textfield-focus-indicator-gap: var(--spectrum-focus-indicator-gap);
 
+	/* border width */
+	--spectrum-textfield-border-width: var(--spectrum-border-width-200);
+
 	/*** Colors ***/
 	--spectrum-textfield-background-color: var(--spectrum-gray-25);
 
+	/* border  colors */
+	--spectrum-textfield-border-color: var(--spectrum-gray-300);
```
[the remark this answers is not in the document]
rise-erpelding: This one was the same issue as the border width, should be fixed now!

components/textfield/index.css
rise-erpelding: Taking a quick glance, textarea looks pretty similar (and was done by the same designer), and we have a card in the backlog for [textarea](https://jira.corp.adobe.com/browse/CSS-768), the main differences are the minimum width and height/minimum height.

It also looks like they were designed separately for S1 but are still both part of the textfield component, so I suspect when we get to migrating textarea it would just consist of making any necessary adjustments within the same textfield component.

components/textfield/stories/template.js
rise-erpelding: I think I would leave that on the implementations to add since there doesn't seem to be any specific CSS needed if the `required` attribute is on. There is [some documentation](https://spectrum.corp.adobe.com/page/text-field/#Required-or-optional) on what styling requirements there are for required text, I think the only thing that might be relevant to text field would be the asterisk in the field label, but that's in a different component.

components/textfield/stories/textfield.stories.js
rise-erpelding: Yes, I don't think I've used it before but it looks like you can use `pattern` to trigger `:valid` or `:invalid` pseudoclass which potentially could be neat.

components/textfield/index.css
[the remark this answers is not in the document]
marissahuysentruyt: sounds good!

components/textfield/index.css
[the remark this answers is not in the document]
marissahuysentruyt: Beautiful, thanks for re-explaining that for me. Makes sense, and I'll be sure to use it in the future then 😬

components/textfield/stories/template.js
marissahuysentruyt: Oh right, in the field label component. Got it.

components/textfield/index.css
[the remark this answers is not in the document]
marissahuysentruyt: That's it!

components/textfield/index.css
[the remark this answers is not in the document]
marissahuysentruyt: Excellent detective skills 🕵️‍♀️

components/textfield/stories/template.js
```diff
@@ -148,3 +173,130 @@ export const Template = ({
...
+		testHeading: "Disabled",
+		isDisabled: true,
+	},
+	{
+		testHeading: "Required",
+		isRequired: true,
+	},
+	{
+		testHeading: "Read-only",
+		value: "username@reallylongemail.com",
```
marissahuysentruyt, line 299: Does this need to be changed to the new "UsernameWiderThanInput@ReallyLongEmail.com" at all? It's not really necessary I guess, but maybe it would cut down on VRT diffs?
Screenshot 2024-07-25 at 12 19 14 PM

rise-erpelding: I'll change it! It doesn't look like `main` is applying a value at all, which I must have disagreed with, because I added one.

components/textfield/stories/template.js
```diff
@@ -234,10 +234,6 @@ export const TextFieldGroup = Variants({
 		labelText: "Username",
 		value: "UsernameWiderThanInput@ReallyLongEmail.com"
 	},
-	{
-		testHeading: "With placeholder",
-		placeholder: "Username",
```
rise-erpelding: I don't think there's much reason to test the placeholder here, so I removed it after initially adding it.

I did leave it in Storybook as a control but I'm open to hearing an argument for removing it.

components/textfield/index.css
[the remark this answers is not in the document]
rise-erpelding: I did ask design about this and looking back through Slack and the Jira ticket I don't think I see a solid answer, I'm going to err on the side of keeping the legacy style (`padding-inline-end` equal to half the border radius) for now.

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
...
 	/* font styles */
 	--spectrum-textfield-font-family: var(--spectrum-sans-font-family-stack);
 	--spectrum-textfield-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-textfield-font-style: var(--spectrum-default-font-style);
+	--spectrum-textfield-placeholder-font-size: var(--spectrum-font-size-100);
+	--spectrum-textfield-line-height: var(--spectrum-line-height-100);
+	--spectrum-textfield-line-height-cjk: var(--spectrum-cjk-line-height-100);
+
+	/* set input line-height to the height of the textfield - prevents the cutting off of diacritics in some languages */
+	--spectrum-textfield-input-line-height: var(--spectrum-textfield-height);
```
rise-erpelding, line 57: Noting that because of our previous diacritic issues, we use a line height set to the textfield height for the single line (text field) input WITHOUT an option to use a mod, and the `--spectrum-textfield-line-height` token (and mod) are used for the multiline (text area) input.

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
...
 	/* default icon spacing - valid */
-	--spectrum-textfield-icon-spacing-inline-end-quiet-valid: var(--spectrum-field-edge-to-validation-icon-quiet);
+	--spectrum-textfield-icon-spacing-inline-start-valid: var(--spectrum-text-to-visual-100);
+	--spectrum-textfield-icon-spacing-inline-end-valid: var(--spectrum-field-edge-to-validation-icon-medium);
+	--spectrum-textfield-icon-spacing-block-valid: var(--spectrum-field-top-to-validation-icon-medium);
 
 	/* font styles */
 	--spectrum-textfield-font-family: var(--spectrum-sans-font-family-stack);
 	--spectrum-textfield-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-textfield-font-style: var(--spectrum-default-font-style);
```
rise-erpelding, line 51: You'll see that `font-style` was added in multiple places because it wasn't here before. I'm not sure if there's an argument for keeping it out of the CSS since it wasn't here before, but since it was in the spec, I added it.

I would also love to hear any feedback on whether we really need to be offering mods for all of these and whether the input and character count should really have different custom properties (is it useful to have `--spectrum-textfield-font-family` and also `--spectrum-textfield-character-count-font-family` for instance). We set fonts on the character count, placeholder, and input, and that all feels a little heavy-handed.

components/textfield/dist/metadata.json
```diff
@@ -132,21 +135,17 @@
...
     "--mod-textfield-icon-spacing-inline-end-invalid",
-    "--mod-textfield-icon-spacing-inline-end-quiet-invalid",
-    "--mod-textfield-icon-spacing-inline-end-quiet-valid",
     "--mod-textfield-icon-spacing-inline-end-valid",
     "--mod-textfield-icon-spacing-inline-start-invalid",
     "--mod-textfield-icon-spacing-inline-start-valid",
     "--mod-textfield-label-spacing-block",
-    "--mod-textfield-label-spacing-block-quiet",
     "--mod-textfield-label-spacing-inline-side-label",
+    "--mod-textfield-line-height",
```
rise-erpelding, line 143: Only using this for multiline, the single-line textfield does not allow mods.

components/textfield/index.css
```diff
@@ -613,12 +558,17 @@
 
 /*** Text Area ***/
 .spectrum-Textfield--multiline {
-	--spectrum-textfield-input-line-height: normal;
-
 	.spectrum-Textfield-input {
+		line-height: var(--mod-textfield-line-height, var(--spectrum-textfield-line-height));
 		min-inline-size: var(--mod-text-area-min-inline-size, var(--spectrum-text-area-min-inline-size));
 		min-block-size: var(--mod-text-area-min-block-size, var(--spectrum-text-area-min-block-size));
```
rise-erpelding, line 564: Does it feel ok to have these custom props named like this, or do we want to rename them to have "textfield" in the name, something more like `--spectrum-textfield-text-area-inline-size`

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
  */
 
 .spectrum-Textfield {
```
rise-erpelding, line 14: Most of this top selector is reorganization, with some light renaming, changing of values, plus removal of the quiet variant.

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
...
-
-	--spectrum-textfield-border-color: var(--spectrum-gray-500);
-	--spectrum-textfield-border-color-hover: var(--spectrum-gray-600);
-	--spectrum-textfield-border-color-focus: var(--spectrum-gray-800);
-	--spectrum-textfield-border-color-focus-hover: var(--spectrum-gray-900);
-	--spectrum-textfield-border-color-keyboard-focus: var(--spectrum-gray-900);
-
-	--spectrum-textfield-border-width: var(--spectrum-border-width-100);
+	--spectrum-textfield-border-width: var(--spectrum-border-width-200);
+	--spectrum-textfield-corner-radius: var(--spectrum-corner-radius-medium-size-medium);
```
rise-erpelding, line 22: Happy to change this from `corner-radius` to `border-radius` if others have a strong opinion about it

components/textfield/index.css
```diff
@@ -304,12 +274,8 @@
 		grid-column: 1 / span 1;
 	}
 
-	.spectrum-Textfield--quiet & {
-		margin-block-end: var(--mod-textfield-label-spacing-block-quiet, var(--spectrum-textfield-label-spacing-block-quiet));
-	}
-
 	.is-disabled & {
-		color: var(--spectrum-textfield-text-color-disabled);
+		color: var(--highcontrast-textfield-text-color-disabled, var(--mod-textfield-text-color-disabled, var(--spectrum-textfield-text-color-disabled)));
```
rise-erpelding, line 278: There were a few disabled selectors like this one where there was no `--mod` being offered, I added one back in since I didn't see a reason not to.

components/textfield/index.css
aramos-adobe: I think it's fine. Design definition has been 'corner-rounding' as of late so this change is consistent.

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
...
-	--spectrum-textfield-icon-spacing-inline-end-quiet-valid: var(--spectrum-field-edge-to-validation-icon-quiet);
+	--spectrum-textfield-icon-spacing-inline-start-valid: var(--spectrum-text-to-visual-100);
+	--spectrum-textfield-icon-spacing-inline-end-valid: var(--spectrum-field-edge-to-validation-icon-medium);
+	--spectrum-textfield-icon-spacing-block-valid: var(--spectrum-field-top-to-validation-icon-medium);
 
 	/* font styles */
 	--spectrum-textfield-font-family: var(--spectrum-sans-font-family-stack);
 	--spectrum-textfield-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-textfield-font-style: var(--spectrum-default-font-style);
+	--spectrum-textfield-font-size: var(--spectrum-font-size-100);
```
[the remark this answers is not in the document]
marissahuysentruyt: Personally, I wouldn't think we would want `--spectrum-textfield-font-family` AND `--spectrum-textfield-character-count-font-family`. Like you said, that feels too heavy-handed to me, but I don't have any other reason except that I can't see and instance where those 2 font families would be different, in the same component. That just seems odd in my mind.

components/textfield/index.css
marissahuysentruyt: Oh yeah...that's a tricky one. My personal preference is that I like what you have. It's very clear. Something about "textfield-text-area" is confusing to me. 

I wonder if we could refactor this to use mods at all? I'm thinking of meter, where the meter-specific properties (`--spectrum-meter-*`) are actually just `--mod-progress-bar-*` properties. Would that work at all? It might not- I haven't actually tried it yet.

https://github.com/adobe/spectrum-css/blob/spectrum-two/components/meter/index.css

components/textfield/index.css
marissahuysentruyt: Just to clarify- I am not supposed to see a difference then in the text field, when switching between locales, correct? I don't see that right now (all the line-heights for text field are set to `--spectrum-textfield-height`) and I believe that is expected.

marissahuysentruyt (changes requested):
Ah yes, the "healthy-chicken-clap" changeset! 😆 So glad we can revisit these and get them all up-to-date again!

I think overall, this is in really great shape still. The only real reason I'm "requesting changes" is to add the hover test case to the test file and maybe some documentation regarding the required state, but everything else I could probably see an argument for not addressing. 

A couple more questions I couldn't really find a good spot for: 
`displayLabel = false`: do you think that should be changed to `true` in the template.js args? I think the Figma component has the label displayed by default, so should we update that at all?

`required`: do you think we should add any documentation about the `required` attribute being added with the control? When a user toggles that required control, visually they don't see anything, so could we add something about the `required` attribute in the control description? We could add a required story too if you wanted. 🤷‍♀️ 

Nice work on this!

components/textfield/index.css
rise-erpelding: Correct, line height should be the same for textfield regardless of locale, but will differ for cjk in the textarea.

components/textfield/index.css
rise-erpelding: Hmmm...

Were you thinking something like this?

```css
.spectrum-Textfield--multiline {
	.spectrum-Textfield-input {
		line-height: var(--mod-textfield-line-height, var(--spectrum-textfield-line-height));
		--mod-textfield-min-width: var(--spectrum-text-area-min-inline-size);
		--mod-textfield-min-height: var(--spectrum-text-area-min-block-size);
        }
}
```

I realized looking through the code, then the spec, that there is no existing `--mod-textfield-min-height` because textfield doesn't have one. I'd lean against adding one because it's not in the spec. I'd also lean against something like:

```css
--mod-textfield-min-width: var(--spectrum-text-area-min-inline-size);
min-block-size: var(--mod-text-area-min-block-size, var(--spectrum-text-area-min-block-size));
```

Because mixing the setting of a mod with the setting of a style property feels weird? Thoughts? Did you have something else in mind that I'm not thinking of?

components/textfield/index.css
```diff
@@ -334,21 +305,25 @@
 	font-size: var(--mod-textfield-character-count-font-size, var(--spectrum-textfield-character-count-font-size));
 	font-family: var(--mod-textfield-character-count-font-family, var(--spectrum-textfield-character-count-font-family));
 	font-weight: var(--mod-textfield-character-count-font-weight, var(--spectrum-textfield-character-count-font-weight));
+	font-style: var(--mod-textfield-character-count-font-style, var(--spectrum-textfield-character-count-font-style));
+	line-height: var(--mod-textfield-character-count-line-height, var(--spectrum-textfield-character-count-line-height));
 	grid-row: 1;
 	grid-column: 2 / span 1;
```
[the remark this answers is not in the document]
rise-erpelding: I made some massive additions to the tests and thought it would be nice to add character count without label should that combination ever exist, this is what removing them does in that case, it shifts the character count over to the start side:

image

I wonder if it's worth keeping it to cover that situation? Or is there a better way to address it? Is it worthwhile to leave a comment in the code that that's what it's there for, maybe?

rise-erpelding:
@marissahuysentruyt 

> 
> A couple more questions I couldn't really find a good spot for: `displayLabel = false`: do you think that should be changed to `true` in the template.js args? I think the Figma component has the label displayed by default, so should we update that at all?
> 

I updated this!

> 
> `required`: do you think we should add any documentation about the `required` attribute being added with the control? When a user toggles that required control, visually they don't see anything, so could we add something about the `required` attribute in the control description? We could add a required story too if you wanted. 🤷‍♀️
> 

This was a good catch! I'd previously had this conversation with the designer and subsequently realized it was just an issue in the template where I wasn't passing that required prop down to FieldLabel. I'm not sure if I only fixed it in my head or if I made the change and it was lost in a rebase, but in any case we should have asterisks now! I also added a description for the control and made some "Required" stories. According to the guidelines, you can use the asterisk or add "(required)" into the label. IMO, I don't think it's worthwhile to test "(required)" or make it part of controls, but I did address it and show it on the Docs page, let me know if you think that's sufficient.

components/textfield/index.css
```diff
@@ -12,56 +12,74 @@
...
 
 	/* character count */
+	--spectrum-textfield-character-count-color: var(--spectrum-neutral-subdued-content-color-default);
 	--spectrum-textfield-character-count-font-family: var(--spectrum-sans-font-family-stack);
 	--spectrum-textfield-character-count-font-weight: var(--spectrum-regular-font-weight);
+	--spectrum-textfield-character-count-font-style: var(--spectrum-default-font-style);
+	--spectrum-textfield-character-count-font-size: var(--spectrum-font-size-100);
+	--spectrum-textfield-character-count-line-height: var(--spectrum-line-height-100);
+	--spectrum-textfield-character-count-line-height-cjk: var(--spectrum-cjk-line-height-100);
 	--spectrum-textfield-character-count-spacing-inline: var(--spectrum-spacing-200);
```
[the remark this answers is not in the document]
rise-erpelding: I'd maybe err on the side of leaving it, what do you think? Playing with it, I can see that the label is stretching appropriately when the text is longer:

image
image

marissahuysentruyt (changes requested):
I love the new icons- they're so fresh! This is well on its way- I only have very small questions and a small suggestion that we should add! (about the "(required)" 🥳🥳🥳

components/textfield/stories/textarea.template.js
```diff
@@ -78,28 +78,25 @@ export const TextAreaOptions = (args, context) => Container({
...
 	content: html`
 		${Container({
 			withBorder: false,
-			containerStyles: {
-				"gap": "8px",
-			},
-			heading: "Default",
-			content: Template({...args, isKeyboardFocused: true}, context)
+			heading: "Required with (required) label",
+			content: Template({...args, isRequired: true, isRequiredWithoutAsterisk: true, labelText: "Interests", value: "", helpText: "Describe the interests you'd like to explore through our tutorials."}, context),
```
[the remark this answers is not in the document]
rise-erpelding: Thanks! I don't love all the repetition between the two components, it's really easy to miss things like that.

components/textfield/index.css
```diff
@@ -12,51 +12,69 @@
...
+	--spectrum-textfield-character-count-spacing-block: var(--spectrum-component-bottom-to-text-100);
 	--spectrum-textfield-character-count-spacing-inline-side: var(--spectrum-side-label-character-count-to-field);
+	--spectrum-textfield-character-count-spacing-block-side: var(--spectrum-component-top-to-text-100);
 
 	/* focus indicator */
 	--spectrum-textfield-focus-indicator-width: var(--spectrum-focus-indicator-thickness);
 	--spectrum-textfield-focus-indicator-gap: var(--spectrum-focus-indicator-gap);
 
+	/*** Colors ***/
+	--spectrum-textfield-background-color: var(--mod-textfield-background-color, var(--spectrum-gray-25));
```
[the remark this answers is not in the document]
rise-erpelding: Oop, good catch, I skimmed over this one assuming that all the tokens would carry over from textfield, so I will ask.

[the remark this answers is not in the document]
rise-erpelding: I think I saw this updated in Figma to `gray-25` so this does not need any changing!

jawinn (commented):
This is looking really good. I took a first pass and started going through validation steps. I'll go through the rest next week.

Textfield:
- Docs: What do you think about making "Invalid" and "Invalid, Focused" their own invalid story instead of being under default? This feels like a better way to organize it, since Invalid is not default. It also could have Invalid, Keyboard Focused to be consistent with the Default story.
- Storybook controls that would be nice to have descriptions for: Valid, Invalid, Loading, Display field label, Loading (?)
- **Loading**: what's up with this storybook control? It shows the loader but aligns weirdly; it's not aligned well in prod either. I also don't see any documentation about this option. Is this something legacy that needs to be removed for S2, or something that needs to be added? 🤔 
Screenshot 2025-03-06 at 4 41 19 PM
- Chromatic template suggestion: it might be worth adding Disabled + Hovered to the tests as well, to ensure that continues to not show a color change on hover.

components/textfield/stories/textarea.stories.js
```diff
@@ -149,7 +152,7 @@ Sizing.parameters = {
  * displays the focus indicator. This indicator should not appear on focus from a click or tap.
  * The example below has this class applied on first load for demonstration purposes.
  */
-export const KeyboardFocus = KeyboardFocusTemplate.bind({});
+export const KeyboardFocus = Template.bind({});
```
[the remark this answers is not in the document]
marissahuysentruyt: @rise-erpelding Do you think it's better to keep the `KeyboardFocus` stories separated because they have some documentation regarding some implementation details, like the proper class name? Or could we move that documentation up with the default stories for textfield & textarea?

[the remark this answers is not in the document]
rise-erpelding: Also realizing that we need to take `:focus-visible` out to keep the focus outline from appearing on mouse focus, following the work done in #2306.

rise-erpelding:
Latest updates should address the most recent comments/changes requested! They include:
* Docs page:
    * Removing the keyboard focused story and combining its comments into the default-level story.
    * Removing invalid states from the default-level story and making them their own story.
    * Minor reformatting of JSDoc comments, including addition of line breaks for all long comments (there had already been a few with line breaks, so this adds consistency), removal of curly quotes in favor of straight quotes, and addition of references to textfield from textarea and vice versa.
    * Removal of wrapper and container spacing styles, stories with multiple containers displayed in row rather than column direction
* Storybook controls
    * Hiding isLoading control because it looks unexpected and is not in the design spec (I will create a ticket for this)
    * Add descriptions for select controls
* Testing template: add disabled & hover tests
* CSS: removal of `:focus-visible` selectors as they cause the focus outline to appear on mouse focus

marissahuysentruyt (dismissed):
Thanks for resurrecting this one and getting it all updated! I think things are looking good from my perspective! 

computer-kid-approuve

jawinn (dismissed):
Nice work on this. It's super close to approval for me.

**Generic selectors question**:
Lets see what the consensus on this is. Hopefully it won't prove to be too difficult to implement considering how it would change the specificity of a lot of selectors.

**High contrast:**
Looking at the component in high contrast mode in AssistivLabs Edge, there are a few differences that I'm not expecting.

1. Two of the disabled inputs are showing a different background color (that oddly looks like the ButtonFace color; purple in this color scheme). I'd expect these to have the same background color.
2. Focus + hover and Read-only is showing a different border color. I'd expect focused to show the highlight color, and read-only to look like the default.
3. It looks like an automatic text readability backplate is being rendered below the text. This is visible on the Disabled + Hover and Disabled.

Screenshot 2025-03-20 at 3 16 30 PM

components/textfield/dist/metadata.json
```diff
@@ -1,59 +1,68 @@
...
+    ".is-disabled:hover .spectrum-Textfield.is-readOnly .spectrum-Textfield-input",
+    ".is-disabled:hover .spectrum-Textfield.is-readOnly .spectrum-Textfield-input::placeholder",
+    ".is-disabled:hover .spectrum-Textfield.is-readOnly:hover .spectrum-Textfield-input",
+    ".is-disabled:hover .spectrum-Textfield.is-readOnly:hover .spectrum-Textfield-input::placeholder",
     ".is-focused .spectrum-Textfield-input",
     ".is-focused .spectrum-Textfield-input::placeholder",
     ".is-focused .spectrum-Textfield-input:hover",
     ".is-focused .spectrum-Textfield-input:hover::placeholder",
+    ".is-focused:hover .spectrum-Textfield-input",
+    ".is-focused:hover .spectrum-Textfield-input::placeholder",
```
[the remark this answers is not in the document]
marissahuysentruyt: Looks like Rise scoped all of this generic classes with `.spectrum-Textfield` 👍

components/textfield/index.css
```diff
@@ -611,24 +616,17 @@
...
+		.spectrum-Textfield-input::placeholder {
+			--highcontrast-textfield-text-color-valid: CanvasText;
+			--highcontrast-textfield-text-color-invalid: CanvasText;
+			--highcontrast-textfield-text-color-default: CanvasText;
+			--highcontrast-textfield-text-color-hover: CanvasText;
+			--highcontrast-textfield-text-color-focus: CanvasText;
+			--highcontrast-textfield-text-color-focus-hover: CanvasText;
+			--highcontrast-textfield-text-color-keyboard-focus: CanvasText;
+			--highcontrast-textfield-text-color-disabled: GrayText;
+			--highcontrast-textfield-text-color-readonly: CanvasText;
```
rise-erpelding, line 629: Consolidating because placeholder and input value styles should be the same

components/textfield/index.css
```diff
@@ -345,11 +345,15 @@
 	grid-row: 2;
 	grid-column: 1 / span 2;
 
+	/* prevents unexpected color changes for border in high contrast mode */
+	forced-color-adjust: none;
```
rise-erpelding, line 349: Forces the input to use the styles from CSS and not the default, so that we won't get that border flash. But since WHCM is now relying on the styles set for the input, we have to set the highcontrast background color or it turns white in WHCM.

components/textfield/index.css
```diff
@@ -76,6 +86,8 @@
 	--spectrum-textfield-text-color-readonly: var(--spectrum-neutral-content-color-default);
 
 	/* Disabled Colors */
+	--spectrum-textfield-background-color-disabled: var(--highcontrast-textfield-background-color-disabled, var(--mod-textfield-background-color-disabled, var(--spectrum-gray-25)));
+	--spectrum-textfield-border-color-disabled: var(--highcontrast-textfield-border-color-disabled, var(--mod-textfield-border-color-disabled, var(--spectrum-disabled-border-color)));
```
marissahuysentruyt, line 90: So this disabled border color set according to the specs, but I think the read-only styles are overriding it somewhere. When I toggle the disabled control, the entire textfield disappears.

components/textfield/index.css
```diff
@@ -494,48 +506,29 @@
...
-		}
-	}
-
 	/****** Input ReadOnly ******/
 	.spectrum-Textfield.is-readOnly &,
 	.spectrum-Textfield.is-readOnly:hover &,
 	&:read-only {
 		background-color: transparent;
-		border-color: transparent;
+		border-color: var(--highcontrast-textfield-border-color-readonly, transparent);
```
marissahuysentruyt, line 514: Maybe we need some `:not(:is-disabled)` or something here?

components/textfield/index.css
marissahuysentruyt: https://github.com/user-attachments/assets/78762092-92f0-4922-9505-b6bc22773354

components/textfield/index.css
```diff
@@ -439,39 +443,46 @@
...
 		padding-inline-end: calc(var(--mod-textfield-icon-spacing-inline-start-invalid, var(--spectrum-textfield-icon-spacing-inline-start-invalid)) + var(--mod-textfield-icon-size-invalid, var(--spectrum-textfield-icon-size-invalid)) + var(--mod-textfield-icon-spacing-inline-end-invalid, var(--spectrum-textfield-icon-spacing-inline-end-invalid)) - var(--mod-textfield-border-width, var(--spectrum-textfield-border-width)));
 		color: var(--highcontrast-textfield-text-color-invalid, var(--mod-textfield-text-color-invalid, var(--spectrum-textfield-text-color-invalid)));
 		border-color: var(--highcontrast-textfield-border-color-invalid-default, var(--mod-textfield-border-color-invalid-default, var(--spectrum-textfield-border-color-invalid-default)));
 	}
 
 	/* Invalid hover */
-	.is-invalid:hover:not(.is-disabled) &,
-	.is-invalid &:hover:not(.is-disabled) {
+	.spectrum-Textfield.is-invalid:hover:not(.is-disabled) &,
+	.spectrum-Textfield.is-invalid &:hover:not(.is-disabled) {
```
marissahuysentruyt, line 462: Now that we have the `is-hover` class, do we need to add that selector here at all?  Invalid + Disabled + hover looks right on the Chromatic tests, but then if I actually hover over it with a cursor, I get the border color to change.

Or maybe it just needed to change a bit: .spectrum-Textfield.is-invalid:not(.is-disabled)&:hover? I might not be in the selector...

marissahuysentruyt: https://github.com/user-attachments/assets/58d52fe7-a734-443a-a112-df8ab11d5d1d

components/textfield/dist/metadata.json
```diff
@@ -65,37 +35,68 @@
...
     ".spectrum-Textfield.is-disabled .spectrum-FieldLabel",
+    ".spectrum-Textfield.is-disabled .spectrum-HelpText .spectrum-HelpText-text",
+    ".spectrum-Textfield.is-disabled .spectrum-Textfield-characterCount",
     ".spectrum-Textfield.is-disabled .spectrum-Textfield-input",
     ".spectrum-Textfield.is-disabled .spectrum-Textfield-input::placeholder",
+    ".spectrum-Textfield.is-disabled .spectrum-Textfield-input:read-only::placeholder",
     ".spectrum-Textfield.is-disabled .spectrum-Textfield-validationIcon",
+    ".spectrum-Textfield.is-disabled .spectrum-Textfield.is-readOnly .spectrum-Textfield-input::placeholder",
+    ".spectrum-Textfield.is-disabled .spectrum-Textfield.is-readOnly:hover .spectrum-Textfield-input::placeholder",
+    ".spectrum-Textfield.is-disabled.is-readOnly:hover .spectrum-Textfield-input",
```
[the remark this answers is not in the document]
rise-erpelding: could've sworn I staged and committed the metadata.json 🤷‍♀️

components/textfield/index.css
marissahuysentruyt: @rise-erpelding I noticed a similar thing with the invalid+disabled+otherStates. When I toggled random controls, I happened to notice the first one, and then dug some more (with the controls and the keyboard). I'm decently confident that your CSS is right, and that this is just a side effect of/issue with storybook controls, so feel free to disregard! I'm going to approve anyways, since I can't actually recreate them outside of the storybook controls 🤞 

Invalid+Disabled+Focus (I think this might just be a storybook only thing- I can't actually click or tab to this)
Screenshot 2025-04-01 at 2 23 29 PM

Invalid+Disabled+KeyboardFocus (same here- I think this is just storybook-only, where the controls create something that doesn't actually trigger/happen with a mouse/keyboard)
Screenshot 2025-04-01 at 2 16 25 PM

Invalid+Disabled or Readonly+FocusHover or KeyboardFocus (same with these)
Screenshot 2025-04-01 at 2 12 07 PM
Screenshot 2025-04-01 at 2 20 30 PM

marissahuysentruyt (approved):
thumbs-up-good-job

Thanks for sticking with this one! Let's get it merged!

jawinn (approved):

PR #2861, opened 2024-06-24
fix(dialog): implement dialog button groups

marissahuysentruyt:
## Description
Currently on the [static docs site](https://opensource.adobe.com/spectrum-css/dialog.html), the dialogs don't really follow the guidelines and documentation according to the [internal/beta site](https://spectrum-contributions.corp.adobe.com/page/standard-dialog-beta/#Dismissible): 

> When not dismissible, the dialog needs to have one or more buttons in the footer for users to perform an action (including closing the dialog) and the background overlay cannot be clicked.

This PR implements the missing button groups on non-dismissible dialogs on the docs page. It also adds/removes classes in `dialog.yml` to implement the button groups.

### Jira
[CSS-815](https://jira.corp.adobe.com/browse/CSS-815) 

### Validation steps

- [ ] Pull down the branch
- [ ] Run `yarn run dev` and visit [the dialog docs page](http://localhost:3000/dialog.html)
- [ ] Verify all dialog variants are closed on page load
- [ ] Test all closed dialogs on the page: 
    - opening the dialog should add the `is-open` class to `spectrum-Modal` and `spectrum-Modal-wrapper`
    - once the dialog is open, any buttons (like the close button or a button group) should be clickable/focusable
    - ensure the small, medium and large dialogs now have a `ButtonGroup`
    - closing the dialog with one of the new buttons should remove `is-open` from those same modal elements 

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/dialog/metadata/dialog.yml
+++ b/components/dialog/metadata/dialog.yml
@@ -14,13 +14,22 @@ examples:
     markup: |
       <button class="spectrum-Button spectrum-Button--sizeM spectrum-Button--outline spectrum-Button--staticWhite spectrum-CSSExample-overlayShowButton" onclick="openDialog(this.nextElementSibling)"><span class="spectrum-Button-label">Open Small Dialog</span></button>
 
-      <div class="spectrum-Modal-wrapper spectrum-CSSExample-dialog">
-        <div class="spectrum-Modal is-open">
+      <div class="spectrum-Modal-wrapper spectrum-CSSExample-modal">
+        <div class="spectrum-Modal">
           <div class="spectrum-Dialog spectrum-Dialog--small" role="dialog" tabindex="-1" aria-modal="true">
             <div class="spectrum-Dialog-grid">
               <h1 class="spectrum-Dialog-heading">Disclaimer</h1>
               <hr class="spectrum-Divider spectrum-Divider--sizeM spectrum-Divider--horizontal spectrum-Dialog-divider">
               <section class="spectrum-Dialog-content">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Auctor augue mauris augue neque gravida. Libero volutpat sed ornare arcu. Quisque egestas diam in arcu cursus euismod quis viverra. Posuere ac ut consequat semper viverra nam libero justo laoreet. Enim ut tellus elementum sagittis vitae et leo duis ut. Neque laoreet suspendisse interdum consectetur libero id faucibus nisl. Diam volutpat commodo sed egestas egestas. Dolor magna eget est lorem ipsum dolor. Vitae suscipit tellus mauris a diam maecenas sed. Turpis in eu mi bibendum neque egestas congue. Rhoncus est pellentesque elit ullamcorper dignissim cras lobortis.</section>
+
+              <div class="spectrum-ButtonGroup spectrum-Dialog-buttonGroup spectrum-Dialog-buttonGroup--noFooter">
+                <button class="spectrum-Button spectrum-Button--sizeM spectrum-Button--outline spectrum-Button--secondary spectrum-ButtonGroup-item" type="button" onclick="closeDialog(this.closest('.spectrum-Modal-wrapper'))">
+                  <span class="spectrum-Button-label">Cancel</span>
+                </button>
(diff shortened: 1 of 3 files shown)


components/dialog/metadata/dialog.yml
```diff
@@ -262,9 +289,9 @@ examples:
     demoClassName: spectrum-CSSExample-example--overlay
     markup: |
       <button class="spectrum-Button spectrum-Button--sizeM spectrum-Button--outline spectrum-Button--staticWhite spectrum-CSSExample-overlayShowButton" onclick="openDialog(this.nextElementSibling)"><span class="spectrum-Button-label">Open Fullscreen Takeover Dialog</span></button>
-      <div class="spectrum-Modal-wrapper">
+      <div class="spectrum-Modal-wrapper spectrum-CSSExample-modal">
         <div class="spectrum-Modal spectrum-Modal--fullscreenTakeover" data-testid="modal">
-          <section class="spectrum-Dialog spectrum-Dialog--fullscreenTakeover" role="alertdialog" tabindex="-1" aria-modal="true">
+          <section class="spectrum-CSSExample-dialog spectrum-Dialog spectrum-Dialog--fullscreenTakeover" role="alertdialog" tabindex="-1" aria-modal="true">
```
jawinn, line 294: Do these two examples need the class `spectrum-CSSExample-dialog` added? I had seen that this example class adds a bit of CSS. I disabled those CSS declarations and it still appears the same.

jawinn (changes requested):
Nice work fixing the display of those button groups in the example docs!

components/dialog/metadata/dialog.yml
marissahuysentruyt: You might be right! I was just trying to use the class consistently, but it doesn't need it with the z-index fixed.

jawinn (changes requested):
One last thing I saw. The buttons in the button group don't have a gap between them because the docs page isn't loading the button group CSS. 
Screenshot 2024-06-26 at 11 51 51 AM


I think we need to adjust the dependencies. It should load if buttongroup is added to `devDependencies`.

components/dialog/metadata/dialog.yml
rise-erpelding: It does seem a little funny not to use `.spectrum-CSSExample-dialog` in the Dialog component, but keep it around for styling wrappers in other places such as [alert dialog](https://github.com/adobe/spectrum-css/blob/main/components/alertdialog/metadata/alertdialog.yml#L10) and [menu](https://github.com/adobe/spectrum-css/blob/main/components/menu/metadata/menu.yml#L1351) as well as others.

But, being that we're planning to leave the docs site behind, it may not be worthwhile to look into it.

components/dialog/metadata/dialog.yml
```diff
@@ -14,13 +14,22 @@ examples:
     markup: |
       <button class="spectrum-Button spectrum-Button--sizeM spectrum-Button--outline spectrum-Button--staticWhite spectrum-CSSExample-overlayShowButton" onclick="openDialog(this.nextElementSibling)"><span class="spectrum-Button-label">Open Small Dialog</span></button>
 
-      <div class="spectrum-Modal-wrapper spectrum-CSSExample-dialog">
-        <div class="spectrum-Modal is-open">
+      <div class="spectrum-Modal-wrapper spectrum-CSSExample-modal">
+        <div class="spectrum-Modal">
```
castastrophe, line 18: Curious, why remove is-open?

marissahuysentruyt:
> I think we need to adjust the dependencies. It should load if buttongroup is added to `devDependencies`.

Good catch- totally overlooked that. It's added to `devDependencies` now: https://github.com/adobe/spectrum-css/pull/2861/commits/ba0831bb0f650d8ad5549684ca4a6207893fd81e

Screenshot 2024-07-01 at 10 38 12 AM

components/dialog/metadata/dialog.yml
marissahuysentruyt: Having `is-open` on page load wasn't adding the underlay styles or preventing a user from clicking on things "below" or "under" the dialog's z-index. Removing it is more steps for the user to see that dialog, but keeps the full-fledged styles & behaviors in place when the dialog is triggered (and that `is-open` gets added).

https://github.com/adobe/spectrum-css/assets/69602589/e33ed9fe-2d04-46dc-b301-245b537ccd64

I could probably refactor the `enhancements.js` file for the dialog, but thought this was a quicker fix, and didn't want to spend too much time in there if we're moving away from the docs site.

jawinn (approved):

PR #2883, opened 2024-07-01
docs(popover, progresscircle, radio, rating, sidenav): site docs to storybook

jawinn:
## Description

This continues the migration work to move the documentation from the deprecated docs site to Storybook, covering the components Popover, Radio, Rating, and Sidenav. Variants shown on the docs site for these components have been added to Storybook as new stories if they did not already exist. No custom MDX files were needed for these components.

### Component specific changes and notes:
- **Popover**
  - Docs notes have been moved to relevant stories where possible, and many have been reworded to make them easier to understand. All positions are within one Docs story instead of being spread out amongst different examples.
  - Adds test coverage of dialog style popover found in docs (and SWC)
- **Radio**
  - Separates disabled, read-only, and emphasized options out into their own Docs stories for the user's view, as single radios would not have those options mixed in a group of radios.
  - Also includes a connected fix for **Field group**, to include a missing `role="radiogroup"` and to make use of `aria-readonly` for radios.
- **Rating**
  - Minor cleanup; adjusts some stories and controls to improve documentation and make sure that any docs site documentation is represented.
  - Fix: Disabled in Storybook was not showing filled stars
  - Fix: Wrong focused class was being applied. Focus ring now shows in story and Chromatic template.
- **Sidenav**
  -  Migrates some of the story descriptions from the docs page, and adjusts story content to clarify what is being displayed
  - Adds additional description to document disabled and selected.
  - Fixes some logic for the display of icons and documents this better. Icons should now appear on all first level nav items when the associated control is true.
 
### Other overall Storybook updates:

- **feat: maintain unique name and id args in** `Sizes()` `ArgGrid`
  Radios displayed with the Sizes() template were no longer grouped and showing a selected radio within in group, because they all had the same name. This addition to the ArgGrid() function makes sure that the id and name args used for attributes have a unique string appended when the template is rendered multiple times within the grid.

### Validation steps

- [x] [Popover](https://pr-2883--spectrum-css.netlify.app/preview/?path=/docs/components-popover--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/popover.html)
- [x] [Radio](https://pr-2883--spectrum-css.netlify.app/preview/?path=/docs/components-radio--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/radio.html)
- [x] [Rating](https://pr-2883--spectrum-css.netlify.app/preview/?path=/docs/components-rating--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/rating.html)
- [ ] Rating fix - disabled shows filled stars
- [ ] [Sidenav](https://pr-2883--spectrum-css.netlify.app/preview/?path=/docs/components-side-nav--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/sidenav.html)

### Regression testing

Some expected VRT changes:
- Additional first level icons appearing in sidenav
- Some updated text in sidenav items
- Fixed display of disabled stars in rating (stars should appear)
- Addition of dialog style content to existing popover Default story

Validate:

## To-do list


rise-erpelding (changes requested):
This was a HUGE effort! I think this is pretty close but I have a few component-specific callouts in addition to the comments I left:

## Popover
This is a great improvement! This documentation is really nice and includes all the information from the Docs site, but it works much more nicely and is clearer than the Docs site, plus little things like the vertical transition on open/close work now. ⭐ 

I occasionally saw a hairline/border on the tip within the tip story (not on Docs), but it wasn't consistent. It would disappear sometimes, and I only saw it on my monitor (not my laptop screen). I didn't notice this with the same story on prod. It might be worth investigating and maybe making a bug card for?
Screenshot 2024-07-01 at 2 04 49 PM

### Popover Controls
* There are no controls for nested - this is intentional, right? Would we want to remove controls for the “Positioning options” story too? Do we want to keep that story (with the grid view) in the sidebar? It looks ✨amazing✨ but it keeps crashing my browser tab if I open the story (however it works fine in the Docs page). Maybe keep it on the Docs page and for Chromatic?
* Portions of the "Positioning" options depend on having the tip visible. For instance, without the tip visible, there's no difference between "top-left" and "top-right". This is not a huge deal, but it is a little confusing.

## Rating
This was probably just about the first time I’ve ever looked at this component, and I didn’t realize that if there is a rating selected, there’s an indicator only under the selected (and not disabled/read only) rating if the component is hovered. I think the way the docs site has the selected/not selected variants communicates in a useful way, and Storybook would benefit from having some of those selected/unselected variants, for example Default and Default selected, Disabled and Disabled selected.

## Side nav
> Make sure to use the appropriate variant for the context and user needs. Don’t mix behavior, styles, or variations together in a single navigation menu.

Do you think this is important to retain from the docs site? I didn’t see it communicated in Storybook.

components/radio/stories/radio.stories.js
```diff
@@ -33,7 +37,7 @@ export default {
 			type: { name: "string" },
 			table: {
 				type: { summary: "string" },
-				category: "Component",
+				category: "Content",
```
marissahuysentruyt, line 40: It might be nice to have a `description` here, clarifying that "Name" is referring to the `name` attribute on the input. What do you think?

components/radio/stories/radio.stories.js
```diff
@@ -91,31 +95,132 @@ export default {
 	},
 };
 
-export const Default = (args) => html`
+// ===== Story specific templates ======
+
+// @todo Radio needs more robust coverage of all states, including hover, focus, etc.
+const ChromaticGroupTemplate = (args) => html`
```
marissahuysentruyt, line 101: I already left a similar comment on potentially moving all the markup to template.js. We could do that here too.

components/radio/stories/radio.stories.js
```diff
@@ -91,31 +95,132 @@ export default {
...
+Disabled.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * A radio group has a read-only option for when it's in the disabled state but still needs to be shown.
+ * This allows for content to be copied, but not interacted with or changed.
+ * 
+ * - Read-only radio buttons are disabled, but not all disabled radio buttons are read-only.
+ * - Read-only radio buttons do not have a focus ring, but the button should be focusable.
```
marissahuysentruyt, line 215: I don't understand this line here- the read-only radio buttons don't have a button, right?

Screenshot 2024-07-10 at 11 13 36 AM

components/radio/stories/radio.stories.js
```diff
@@ -91,31 +95,132 @@ export default {
...
+	name: "standard",
+};
+Standard.tags = ["!dev"];
+
+export const Sizing = BasicGroupTemplate.bind({});
+Sizing.args = {
+	name: "sizing",
+};
+Sizing.tags = ["!dev"];
+Sizing.decorators = [withSizingWrapper];
```
marissahuysentruyt, line 183: Do either of these stories need `chromatic: disabledSnapshot: true`? I could see an argument for both remaining unchanged and keeping their snapshots enabled, since the chromatic group doesn't have sizes in it, or an example of wrapping text. Just curious.

components/radio/stories/radio.stories.js
```diff
@@ -91,31 +95,132 @@ export default {
...
+		})}
+	</div>
+`;
+
+// ===== Stories ======
+
+export const Default = (args) => {
+	return window.isChromatic() ? ChromaticGroupTemplate(args) : BasicGroupTemplate(args);
+};
+Default.tags = ["!autodocs"];
```
marissahuysentruyt, line 170: Aaahhh ok I see what you did here. This default story is the one with chromatic attached, but then you've removed it from the autodocs while keeping it in the sidenav. 

Then for the "Standard" story (which is the same as the Default when `isChromatic` is false), you removed it from the sidenav. 

This might be how we get around some of the docs exceptions you were trying to figure out in 2885 right? 🤔

marissahuysentruyt (changes requested):
These components are looking much better! All of the additions are much needed and really amp up how useful storybook can be. And I love it all getting into one place. I did find a few things I think need to be tweaked, and then I mainly left you questions. 👏 👏 👏

Really small, random note: want to add progress circle to the title of this PR?

components/popover/stories/popover.stories.js
```diff
@@ -128,77 +136,118 @@ export default {
...
+			<div style="padding: 16px;">${Story(context)}</div>
 		`,
 	],
 };
 
+/**
+ * Play function that triggers a click on all buttons within the canvas.
+ * Used for initially triggering the appearance of Popover beside the Source button,
+ * that includes the positioning logic that happens on click.
+ */
```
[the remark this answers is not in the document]
jawinn: This was a regular function, so it's fine to use this standard convention for multi-line JavaScript comments. The only reason we need to be careful with them is for the stories themselves, where JS doc descriptions automatically will turn into the user visible description. This is refactored away with the rebase though.

components/popover/stories/popover.stories.js
```diff
@@ -262,3 +312,135 @@ Nested.args = {
...
+ * 
+ * This example uses the Menu component within the popover.
+ */
+export const Standard = Template.bind({});
+Standard.storyName = "Default";
+Standard.args = {
+	isOpen: true,
+	trigger: () => null,
+};
+Standard.tags = ["!dev"];
```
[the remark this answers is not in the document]
jawinn: Since autodocs is included by default, I think it makes sense to not add `autodocs` to every story--- I dug into this again to see _where_ the default is, and it appears that we should make a configuration change that adds the `autodocs` tag to the global settings; it replaced our `autodocs: true` [setting and that setting is going away in Storybook 9.](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#mainjs-docsautodocs-is-deprecated) I'll make a PR for that!

components/popover/stories/popover.stories.js
```diff
@@ -128,77 +136,118 @@ export default {
...
+/**
+ * A popover can be nested within another popover.
+ */
 export const Nested = Template.bind({});
-Nested.play = async ({ canvasElement }) => {
-	const canvas = within(canvasElement);
-	await userEvent.click(canvas.getAllByRole("button")[0]);
-	await userEvent.click(canvas.getAllByRole("button")[1]);
-};
+Nested.play = playToggleButtonClick;
```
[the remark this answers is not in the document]
jawinn: These play functions were removed on main with the Modes work, so those are now removed after the rebase.

components/popover/stories/popover.stories.js
```diff
@@ -34,13 +35,19 @@ const placementOptions = [
...
+ * - Popover's position and distance to its source should be handled by the implementation. Positioning in Storybook is only for demonstration purposes.
+ * - When the `.is-open` class is present, popover is offset from the source by the spacing value defined in `--spectrum-popover-animation-distance`. This
+ * offset is done with a CSS transform and animates with a CSS transition. 
  */
 export default {
 	title: "Popover",
 	component: "Popover",
 	argTypes: {
 		trigger: { table: { disable: true } },
+		triggerId: { table: { disable: true } },
```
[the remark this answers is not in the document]
jawinn: This was just documenting an arg that already existed. It's connected to the trigger/source button and how it toggles the popover open/closed currently.

components/popover/stories/template.js
```diff
@@ -71,13 +71,25 @@ export const Template = ({
...
 					if (position.startsWith("top")) {
 						y = rect.top - popHeight;
-						yOffset = withTip
-							? "- (var(--spectrum-popover-pointer-height) + var(--spectrum-popover-animation-distance) - 1px)"
-							: "- var(--spectrum-popover-animation-distance)";
+						if (isOpen) {
+							yOffset = withTip
+								? "- (var(--spectrum-popover-pointer-height) - 1px)"
+								: "";
+						}
```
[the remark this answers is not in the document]
jawinn: I've dropped this animation fix unfortunately, because it no longer worked after rebase and because we have a followup issue to refactor the positioning.

components/popover/stories/popover.stories.js
```diff
@@ -128,77 +136,118 @@ export default {
...
-				return "Changes side with text direction (like a logical property)";
-			if (option.startsWith("left") || option.startsWith("right"))
-				return "Text direction does not affect the position";
-			return null;
-		};
+/**
+ * Template that displays a Popover with every value of the "position" option.
+ */
+const TipPlacementVariants = (args) => html`
+	<div style="display: grid; gap: 32px; grid-template-columns: repeat(auto-fit, minmax(150px, 1fr)); max-width: 1000px;">
```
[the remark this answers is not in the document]
jawinn: Thanks. All template functions are moved to the `template.js` files now and imported in the stories files.

components/popover/stories/popover.stories.js
```diff
@@ -229,11 +278,12 @@ Nested.args = {
 			triggerId: "trigger-nested-2",
 			isOpen: true,
 			customStyles: {
-				"margin-inline-start": "136px",
+				// @todo This manual test of positioning does not support "Large" scale.
+				"margin-inline-start": "73px",
```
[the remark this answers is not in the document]
jawinn: Yep! I've included a screenshot of that bug to handle in the followup refactor issue CSS-845.

components/radio/stories/radio.stories.js
```diff
@@ -1,9 +1,13 @@
 import { html } from "lit";
 
+import { withSizingWrapper } from "@spectrum-css/preview/decorator";
 import { Template } from "./template";
 
 /**
  * A radio selector allow users to select a single option from a list of mutually exclusive options. All possible options are exposed up front for users to compare.
+ * 
+ * Radio buttons should not be used on their own, they should always be used within a [Field group](?path=/docs/components-field-group--docs). Invalid radio buttons are signified by including
+ * [Help text](?path=/docs/components-help-text--docs) in a Field group. Sample usage for an [invalid radio state](?path=/docs/components-field-group--docs#invalid) is included in the Field group documentation.
```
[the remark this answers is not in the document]
jawinn: I've updated that, and also reached out to the content team for some guidance. It looks like sticking to lowercase for now is best, unless I hear something different after a meeting with them next week.

components/radio/stories/radio.stories.js
jawinn: Good idea. I've added that control description.

components/radio/stories/radio.stories.js
jawinn: I've left these migrated comments for now because after trying to harmonize the differences between what's there and what's documented, it revealed a bigger issue with our read-only states (and within SWC). It was unclear what the correct behavior should be. Updating this documentation is a part of CSS-833: _reevaluate different implementations' use of read-only._

components/radio/stories/radio.stories.js
jawinn: After rebasing with Modes, I decided to disable the snapshots on these for now, and have added a "todo" about implementing the Variants() utility function with increased coverage (similar to what we saw was needed for Checkbox).

jawinn:
Thanks for the thorough reviews! I'm almost done resolving all of the comments here, after that big rebase. I'll re-request a review once they're finished.

To your comments Rise:

> ## Popover
> I occasionally saw a hairline/border on the tip within the tip story (not on Docs), but it wasn't consistent. It would disappear sometimes, and I only saw it on my monitor (not my laptop screen). I didn't notice this with the same story on prod. It might be worth investigating and maybe making a bug card for?

I'll make a bug card for that. I think it's reproducible if you resize the browser height; it happens on every other (odd?) pixel?

> ### Popover Controls
> * There are no controls for nested - this is intentional, right? Would we want to remove controls for the “Positioning options” story too? Do we want to keep that story (with the grid view) in the sidebar? It looks ✨amazing✨ but it keeps crashing my browser tab if I open the story (however it works fine in the Docs page). Maybe keep it on the Docs page and for Chromatic?

I've removed it from the sidebar, since you were having that performance issue. It's probably a lot faster and less problematic now that it's only loading text content and not a Menu in each one, but I think keeping it off the sidebar is probably best.

> ## Rating
> This was probably just about the first time I’ve ever looked at this component, and I didn’t realize that if there is a rating selected, there’s an indicator only under the selected (and not disabled/read only) rating if the component is hovered. I think the way the docs site has the selected/not selected variants communicates in a useful way, and Storybook would benefit from having some of those selected/unselected variants, for example Default and Default selected, Disabled and Disabled selected.

I also just realized that the underline not appearing on the star you hover is a feature and not a bug (highlighting the currently selected star value).

This component definitely needs extra coverage for the hovered with underline state, and could have all its stories combined into one Chromatic-only template, though we were trying to exclude that refactoring from this work, which is why I kept this as a light touch. I've added a todo about condensing and increasing coverage. I don't think we need to show empty stars and some filled stars as separate stories as it's a bit duplicative, if you were suggesting that?

Looking at this did surface a bug in our template that I've fixed in the latest commit; disabled was not showing filled stars! Two classes were being excluded for disabled and readonly that shouldn't have been. Also of note, there is an existing bug related to the hovering in the backlog, CSS-735.

> ## Side nav
> "Make sure to use the appropriate variant for the context and user needs. Don’t mix behavior, styles, or variations together in a single navigation menu." --- Do you think this is important to retain from the docs site? I didn’t see it communicated in Storybook.

I left this out, as it doesn't seem very helpful and is somewhat inaccurate in our context. For example, it's saying not to mix variations, but you can show headings while also showing icons. Where this sentence originated is already covered with more explanation [in the guidelines](https://spectrum.adobe.com/page/side-navigation/#Using-the-right-option).

components/progresscircle/stories/progresscircle.stories.js
```diff
@@ -58,10 +59,32 @@ const ProgressCircleGroup = (args) => html`
...
+	chromatic: { disableSnapshot: true },
+};
+
 export const StaticWhite = ProgressCircleGroup.bind({});
+StaticWhite.storyName = "Static white, indeterminate";
 StaticWhite.args = {
 	staticColor: "white",
+	isIndeterminate: true,
 };
+StaticWhite.decorators = [withSizingWrapper];
```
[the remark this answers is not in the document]
jawinn: Thanks for pointing that one out. This is fixed. I had to make adjustments to how the sizing decorator was updating the args.

components/radio/stories/radio.stories.js
```diff
@@ -91,31 +95,132 @@ export default {
...
+			name: (typeof args.name != "undefined" ? args.name + "-" : "") + "test-5",
 		})}
 	</div>
 `;
+
+const BasicGroupTemplate = (args) => html`
+	<div style="display: flex; flex-direction: column; align-items: flex-start;">
+		${Template({
+			...args,
+			label: "Example label",
```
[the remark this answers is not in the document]
jawinn: I've updated the Default story that appears in Storybook to only have the single radio template, instead of the two radios template; the control to change the label now works with that. The two radio template has set labels to include testing of wrapping text.

components/popover/stories/popover.stories.js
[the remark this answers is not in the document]
jawinn: A PR is up for that: #2924

rise-erpelding (approved):
Thanks for working through all the feedback, Josh! This looks great!

To follow up on a couple of things:
* For Popover, I'm going to note here that hairline bug is definitely visible on prod too, so not a bug introduced here as I'd thought!
* Rating looks good with the changes! I can see how having both unselected and selected can be a little duplicative, and I think it's fine not to have both on the Docs page.

components/sidenav/stories/template.js
```diff
@@ -30,11 +30,13 @@ export const Template = ({
...
-              html`<h2 class="${rootClass}-heading" id="${item.id}-heading">${item.heading}</h2>`
-              :
-              html`
-              <a class="${rootClass}-itemLink">
+            ${item.heading
+              ? html`<h2 class="${rootClass}-heading" id="${item.id}-heading">${item.heading}</h2>`
+              : html`
+              <a
+                class="${rootClass}-itemLink"
+                aria-current=${ifDefined(item.isSelected ? "page" : undefined)}
```
marissahuysentruyt, line 38: great additional with the `aria-current`! 👏 👏 👏

components/rating/stories/template.js
```diff
@@ -65,9 +65,8 @@ export const Template = ({
 					<span
 						class=${classMap({
 							[`${rootClass}-icon`]: true,
-							"is-selected": !isDisabled && idx <= value - 1,
-							"is-currentValue":
-								!isDisabled && !isReadOnly && idx === value - 1,
+							"is-selected": idx <= value - 1,
+							"is-currentValue": idx === value - 1,
```
marissahuysentruyt, line 69: What was the purpose of having the `isDisabled` and `isReadOnly` args? I don't quite understand the `idx` bits, so that might be part of my problem.

Also (because I couldn't drop a comment on the correct lines)- I don't know that `readonly` is available to range inputs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#additional_attributes

marissahuysentruyt (commented):
I think you took care of all of our previous feedback, so now I just looked through everything with the the modes work & some future work we know is happening in mind! Great work!

components/popover/stories/popover.stories.js
```diff
@@ -199,3 +319,49 @@ WithForcedColors.parameters = {
 		modes: disableDefaultModes
 	},
 };
+
+// ********* DOCS ONLY ********* //
```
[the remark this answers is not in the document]
jawinn: Both of those are not docs-only, as they are not excluded from VRTs.

components/popover/stories/popover.stories.js
```diff
@@ -118,12 +128,29 @@ Default.args = {
...
 
-export const WithTip = Variants.bind({});
+export const WithTip = Template.bind({});
+WithTip.storyName = "Default with tip";
 WithTip.args = {
+	...Default.args,
 	withTip: true,
 };
+// This is excluded from the Docs page because the toggle interactions do not work correctly there.
+WithTip.tags = ["!autodocs"];
```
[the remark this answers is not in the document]
jawinn: I've gone ahead and removed the "WithTip" story and have marked nested as `!dev`, which matches with the change on that PR.

components/radio/stories/radio.stories.js
```diff
@@ -87,52 +91,87 @@ export default {
...
+Standard.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+export const Sizing = BasicGroupTemplate.bind({});
+Sizing.args = {
+	name: "sizing",
+};
+Sizing.tags = ["!dev"];
+Sizing.decorators = [withSizingWrapper];
```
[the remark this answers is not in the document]
jawinn: I've added a description including info about medium as the default.

components/rating/stories/rating.stories.js
```diff
@@ -72,17 +77,13 @@ export default {
 export const Default = Template.bind({});
```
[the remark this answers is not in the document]
jawinn: I've added some docs to the main component description and the Default story description to cover this:

"All active stars have the class is-selected applied. Additionally, the current value (the last active star) has the class is-currentValue applied."

and

"A initial value of three is used for the following examples, to demonstrate both active and inactive stars."

components/rating/stories/template.js
jawinn: I've added some docs to the main component description that explains the use of those two classes.

Good callout on the `readonly` attribute being used in this template. I've added a commit that removes that, since it is not supported in HTML. That's popped up in a couple components now.

marissahuysentruyt (approved):
Let's get this merged! Thanks for all your hard work on this!

.storybook/decorators/helpers.js
```diff
@@ -70,3 +70,16 @@ export function fetchContainers(id, isDocs = false) {
 
     return containers;
 }
+
+/**
+ * Object for mapping "size" arg values to their display names.
+ */
+export const sizeDisplayNames = {
```
castastrophe, line 77: This isn't necessary if we use the `Sizes` utility available as an export from `@spectrum-css/preview/decorators`; another reason to use Sizes directly is so that we use consistent grid spacing and layout and typography is loaded in a unified way (instead of loading Typography into components).

.storybook/decorators/sizing.js
```diff
@@ -2,56 +2,74 @@ import { makeDecorator } from "@storybook/preview-api";
 import { html } from "lit";
 import { styleMap } from "lit/directives/style-map.js";
 
+import { Template as Typography } from "@spectrum-css/typography/stories/template.js";
```
castastrophe, line 5: I think we should deprecate the sizing decorator for now and move toward using the Sizes utility. In the future, for docs pages, I think we could make use of a custom doc block for Sizes that will give us a very consistent UX for docs pages.

.storybook/preview.js
```diff
@@ -53,9 +53,9 @@ export const parameters = {
 				"Guides",
 				["Contributing", "*", "Adobe Code of Conduct", "Changelog"],
 				"Components",
-				["*", ["Docs", "Default", "*"]],
+				["*", ["Docs", "Default", "Standard", "Sizing", "*"]],
```
castastrophe, line 56: In this listing, how does Standard differ from Default? Those seems like synonyms to me.

components/popover/stories/popover.stories.js
```diff
@@ -188,14 +202,153 @@ Nested.args = {
...
+	isOpen: true,
+	trigger: () => null,
+	content: [
+		() => Dialog({
+			showModal: false,
+			// @todo replace custom class with 'size: "small"' arg when added Dialog control is merged.
+			customClasses: ["spectrum-Dialog--small"],
+			isDismissable: false,
+			heading: "Example heading",
+			content: [
```
castastrophe, line 224: Might be a todo for later but it seems like the Dialog content should automatically wrap in the Typography component. If we push it through the `renderContent` utility exported from `@spectrum-css/preview/decorators`? We might need to update that function to use Typography styles if the content entry is a string?

components/popover/stories/template.js
```diff
@@ -179,19 +180,21 @@ export const Template = ({
...
-			"align-items": "flex-start",
-		})} class="spectrum-Typography">
+		<div
+			style=${styleMap({
+				"display": "grid",
+				"gap": "16px",
+				"grid-template-columns": "repeat(auto-fit, minmax(232px, 1fr))",
+				"max-width": "1000px",
+			})} 
+			class="spectrum-Typography"
```
castastrophe, line 196:
```suggestion
			class="spectrum-Typography"
```

I don't recommend invoking the spectrum-Typography class manually; this should come from the `Typography` template, but I'm also not sure it's valuable to add it here either.

castastrophe (changes requested):
A few changes noted in progresscircle and popover. I can review the rest at a later date but I'm seeing a few overlaps there too.

.storybook/decorators/sizing.js
jawinn: The `Sizes()` utility is now used in the latest rebased work, instead of the sizing decorator.

components/popover/stories/template.js
jawinn: This is removed.

components/popover/stories/popover.stories.js
jawinn: I'm not sure if this completely still applies after the latest set of changes. Let me know if you think there is something that needs updating for this PR.

.storybook/decorators/helpers.js
jawinn: This commit was dropped.

jawinn:
I've finished reworking these changes after the rebase with the changes to `main`. Popover required a bit of work with the stories and templates to get everything displaying again.
- The sizing decorator is no longer used, and the `Sizes()` utility is used instead.
- I included one small change to the Sizes ArgGrid to help fix the issue with radio buttons having the same name within the Sizing example. Previously this was causing only one radio to be selected at a time, instead of having one selected within each size group. This also makes sure label `for` attributes match up with unique IDs.
- I found and fixed another issue with the Sidenav logic around icon. Icons should now appear on all first level nav items when the associated control is true.

components/sidenav/stories/sidenav.stories.js
```diff
@@ -93,28 +104,28 @@ Multilevel.args = {
 			levelTwoItems: [
 				{
 					id: "section2.1",
-					title: "Section title 1",
+					title: "Section title A",
```
marissahuysentruyt, line 107: Do you want to add any of these `items` changes to the test file for Chromatic? I'm not sure it really makes a difference honestly, I was just thinking. Since we don't have `stateData` for the test cases, maybe it'd be nice to have some of the `current/selected` language you added for the multilevel or heading stories. 

Screenshot 2024-08-28 at 2 03 56 PM

components/sidenav/stories/template.js
```diff
@@ -24,19 +24,22 @@ export const Template = ({
       ...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
     })}>
       ${repeat(items, (item) => item.id, (item) => {
+        // First level nav item with second tier of nav items beneath.
```
marissahuysentruyt, line 27: Great comments!! very helpful.

components/progresscircle/stories/progresscircle.stories.js
```diff
@@ -61,17 +62,60 @@ WithForcedColors.tags = ["!autodocs", "!dev"];
...
+	isIndeterminate: true,
+};
+Indeterminate.tags = ["!dev"];
+Indeterminate.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+export const StaticWhiteDocs = Sizing.bind({});
+StaticWhiteDocs.tags = ["!dev"];
+StaticWhiteDocs.storyName = "Static white, indeterminate";
```
marissahuysentruyt, line 114: Should we also have a non-indeterminate static white story for the docs page? The old docs site features both. What do you think?

Your current `StaticWhite` captures everything (like `Default`) from the test file, so I think that one would have to stay as is. But we could do a little refactoring to have a `StaticWhiteIndeterminateDocs` and `StaticeWhiteDocs`.

components/popover/stories/popover.test.js
```diff
@@ -116,11 +120,38 @@ export const PopoverGroup = Variants({
 				testHeading: `Position: ${position}`,
 				wrapperStyles,
 				position,
-				withTestContainer: true,
 				popoverHeight,
 				popoverWidth,
 			};
 		}),
+		{
+			testHeading: "Dialog style content",
```
marissahuysentruyt, line 128: Want to add a `Nested` test?

marissahuysentruyt (changes requested):
This was another beast of a rebase!! Thanks for sticking with this one! I did note a couple places where we could probably remove older custom templates we were making in favor of the new utilities that have merged. I have a few more

For sidenav- did we need to move this migration note on the docs site to the changelog or is it already there? (I didn't check!) 
"Core token migration work and design updates added a span tag wrapping the link text."

For the `position: left` and possibly the `position: start` & `end` popover tests: should we add `wrapperStyles["align-items"] = "center";? The `position: right` case has that same line. I think that would fix this view up: 

Screenshot 2024-08-28 at 2 39 54 PM

Those are the only 3 that didn't have a specified `wrapperStyles["align-items"]` bit. And there might be a reason we don't need to add that, but then do we remove the same line from `position: right`?

components/radio/stories/radio.stories.js
```diff
@@ -86,13 +92,14 @@ export default {
...
 		actions: {
 			handles: ["click input[type=\"radio\"]"],
-
 		},
 		componentVersion: version,
 	},
 };
 
-export const Default = RadioGroup.bind({});
+export const Default = (args, context) => context.viewMode === "docs"
```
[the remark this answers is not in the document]
jawinn: I had it as a separate "Standard" story previously like your example, but changed to how it is now in order to keep this as the story named "Default" (while also having the "Default" in the sidebar which has the single radio instead of the two radios). A previous comment mentioned the use of "Standard" naming as a synonym for "Default". And this seemed like the simplest way to use the same story name. @castastrophe do you have a preference here?

components/sidenav/stories/sidenav.stories.js
```diff
@@ -76,8 +83,12 @@ WithForcedColors.parameters = {
 };
 
 // ********* DOCS ONLY ********* //
+
+/**
+ * Use a multi-level side navigation when there are multiple layers of hierarchical navigation.
+ */
 export const Multilevel = Template.bind({});
-Multilevel.tags = ["!dev"];
```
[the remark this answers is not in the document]
jawinn: Yeah, I re-enabled them in the sidebar since I saw they couldn't be created with the controls.

components/radio/stories/radio.stories.js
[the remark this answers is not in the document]
marissahuysentruyt: Could we just change the display name of `Standard`? `Standard.storyName = "Default"`?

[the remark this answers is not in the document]
jawinn: I wasn't sure from the Storybook docs whether storyName should be unique; there was some conflicting info, but I think it can be. I'll try putting it back to that.

components/radio/stories/template.js
```diff
@@ -48,9 +53,76 @@ export const Template = ({
...
+			isChecked: true,
+			label: "Initially selected radio button that has wrapping label text",
+			customStyles: {
+				"max-width": "220px",
+			},
+			id: "radio-2-" + (args?.id ?? "default"),
+			name: "radio-example-" + (args?.name ?? "default"),
+		}, context)}
+	</div>
+`;
```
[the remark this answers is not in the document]
jawinn: I've refactored this to use `Container()` for the wrapper in this template.

components/rating/stories/template.js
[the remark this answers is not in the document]
jawinn: Good find on that. I've updated the incorrect class so the focus ring now appears. I also added some docs to the usage notes about how it should be applied. Note: it won't appear on the Docs page annoying because of `updateArgs` not functioning the same there, but it will appear on the Default story and the Chromatic template you screenshotted.

components/progresscircle/stories/progresscircle.stories.js
jawinn: I've added another story for it.

components/popover/stories/popover.test.js
jawinn: It looks liked Nested is already tested through the separate story in Chromatic. I think that makes sense since it's a special case that's manually positioned and not being tested with all the test case combinations.

components/sidenav/stories/sidenav.stories.js
jawinn: I've updated the test case `items[]` to match these changes to `items[]` in the stories.

PR #2885, opened 2024-07-02
fix: chromatic testing preview refresh issue and docs exception

jawinn:
### Storybook: fixes the refresh issue with the chromatic testing preview toolbar item, and excludes Chromatic-only templates from the autodocs
    
Previously when the Chromatic "view testing preview" toolbar item was used, the story would not refresh and display the correct templates until the browser was hard refreshed.
    
This appears to be partly because the custom window.isChromatic function was not working with the dynamic global values (globals.testingPreview).
    
This update makes sure that when the decorator is run, it's working with the current value of `globals.testingPreview`. It also passes through a boolean `isChromaticView` to the story args, which should remove the need to use the window object. 

The `window.isChromatic()` function is left in for the time being for backwards compatibility and consideration of current migrations, but can be refactored away (noted in a `@todo` comment). For this PR I updated **Action button** and **Pagination** as an example of this. I could go through and modify all usage of `window.isChromatic()` as part of this PR if desired; I limited the scope of this in order to get feedback on this first.
    
✨ The function also now excludes Chromatic-only templates from the Docs / autodocs pages by adding a check against `context.viewMode`.

CSS-832

### Validation steps

- [x] Toolbar item in Storybook to view the Chromatic templates ("Show testing preview" and "Default mode") will now refresh the story when it is used. Without a browser refresh. @marissahuysentruyt 
- [x] Chromatic-only templates are not appearing on the docs / autodocs pages (whether "show testing preview" is active or inactive). @marissahuysentruyt 
- [x] Chromatic-only and regular templates are working correctly for Action button and Pagination (which now access `args.isChromaticView`) @marissahuysentruyt 
- [x] Chromatic-only and regular templates are working correctly for other components. Please go through and test a bunch of other components. @marissahuysentruyt  (tested: `accordion`, `picker`, `statuslight`, `coachindicator`, `tray`, `progresscircle`, `menu`, `illustratedmessage`)
- [ ] VRT is working normally. There are no VRT changes from this PR.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.storybook/decorators/testing-preview.js
+++ b/.storybook/decorators/testing-preview.js
@@ -23,7 +23,7 @@ export const withTestingPreviewWrapper = makeDecorator({
 		// Add a boolean to the context that indicates whether the Chromatic view is enabled.
 		context.args.isChromaticView = isChromaticViewEnabled(testingPreview, viewMode === "docs");
 
-		// Backwards compatability with existing use of this custom function.
+		// Backwards compatibility with existing use of this custom function.
 		// @todo This can be removed from window after changing everywhere it is accessed to use args.isChromaticView instead.
 		window.isChromatic = () => context.args.isChromaticView;
 
--- a/.storybook/decorators/testing-preview.js
+++ b/.storybook/decorators/testing-preview.js
@@ -1,9 +1,10 @@
-import { makeDecorator, useEffect } from "@storybook/preview-api";
+import { makeDecorator } from "@storybook/preview-api";
 import isChromatic from "chromatic/isChromatic";
 
 /**
  * @type import('@storybook/csf').DecoratorFunction<import('@storybook/web-components').WebComponentsFramework>
- * @description Lets you preview the Chromatic testing view locally
+ * @description Lets you preview the Chromatic testing view locally. Sets a boolean for context.isChromaticView
+ * 	to be used for conditional rendering within story templates.
  **/
 export const withTestingPreviewWrapper = makeDecorator({
 	name: "withTestingPreviewWrapper",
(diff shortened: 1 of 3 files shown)


jawinn:
After looking at the VRT results, I'm looking into why the Chromatic templates are not displaying for Chromatic.

jawinn:
That issue is resolved and now this is ready for review.

.storybook/decorators/testing-preview.js
```diff
@@ -1,9 +1,10 @@
-import { makeDecorator, useEffect } from "@storybook/preview-api";
+import { makeDecorator } from "@storybook/preview-api";
```
marissahuysentruyt, line 1: Was the `useEffect` somewhat causing the issue where we have to hard refresh to get the updated testing preview value? Is that why it was deleted?

marissahuysentruyt (changes requested):
@jawinn Nice work on this! I'm so excited for this update! 

Looks like things are working as expected and how you outlined in the description 🎉

https://github.com/adobe/spectrum-css/assets/69602589/1c6b61e3-6a40-4c63-bfce-dbdd92e4525e

.storybook/decorators/testing-preview.js
jawinn: It was no longer needed/used after the fix and refactor. It doesn't look like there is any need to do any side effects outside of the normal rendering.

marissahuysentruyt (approved):
It looks like there might be a chromatic test for you to look at for popover. I'm thinking this is the same thing we've seen in other branches/work (I don't see a difference when I open them in tabs and flick back and forth).

This is approved in my eyes, but I'll leave accepting the popover "change" to you, just in case you see something I don't.

jawinn:
Both issues are now resolved in main.

PR #2886, opened 2024-07-03
fix(breadcrumb): disabled behaviors and styles

marissahuysentruyt:
## Description

Disabled breadcrumbs (except for action buttons) in storybook still had pointer events, hover and focus styles. There was no indication that a text-based breadcrumb was disabled. This was partially because the `tabindex` was being set to `0`, which then was adding specific styles. Additionally, the `is-disabled` class was not being added to the proper `.spectrum-Breadcrumb-itemLink` element. This PR should address these a11y concerns and disabled bugs.

### Validation steps

@jawinn 

- [x] Verify that the disabled breadcrumb on the [breadcrumb storybook page](https://pr-2886--spectrum-css.netlify.app/preview/?path=/docs/components-breadcrumbs--docs) has no hover/focus styles, and no pointer events. Additionally, the breadcrumb item should be left out of the tab order
- [x] Verify that the parent `.spectrum-Breadcrumbs-item` does not have the `is-disabled` class
- [x] With the project running locally, (`yarn start`), find the breadcrumb storybook docs or default page
- [x] In `breadcrumb/stories/breadcrumb.stories.js`, add an additional disabled breadcrumb item (or feel free to experiment with ones already there). In the browser inspector, verify the following:
    - [x] when `isDisabled: false`, the `tabindex` of your `.spectrum-Breadcrumbs-itemLink` breadcrumb is `0`. It should be focusable, clickable, and have a hover state. The `is-disabled` class should not be applied at this point and there should not be a `disabled` attribute
    - [x] when `isDisabled: true`, the `tabindex` should not be defined on that element. It should not be focusable, clickable, or have a hover state. The `is-disabled` class and `aria-disabled="true"` attribute should be added to `.spectrum-Breadcrumbs-itemLink`

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/breadcrumb/stories/template.js
+++ b/components/breadcrumb/stories/template.js
@@ -1,7 +1,7 @@
 import { html } from "lit";
 import { classMap } from "lit/directives/class-map.js";
 import { when } from "lit/directives/when.js";
-
+import { ifDefined } from "lit/directives/if-defined.js";
 import { Template as ActionButton } from "@spectrum-css/actionbutton/stories/template.js";
 import { Template as Icon } from "@spectrum-css/icon/stories/template.js";
 
@@ -28,7 +28,6 @@ export const Template = ({
 				return html` <li
 					class=${classMap({
 						[`${rootClass}-item`]: true,
-						"is-disabled": isDisabled,
 						"is-dragged": isDragged && item.isDragged,
 					})}
 				>
@@ -48,9 +47,13 @@ export const Template = ({
 								idx !== arr.length - 1,
 								() =>
 									html`<div
-										class="${rootClass}-itemLink"
+										class=${classMap({
+											[`${rootClass}-itemLink`]: true,
+											"is-disabled": isDisabled,
+										})}
+										aria-disabled=${ifDefined(isDisabled ? "true" : undefined)}
 										role="link"
-										tabindex="0"
+										tabindex=${ifDefined(!isDisabled ? "0" : undefined)}
 									>
 										${label}
 									</div>`,
--- a/.changeset/modern-ladybugs-wash.md
+++ b/.changeset/modern-ladybugs-wash.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/breadcrumb": patch
+---
+
+Ensures disabled breadrumb items behave as expected. They are conditionally left out of tab order depending on their `isDisabled` value. Uses [aria-disabled="true"] and applies `is-disabled` class to `.spectrum-Breadcrumb-itemLink`.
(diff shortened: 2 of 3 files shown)


components/breadcrumb/stories/template.js
```diff
@@ -48,9 +47,13 @@ export const Template = ({
 								idx !== arr.length - 1,
 								() =>
 									html`<div
-										class="${rootClass}-itemLink"
+										class=${classMap({
+											[`${rootClass}-itemLink`]: true,
+											"is-disabled": isDisabled,
+										})}
+										?disabled=${isDisabled}
```
jawinn, line 54:
```suggestion
										aria-disabled=${ifDefined(isDisabled ? "true" : undefined)}
```
The `disabled` attribute does not apply to a div. On the docs site, this has `aria-disabled`, so using that instead would match the existing markup. _Though I wonder why a [disabled anchor tag](https://www.scottohara.me/blog/2021/05/28/disabled-links.html) is not being used instead of a div in the original markup._

components/breadcrumb/stories/template.js
```diff
@@ -48,9 +47,13 @@ export const Template = ({
...
 										role="link"
-										tabindex="0"
+										tabindex="${isDisabled ? "" : 0}"
```
jawinn, line 56:
```suggestion
										tabindex=${ifDefined(!isDisabled ? "0" : undefined)}
```
So the attribute does not render for non-disabled.

jawinn (changes requested):

components/breadcrumb/stories/template.js
marissahuysentruyt: Ah yes, this is what I was trying to do! (get it to not render) I forgot about `ifDefined`. Thanks!!

components/breadcrumb/stories/template.js
marissahuysentruyt: Yes I wondered about the `role="link"` divs too.

For now, I [changed `disabled` in the template to `aria-disabled`](https://github.com/adobe/spectrum-css/pull/2886/commits/55c1d5f688a8808e023f3e53d07cccf0fc70c0a6). What's our preference on keeping that `aria-disabled` attribute selector in the CSS? Or should I just revert[ that commit ](https://github.com/adobe/spectrum-css/pull/2886/commits/f023f57b643a1f2a50415c6fbc7b232684a8700f) and keep the CSS untouched?
```
&.is-disabled,
&[aria-disabled] // I haven't actually made this change yet
```

jawinn: I think adding a selector for aria-disabled might be a nice addition, instead of [disabled]. `aria-disabled` can equal false, so make sure that's `&[aria-disabled="true"]

marissahuysentruyt: Lovely! here's the change: https://github.com/adobe/spectrum-css/pull/2886/commits/bf9b82eaa8f45364595657f0a4b8b9e4c93c4538

jawinn (approved):
Looks good. I would just add a changeset with a patch bump noting the CSS addition.

pfulton (approved):
Thanks for taking care of this one!

PR #2912, opened 2024-07-16
fix(storybook): snapshot padding and minor prep for s2

castastrophe:
### New features for @spectrum-css Storybook

- Migrate test functions to stories/<component>.test.js
- Harden the testing grid utilities in the decorators by adding detailed comments and tightening up grid styles
- Rename .storybook/decorators/overlay.js to underlay.js (which is what it is)
- Expand test coverage

Review the updated baselines [here](https://www.chromatic.com/review?appId=64762974a45b8bc5ca1705a2&number=17&type=unlinked&view=changes).

### Story updates

- Note that ids now have a default fallback value using the random generator if not provided

### Regression testing

Validate:

## To-do list


components/menu/stories/menu.stories.js
```diff
@@ -211,9 +206,9 @@ TraySubmenu.argTypes = {
 };
 TraySubmenu.args = {
 	selectionMode: "multiple",
-	customStyles: {
-		"--mod-menu-inline-size": "100%",
-	},
+	// customStyles: {
```
pfulton, line 209: Should we just delete this `customStyles` object now?

components/actionbar/stories/actionbar.stories.js
```diff
@@ -80,12 +80,10 @@ export default {
 
 export const Default = ActionBarGroup.bind({});
 Default.args = {
-	customPopoverStyles: {
-		// Prevent the popover from being rendered offscreen
-		"--spectrum-popover-animation-distance": "-10px",
-		"inline-size": "600px",
-		"max-inline-size": "100%",
-	}
+	// customPopoverStyles: {
```
pfulton: Should we delete these commented out blocks now or were we keeping them around for when we revisit popover?

components/actionbar/stories/actionbar.stories.js
```diff
@@ -104,12 +102,10 @@ export const Emphasized = Template.bind({});
 Emphasized.tags = ["autodocs", "!dev"];
 Emphasized.args = {
 	isEmphasized: true,
-	customPopoverStyles: {
-		// Prevent the popover from being rendered offscreen
-		"--spectrum-popover-animation-distance": "-10px",
-		"inline-size": "600px",
-		"max-inline-size": "100%",
-	}
+	// customPopoverStyles: {
```
pfulton: Should we delete these commented out blocks now or were we keeping them around for when we revisit popover?

pfulton (commented):
I think everything looks good here. I left a few questions that are not blockers for merging—just wondering about keeping blocks of commented out code around.

components/menu/stories/menu.stories.js
castastrophe: Good point, I've updated this

components/actionbar/stories/actionbar.stories.js
castastrophe: Great minds!  I had just deleted this after your review.

.changeset/giant-rats-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/popover": patch
+---
+
+Remove internal-only --flow-direction variable
```
pfulton, line 5: Thanks for this!

pfulton (approved):
I think things look good here.

components/popover/stories/template.js
jawinn, line 1: The Popover changes feel like they should be their own PR. There is a lot changing and to validate.

With no longer using absolute positioning, this causes display issues in other component stories that are using Popover when an element is clicked. I happened to look at Coach mark. I'm not sure what other components may be affected:
Screenshot 2024-07-31 at 12 04 37 PM

We could possibly look at removing the on click open toggle entirely. What do you think about that?

On the Default story, we also aren't able to preview the transition animation on open/close (should be the opacity transition and also the translateY from 0 to the popover-animation-distance as seen [here](https://opensource.adobe.com/spectrum-css/popover.html)). I thought that was one reason for continuing to have the on click.

I also saw that the positioning chromatic view has some issues in RTL:
Screenshot 2024-07-31 at 2 10 52 PM .
Is it necessary to show the source button at all in this chromatic template, since the positioning is not a part of the component CSS? And is it necessary to show a Menu within each one, instead of the basic text content used before?

jawinn (changes requested):

jawinn (commented):
I looked over all the changed files and this is looking really good. The only main things I saw, as noted in my review, were the changes in display for `Modal` and a few situations where the refactored `Popover` positioning causes some issues.

.storybook/package.json
```diff
@@ -58,7 +58,6 @@
 		"lit": "^3.1.3",
 		"lodash-es": "^4.17.21",
 		"postcss": "^8.4.40",
-		"postcss-pseudo-classes": "^0.4.0",
```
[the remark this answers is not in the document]
castastrophe: There is no postcss configuration that is local to the .storybook directory anymore; the config is now being inherited from the root postcss.config so only the root-level dependencies are used rather than needing any storybook-scoped dependencies.

.storybook/preview.js
```diff
@@ -37,8 +37,8 @@ setConsoleOptions({
 });
 
 /** @type import('@storybook/types').StorybookParameters & import('@storybook/types').API_Layout */
-export const parameters = {
-	layout: "padded",
+const parameters = {
+	layout: "centered",
```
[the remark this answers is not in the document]
castastrophe: Yes, it improved how elements filled their containers as I was reviewing the changes.

components/icon/stories/icon.stories.js
```diff
@@ -123,71 +115,58 @@ WithForcedColors.parameters = {
...
-				})}
-			>
-				${uiIconSizes[iconName.replace(/(Left|Right|Up|Down)$/, "")]?.map((iconSize) =>
-					Template({ ...args, setName: "ui", iconName: iconName + iconSize })
-				)}
-				${when(uiIconSizes[iconName]?.length == 0, () =>
-					Template({ ...args, setName: "ui", iconName })
-				)}
-			</div>`
+const IconListTemplate = (args, iconsList = [], context) => html`
```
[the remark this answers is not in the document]
castastrophe: For icon I tried to minimize the work being done for icon because it's so integral. I would definitely like to clean it up in its own PR soon.

components/modal/stories/modal.stories.js
```diff
@@ -31,26 +34,44 @@ export default {
...
-		(story) => window.isChromatic() ? html`<div style="width: 1200px; height: 800px;">${story()}</div>` : story()
+		withUnderlayWrapper,
 	],
 };
 
 export const Default = ModalGroup.bind({});
-Default.args = {};
+Default.args = {
+	content: [
+		(passthroughs, context) => Dialog({
```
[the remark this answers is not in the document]
castastrophe: > We're also missing testing of the variants ("Sizing preference"). And changing this to fullscreen currently does not display correctly with the Dialog

We weren't testing those variants to begin with and the complexity of supporting them didn't make sense based on the goals for this PR. We can add a backlog story to add those later?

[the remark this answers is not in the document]
jawinn: Adding coverage for sizing preference can definitely be done in a followup. But I don't think we should be displaying and testing Dialog here instead of just Modal.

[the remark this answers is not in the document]
castastrophe: Modal doesn't snapshot well without something inside of it - do you have a suggestion for what to put inside Modal to render it in a way we can test?

components/popover/stories/template.js
castastrophe: Going to resolve this as I did some rework to simplify and harden the popover experience in demo & VRT mode so we'll need to re-review those aspects.

components/modal/stories/modal.stories.js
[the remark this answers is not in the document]
jawinn: In #2854 I added an option to display the Underlay like Dialog does. That way the background and rounded corners are visible.

castastrophe:
With some of these pivots after code review, the previously accepted baselines are making it challenging so I chose to close this PR and open a fresh one to see if I can trigger a fresh compare against main: https://github.com/adobe/spectrum-css/pull/2982

PR #2915, opened 2024-07-18
fix(storybook): some inline stories in docs cutting off

jawinn:
## Description

Adds some CSS to Storybook to adjust inline stories on the Docs pages. This stops them from cutting off content, by adjusting the overflow property. This was affecting components such as in-field button, color loupe, etc.

The display difference between regular stories and inline stories is because [Storybook sets an inline style with `overflow: auto` in its InlineStory function](https://github.com/storybookjs/storybook/blob/535f7b9ba3849719e10e4b577f78660b77cdcaec/code/ui/blocks/src/components/Story.tsx#L80C63-L80C74). This appears to be originally added in Storybook 6.1 to help fixed elements from leaking out; that does not appear to be an issue with our fixed elements after this fix (is it possibly not an issue anymore in Storybook due to other changes?).

CSS-816

### Validation steps

- [x] The following components are not cut off on their Docs pages:
  - Action bar
  - Color area
  - Color handle (previously was cutting off 3/4 of the circle)
  - Color loupe
  - Color slider
  - Dial
  - Drop indicator
  - Floating action button
  - In-field button
  - Tooltip
- [x] Components with `position: fixed` still display the same in Docs and their story: action bar, alert dialog, modal, tray, underlay

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.storybook/assets/base.css
+++ b/.storybook/assets/base.css
@@ -34,7 +34,7 @@ svg:has(symbol):not(:has(use)) {
 	display: none;
 }
 
-/* set back to default h2 styles */
+/* Make sure heading provided by <Stories/> doc block uses styles consistent with the default H2. */
 #variants {
 	font-size: 24px;
 	margin-block-end: 8px;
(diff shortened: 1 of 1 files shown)


rise-erpelding (approved):
Thanks for digging into this and figuring out what was causing the cutoff! I looked over all the components that use fixed position and don't see any issues with those, and I can see that the cutoff issue is resolved in all the components that were experiencing it before, so I think this fix is good to go!

pfulton (approved):
This looks much better. Thanks for digging into this!

castastrophe (approved):

PR #2925, opened 2024-07-23
docs(swatch,swatchgroup,table,tabs): docs migrations to storybook

marissahuysentruyt:
## Description

This PR continues migrating documentation from the static docs site into Storybook. The focus is on `swatch`, `swatchgroup`, `table` and `tabs`. All variants of the components should now be displayed on the Storybook `Docs` page. All components have `autodocs` and/or `!dev` stories now. Migration notes for `tables` and `tabs` have been added to the corresponding `changelog`s.

[Jira](https://jira.corp.adobe.com/browse/CSS-793)

No MDX pages were added for any components.

**This PR doesn't need a changeset since it's docs-only.**

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [ ] The [swatch storybook docs](https://pr-2925--spectrum-css.netlify.app/preview/?path=/docs/components-swatch--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/swatch.html)
- [ ] Verify all variants have been covered and render when using [the Chromatic testing preview](https://pr-2925--spectrum-css.netlify.app/preview/?path=/story/components-swatch--default&globals=testingPreview:!true)
- [ ] The [swatch group storybook docs](https://pr-2925--spectrum-css.netlify.app/preview/?path=/docs/components-swatch-group--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/swatchgroup.html).
- [ ] Verify all variants have been covered and render when using [the Chromatic testing preview](https://pr-2925--spectrum-css.netlify.app/preview/?path=/story/components-swatch-group--default&globals=testingPreview:!true)
- [ ] The [table storybook docs](https://pr-2925--spectrum-css.netlify.app/preview/?path=/docs/components-table--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/table.html)
- [ ] Verify all variants have been covered and render when using [the Chromatic testing preview](https://pr-2925--spectrum-css.netlify.app/preview/?path=/story/components-table--default&globals=testingPreview:!true)
- [ ] The [tabs storybook docs](https://pr-2925--spectrum-css.netlify.app/preview/?path=/docs/components-tabs--docs) include all necessary info when compared to the [corresponding docs site page](https://opensource.adobe.com/spectrum-css/tabs.html)
[...]

PR #2933, opened 2024-07-25
fix(swatch): prepares swatch for new stories

marissahuysentruyt:
- adds the background property for .is-image class for gradient and/or image fills
- adds z-index for mixedValueIcon so it sits "on top" of background color

### Regression testing

Validate:

## To-do list


components/swatch/index.css
```diff
@@ -150,6 +150,11 @@
 	/* Swatch fill: Image, Gradient, SVG */
 	&.is-image {
 		.spectrum-Swatch-fill {
+			/* Undefined variables allows custom stylesheet or JS to pass the value to this element */
```
rise-erpelding, line 153:
```suggestion
			/* Undefined variables allow custom stylesheet or JS to pass the value to this element */
```

rise-erpelding (commented):
Thanks for walking me through the changes with your validation instructions! This looks pretty good but I had a few follow-up questions/comments:
* I can't use the controls to switch the gradient on/off or adjust the gradient colors because they weren't added as `argTypes`. The stories also display in the sidebar, and I'm not sure that's necessarily what we want to do. But these are temporary changes added in the validation anyway, so I'm not sure if that's relevant here, but I wanted to note it anyway in case the same code is being used in a different PR.
* Everything seems to work as expected, except that I don't see the opacity checkerboard in the background if I set gradient stops to transparent. Is there a reason the `background` property isn't set on the `::before` element since that's where the `background-color` was set before?
Screenshot 2024-07-30 at 7 42 48 AM

marissahuysentruyt:
During review, we found I was misusing the `is-image` class in #2925, and completely missing `.spectrum-Swatch-image`. After correcting the template in that branch, these CSS changes were no longer necessary.

PR #1501, opened 2022-08-18
feat(swatch)!: migrate swatch to core tokens

yosevu:
## Description

- Update windows high contrast mode
- Update YAML examples
  - Rename `--spectrum-picked-color` to `--mod-swatch-fill-background-color`
  - Add `is-image` class to Image, Gradient, and Mixed Value swatches for transparent backgrounds 

Note: There are currently [two open questions](https://adobedesign.slack.com/archives/G019JTYMT6H/p1660685017664239) about border colors and the XS workflow icon size in Slack, however, these may not be blockers for this work.

## How and where has this been tested?
- Tested locally referencing the Swatch XD file.
- Tested WHCM with forced colors in Chrome Emulation
- Medium and Large scales

## To-do list

- [x] If my change impacts other components, I have tested to make sure they don't break.
- [x] If my change impacts documentation, I have updated the documentation accordingly.
- [x] I have read the [CONTRIBUTING document](/.github/CONTRIBUTING.md).
- [x] I have tested these changes in Windows High Contrast mode.

- [ ] This pull request is ready to merge.

PR #2954, opened 2024-07-30
feat(swatch): add support for lightBorder

marissahuysentruyt:
## Description

During the documentation migration work in #2925, it was noticed that the "light border" variant didn't have corresponding CSS. After some digging, I found that we did support it at one point, but it may have accidentally gotten deleted. When reading the Slack messages linked in #1501, it sounds like a light border variant should have been supported, but I believe that was one of the "open questions" related to that PR. This PR re-implements the modifier class `.spectrum-Swatch--lightBorder` based on a new conversation with design.

I restarted this conversation with Miruna and got clarification from her:

me: 
...I've been migrating a bunch of documentation from our [static site](https://opensource.adobe.com/spectrum-css/swatch.html) to storybook. I saw on the static site there was a "light border" variant that we don't currently support, and also found on the [Spectrum guidance page ](https://spectrum.corp.adobe.com/page/swatch-group/#Border-only-for-low-contrast-swatches) that there should be a border around a swatch with low contrast to its background. Right now, we don't have any CSS for this. I was just going to add it during my docs migration, but wanted to double check first that we still should be supporting this, or if I just found outdated info.

Miruna: 
> ...yes the border with low contrast (20% opacity) should be applied to Swatch only when they are in a Swatch group. This helps with comparing colors in a Swatch group better than the 42% opacity border that is applied to the single swatch...I think for Spectrum 1 the border was applied individually by product teams cause it was depending on the color of the Swatch ( it was not applied to all colors in a Swatch group, but dont quote me on that tho lol). But we decided that for S2 we are gonna be more restrictive with this and give the guidance to have a border for single swatch or swatch group cause its an important spectrum component so there should be a design POV in place

me: For the lightBorder, do the values change at all between light/dark modes?

Miruna: 
> it just stays linked to gray-1000 20% for both

This PR does not address any S2 design recommendations, but does set up CSS support for swatches to have a light border. Further implementation of `--lightBorder` should be seen in #2925 after this PR is merged. 

[Jira ticket](https://jira.corp.adobe.com/browse/CSS-855)

### Validation steps

- [x] Double check with design that lightBorder should be supported, and the value does not change between light and dark modes @marissahuysentruyt
- [x] Visit [the swatch docs page](https://pr-2954--spectrum-css.netlify.app/preview/?path=/docs/components-swatch--docs)
- [x] Inspect the swatch and find `.spectrum-Swatch`. Verify the swatch has its default border (`rgba(0, 0, 0, 0.51)`) in light mode.
- [x] Toggle the light and dark modes in the toolbar switcher. The default border should change based on the color mode as before. (`rgba(255, 255, 255, 0.51)` is for dark/darkest mode)
- [x] Add the `.spectrum-Swatch--lightBorder` class to the element. Verify the border has changed to `rgba(0, 0, 0, 0.2)`. 
- [x] Toggle light and dark modes in the toolbar switcher. The `spectrum-Swatch--lightBorder` class should  mimic the border behavior from before, but now uses `--spectrum-Swatch-border-color-light` that has a 20% opacity (instead of 51%). You may have to re-add the `--lightBorder` class each time you change color modes. 😩

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/pretty-socks-notice.md
+++ b/.changeset/pretty-socks-notice.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/swatch": minor
+---
+
+Reimplements support for the `spectrum-Swatch--lightBorder` class for swatches specifically in a swatch group.
--- a/components/swatch/index.css
+++ b/components/swatch/index.css
@@ -40,12 +40,14 @@
 	/* Light theme placeholder tokens */
 	.spectrum--light & {
 		--spectrum-swatch-border-color: rgba(0, 0, 0, 51%);
+		--spectrum-swatch-border-color-light: rgba(0, 0, 0, 20%);
 	}
 
 	/* Dark and Darkest theme placeholder tokens */
 	.spectrum--dark &,
 	.spectrum--darkest & {
 		--spectrum-swatch-border-color: rgba(255, 255, 255, 51%);
+		--spectrum-swatch-border-color-light: rgba(255, 255, 255, 20%);
 	}
 }
 
@@ -279,6 +281,14 @@
 
 			/* Undefined variable allows custom stylesheet or JS to pass the value to this element */
 			background-color: var(--spectrum-picked-color, transparent);
+		}
+	}
+}
+
+.spectrum-Swatch--lightBorder {
+	.spectrum-Swatch-fill {
+		&::before {
+			box-shadow: inset 0 0 0 var(--mod-swatch-border-thickness, var(--spectrum-swatch-border-thickness)) var(--highcontrast-swatch-border-color-light, var(--mod-swatch-border-color-light, var(--spectrum-swatch-border-color-light)));
 		}
 	}
 }
--- a/components/swatch/metadata/mods.md
+++ b/components/swatch/metadata/mods.md
@@ -2,6 +2,7 @@
 | -------------------------------------------- |
 | `--mod-animation-duration-100`               |
 | `--mod-swatch-border-color`                  |
+| `--mod-swatch-border-color-light`            |
 | `--mod-swatch-border-color-selected`         |
 | `--mod-swatch-border-radius`                 |
 | `--mod-swatch-border-thickness`              |


.changeset/pretty-socks-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/swatch": minor
+---
+
+Reimplements support for a `--lightBorder` class for swatches specifically in a swatch group.
```
rise-erpelding, line 5: 💅 🧦 👀  Thanks for adding this!

components/swatch/index.css
```diff
@@ -283,6 +283,21 @@
 	}
 }
 
+/* the lightBorder does not change values between light/dark modes */
+.spectrum-Swatch--lightBorder {
+	.spectrum--light &,
+	.spectrum--dark &,
+	.spectrum--darkest & {
+		--spectrum-swatch-border-color: rgba(0, 0, 0, 20%);
+	}
```
rise-erpelding, line 292: I'm impartial because I think this does the job... but because I am always curious about the historical context, I went digging and found [the PR](https://github.com/adobe/spectrum-css/pull/1501/files#diff-cc50ec1bfa218207591ac782352b4f5894b71d81450b2c74fa9004ca1dfca665) where `--lightBorder` disappears, and the CSS was originally like this:

```
.spectrum-Swatch--lightBorder {
  .spectrum-Swatch-fill {
    &:before {
      box-shadow: inset 0 0 0 var(--spectrum-swatch-fill-border-width) var(--spectrum-swatch-fill-border-color-light);
    }
  }
```
(note that we don't use `--spectrum-swatch-fill-border-width` anymore though)

And then the `--spectrum-swatch-fill-border-color-light` prop was defined ahead of that, kind of like how we usually define properties for hover, focus, etc.

Given how the code is in `main` now, it could fit up with the other placeholder token definitions something like:

```
	/* Light theme placeholder tokens */
	.spectrum--light & {
		--spectrum-swatch-border-color: rgba(0, 0, 0, 51%);
		--spectrum-swatch-fill-border-color-light: rgba(0, 0, 0, 20%);
	}

	/* Dark and Darkest theme placeholder tokens */
	.spectrum--dark &,
	.spectrum--darkest & {
		--spectrum-swatch-border-color: rgba(255, 255, 255, 51%);
		--spectrum-swatch-fill-border-color-light: rgba(0, 0, 0, 20%);
	}
```

Thoughts? Feel free to tell me I'm wrong, I'm not sure if one is better than the other! You've definitely spent more time looking at the swatch CSS than I have at this point!

rise-erpelding (approved):
This is good and it works! I left some questions about the code, but another question I have is about what color this light border should be. In the PR description, you note that Miruna said `gray-1000` 20% for both dark and light mode, but I _think_ this might be a color token that only exists in [Spectrum 2](https://github.com/adobe/spectrum-tokens/blob/%40adobe/spectrum-tokens%4013.0.0-beta.12/packages/tokens/src/color-palette.json#L382)? But in that case, it does have different rgb values for dark and light, so maybe `rgba(0, 0, 0, 20%)` wouldn't be used for both dark and light mode?

Clarification on the color would be great, I think that's about the only thing holding me back from approval here!

components/swatch/index.css
marissahuysentruyt: I don't know why I didn't think to put it at the top with the other light/dark/darkest variables. 🤦‍♀️ I like that organization though! https://github.com/adobe/spectrum-css/pull/2954/commits/ff94816cddad501e88340a04500d1e214cba86cf

As for the variable color change, you're definitely right. I misunderstood. I knew that token was new, but I didn't dig through the tokens dist. Now that I've re-read Miruna's response, the color will stay attached to `gray-1000`, but gray-1000 changes from light to dark. All I did was unattach gray-1000 for dark mode, which is wrong. Thanks for pointing that out!! https://github.com/adobe/spectrum-css/pull/2954/commits/f7559114fd4eefb4111e75c082054fa0d10a373e

components/swatch/index.css
```diff
@@ -40,12 +40,14 @@
 	/* Light theme placeholder tokens */
 	.spectrum--light & {
 		--spectrum-swatch-border-color: rgba(0, 0, 0, 51%);
+		--spectrum-swatch-border-color-light: rgba(0, 0, 0, 20%);
```
jawinn, line 43: Based on some previous Slack discussions about variable naming, I think it might be a good idea to swap the terms around and rename the variable to something like:
`--spectrum-swatch-light-border-color`

jawinn (changes requested):

.changeset/pretty-socks-notice.md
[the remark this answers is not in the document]
marissahuysentruyt: thanks!

jawinn (approved):

rise-erpelding (approved):
Looks great in dark and light mode! 🎉  Thanks @marissahuysentruyt!

pfulton (approved):

PR #2955, opened 2024-07-30
fix(card)!: change cover photo background color and include mod

jawinn:
## Description

Updates the background color behind the default variant's cover photo to match the design spec and adds a mod property so it can be customized.

1. BREAKING CHANGE: The default variant's cover photo now uses the same grey background color as the preview area for the quiet variant, which is visible when the image does not take up the entire space. This intended background color for non-quiet `--spectrum-background-base-color` was confirmed by the design team.

   Spectrum CSS currently always uses `background-size: cover` on a background image for this variant, but SWC uses an actual image with `object-fit`, so the use case of an image that did not fill the whole area was not present or testable on the CSS side. I created a followup Jira issue CSS-878 for getting this markup more in sync and adding it to VRTs.

2. This provides the existing mod `--mod-card-preview-background-color` for the area behind the image in the default variant. Which is the same mod already used for the area behind the image in the existing quiet variant. This resolves an issue reported in Slack about needing to customize this background color behind the cover photo (`.spectrum-Card-coverPhoto`).

### Screenshots

Before:
Screenshot 2024-08-09 at 3 39 08 PM

After:
Screenshot 2024-08-09 at 3 37 03 PM

S1 design spec for quiet. This same background is what should apply to the default variant, as confirmed by design.
Screenshot 2024-08-08 at 3 36 18 PM

And [the existing guidelines](https://spectrum.adobe.com/page/cards/#Preview-aspect-ratio) had an image with this use case including the darker background.

### Validation steps

- [x] Set `--mod-card-preview-background-color: red` in the inspector on `.spectrum-Card-coverPhoto` (Default card story) and turning off the background image, should show red in the background of the preview image area only:
Screenshot 2024-07-30 at 1 13 56 PM
- [x] There are no regressions in VRT
- [x] Set `background-size: contain; background-repeat: no-repeat;` in the inspector on `.spectrum-Card-coverPhoto`. The grey background should appear as it does in the "After" screenshot above.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/little-dodos-peel.md
+++ b/.changeset/little-dodos-peel.md
@@ -1,0 +1,7 @@
+---
+"@spectrum-css/card": major
+---
+
+BREAKING CHANGE: The card component's default variant now uses the same grey background color behind the cover photo that is used behind the image for the quiet variant. This background was only visible when the image did not take up the entire space. The intended background color of `--spectrum-background-base-color` for non-quiet variants was confirmed by the design team.
+
+This also provides the existing mod custom property `--mod-card-preview-background-color` for customizing this area behind the image for the default variant.
--- a/components/card/index.css
+++ b/components/card/index.css
@@ -227,7 +227,8 @@
 
 	background-size: cover;
 	background-position: center center;
-	background-color: var(--mod-card-background-color, var(--spectrum-card-background-color));
+
+	background-color: var(--mod-card-preview-background-color, var(--spectrum-card-preview-background-color));
 	border-block-end-color: var(--mod-card-border-color, var(--spectrum-card-border-color));
 }
 


components/card/index.css
```diff
@@ -227,7 +227,8 @@
 
 	background-size: cover;
 	background-position: center center;
-	background-color: var(--mod-card-background-color, var(--spectrum-card-background-color));
+	background-color: var(--mod-card-preview-background-color, var(--mod-card-background-color, var(--spectrum-card-background-color)));
```
jawinn, line 230: Looking at what's [suggested in this guideline screenshot](https://spectrum.adobe.com/page/cards/#Preview-aspect-ratio) but not defined in tokens for non-quiet variants, this really should be: `var(--mod-card-preview-background-color,var(--spectrum-card-preview-background-color))`. Though I am hesitant to make that change to the existing behavior. Thoughts on whether that could be done? It would change the default background behind the cover photo image to be darker:
Screenshot 2024-07-30 at 1 20 44 PM

marissahuysentruyt (commented):
This is looking good! Everything validates as you outlined: 

Screenshot 2024-08-01 at 12 47 13 PM

I even set the `--mod-card-background-color` to make sure the preview property was affecting the right area.

Screenshot 2024-08-01 at 12 50 53 PM

I left a question, but this seems good to me. Do we need a changeset at all?

components/card/index.css
jawinn: > would it break things or cause issues downstream
Yes, that is the concern.

marissahuysentruyt: Got it. I don't feel like I have enough of an opinion to tell you one way or another, but I did see [this screenshot back on the guidelines page](https://spectrum.adobe.com/page/cards/#Customizable) you mentioned: 

Screenshot 2024-08-02 at 11 55 09 AM

It's not a quiet variant, and still has the gray background. But without actually seeing that image, who knows- it could have the gray background itself. Who would we ask?

jawinn: After confirming with the design team that it should have the darker grey background, I've gone ahead and made the breaking change suggested in the original comment here. I've updated everything accordingly to fully communicate the effects of this.

jawinn:
I confirmed with the design team that this should have the same grey background as used on the quiet variant. I've gone ahead and modified this change to be the breaking change discussed in the comments. I've updated the PR and changeset to document this as fully as I could, with screenshots, updated validation insstructions, etc.

This is ready for a re-review.

components/card/index.css
```diff
@@ -506,3 +507,9 @@
 		border-radius: 0;
 	}
 }
+
+@media (forced-colors: active) {
+	.spectrum-Card-coverPhoto {
+		border-block-end: var(--mod-card-border-width, var(--spectrum-card-border-width)) solid CanvasText;
```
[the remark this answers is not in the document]
marissahuysentruyt: Maybe it's not necessary on that asset preview story. Adding that border does change the height of the cards, maybe because they're pulling in the `asset` component, so maybe it's more work outside of this branch. Feel free to disregard!

marissahuysentruyt (approved):
Everything works and validates as you explained- thanks for the updated instructions! I went down sort of a rabbit hole with the asset preview variant, but I wouldn't mind hearing your thoughts on that one 😆 

Things look great, everything checks out with design, so this is approved!

components/card/index.css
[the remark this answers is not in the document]
jawinn: I just happened to notice on the prod docs site that Asset (and default) look fine and show a border, because there is an `<hr>` Divider in the markup! That `<hr>` appears to be missing from our Storybook. I'm going to make a followup Jira issue for that and remove the high contrast change from this PR.

.changeset/little-dodos-peel.md
```diff
@@ -0,0 +1,10 @@
+---
+"@spectrum-css/card": major
+---
+
+BREAKING CHANGE: The card component's default variant now uses the same grey background color behind the cover photo that is used behind the image for the quiet variant. This background was only visible when the image did not take up the entire space. The intended background color for non-quiet `--spectrum-background-base-color` was confirmed by the design team.
```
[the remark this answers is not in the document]
jawinn: Updated. Thanks.

PR #2961, opened 2024-07-31
fix(breadcrumb): center breadcrumb itemLink

marissahuysentruyt:
## Description

During the [SWC implementation of breadcrumbs](https://github.com/adobe/spectrum-web-components/pull/4578), it was noticed that [there might be some misalignment of the breadcrumb items](https://adobedesign.slack.com/archives/CS22GNRU2/p1721217941481189?thread_ts=1720625760.158419&cid=CS22GNRU2), specifically for large/mobile scale: 

https://github.com/user-attachments/assets/33dc71fb-2f5a-4fee-8985-538457d914ac

In the video above, when breadcrumbs is toggled to the mobile scale, the margin around the separator icon increases from 19px to 25px. That margin causes the `spectrum-Breadcrumbs-item` element to increase in height in comparison to the last breadcrumb item, which has `display: none` set on its corresponding separator icon. You can see in the screenshots below that the last breadcrumb item is shorter in height without the additional margin increase from the icon:

Screenshot 2024-08-01 at 9 11 07 AM
Screenshot 2024-08-01 at 9 11 24 AM

Because of the increase in height, as well as the margins placed on the `spectrum-Breadcrumbs-itemLink` element, you can see that element increases in height, filling any remaining space. Although the `spectrum-Breadcrumbs` parent element has `align-items: center`, that was only affecting the `spectrum-Breadcrumbs-item` element, and not the `spectrum-Breadcrumbs-itemLink` element that houses the text node. As the item link element grows, you can see the visual misalignment of text to the separator icon because the text node was not centered in the item link element. This also leads to the first few breadcrumbs' text to look misaligned to the last breadcrumb text:

Screenshot 2024-08-01 at 9 16 54 AM
Screenshot 2024-08-01 at 9 16 29 AM

Adding `align-self: center` to `spectrum-Breadcrumbs-itemLink` resolves the height increase and corresponding text node misalignment. Additionally, `align-self: center` was added to the `spectrum-ActionButton` in use cases where an icon is used.

before:
Screenshot 2024-08-01 at 9 27 24 AM

after: 
Screenshot 2024-08-01 at 9 27 46 AM

The example below is from the Spectrum CSS Storybook, toggling between desktop and mobile scale: 

https://github.com/user-attachments/assets/aaa3105b-14bd-454c-80bf-20cc86203b72

This example below uses the preview URL from the SWC breadcrumb PR, adding `align-self: center` in the inspector, and toggling it on and off:

https://github.com/user-attachments/assets/055f48a9-c6bd-4b3c-a498-c30819726601

### Validation steps

- Visit the [breadcrumbs default story page](https://pr-2961--spectrum-css.netlify.app/preview/?path=/story/components-breadcrumbs--default)
- Toggle the scale to "large" for mobile viewing, and turn on both the border and measurement tools

Screenshot 2024-08-02 at 10 16 12 AM

- Inspect one of the `spectrum-Breadcrumbs-itemLink` elements. 
- Toggle the `align-self: center` in the inspector. You should see the borders of the item link elements change, and the text within them move up, away from the center line of the container

Screenshot 2024-08-02 at 10 12 16 AM

- With `align-self: center` turned off, measure the heights of `spectrum-Breadcrumbs-itemLink`. You should see a height difference with the first two items, in comparison to the last item/current breadcrumb. The first two items will be taller
- Turn on `align-self: center` in the inspector, and remeasure all breadcrumb item links. All should be the same height now
- In the Storybook controls, change the variant to compact or multiline. Repeat the process of checking that `align-self: center` better aligns the text nodes to the separator icons, and keeps the height of `spectrum-Breadcrumbs-itemLink` the same across all breadcrumb items (note that the final breadcrumb item in the multi-line variant _should_ have a different height)

#### Additional validation
- Pull down the branch and run locally
- In `breadcrumbs/breadcrumbs.stories.js`, add the following to ` Default.args.items` (add this before the last item): 
```
  {
    iconName: "FolderOpen",
    isDisabled: true,
  },
```
- Visit [the breadcrumbs default page](http://localhost:8080/?path=/story/components-breadcrumbs--default&globals=scale:large;measureEnabled:!true;outline:!true) (with the large scale, borders on, and measurement on)
- Back in the browser, you should now have additional breadcrumb with a disabled action button and folder icon
- Inspect the action button and toggle the `align-self: center` property. You should see the same behavior as with the text nodes, where `align-self: center` turned on ensures the action button is centered with the separator, and that property turned off leaves the action button unaligned.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.changeset/gold-spies-learn.md
+++ b/.changeset/gold-spies-learn.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/breadcrumb": patch
+---
+
+Aligns breadcrumb item text node to center of container. This should ensure the text and separator icon are center-aligned.
--- a/components/breadcrumb/index.css
+++ b/components/breadcrumb/index.css
@@ -176,6 +176,8 @@
 	display: inline-flex;
 	white-space: nowrap;
 
+	align-items: center;
+
 	font-family: var(--mod-breadcrumbs-font-family, var(--spectrum-breadcrumbs-font-family));
 	font-size: var(--mod-breadcrumbs-font-size, var(--spectrum-breadcrumbs-font-size));
 	font-weight: var(--mod-breadcrumbs-font-weight, var(--spectrum-breadcrumbs-font-weight));


rise-erpelding (commented):
Thank you for a great explanation of why this is happening and what it takes to fix it!

One really small thing I'm still questioning is whether we definitely need that `align-self: center;` on `> .spectrum-ActionButton` for the icon breadcrumb, the icon looks misaligned with or without it (pretty sure it's probably centered the way we'd want it, but I think it maybe looks a little wonky just because it's a different size than the text completely), and it's hard to gauge by measuring the size differences in the `.spectrum-Breadcrumbs-item` children as we did in the validation instructions for the text breadcrumbs... is there a better way to measure that this is centered that I maybe missed?

I'm also curious, if we do want to make sure that the action button is center-aligned (which most likely we do), can we take care of it by doing an `align-items: center;` on the parent (`.spectrum-Breadcrumbs-item`) rather than the individual children (`.spectrum-Breadcrumbs-itemLink` and `.spectrum-Breadcrumbs-item > .spectrum-ActionButton`), or would that cause some kind of unintended side effect?

marissahuysentruyt:
>  can we take care of it by doing an `align-items: center;` on the parent (`.spectrum-Breadcrumbs-item`) rather than the individual children (`.spectrum-Breadcrumbs-itemLink` and `.spectrum-Breadcrumbs-item > .spectrum-ActionButton`), or would that cause some kind of unintended side effect?

@rise-erpelding- I struggled with this. [I pushed up the `align-items: center`](https://github.com/adobe/spectrum-css/pull/2961/commits/08319c04637ae03de9b770fcf590bbb90784dce4) for `spectrum-Breadcrumbs-item`. Personally, I think that's the best way to do this in the long run. The reason I went with the `align-self` was because I was specifically trying to find a fix just for the mobile scale (which was where this was noticed originally). I guess I was concerned that `align-items: center` on `Breadcrumb-item` was moving the separators in desktop, centering them. I don't think that's a bad thing- it's still technically got all of their margin, just centered within `spectrum-Breadcrumbs-item` now. But if I look at [the S1 design file](https://www.figma.com/design/MPtRIVRzPp2VHiEplwXL2X/S-%2F-Desktop?m=auto&node-id=699-2159&t=I8pzPw1sFS1BwofJ-1), and [even the S2 file](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?m=auto&node-id=10711-11033), they're not technically centered? I thought I read somewhere that the designs (at least for S2 right now) were geared towards desktop, so I was trying not to touch desktop and just fix it for mobile (although the action button is now centered in desktop with that too. 🤦‍♀️ ). What do you think? I'm definitely open to the `align-items: center` fix, but maybe I got hung up on the "mobile" part. 

> is there a better way to measure that this is centered that I maybe missed?

I'm not sure. In Chrome, if I hover over `align-items: center` in the inspector, it will give me a grid line. Firefox does something sort of similar, but just extends the bounding boxes (it's definitely more "by eye" in Firefox). 

Screenshot 2024-08-08 at 10 33 32 AM

rise-erpelding (approved):
Ok, now that I'm hovering in the right spot I can see it centering the items!

image

I was on the fence about `align-self` vs. `align-items` but after hearing your thoughts I agree with you and think `align-items` is a better way to go in this case. Nice work and thanks again for documenting all the things so clearly on this bug!

PR #2963, opened 2024-08-01
docs: clarify usage of very low opacity on some inputs

jawinn:
## Description

Adds some comments to clarify existing CSS that uses a low opacity (0.0001 and 0.000001) for visually hidden inputs. And explains why they are hidden this way with sizing and positioning for touch users, instead of using a screenreader only (.sr-only) technique.

**Reference info:**
- [Inclusively Hiding & Styling Checkboxes and Radio Buttons: Hiding the checkboxes inclusively](https://www.sarasoueidan.com/blog/inclusively-hiding-and-styling-checkboxes-and-radio-buttons/#:~:text=navigating%20by%20touch.-,Hiding%20the%20checkboxes%20inclusively,-Touch%20interface%20screen) (2020)  - Why this technique appears to be used and why a visually-hidden or sr-only class would not be an improvement over the current way that these inputs are hidden.
- [Inclusive Considerations When Restyling Form Controls](https://24ways.org/2018/inclusive-considerations-when-restyling-form-controls/) (2018) - Source code that referenced the ChromeVox issue as being the reason that a very small but not zero opacity is used. This issue is also mentioned in a few old tangential GitHub issues and on StackOverflow. I wasn't able to find any other reasoning for why this technique would be used, as `opacity: 0` elements are included in the accessibility tree.

At some point in the future we may want to look at dropping the very low opacity in favor of `opacity: 0`. I cannot find any official change log available that references exactly when the ChromeVox issue was fixed.

CSS-814

### Validation steps

- [ ] Validate that explanation and wording are correct 

## To-do list

### DIFF
--- a/components/checkbox/index.css
+++ b/components/checkbox/index.css
@@ -399,6 +399,10 @@
 	}
 }
 
+/**
+ * The input is visually hidden (with opacity), and remains accessible using assistive tech.
+ * It is still sized and positioned where it would naturally be present so that touch users can find it with haptics.
+ */
 .spectrum-Checkbox-input {
 	color: var(--mod-checkbox-control-color-default, var(--spectrum-checkbox-control-color-default));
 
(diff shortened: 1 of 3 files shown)


pfulton (approved):
Thanks for leaving these comments! This will come in handy in the future when we look back to refactor these :-)

marissahuysentruyt (approved):
Looks good- thanks for researching this and leaving some explanations!

PR #2980, opened 2024-08-07
fix(menu): do not rely on source order of icon dependency css

jawinn:
## Description

Previously, menus were not displaying correctly if the CSS for icon was loaded after the CSS for menu, because `.spectrum-Icon` had the same specificity as some of the icon classes in Menu. This bug was visible on the theme/scale/direction pickers at the top of the docs site, for some components (e.g. [combobox](https://opensource.adobe.com/spectrum-css/combobox.html) and coach mark).

**Before (bug):**
before-bug
**After:**
after

This update increases the specificity on some icon selectors so the icons should always override the styles coming from `.spectrum-Icon`.

It also refactors a little bit of messy CSS by consolidating some of the same selectors that were spread out in different parts of the file.

CSS-834

### Validation steps

- [x] The theme/scale/direction pickers on all of the docs site pages are now displaying correctly (such as combobox, coach mark, popover, etc).
- [x] Menu still displays the same. Including hover and focus states, drill-in variant, disabled state, etc.
- [x] VRT shows no changes

### Regression testing

Validate:

## To-do list


components/menu/index.css
```diff
@@ -291,17 +291,46 @@
...
 }
 
-.spectrum-Menu-itemIcon {
+.spectrum-Menu-itemIcon.spectrum-Icon {
 	fill: var(--highcontrast-menu-item-color-default, var(--mod-menu-item-label-icon-color-default, var(--spectrum-menu-item-label-icon-color-default)));
 	color: var(--highcontrast-menu-item-color-default, var(--mod-menu-item-label-icon-color-default, var(--spectrum-menu-item-label-icon-color-default)));
 }
 
-.spectrum-Menu-checkmark {
-	align-self: center;
```
jawinn, line 300: This is overwritten by later CSS in the file that sets it to `align-self: start;`

components/menu/index.css
```diff
@@ -291,17 +291,46 @@
...
+}
+
+.spectrum-Menu-itemIcon:not(.spectrum-Menu-chevron, .spectrum-Menu-checkmark) {
+	block-size: var(--mod-menu-item-icon-height, var(--spectrum-menu-item-icon-height));
+	inline-size: var(--mod-menu-item-icon-width, var(--spectrum-menu-item-icon-width));
+}
+
+.spectrum-Menu-itemIcon--workflowIcon {
+	/* Always provide space at the end of a workflow icon. */
+	margin-inline-end: var(--mod-menu-item-label-text-to-visual, var(--spectrum-menu-item-label-text-to-visual));
```
jawinn, line 333: All these additions are moved up from below. The only changes should be the addition of `.spectrum-Icon` to a couple selectors.

components/menu/index.css
```diff
@@ -545,8 +574,6 @@
 .spectrum-Menu-chevron {
 	grid-area: chevronArea;
 	align-self: center;
-	block-size: var(--spectrum-menu-item-checkmark-height);
-	inline-size: var(--spectrum-menu-item-checkmark-width);
 }
```
marissahuysentruyt, line 577: So the `block-size` and `inline-size` were actually styles for the icon (`spectrum-Icon`). which is why the grid area and align-self can be left separately?

components/menu/index.css
```diff
@@ -545,8 +574,6 @@
 .spectrum-Menu-chevron {
 	grid-area: chevronArea;
 	align-self: center;
-	block-size: var(--spectrum-menu-item-checkmark-height);
-	inline-size: var(--spectrum-menu-item-checkmark-width);
```
marissahuysentruyt, line 549: So were the `block-size` and `inline-size` actually styles for the icon (`spectrum-Icon`), which is why the grid area and align-self can be left separately?

jawinn: `block-size` and `inline-size` were just moved up to the existing chevron selector (see `.spectrum-Menu-chevron.spectrum-Icon`).

I left the grid and alignment where they were since all the  other grid and alignment styles are still there next each other, and I didn't want to get carried away with refactoring.

rise-erpelding (approved):
This looks great, solves the issue, and doesn't seem to introduce any unwanted changes! Thanks for digging into the specificity here and finding the solution to this odd bug!

components/menu/index.css
```diff
@@ -291,17 +291,46 @@
 	}
 }
 
-.spectrum-Menu-itemIcon {
+.spectrum-Menu-itemIcon.spectrum-Icon {
```
castastrophe, line 294:
```suggestion
.spectrum-Menu-itemIcon.spectrum-Icon {
```

I'm not sure this is going to translate to SWC since the icon will be pulled in as a nested web component in the shadow dom.  That's why we've moved generally toward not nesting the class name of components and using mods passthroughs instead. If icon is overriding these styles do we need to use a fill and color passthrough?

castastrophe (changes requested):

components/menu/index.css
jawinn: We still need to increase the specificity in order to solve the source order issue. Using mod passthroughs would not address that. The issue stems from the same specificity of the single class names like this one, and `.spectrum-Icon`. And if the icon CSS is included afterwards, it is overriding things it shouldn't. 

What do you think about using `.spectrum-Menu .spectrum-Menu-itemIcon` instead?

jawinn: I'll work on a different method using mod passthroughs. It'll need an additional mod or two added to Icon, but I think it will work.

jawinn: I've updated this to use `.spectrum-Menu .thing` instead of `.thing.spectrum-Icon`, as SWC should convert that just fine, correct? That seems in line with most other CSS and doesn't require having to add a mod to the icon component for every property that gets changed by menu's icon classes, now and into the future.
2ec451d2c85c3e1e7973b7aa29de1c8da2097359

components/menu/metadata/metadata.json
```diff
@@ -5,6 +5,9 @@
     ".is-selectableMultiple .spectrum-Menu-itemCheckbox",
     ".js-focus-within .spectrum-Menu .spectrum-Menu-item--collapsible.is-open[focus-within]",
     ".spectrum-Menu",
+    ".spectrum-Menu .spectrum-Icon",
```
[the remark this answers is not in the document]
castastrophe: (This is why the CI build is failing)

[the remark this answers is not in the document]
jawinn: Ah, yeah the updated metadata needed to get pushed up. I just pushed that up and rebased again with `main`.

castastrophe (approved):

PR #3023, opened 2024-08-23
fix(progressbar, search, pickerbutton): migrate --mod variables to index.css

jawinn:
## Description

Part of the larger effort to remove `--mod` property references from our themes files. Moves the mods for:
- progressbar
- search
- pickerbutton

### Validation steps

  - [ ] No `--mod` prefixed properties exist in the `themes/spectrum-two.css` files for progressbar, search, and pickerbutton
  - [ ] Mods are moved to index.css and are applied so that they still work the same as before

### Regression testing

Validate:

## To-do list


components/search/themes/spectrum-two.css
```diff
@@ -87,7 +87,6 @@
 	}
 
 	.spectrum-Search--quiet {
-		--spectrum-search-quiet-button-offset: calc((var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2) - (var(--mod-workflow-icon-size-100, var(--spectrum-workflow-icon-size-100)) / 2));
```
castastrophe, line 90: Is this already moved into the search component's index.css so that we're maintaining this offset calculation?

castastrophe (changes requested):
Looks great!  Can you add your changeset notes to this PR as well?

components/search/themes/spectrum-two.css
jawinn: Yeah, I saw it was already being overridden by this existing line in index.css:
https://github.com/adobe/spectrum-css/blob/61868f94cac1a0f9918fa6c60679c2970750ae16/components/search/index.css#L160

I noticed that this is slightly different than main though:
`--spectrum-search-quiet-button-offset: calc((var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2) - (var(--mod-workflow-icon-size-100, var(--spectrum-search-icon-size)) / 2));` 
as compared to 
`--spectrum-search-quiet-button-offset: calc((var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2) - (var(--mod-workflow-icon-size-100, var(--spectrum-workflow-icon-size-100)) / 2));` 

I think that was to remove the global token reference from the index.css? It results in slightly different behavior, because `--spectrum-search-icon-size` changes per t-shirt size (and the mod name should be `--mod-search-icon-size`). Using `--spectrum-search-icon-size` looks like it would be more correct and a fix...but the more I look at the quiet close button positioning, and look at React Spectrum's version, I think the positioning is incorrect either way; I can open a bug issue to look into fixing that in main. For now, should I keep this as is, but rename the mod?

jawinn: I've changed the mod name for now here: [948da81](https://github.com/adobe/spectrum-css/pull/3023/commits/948da81cf75ca01b542147f3de81f3a9cd0ab638)

I've created a Jira issue CSS-914 and started a discussion with design to confirm about the correct alignment of the clear button for quiet. Whether there are changes or not, it looks like this needs a bugfix on `main` first.

castastrophe (approved):

PR #3024, opened 2024-08-23
build: build robust metadata reports for every component

castastrophe:
## Description

To support a need for standardized reporting on the features available in each component, this PR creates a new `components/*/metadata.json` output. After building every component, the build script will evaluate the content of the compiled `dist/index.css` file and report on the following:

- `selectors`: all selectors created and used in this component
- `component`: all component-named custom properties used
- `modifiers`: the modifiers used that match the component's naming convention (separate from passthrough modifiers or global modifiers)
- `theming-layer`: the --system prefixed variables generated by the theming tool
- `high-contrast`: the available high contrast variable hooks
- `global`: references to any potential global design tokens
- `passthroughs`: modifiers that did not match the standard naming convention of the component and are _likely_ passthroughs

These values are sorted once collected to ensure a standardized format even when styles are reordered; this will prevent unnecessary diffs to this file when linting or structural updates occur in a component.

To the goal of creating this new output, the `tasks/component-builder.js` script was updated and optimized to improve readability of the file and reduce complexity.

### Validation steps

- [x] `yarn builder actionbutton`: expect to see the creation of a `components/actionbutton/metadata.json file`
- [x] Expect to see no changes to the compiled `*.css` assets (`yarn compare`)
- [x] Expect to see a subtle reordering in some of the `metadata/mods.md` files but no removals (mods not using the component name in the string are moved to the bottom).

## To-do list


tasks/component-builder.js
```diff
@@ -104,81 +42,115 @@ async function extractProperties(
...
+	const baseSelectors = [".spectrum", ".spectrum--express", ".spectrum--legacy"];
 	const selectors = new Set();
 	const root = postcss.parse(content);
-	root.walkRules(rule => {
-		if (rule.selector) selectors.add(rule.selector);
+	root.walkRules((rule) => {
+		if (rule.selectors) {
+			rule.selectors.forEach((selector) => {
+				// If the selector is not a base selector, add it to the set
+				if (baseSelectors.some((base) => selector !== base)) {
```
rise-erpelding, line 63: I can see that this [isn't excluding selectors](https://github.com/adobe/spectrum-css/blob/ab74143c4d5ca11fa3cbb156cf9d07a972311766/components/actionbutton/metadata/metadata.json#L4) like `.spectrum`, should it be?

rise-erpelding (approved):
This seems to be working great! I played around with adding/removing selectors, custom properties, etc. and the script seems to be catching all of those changes every time!

I have a question in addition to the one I left in the comment, but these might not necessarily block merging:

I did notice that if I manually manipulated the contents of the `metadata.json` file, it doesn't update the file again until I run `nx reset`, that's probably fine but want to call it out just in case that's not desired behavior.

tasks/component-builder.js
castastrophe: Hmm, good question. I'm torn because they are selectors that are present in the file so it's accurate that they're listed but I'm not sure how helpful that inclusion is. What are your thoughts?

rise-erpelding: I think that's ok if they're listed! Moreso I was just confused by the comment here, I was expecting it not to include those `baseSelectors`.

cdransf (approved):

tasks/component-builder.js
castastrophe: Oh!  I was using a some selector instead of every when checking for !== so it wasn't excluding them like the code expected.  I fixed it!

PR #3047, opened 2024-08-28
test: always run Chromatic against main

pfulton:
Updates our GitHub Action for VRT to always run on pushes to `main`, as suggested in the Chromatic docs:
https://www.chromatic.com/docs/github-actions/#maintain-a-clean-main-branch

> A clean main branch is a development best practice and highly recommended for Chromatic. This means testing your main branch to ensure builds are passing. It’s important to note that baselines will not persist through branching and merging unless you test your main branch.

> If the builds result from direct commits to main, you must accept changes to keep the main branch clean. If they’re merged from feature-branches, you must ensure those branches are passing before you merge into main.

### Validation steps
Unfortunately, I think we'll only know upon merging this into `main`.

## To-do list

- [X] ✨ This pull request is ready to merge. ✨

### DIFF
--- a/.github/workflows/vrt.yml
+++ b/.github/workflows/vrt.yml
@@ -1,6 +1,9 @@
 name: Visual regression testing
 
 on:
+    push:
+      branches:
+        - main
     workflow_dispatch:
     workflow_call:
         inputs:


.github/workflows/vrt.yml
```diff
@@ -1,6 +1,9 @@
 name: Visual regression testing
 
 on:
+    push:
+      branches:
+        - main
```
pfulton, line 6: We should probably see if we can fine tune this to ignore `release` commits 🤔

castastrophe: I don't think it hurts to try this and see if releases are causing noise. I think releases won't trigger Chromatic because they don't alter files that chromatic is watching. We'll have to keep an eye on that to prove out that theory.

castastrophe (approved):

.github/workflows/vrt.yml
pfulton: You're probably right! I bet we'll know soon enough 😆

PR #3049, opened 2024-08-29
fix(button, SWC-342): update content-color for primary s2 button

TarunAdobe:
## Description

This PR fixes SWC-342. Updates the content-color-default for primary button.

### Validation steps

 - [ ] No expected changes to `metadata.json`

### Regression testing

Validate:

## Screenshots

Previously:

image

## To-do list

### DIFF
--- a/components/button/themes/express.css
+++ b/components/button/themes/express.css
@@ -37,12 +37,5 @@
 			--spectrum-button-background-color-down: var(--spectrum-neutral-background-color-down);
 			--spectrum-button-background-color-focus: var(--spectrum-neutral-background-color-key-focus);
 		}
-
-		&.spectrum-Button--primary {
-			--spectrum-button-content-color-default: var(--spectrum-white);
-			--spectrum-button-content-color-hover: var(--spectrum-white);
-			--spectrum-button-content-color-down: var(--spectrum-white);
-			--spectrum-button-content-color-focus: var(--spectrum-white);
-		}
 	}
 }
--- a/components/button/themes/express.css
+++ b/components/button/themes/express.css
@@ -37,5 +37,12 @@
 			--spectrum-button-background-color-down: var(--spectrum-neutral-background-color-down);
 			--spectrum-button-background-color-focus: var(--spectrum-neutral-background-color-key-focus);
 		}
+
+		&.spectrum-Button--primary {
+			--spectrum-button-content-color-default: var(--spectrum-white);
+			--spectrum-button-content-color-hover: var(--spectrum-white);
+			--spectrum-button-content-color-down: var(--spectrum-white);
+			--spectrum-button-content-color-focus: var(--spectrum-white);
+		}
 	}
 }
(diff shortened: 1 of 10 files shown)


Rajdeepc:
Nice! Do we have a Figma link from where we can check the actual tokens for S2 dark|light?

jawinn (changes requested):
It doesn't seem like we should be adding new custom tokens for this.

In Spectrum 2, the token used for the content color in the Primary button has changed to `gray-25`:
Screenshot 2024-09-04 at 2 55 28 PM

But I thought we weren't getting into individual component level changes as part of this work? If we are and this particular value is to be changed before the full S2 migration of button, would it make sense to use the new `gray-25` in `spectrum-two.css` and keep `--spectrum-white` in `spectrum.css`? It seems like the S1 value should stay as `--spectrum-white` unless that's a breaking spec change that is also being made to `main`.

components/button/themes/spectrum-two.css
```diff
@@ -138,10 +138,10 @@
...
 			--spectrum-button-border-color-focus: transparent;
 
-			--spectrum-button-content-color-default: var(--spectrum-white);
-			--spectrum-button-content-color-hover: var(--spectrum-white);
-			--spectrum-button-content-color-down: var(--spectrum-white);
-			--spectrum-button-content-color-focus: var(--spectrum-white);
+			--spectrum-button-content-color-default: var(--spectrum-gray-25);
+			--spectrum-button-content-color-hover: var(--spectrum-gray-25);
+			--spectrum-button-content-color-down: var(--spectrum-gray-25);
+			--spectrum-button-content-color-focus: var(--spectrum-gray-25);
```
TarunAdobe, line 144: I've only changed the S2 mappings for the `button-content-color` as having `--spectrum-white` was resulting in a inaccessible state for the spectrum-two primary button in dark mode. Let me know if it works now? @jawinn

rise-erpelding (changes requested):
This works as expected in S2, but is causing some issues in Spectrum 1/Legacy and Express:
image

I think we'd need to set the button-content-color (back to `--spectrum-white`) for those themes to get those back to what they were before.

components/button/themes/spectrum-two.css
jawinn: Like Rise mentioned, we'd need to set these to `--spectrum-white` in the Spectrum 1/Legacy `spectrum.css` to maintain existing S1 values, since the value is changing for Spectrum 2.

components/button/themes/express.css
```diff
@@ -37,5 +37,12 @@
 			--spectrum-button-background-color-down: var(--spectrum-neutral-background-color-down);
 			--spectrum-button-background-color-focus: var(--spectrum-neutral-background-color-key-focus);
 		}
+
```
rise-erpelding, line 40: @TarunAdobe I should have been clearer in the last feedback I left, sorry! This `themes/express.css` file imports `themes/spectrum.css` so the styles from `spectrum.css` are inherited by default here, I don't think anything needs to be added to this theme for this issue.

rise-erpelding (approved):

PR #3051, opened 2024-08-29
fix: adding max height and full height to picker menu component

aramos-adobe:
## Description

The picker menu list can get very extensive. Why don't we add the ability to set a fixed height to scroll or extend the menu to the height of the viewport where a user can select their option.

- Added max-block-size capabilities
- Added full height capabilities

### Regression testing

Validate:

## To-do list


Rajdeepc:
Looking nice! Just an FYI. This issue was also being reported while using Tray component in silo. So would it also be okay to add this property in tray.css too so that users can have control on the max-height?
cc @pfulton @rubencarvalho

components/picker/stories/template.js
```diff
@@ -136,13 +138,10 @@ export const Template = ({
 				isDisabled,
 				isReadOnly,
 				customClasses,
-				customStyles: {
```
rubencarvalho, line 139: Where did these go? 😛

components/picker/index.css
```diff
@@ -491,3 +491,10 @@
 		color: var(--highcontrast-picker-content-color-disabled, var(--mod-picker-font-color-disabled, var(--spectrum-picker-font-color-disabled)));
 	}
 }
+
+/* Fixed menu height */
+.spectrum-Picker--menu-height-fixed {
+	& + .spectrum-Popover.is-open {
+		max-block-size: var(--mod-picker-menu-height-fixed, 200px);
```
rubencarvalho, line 498: Just for context, did we confirm this default value with design?

.changeset/unlucky-wasps-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/picker": major
```
rubencarvalho, line 2: Should this be a major bump? Seems like this feature is non-breaking.

components/picker/stories/template.js
```diff
@@ -39,6 +40,7 @@ export const Picker = ({
 				["is-open"]: isOpen,
 				["is-loading"]: isLoading,
 				["is-keyboardFocused"]: isKeyboardFocused,
+				[`${rootClass}--menu-height-fixed`]: menuHeight,
```
rubencarvalho, line 43: Have we considered the pros and cons of not having a dedicated API for this feature? For instance, instead of having a specific menu-height class, we could use a generic mod variable applied to the base class. This mod variable would function as a feature toggle, although customers would need to define its value themselves (because we would not have a default fallback).

components/picker/stories/picker.stories.js
```diff
@@ -52,6 +52,16 @@ export default {
 		},
 		isQuiet,
 		isOpen,
+		menuHeight: {
+			name: "Max Menu Height",
```
rubencarvalho, line 56: I would recommend renaming it to "Menu max block size" (doesn't sound as nice, but I would be for accuracy) - what do you think?

components/picker/stories/template.js
rubencarvalho: But I guess this is a more holistic API design question, maybe the CSS team can help understand when a `mod` is enough vs a dedicated toggleable class feature.

components/picker/index.css
```diff
@@ -491,3 +491,10 @@
 		color: var(--highcontrast-picker-content-color-disabled, var(--mod-picker-font-color-disabled, var(--spectrum-picker-font-color-disabled)));
 	}
 }
+
+/* Fixed menu height */
+.spectrum-Picker--menu-height-fixed {
+	& + .spectrum-Popover.is-open {
```
rubencarvalho, line 497: Do we need the `.is-open` here? Can't the popover `max-block-size` be defined independently of being open?

components/picker/metadata/metadata.json
```diff
@@ -124,6 +125,7 @@
     "--mod-picker-icon-color-key-focus",
     "--mod-picker-inline-size",
     "--mod-picker-line-height",
+    "--mod-picker-menu-height-fixed",
```
rubencarvalho, line 128: Same comment as the other one, around the naming 😛

components/picker/stories/template.js
aramos-adobe: @rubencarvalho I removed the inline styling of the Picker because it was displaying the label and button inline and overriding the flex-box styling and disrupting the text wrap especially when the label was displayed left. So I "turned it off" by removing this block. I can bring it back if needed

rubencarvalho: Oh, thanks for the context! I thought it was a forgotten edit 😄

components/picker/metadata/metadata.json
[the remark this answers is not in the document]
pfulton: For now, this is our approach. We may revisit it in the near future, but currently we will move forward with this approach (so it's valid here).

components/picker/index.css
aramos-adobe: For specificity, I think it's necessary. It also gives other devs more context.

components/picker/metadata/metadata.json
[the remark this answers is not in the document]
aramos-adobe: `max-block-size: var(--spectrum-popover-max-block-size);`

This would require us defining a global variable that can be shared.

Rajdeepc:
@pfulton Would you let me know what has been decided here? I see its still open!

components/picker/stories/template.js
castastrophe: I agree, this feels like a good use-case to add a mod rather than new API.

castastrophe (changes requested):

PR #3056, opened 2024-08-30
fix(modal, SWC-344): update background-color for s2 modal

Rajdeepc:
## Description

This PR fixes SWC-344. Updates the `--spectrum-modal-background-color-quiet` for modal

### Regression testing

Validate:

## Screenshots

Previously:
Screenshot 2024-08-30 at 3 06 15 PM
Updated:
Screenshot 2024-08-30 at 3 21 51 PM

## To-do list

### DIFF
--- a/components/modal/themes/spectrum-two.css
+++ b/components/modal/themes/spectrum-two.css
@@ -21,7 +21,7 @@
 		--spectrum-modal-max-height: 90vh;
 		--spectrum-modal-max-width: 90%;
 
-		--spectrum-modal-background-color: var(--spectrum-gray-75);
+		--spectrum-modal-background-color: var(--spectrum-modal-background-color-quiet);
 		--spectrum-modal-confirm-border-radius: var(--spectrum-corner-radius-100);
 		--spectrum-modal-confirm-exit-animation-duration: var(--spectrum-animation-duration-100);
 		--spectrum-modal-confirm-entry-animation-duration: var(--spectrum-animation-duration-500);
--- a/.changeset/good-parrots-attack.md
+++ b/.changeset/good-parrots-attack.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/modal": major
+---
+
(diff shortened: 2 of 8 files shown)


components/modal/themes/spectrum-two.css
```diff
@@ -21,7 +21,7 @@
 		--spectrum-modal-max-height: 90vh;
 		--spectrum-modal-max-width: 90%;
 
-		--spectrum-modal-background-color: var(--spectrum-gray-75);
```
jawinn, line 24: If we're changing this value for Spectrum 2, shouldn't this be set back to the existing S1 value for S1/Legacy in `spectrum.css`?

I was also wondering if the design team has a new token that should be used for this background color, rather than us creating a custom token that would only be used temporarily. That might be worth confirming with the designers in `#spectrum_web_alliance`. Looking at the S2 Figma for the dialogs, it seems like it could be `background-layer-2-color`, but that has different values than I'm seeing here listed for both S1 and S2:

Screenshot 2024-09-05 at 10 45 29 AM

jawinn (changes requested):

components/modal/themes/spectrum-two.css
```diff
@@ -21,7 +21,7 @@
...
+		--spectrum-modal-background-color: var(--spectrum-modal-background-color-quiet);
```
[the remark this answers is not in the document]
Rajdeepc: No this is not needed now! Thanks for pointing it out!

components/modal/themes/spectrum-two.css
Rajdeepc: Thanks @jawinn The new token is `--spectrum-background-layer-2-color`

jawinn: Looks good. Thanks for confirming that with the design team.

jawinn (approved):
Thanks for confirming that with the design team.

rise-erpelding (approved):

PR #3086, opened 2024-09-09
feat(button): add label text wrap option

aramos-adobe:
## Description

- This change makes buttons _not_ wrap label text by default.
- Introducing a class to make label text wrap if dev chooses to do so.
- Added label wrap class in Text Overflow story in Storybook

### Regression testing

Validate:

## To-do list


aramos-adobe:
Screenshot 2024-09-11 at 9 27 06 AM
Screenshot 2024-09-11 at 9 27 15 AM
Screenshot 2024-09-11 at 9 27 29 AM

.changeset/mean-shrimps-press.md
jawinn, line 1: I see two changesets here with similar content; it looks like we only need one?

components/button/stories/template.js
```diff
@@ -158,13 +160,15 @@ export const TextOverflowTemplate = (args) => Container({
       customStyles: {
         "max-inline-size": "480px",
       },
+      isLabelWrapped: true,
```
jawinn, line 163:
```suggestion
```
It looks like this was an older variable naming?

components/button/stories/template.js
```diff
@@ -158,13 +160,15 @@ export const TextOverflowTemplate = (args) => Container({
...
       label: "An example of text overflow behavior when there is no icon. When the button text is too long for the horizontal space available, it wraps to form another line.",
     })}
     ${Template({
       ...args,
       customStyles: {
         "max-inline-size": "480px",
       },
+      isLabelWrapped: true,
```
jawinn, line 171:
```suggestion
```

components/button/stories/button.stories.js
```diff
@@ -58,6 +58,13 @@ export default {
 		isActive,
 		isPending,
 		staticColor,
+		noWrap: {
+			name: "Disable Label Wrap",
+			type: { name: "boolean" },
+			table: {
+				category: "Content",
```
jawinn, line 65:
```suggestion
				category: "Advanced",
```
I would categorize this as advanced, since it's not a regular option from the spec and isn't typically recommended. Otherwise it would be in the "Component" category. "Content" is typically for controls that are for modifying the content displayed within the story (e.g. text, image source, icon name, etc).

jawinn (changes requested):
I think the `nowrap` CSS should work here. This and the added control look good!

Could you make sure that we have this included in the VRTs, by adding a `noWrap: true` test in the `button.test.js` file? It likely will need a set width on the wrapperStyles, or a min-inline-size like the text overflow examples in order for wrapping to happen. When added it'll be viewable on the Default story with "Show testing preview" turned on.

components/button/stories/button.stories.js
```diff
@@ -58,6 +58,13 @@ export default {
 		isActive,
 		isPending,
 		staticColor,
+		noWrap: {
+			name: "Disable Label Wrap",
```
jawinn, line 62:
```suggestion
			name: "Disable label wrap",
```
Use [sentence case](https://spectrum.adobe.com/page/grammar-and-mechanics/#Sentence-case) on this. That's something we've been working on making more consistent across the docs lately.

components/button/stories/template.js
aramos-adobe: Yes I need to get rid of this. This was from a previous approach of making the button label wrap by default.

components/button/index.css
```diff
@@ -294,6 +294,17 @@ a.spectrum-Button {
 	}
 }
 
+/* Disable button label wrap  */
+.spectrum-Button--noWrap {
+	justify-content: flex-start;
```
jawinn, line 299: With the `overflow` and `text-overflow` additions, is this addition still needed? With short text plus the minimum width and for buttons with some other set width, the text will no longer be centered:
Screenshot 2024-09-16 at 3 10 40 PM

.changeset/green-maps-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/button": minor
+---
+
+Adding the ability to disable label wrap in button component
```
jawinn, line 5:
```suggestion
Adds the ability to disable text wrapping within the button component's label by applying the class `.spectrum-Button--noWrap` to `.spectrum-Button`.
```
Changing the tense and adding some more detail here, since this will go in the change log and be referenced by SWC and others.

components/button/index.css
```diff
@@ -294,6 +294,17 @@ a.spectrum-Button {
...
+	
+	.spectrum-Button-label {
+		white-space: nowrap;
+		overflow: hidden;
+		text-overflow: ellipsis;
```
jawinn, line 304: Now that this is working like the text overflow behavior of Action button, it creates accessibility issues when there is text overflow, since the text can be sometimes unreadable. Does that mean we need to provide similar guidelines about a tooltip? We'd be asking more of the implementation for this outside the spec feature: https://spectrum.adobe.com/page/action-button/#Text-overflow

components/button/stories/button.test.js
```diff
@@ -66,6 +66,27 @@ export const ButtonGroups = Variants({
...
+			testHeading: "Disable label wrapping with workflow icon",
+			customStyles: {
+				"max-inline-size": "120px",
+			},
+			iconName: "Star",
+			label: "Be a premium member",
+			withStates: false,
+			noWrap: true,
+			Template,
+		},
```
jawinn, line 89: Thanks for making sure this is represented in the VRTs 👍

jawinn (changes requested):

components/button/index.css
aramos-adobe: @jawinn When the max-inline-size is set all the button content gets cut off from left and right edges of the container. I was thinking of doing a flex-start and overflow additions it'll show the dev the text is too long and they need to make an adjustment character wise. 

After some thought, I'm going to remove this because it's not a design specification and I don't want teams to get mislead in anyway.

jawinn: I do think what you had, `.spectrum-Button-label { overflow: hidden; text-overflow: ellipsis; }` is a good idea. I think we _should_ handle the text overflow for cases where the button's width is constrained, as it gets a bit broken as seen in the Chromatic testing view right now. Does it work okay with just those two properties?

aramos-adobe: @jawinn Yes, they work with just these two properties.

components/button/index.css
aramos-adobe: @jawinn I tried it and it added the test case. 

Tooltip only shows when `nowrap` is true and `hideLabel` is false. 
If the text is not "ellipsized" or visible I added a class to hide the tooltip: `spectrum-Tooltip--hide`.

components/button/stories/button.stories.js
```diff
@@ -286,3 +295,18 @@ WithWrapping.args = {
 WithWrapping.parameters = {
 	chromatic: { disableSnapshot: true },
 };
+
+/**
+ * The normal behavior for lengthy text in the given horizontal space available is that it will wrap to form another line. By using the `noWrap` option, the lengthy button text will horizontally expand the entirety of the container. Be sure to define the width of the button container using this option. Please note: this can cause undesired overflow experiences and to help prevent this, the overflowing text will attempt to hide by showing an ellipsis (...).
```
jawinn, line 300:
```suggestion
The normal behavior for lengthy text in the given horizontal space available is that it will wrap to form another line. By using the `.spectrum-Button--noWrap` class, the lengthy button text will not cause a line break and the width of the button will expand until it reaches its maximum width.

Please note: this can cause undesired overflow experiences and to help prevent this, the overflowing text will attempt to hide by showing an ellipsis (...). This is demonstrated in the last two examples below, by constraining the maximum width of the button.

This option is not part of the design spec, so please use carefully, with consideration of the overflow behavior and the readability of the button's content.
```

components/button/stories/button.stories.js
```diff
@@ -286,3 +295,18 @@ WithWrapping.args = {
 WithWrapping.parameters = {
 	chromatic: { disableSnapshot: true },
 };
+
+/**
+ * The normal behavior for lengthy text in the given horizontal space available is that it will wrap to form another line. By using the `noWrap` option, the lengthy button text will horizontally expand the entirety of the container. Be sure to define the width of the button container using this option. Please note: this can cause undesired overflow experiences and to help prevent this, the overflowing text will attempt to hide by showing an ellipsis (...). 
+ * */
+
+export const DisableWrapping = TextWrapTemplate.bind({});
```
jawinn, line 303: What do you think about showing a third option which demonstrates that this can escape its container? I experimented in the inspector a little bit:
Screenshot 2024-09-19 at 3 10 34 PM
Here I'm setting the width on the inner container (adjustable through `wrapperStyles`), and the buttons with ellipsis are `max-inline-size: 100%`.

This makes me wonder if we need a new mod for `max-inline-size`, if we're expecting users in SWC to be able to want to change that property too (@pfulton)?

jawinn (changes requested):

pfulton (approved):
Thanks!!

jawinn (approved):

aramos-adobe:
<iframe src="https://giphy.com/embed/gKfyusl0PRPdTNmwnD" width="480" height="250" style="" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/pleading-vincent-van-gogh-please-god-gKfyusl0PRPdTNmwnD">via GIPHY</a></p>

aramos-adobe:
image

PR #2218, opened 2023-10-18
refactor(treeview): fixes for sizes, height, and indicator alignment

jawinn:
## Description

**Tree view fixes**
A refactor to fix these various issues discovered with Tree view:
- Incorrect indentation at various t-shirt sizes.
- Tree view items and their link element had irregular heights depending on whether they had a chevron or not. 
- Tree view items were not the correct component heights at various t-shirt sizes.
- Tree view item indicator (chevron) was not square and not the right size at various t-shirt sizes.
- Font-size was not changing. Now uses font-size tokens, confirmed with design.
- Standalone was not showing rounded corners (now renamed to "detached" per guidelines).
- Selected Thumbnail items were showing focus styles instead of selected styles; examples now include the "layer" variant of Thumbnail.
- "Section" had heading divs that were a child of `ul`, which was not valid markup. This markup has been adjusted and an additional class added to help display this properly within an `li`.

CSS-592

**Storybook additions** 
Also adds a long text truncation example to a story, and an additional story for the missing "Flat" markup example from the docs page. The selected state for the thumbnail variant has also been added to the examples.

**Display property change** 
For hidden items, `display: none` is now used instead of `visibility: hidden`, to improve performance. Addresses #2201 

CSS-603

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation instructions

- [x] The treeview item _and_ treeview itemLink (`<a>` tag) are 32px height at medium, 24px at small, 40px at large, and 48px at extra large
- [x] "Folders and Files" Story: Nested items maintain indentation at large and extra large sizing (in previous behavior, indentation kept decreasing as t-shirt size increased, till there was none left)
- [x] The Treeview indicator (chevron) element is now square, aligned properly to the left of the item without escaping it, and is the width and height of the treeview item (same heights mentioned in first validation item)
- [x] All stories look to render correctly at all t-shirt sizes, including selected item, hover, and tab focus (note, Storybook psuedo addon may cause an issue with focus on selected item; refreshing the page usually fixes)
- [x] New Storybook "Flat markup" story appears and indentation looks correct (+ at different sizes)
- [x] The Thumbnail variant looks correct, and like the design on the Xd file from the guidelines: the height of the tree view items with the thumbnail, per that file, should be 6px taller than default for small (both medium and large scale), and 2px taller at medium when at large scale (test in Storybook).
[...]

PR #3089, opened 2024-09-09
fix(treeview): display corner rounding for detached treeview

rise-erpelding:
## Description

The detached treeview variant was inadvertently removed in a refactor (#2218), this returns it to its original styling as seen before #2218, with corner rounding on items visible in the detached variant.

[CSS-871](https://jira.corp.adobe.com/browse/CSS-871)

### Validation steps
- [x] Confirm that the treeview detached variant has corner rounding displayed as before, either using the [Storybook controls](https://pr-3089--spectrum-css.netlify.app/preview/?path=/story/components-tree-view--default&args=variant:detached) or viewing the Detached variant on the [Docs page](https://pr-3089--spectrum-css.netlify.app/preview/?path=/docs/components-tree-view--docs#detached) (I could not find any guidance in design files, but the [documentation](https://spectrum.adobe.com/page/tree-view/#Detached) does show this variant) @marissahuysentruyt 
- [x] Confirm that [testing preview](https://pr-3089--spectrum-css.netlify.app/preview/?path=/story/components-tree-view--default&globals=testingPreview:!true)/[VRTs](https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=2904) add coverage for detached variant @marissahuysentruyt 

### Regression testing

Validate:

- [x] Any VRT changes have been accepted (by reviewer and/or PR author).

## Screenshots

Before:
image

After:
image

## To-do list

### DIFF
--- a/components/treeview/stories/template.js
+++ b/components/treeview/stories/template.js
@@ -25,6 +25,7 @@ export const TreeViewItem = ({
 	iconSet,
 	thumbnail,
 	items,
+	variant,
 	customClasses = [],
 } = {}, context = {}) => {
 	if (type === "heading") {
@@ -93,14 +94,16 @@ export const TreeViewItem = ({
 						customClasses: [`${rootClass}-itemIcon`],
 					}, context)
 				)}
-				${when(thumbnail, () =>
+				${when(variant === "thumbnail", () =>
 					Thumbnail({
+						imageURL: "example-card-landscape.png",
 						...thumbnail,
 						size: size == "s"  ? "200"
 							: size == "m"  ? "200"
 							: size == "l"  ? "400"
 							: size == "xl" ? "600"
 							: "300",
+						
 						isLayer: true,
 						isSelected,
 						isDisabled,
@@ -117,6 +120,7 @@ export const TreeViewItem = ({
 				Template({
 					items: items,
 					size,
+					variant,
 					rootClass: "spectrum-TreeView",
 					customClasses: ["is-opened"],
 				}, context)
@@ -141,7 +145,7 @@ export const Template = ({
 			[rootClass]: true,
 			[`${rootClass}--size${size?.toUpperCase()}`]:
 				typeof size !== "undefined",
-			[`${rootClass}--${variant}`]: typeof variant !== "undefined",
+			[`${rootClass}--${variant}`]: typeof variant !== "undefined" && variant !== "default",
 			[`${rootClass}--quiet`]: isQuiet,
 			...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
 		})}
@@ -155,6 +159,7 @@ export const Template = ({
 			(item) => TreeViewItem({
 				...item,
 				size,
+				variant,
 			}, context),
 		)}
 	</ul>
--- a/components/treeview/stories/treeview.stories.js
+++ b/components/treeview/stories/treeview.stories.js
@@ -15,7 +15,14 @@ export default {
 	component: "TreeView",
 	argTypes: {
 		items: { table: { disable: true } },
-		variant: { table: { disable: true } },
+		variant: {
+			name: "Variant",
+			type: { name: "string" },
+			category: "Component",
+			defaultValue: { summary: "default" },
+			options: ["default", "detached", "thumbnail"],
+			control: "select",
+		},
 		size: size(["s", "m", "l", "xl"]),
 		isQuiet,
 	},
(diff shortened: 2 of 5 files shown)


components/treeview/stories/treeview.stories.js
```diff
@@ -15,14 +15,22 @@ export default {
 	component: "TreeView",
 	argTypes: {
 		items: { table: { disable: true } },
-		variant: { table: { disable: true } },
+		variant: {
+			name: "Variant",
+			type: { name: "string" },
+			category: "Component",
+			defaultValue: { summary: "default" },
+			options: ["default", "detached"],
```
rise-erpelding, line 23: Detached is the only non-default variant that I've included in the controls. Thumbnail is also considered a variant but that felt a little tricky since the thumbnails appear in the individual items. I'm leaving it out but I'm happy to hear any discussion or ideas related to this.

marissahuysentruyt: Also, table does the thumbnails as a control. Would anything in the [table template file](https://github.com/adobe/spectrum-css/blob/rise-erpelding/css-871-fix-detached-variant/components/table/stories/template.js) be helpful? That control for table thumbnails is a boolean, but maybe it could be adapted for use here?


https://github.com/user-attachments/assets/7252529c-c2d2-4dc9-8a46-ff871e438819

marissahuysentruyt (commented):
This is looking pretty good! Everything seems to work as you described, I see the detached variants on the docs site, in VRTs, the chromatic view, and can create it with the variant control. I left just a couple of thoughts/questions for you. 

I also didn't know where to put this question, but what does "outside of a panel" mean? Are there supposed to be left and right borders for the default treeview, sort of like the other images on the [guidance site](https://spectrum.adobe.com/page/tree-view/#Detached)? It looks like those borders are getting removed in S2, so I might have just missed that this component was already migrated!

rise-erpelding:
> I also didn't know where to put this question, but what does "outside of a panel" mean? Are there supposed to be left and right borders for the default treeview, sort of like the other images on the [guidance site](https://spectrum.adobe.com/page/tree-view/#Detached)? It looks like those borders are getting removed in S2, so I might have just missed that this component was already migrated!

Yeah, that's a great question. I'd guess that in the wild, it should have those borders or be inside of a container that has visible border lines to indicate it being inside of the container, and then the detached variant would maybe not be in that container? None of the designs have borders like that (looking back at S1 in Figma and also the older XD designs), so that makes me think those would be part of the container treeview is in.

However, I'm also wondering how heavily this component is used--[SWC](https://github.com/adobe/spectrum-web-components/tree/main/packages) doesn't appear to use it, and the only reason this detached variant is being fixed is because I was migrating the documentation and noticed that the variant didn't look right, not because someone happened to be using it downstream and noticed it was wrong. I think we could handle anything else missing (for instance, the drag icons) from this when it gets migrated to S2 unless there's a compelling reason to do otherwise.

components/treeview/stories/treeview.stories.js
```diff
@@ -15,14 +15,22 @@ export default {
...
 	argTypes: {
 		items: { table: { disable: true } },
-		variant: { table: { disable: true } },
+		variant: {
+			name: "Variant",
+			type: { name: "string" },
+			category: "Component",
+			defaultValue: { summary: "default" },
+			options: ["default", "detached", "thumbnail"],
+			control: "inline-radio",
```
marissahuysentruyt, line 24: This won't prevent me from approving, I just wanted to bring it up. Most of our instances of `inline-radio`s are 2 options (although one control in typography and tooltip don't do that). Should we do a select instead? 

I don't have particularly strong feelings either way, I feel like I read that comment somewhere to change from an inline-radio to a select once we added more options.

components/treeview/stories/template.js
```diff
@@ -93,14 +94,16 @@ export const TreeViewItem = ({
 						customClasses: [`${rootClass}-itemIcon`],
 					}, context)
 				)}
-				${when(thumbnail, () =>
+				${when(variant === "thumbnail", () =>
```
marissahuysentruyt, line 97: 👏👏👏 Glad that worked!

marissahuysentruyt (approved):
Everything checks out according to the validation instructions! We even got the thumbnail variant into the controls 🎉

I left a comment about the inline-radio vs. select, but I could go either way on that, so I'm approving.

approved-target-lady

components/treeview/stories/treeview.stories.js
rise-erpelding: Select feels more logical/natural to me, just pushed that change!

PR #3123, opened 2024-09-17
chore(stepper): update vrt snapshots

castastrophe:
## Description

Update VRT to show "invalid" as a top-level test instead of a state because it's beneficial for us to see an "invalid" snapshot in combination with all other states (disabled, hovered, focused, etc.).

Moved "hide-stepper" to a state because it's most helpful to see it as a single snapshot for default and quiet variants.

### Validation steps

- [x] Expect visual regression tests to show a change to the testing grid for the Stepper component

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/stepper/stories/stepper.test.js
+++ b/components/stepper/stories/stepper.test.js
@@ -12,14 +12,27 @@ export const StepperGroup = Variants({
 			isQuiet: true,
 		},
 		{
-			testHeading: "Hide stepper",
-			hideStepper: true,
+			testHeading: "Invalid",
+			isInvalid: true,
+		},
+		{
+			testHeading: "Quiet + invalid",
+			isQuiet: true,
+			isInvalid: true,
 		},
 	],
 	stateData: [
 		{
+			testHeading: "Hide stepper",
+			hideStepper: true,
+		},
+		{
 			testHeading: "Disabled",
 			isDisabled: true,
+		},
+		{
+			testHeading: "Hovered",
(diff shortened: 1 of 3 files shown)


components/stepper/stories/stepper.test.js
```diff
@@ -12,46 +12,43 @@ export const StepperGroup = Variants({
...
-			isInvalid: true,
+			testHeading: "Keyboard-focused",
 			isKeyboardFocused: true,
 		},
 		{
-			testHeading: "Invalid + disabled + focused",
-			isInvalid: true,
+			testHeading: "Disabled + hovered",
 			isDisabled: true,
-			isFocused: true,
```
marissahuysentruyt, line 50: Did we want to keep `Disabled + focused`? It looks maybe the `Disabled + hover` replaced this on accident.

marissahuysentruyt (commented):
I think these changes make sense! I like adding in the hover & disabled-hover test options. Those didn't even cross my mind when I migrated the docs. 🤦‍♀️

My only comment is about making sure we want to remove the `disabled-focused` case we had. I took a peek at the VRTs and things look good over there as far as I can tell.

components/stepper/stories/stepper.test.js
castastrophe: I'll add back the stepper + focus; I’d removed it because focus doesn’t really have distinct styles so I thought we might get the same validation from hover.

PR #3134, opened 2024-09-19
docs(menu): docs site to storybook

rise-erpelding:
## Description

- Makes some adjustments to the template to fix existing issues - the switches arg (`hasActions`) and `selectionMode` weren't being passed down through `MenuGroup` causing selected items to not show up as selected.
- Updates test coverage, with `include` and `ignore` we now have the ability to better control which combinations of variants/states are shown
- Displays docs site component variants on the Storybook docs page
    - Also adds additional stories/text from the [design system site](https://spectrum.adobe.com/page/menu/) to give a more thorough explanation of the component

This is a docs-only PR so no changeset is needed.

CSS-934

### Validation steps

- [ ] The Menu Storybook Docs page contains all the relevant information from the [Docs site](https://opensource.adobe.com/spectrum-css/menu.html)
- [ ] Grammar, organization, and spacing of the content on the Menu Docs page is accurate/makes sense
- [ ] Testing coverage is appropriate and covers all variants of menus/menu items, which would probably consist of something like:
    - [ ] Default, Hover, Focus, Disabled, and Active states for menu items with:
        - [ ] No icon
        - [ ] Icon
        - [ ] Description
        - [ ] Value
    - [ ] Open states for Collapsible and Drill-in Submenu items
    - [ ] Default, Hover, Focus, Disabled, and Active states for menu items with:
        - [ ] Single selection
        - [ ] Multi selection with checkboxes
        - [ ] Multi selection with switches

### Regression testing

Validate:

## To-do list


components/menu/stories/menu.stories.js
```diff
@@ -237,13 +249,16 @@ TraySubmenu.parameters = {
 };
 
 export const MenuItem = MenuItemGroup.bind({});
+MenuItem.tags = ["!autodocs"];
 MenuItem.argTypes = {
 	isDisabled,
 	isActive,
 	isFocused,
 	isHovered,
-	isOpen,
```
rise-erpelding, line 245: `isOpen` is used with menu items, but only for Collapsible and submenus, which we're not demonstrating in this story, it didn't feel like a good fit to keep this state in the controls

components/menu/stories/menu.stories.js
```diff
@@ -375,9 +402,134 @@ Collapsible.args = {
...
+Sizing.tags = ["!dev"];
+Sizing.args = {
+	items: [
+		{
+			idx: 1,
+			label: "Menu item",
+		},
+		{
+			idx: 2,
+			label: "Menu item with icon",
```
rise-erpelding, line 432: Do we like the menu being displayed like this with each item having some different kind of content (one with an icon, one with value, one with description)? I was trying to avoid needing so many stories, but can add more to display the content differently if this feels confusing.

components/menu/stories/template.js
```diff
@@ -396,3 +398,293 @@ export const Template = (
...
+				],
+				...args
+			}, context),
+			(args, context) => Popover({
+				isOpen: true,
+				position: "end-top",
+				customStyles: {
+					"--mod-popover-animation-distance": "-4px",
+					top: "-105px",
+					"inline-size": "120px",
```
rise-erpelding, line 657: I don't think we handle things like the positioning of the menu next to the submenu in our CSS, so this CSS is making some of those adjustments.

jawinn (changes requested):
Looking great so far. The documentation of the truncation class is the main thing I see here that needs adjustment.

components/menu/stories/menu.stories.js
rise-erpelding: I think another item with both icon and description is a good idea!

If the story name changes to "Default with sizing", it'll show up after "Collapsible" instead of being the first story because of how Storybook alphabetizes stories. I think I kind of prefer keeping it as "Default" for that reason, but what do you think?

components/menu/stories/template.js
```diff
@@ -392,3 +398,293 @@ export const Template = (
...
+						shouldTruncate: group.shouldTruncate || false,
+						items: group.items,
+					})}
+				`
+			})}	
+		`)
+	});
+};
+
+export const SelectionGroup = (args, context) => {
```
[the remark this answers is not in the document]
rise-erpelding: Added these in, let me know what you think:
image

components/menu/stories/menu.stories.js
```diff
@@ -375,9 +402,134 @@ Collapsible.args = {
...
+export const TextOverflow = OverflowGroup.bind({});
+TextOverflow.tags = ["!dev"];
+TextOverflow.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+TextOverflow.args = {
+	customStyles: {
+		"max-inline-size": "150px",
+	}
+};
```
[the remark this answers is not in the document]
rise-erpelding: I incorporated the mention of the `--truncate` class and mentioned that `inline-size`/`max-inline-size` needs to be set, I think this now includes the relevant information from the docs site. I'm not really clear on why there was text on truncation for the drill-ins, but added examples to show truncation with drill-ins as well as values, icons, and section headings for good measure.
image

components/menu/stories/menu.stories.js
jawinn: 👍 I think keeping as "Default" is fine then.

components/menu/stories/menu.stories.js
```diff
@@ -375,9 +406,147 @@ Collapsible.args = {
...
+ * text to a new line.
+ *
+ * Menu item labels and headings can be truncated with an ellipsis by using the `spectrum-Menu-itemLabel--truncate`
+ * class. This class should be applied to the item label or section heading that requires truncation within a menu with
+ * a set `inline-size` or `max-inline-size`. When text is truncated, it should be revealed with a tooltip on hover
+ * (not displayed here).
+ *
+ * Truncation can be used in conjunction with icons, values, and drill-ins in a menu item. Note that descriptions
+ * within a menu item always wrap by default and do not have a truncation option.
+ */
```
jawinn, line 530:
```suggestion
/**
 * When a menu item’s label or description exceed the available horizontal space, the default behavior is to wrap the
 * text to a new line.
 *
 * Menu item labels and headings can be truncated with an ellipsis by using the `spectrum-Menu-itemLabel--truncate`
 * class. Truncation can only occur if the menu has
 * a set `inline-size` or `max-inline-size`, or it is constrained by the width of its parent element(s). For demonstration purposes, a `max-inline-size` is set on the following menu examples. When text is truncated, it should be revealed with a tooltip on hover
 * (not displayed here).
 *
 * Truncation can be used in conjunction with icons, values, and drill-ins in a menu item. Note that descriptions
 * within a menu item always wrap by default and do not have a truncation option.
 */
```
It doesn't necessarily need an `inline-size` or `max-inline-size`. It could also be constrained by the width of its parent. We could also drop that sentence entirely instead of trying to explain all situations where text overflow would occur. I think it's worth noting when we are setting inline styles to show certain behavior in the docs, so I also added a sentence for that.

jawinn (approved):
Looks great! Approved, with one last suggestion for the truncation description.

components/menu/stories/menu.stories.js
```diff
@@ -190,12 +86,132 @@ export default {
...
+ * When a menu is displayed within a tray, a submenu will replace the tray content when the parent menu item is
+ * selected. A submenu displays a back button (labeled by the title of the parent item) at the top of the tray to
+ * return the user to the previous level of the menu. The back arrow size scale used with the various menu sizes are
+ * small: 200, medium: 300, large: 400, and extra large: 500.
+ */
 export const TraySubmenu = MenuTraySubmenu.bind({});
+TraySubmenu.storyName = "Submenu in tray";
 TraySubmenu.argTypes = {
 	selectionMode: { table: { disable: true } },
 	hasDividers: { table: { disable: true } },
```
marissahuysentruyt, line 217: This is a very interesting way to do this for each variant. I think this is the first time I'm seeing changes to the `argTypes` within a specific story. Huh!!

components/menu/stories/menu.test.js
```diff
@@ -84,7 +85,7 @@ export const MenuWithVariants = Variants({
 					items: [{ label: "Defaults to not visible within closed item" }],
 				},
 				{
-					label: "Watches",
+					label: "Watches and longer truncated label that is really really much longer",
```
marissahuysentruyt, line 88: "watches" 🤭 Some of the old content is really funny

components/menu/stories/menu.test.js
```diff
@@ -196,23 +218,44 @@ export const MenuItemGroup = Variants({
...
+			include: ["Truncation", "Text wrapping"],
+		},
+		{
+			testHeading: "Without description",
+			description: undefined,
+			include: ["Drill-in", "Truncation", "Text wrapping"],
+		},
 		// {
 		// 	testHeading: "Highlighted",
 		// 	isHighlighted: true,
```
marissahuysentruyt, line 261: What is this `isHighlighted`? Maybe that's like...focused? Or selected?

components/menu/stories/template.js
```diff
@@ -217,6 +219,7 @@ export const MenuItem = (
 
 export const MenuGroup = (
 	{
+		hasActions = false,
```
marissahuysentruyt, line 222: Does this need to be in the `argTypes` at all? I saw it in [the `spectrum-two` menu](https://github.com/adobe/spectrum-css/blob/10392df161f93293cf04b38f2d49009c3fb6d998/components/menu/stories/menu.stories.js#L45), so it might not need to move into this branch. (or it's just an arg for the MenuItem)

marissahuysentruyt (approved):
I think Josh caught lots of good things, and it looks like you've taken care of all of his comments/suggestions! I left a couple of questions in the code (nothing to stop me from approving). The only thing I think I saw in the s2 menu was this [`max-inline-size` arg,](https://github.com/adobe/spectrum-css/blob/10392df161f93293cf04b38f2d49009c3fb6d998/components/menu/stories/menu.stories.js#L35) but that may not need to be migrated over just yet. It might be specific to s2.

The S2 menu work was already a beast, and then you you took this on too- phew! You're officially our menu expert!

components/menu/stories/menu.test.js
rise-erpelding: I wasn't sure on this one either, so I left it be.

components/menu/stories/template.js
rise-erpelding: Yeah, it's just an `argType` for `MenuItem`. Menu is weird because it has states/args that apply to the full menu, but then also ones for the individual items, and in a lot of cases there's not a lot of overlap.

PR #3143, opened 2024-09-19
fix(storybook): updates to pass the correct content types to the dialog and popover components

cdransf:
## Description

- [x] Returns the html passed to the popover rather than simply passing html`` to resolve `Error: expected template strings to be an array` and provide the data in the expected type/shape.
- [x] Passes a simplified string to the documented Dialog component (errors were caused by nested `Typography` components and inconsistent example data.

### Validation steps

1. Check out the branch and run Storybook locally (or navigate to the Storybook URL for this PR).
2. Open the docs section of the Dialog component and verify that all examples render.
3. Open the docs section of the Popover component and verify that all examples render.

### Regression testing

Validate:

## Screenshots

Screenshot 2024-09-19 at 2 47 15 PM
Screenshot 2024-09-19 at 2 47 02 PM
Screenshot 2024-09-19 at 11 48 04 AM

## To-do list

### DIFF
--- a/components/button/stories/template.js
+++ b/components/button/stories/template.js
@@ -110,7 +110,7 @@ export const Template = ({
 export const ButtonsWithIconOptions = ({
 	iconName,
 	...args
-}) => Container({
+}, context = {}) => Container({
 	withBorder: false,
 	direction: "row",
 	wrapperStyles: {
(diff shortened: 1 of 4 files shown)


rise-erpelding (commented):
Hey, this fixes the issue with this story on the docs page for popover, which is awesome!

I don't love that we need to remove the `inline: false` on dialog but agree that it's probably the best possible option and I like that you left it in as a TODO comment for us to handle later.

A couple of other things I noticed about dialog that maybe warrant some further discussion:
* We can't dismiss disimissible dialogs now, which is probably an ok trade-off
* The story fits differently into its container without `inline: false`, like it's kind of wedged in between the boundaries of its container:
image
We could adjust the height to give those stories more space, **but** I also noticed comparing with Firefox that the stories are all stuck to the top of the frame in prod, instead of having some top padding:
image

Given that, maybe it's ok that the stories look wedged in now and we might not want to give them more space? I'd love to hear your thoughts, and if anyone else wants to weigh in with their opinions!

cdransf:
> Hey, this fixes the issue with this story on the docs page for popover, which is awesome!
> 
> I don't love that we need to remove the `inline: false` on dialog but agree that it's probably the best possible option and I like that you left it in as a TODO comment for us to handle later.
> 
> A couple of other things I noticed about dialog that maybe warrant some further discussion:
> 
> * We can't dismiss disimissible dialogs now, which is probably an ok trade-off
> * The story fits differently into its container without `inline: false`, like it's kind of wedged in between the boundaries of its container:
>   image
>   We could adjust the height to give those stories more space, **but** I also noticed comparing with Firefox that the stories are all stuck to the top of the frame in prod, instead of having some top padding:
>   image
> 
> Given that, maybe it's ok that the stories look wedged in now and we might not want to give them more space? I'd love to hear your thoughts, and if anyone else wants to weigh in with their opinions!

Good call outs! I'm inclined to address those changes once we fix the issues caused by restoring `inline: false` since we can attribute the changes to removing that.

I'm also seeing `TypeError: updateArgs is not a function. (In 'updateArgs({ isPending: true })', 'updateArgs' is undefined)` when I click `Cancel` on `main` and this branch.

rise-erpelding (approved):
I feel good about Dialog/Popover now that I've heard some of your thoughts on it!

And I'm seeing the fix you added with button group, nice work!

cdransf:
> I feel good about Dialog/Popover now that I've heard some of your thoughts on it!
> 
> And I'm seeing the fix you added with button group, nice work!

Thank you! I was debating taking a look at the dismiss issue too, but we know what _will_ fix it since it's failing silently, so we can run with it. ✨

PR #3154, opened 2024-09-23
fix(slider ramp): correct handle and ramp vertical alignment

cdransf:
## Description

This applies a flex layout to the spectrum slider controls to consistently align the enclosed handle and ramp, while also removing the margins that might otherwise interfere with alignment. This aims to resolve the issues with the calc-based approach that exhibited variations in vertical alignment.

Checked locally in Storybook.

### Validation steps

1. Pull down the branch and run Storybook locally or access the Storybook URL for the PR.
2. Open the Slider component and select `Default`.
3. Verify that the handle is vertically aligned over the track.

### Regression testing

Validate:

## Screenshots

Screenshot 2024-09-30 at 3 56 14 PM

## To-do list

### DIFF
--- a/components/slider/index.css
+++ b/components/slider/index.css
@@ -152,6 +152,11 @@
 	position: relative;
 	z-index: auto;
 
+	&:not(:has(.spectrum-Slider-ticks)) {
+		display: flex;
+		align-items: center;
+	}
+
 	/* These calculations prevent the track from spilling outside of the control */
 	inline-size: calc(100% - calc(var(--mod-slider-controls-margin, var(--spectrum-slider-controls-margin))) * 2);
 	margin-inline-start: var(--mod-slider-controls-margin, var(--spectrum-slider-controls-margin));
--- a/components/slider/index.css
+++ b/components/slider/index.css
@@ -116,6 +116,10 @@
 	min-inline-size: var(--mod-slider-min-size, var(--spectrum-slider-min-size));
 
 	user-select: none;
+
+	&:not(.spectrum-Slider--sideLabel) .spectrum-Slider-labelContainer + .spectrum-Slider-controls:has(.spectrum-Slider-ramp) {
+		margin-block-start: calc(var(--mod-slider-ramp-track-height, var(--spectrum-slider-ramp-track-height)) / 2);
+	}
 }
 
 .spectrum-Slider--sideLabel {
@@ -416,6 +420,7 @@
 	&:lang(ko) {
 		line-height: var(--mod-slider-cjk-line-height, var(--spectrum-slider-cjk-line-height));
 	}
+
 }
 
 .spectrum-Slider-label {
--- a/components/slider/index.css
+++ b/components/slider/index.css
@@ -289,7 +289,6 @@
 }
 
 .spectrum-Slider-ramp {
-	margin-block-start: calc(var(--mod-slider-ramp-track-height, var(--spectrum-slider-ramp-track-height)) / 2);
 	block-size: var(--mod-slider-ramp-track-height, var(--spectrum-slider-ramp-track-height));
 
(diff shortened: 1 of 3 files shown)


components/slider/index.css
```diff
@@ -284,7 +289,6 @@
 }
 
 .spectrum-Slider-ramp {
-	margin-block-start: calc(var(--mod-slider-ramp-track-height, var(--spectrum-slider-ramp-track-height)) / 2);
```
marissahuysentruyt, line 287: Do we need to preserve this spacing between the ramp container and the labels?  Without it, both containers touch. To me, it looks like this spacing will need to be defined in S2 as well (just based on the "spacing (field label to slider control" and the "spacing (top/bottom edge to label)" headings). Maybe instead of removing it completely, we can move it elsewhere? Into a different selector block? I don't even know if that's possible, you probably know better than me at this point, but I was just trying think of ways to avoid any VRT changes.

rise-erpelding (approved):
Looked at all sizes, small and XL particularly are a great improvement!
image
image

components/slider/index.css
cdransf: Alrighty! I've added the following to the root `.spectrum-Slider` class:

```css
&:not(.spectrum-Slider--sideLabel) .spectrum-Slider-labelContainer + .spectrum-Slider-controls:has(.spectrum-Slider-ramp) {
  margin-block-start: var(--mod-spectrum-slider-value-side-padding-block, var(--spectrum-slider-value-side-padding-block));
}
```

Adding the `8px` gap called for in the Figma for S2 to the `.spectrum-Slider-controls` provided the component is _not_ in the `side label` configuration but _is_ using the `ramp` design. I'm wondering if that selector is too verbose or perhaps too fragile, but it does need to balance all of those different cases and preserve the spacing without being applied to the `.spectrum-Slider-ramp` class directly as that does cause the misalignment with the slider control to crop back up.

components/slider/index.css
```diff
@@ -22,6 +22,7 @@
 	--spectrum-slider-handle-border-width-down: var(--spectrum-slider-handle-border-width-down-medium);
 	--spectrum-slider-label-top-to-text: var(--spectrum-component-top-to-text-75);
 	--spectrum-slider-control-to-field-label: var(--spectrum-slider-control-to-field-label-medium);
+	--spectrum-slider-value-side-padding-block: var(--spectrum-spacing-100);
```
marissahuysentruyt, line 25: Does this `--spectrum-slider-value-side-padding-block` actually need to be that whole `calc` from the `margin-block-start` instead of just `var(--spectrum-spacing-100)`? My thought is that the margin we removed was being calculated with `--spectrum-slider-ramp-track-height`, which is 16px for desktop and 20px for mobile. So I believe the margin was supposed to change when we toggle the "large" platform scale, so it would need to be 10px on mobile instead of 8px.

Screenshot 2024-10-01 at 1 21 20 PM
Screenshot 2024-10-01 at 1 12 21 PM

We should probably double check with Patrick & Cassondra (maybe even design) because it does look like this spacing might be staying at 8px for the ramp slider in S2 (regardless of desktop vs mobile). Am I [looking at everything right](https://opensource.adobe.com/spectrum-css/preview/index.html?path=/story/components-slider--default&args=variant:ramp)? Is preserving that padding now worth it, if it's going to change (basically to how you have it currently with `spacing-100`) soon-ish?

marissahuysentruyt (approved):
Thanks for digging into this one! When I wrote up that bug card, I had no idea how many scenarios we'd be trying to account for, so thanks for the persistence! 

I'm going to approve, but I do believe there are some selectors and mods in the metadata files that were built that no longer apply (since we removed that created custom property). I'm assuming those should probably be removed.

PR #3160, opened 2024-09-24
docs(actiongroup): migrating docs to storybook

aramos-adobe:
## Description

Adding horizontal, vertical, and justified layouts

### Regression testing

Validate:

## To-do list


rise-erpelding (changes requested):
This looks really nice! I really like the way that you've grouped things (with the default, icon-only, quiet variants all together) to keep the number of variants from feeling overwhelming or confusing. I also really like that there are lots of examples (particularly in the sizing) of icons & text together in an action button. And the Overflow section and testing coverage is such a great addition! Overall this is such a huge improvement over the old docs site.

There were two small things I noticed that I think could use improvement, but I'm happy to hear arguments to the contrary:
- We're not showing the emphasized variant for action group. Since we generally do show it for components that have it, and there's a control for it, and we're testing it, it seems like something we ought to be displaying on our Docs page.
- For testing coverage, there are lots of variants, I don't know that every single combination needs to be tested (for instance icon-only compact, or compact quiet, or justified compact, etc.), but one case that I think would be beneficial would be more for vertical action groups, particularly vertical compact.

components/actiongroup/stories/actiongroup.test.js
```diff
@@ -25,5 +25,36 @@ export const ActionGroups = Variants({
...
+					label: "Cut",
+				},
+				{
+					iconName: "Move",
+					iconSet: "workflow",
+					label: "Move",
+				},
+			]
+		}
 	],
```
marissahuysentruyt, line 59: I think we could add some coverage for the icon-only and text-only options. Tabs has those same 3 options (text-only, text + icon, icon-only), so [we used the `stateData`](https://github.com/adobe/spectrum-css/blob/7e8f6114aef049b82904cebf6a112aa5a2be0feb/components/tabs/stories/tabs.test.js#L86) to cover all of them.

Screenshot 2024-10-02 at 1 59 51 PM

components/actiongroup/stories/template.js
```diff
@@ -43,9 +43,141 @@ export const Template = ({
 					isQuiet: areQuiet,
 					isEmphasized: areEmphasized,
 					customClasses: [`${rootClass}-item`],
+					size: size
```
marissahuysentruyt, line 46: Ohhh great catch here! 👏👏👏

components/actiongroup/stories/actiongroup.stories.js
```diff
@@ -65,27 +85,145 @@ export default {
...
+			label: "",
+		},
+	]
+};
+
+JustifiedIconOnlyCompact.storyName = "Justified (compact, icon-only)";
+
+
+/**
+ * When space is limited in an action group, there are 2 options for the group's overflow behavior: wrap or collapse. By default, an action group is set to wrap, meaning that the action buttons inside the group wrap to form another line. Alternatively, an action group can be set to collapse inside a More (...) action button.
```
marissahuysentruyt, line 228: Do you think we need backticks or something for the 'More (...)' part? I don't think it'd be recommended to use quotes, but maybe backticks would draw attention to that. Or we could wrap it in `**`/a `<strong>` tag like it is on the guidance site? Any thoughts on that?

components/actiongroup/stories/actiongroup.stories.js
```diff
@@ -1,11 +1,13 @@
...
 import pkgJson from "../package.json";
 import { ActionGroups } from "./actiongroup.test.js";
+import { OverflowOption, TreatmentTemplate } from "./template.js";
 
 /**
- * The action group component is a collection of action buttons.
+ * An action group is a grouping of [action buttons](?path=/docs/components-action-button--docs&globals=testingPreview:!true) that are related to each other.
  */
 export default {
 	title: "Action group",
```
marissahuysentruyt, line 13: oooo! What do you think about adding an icon-only control? Similar to [tabs](https://github.com/adobe/spectrum-css/blob/7e8f6114aef049b82904cebf6a112aa5a2be0feb/components/tabs/stories/tabs.stories.js#L74)? That would give users the option to see their icon-only action buttons. 

Screenshot 2024-10-02 at 2 12 43 PM

marissahuysentruyt (changes requested):
Nice work, this one is coming along!

I think I agree with Rise, that if we're testing the `isEmphasized`, maybe we should showcase it on the docs page. (which I think you actually just pushed up a fix for). I had some similar thoughts around the `isDisabled`. 

I had some comments on adding controls too, so let me know what you think!

rise-erpelding (approved):
I can see lots of those extra suggestions were added--this looks great!

components/actiongroup/stories/actiongroup.stories.js
```diff
@@ -43,16 +46,45 @@ export default {
 			},
 			control: "boolean",
 		},
+		iconOnly: {
+			name: "Icon Only",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Advanced",
```
marissahuysentruyt, line 54: Does `iconOnly` belong in "Advanced", or is it just "Component?" I don't really know or have much of an opinion, just curious. Some other components that have some similar icon controls, like contextual help, tabs, and menu (technically menu item) have it in "Component" or "Content." Just a thought for you!

marissahuysentruyt (changes requested):
Thanks for making those updates! I found a few more updates I think we should make 😊

marissahuysentruyt (approved):
@aramos-adobe thanks for making all of the updates! This looks fantastic! I think all of the additions gives users so much context and explanation for this component now, in comparison to the current docs site! 

I did a quick check of the CHANGELOG and I think the migration guide is all included, so boom- approved!

PR #3164, opened 2024-09-24
fix: blank field group docs and read-only radio alignment

jawinn:
## Description

Includes two small fixes:

1. The field group "Docs" page was blank. It was missing the default arg value for `items` in its Template, causing nothing to render within the field group markup.
2. The read-only radios on the field group Docs page were displaying on the completely wrong side. This issue also existed on the Chromatic template for radio. This was because of `margin-inline-start` being set to `auto` instead of `0` for read-only. In a flex container, `auto` on the margin of a flex item will automatically extend its specified margin to occupy the extra space.

### Validation steps

- [x] Stories on the field group "Docs" page in Storybook are now appearing @marissahuysentruyt 
- [x] Read-only radios in field group are no longer aligned to the right. Fix is visible in VRT changes for field group. @marissahuysentruyt 

### Regression testing

Validate:

## Screenshots

**Read-only bug. Previous behavior:**
Screenshot 2024-09-24 at 4 26 20 PM

Screenshot 2024-09-24 at 4 25 33 PM

**Fixed behavior:**
Screenshot 2024-09-24 at 4 45 58 PM

Screenshot 2024-09-24 at 4 47 28 PM

## To-do list

### DIFF
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -62,6 +62,27 @@ export default {
 		inputType: "radio",
 		labelPosition: "top",
 		layout: "vertical",
+		label: "Select one of the following options:",
+		items: [
+			(passthroughs, context) => Radio({
+				...passthroughs,
+				id: "apple",
+				label: "Apples are best",
+				customClasses: ["spectrum-FieldGroup-item"],
+			}, context),
+			(passthroughs, context) => Radio({
+				...passthroughs,
+				id: "banana",
+				label: "Bananas forever",
+				customClasses: ["spectrum-FieldGroup-item"],
+			}, context),
+			(passthroughs, context) => Radio({
+				...passthroughs,
+				id: "pear",
+				label: "Pears or bust",
+				customClasses: ["spectrum-FieldGroup-item"],
+			}, context),
+		],
 		isInvalid: false,
 		isRequired: false,
 		isReadOnly: false,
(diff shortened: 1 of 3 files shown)


components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -62,6 +62,27 @@ export default {
...
 		labelPosition: "top",
 		layout: "vertical",
+		label: "Select one of the following options:",
+		items: [
+			(passthroughs, context) => Radio({
+				...passthroughs,
+				id: "apple",
+				label: "Apples are best",
+				customClasses: ["spectrum-FieldGroup-item"],
+			}, context),
```
marissahuysentruyt, line 72: We'll have to edit some of the `items` for certain stories. Right now, all of the checkbox variants are showing radio buttons instead of checkboxes.

Screenshot 2024-09-26 at 9 14 19 AM

marissahuysentruyt (changes requested):
Like you said, 2 for 1! The radio `margin-inline-start` fix realigns everything properly for ltr and rtl. The only we should do is pass some checkboxes to the checkbox stories! Thanks for cleaning this up!

components/fieldgroup/stories/fieldgroup.stories.js
marissahuysentruyt: Also, does the toggle for `inputType` need to do anything else but change the `role` from `radiogroup` to `group`? It's doing that right now, all is good- I just expected it to change the component shown, but that seems like work for another PR.

rise-erpelding (commented):
Now that you've brought the Docs back on FieldGroup, there are lots of other things going on here with the controls particularly that have my attention. Here's what I've noticed (I already mentioned some of these below):

- We have a stray `label` control in the controls
- We lost the Help text, which maybe is ok (although we are testing it here in Chromatic, so it seems like we ought to keep it), but since that was the only thing that visibly changed when the `invalid` control is turned on, now the `invalid` control doesn't appear to do anything.
- We gained a `required` control at some point after the docs to storybook for this was completed and that doesn't seem to do anything, I think maybe we just need to pass that arg down to the FieldLabel in the template.
- Our controls for checkbox/radio don't do anything, we can only see radios
- The testing preview _does_ show checkboxes and radios but shows checkboxes only with horizontal positioning (and doesn't show radios), do we think this is sufficient to test them? It feels to me like we could be missing something by not testing radios _and_ checkboxes in **all** positions, but I could be convinced otherwise.

Addressing these separately would definitely be ok as far as I'm concerned since these aren't things that broke in this PR and they weren't in the original scope of the work, so I don't know that all of this needs to be fixed before we merge, but definitely all things we should address with fieldgroup at some point!

components/fieldgroup/stories/fieldgroup.stories.js
jawinn: This is beyond the scope of this PR, but thanks for pointing out these regressions. I've added this to follow-up **CSS-969**.

jawinn:
> Now that you've brought the Docs back on FieldGroup, there are lots of other things going on here with the controls particularly that have my attention...

Thanks for calling these out. I've added this to the followup Jira issue **CSS-969**, since as you mentioned this is outside the scope of this PR.

rise-erpelding (approved):

marissahuysentruyt (approved):
Thanks for creating the follow up card!

PR #3175, opened 2024-09-27
docs(switch): docs site to storybook

rise-erpelding:
## Description
Updates the Switch Docs page in Storybook to align more closely with [Docs site](https://opensource.adobe.com/spectrum-css/switch.html), incorporating [guidelines](https://spectrum.adobe.com/page/switch/) to add extra information.

Testing coverage was examined but no changes were made.

Docs-only, no changeset necessary.

[CSS-940](https://jira.corp.adobe.com/browse/CSS-940)

### Validation steps
- [ ] The Switch storybook docs page contains all the relevant information from the [docs site](https://opensource.adobe.com/spectrum-css/switch.html)
- [ ] Grammar, capitalization, and organization of content is accurate/makes sense.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/switch/stories/switch.stories.js
+++ b/components/switch/stories/switch.stories.js
@@ -1,11 +1,12 @@
+import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isChecked, isDisabled, isEmphasized, size } from "@spectrum-css/preview/types";
 import pkgJson from "../package.json";
 import { SwitchGroup } from "./switch.test.js";
-import { Template } from "./template.js";
+import { DocsSwitchGroup, Template } from "./template.js";
 
 /**
- * A switch is used to turn an option on or off. Switches allow users to select the state of a single option at a time.
+ * A switch is used to turn an option on or off. Switches allow users to select the state of a single option at a time and are best used for communicating activation.
  */
 export default {
 	title: "Switch",
@@ -37,6 +38,7 @@ export default {
 };
 
 export const Default = SwitchGroup.bind({});
+Default.tags = ["!autodocs"];
 Default.args = {};
 
 // ********* VRT ONLY ********* //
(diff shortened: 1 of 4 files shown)


components/switch/stories/switch.stories.js
```diff
@@ -50,31 +52,78 @@ WithForcedColors.parameters = {
...
 };
 
-export const Disabled = Template.bind({});
-Disabled.tags = ["!dev"];
-Disabled.args = {
-	isDisabled: true
+export const Sizing = (args, context) => Sizes({
+	Template,
+	...args,
+}, context);
```
castastrophe, line 111:
```suggestion
export const Sizing = (args, context) => Sizes({
	Template,
	withHeading: false,
	withBorder: false,
	...args,
}, context);
```

Using the no heading, no border variant reduces duplication for the docs page so only 1 sizing heading appears.

castastrophe (changes requested):
I noticed you'd left off the migration guide, was that intentional?

marissahuysentruyt (changes requested):
This is looking great! I had a few very minor thoughts about headings & borders for some of the containers, but overall, I see all of the documentation! I had two other thoughts I wanted to run by you.

Thing 1: Could we move some of the "Migration Guide" into the changelog? Some of it looks like it is already there (like the t-shirt sizing and the focus ring), but maybe making sure the name change from Toggle to Switch is more obvious in the changelog would be nice. I don't think the section about quiet/emphasized is in the changelog either.

Thing 2: This is completely an opinion here, but is the word "checked" in our controls for switch the "correct" word? The `checked` attribute gets added, and the switch is just a checkbox input to begin with, so I get why we're using the word "checked." However, in the docs, we don't really reference the component as being "checked" at all (I don't think), and instead we reference it being selected or not. 

Maybe we could just change the display name of the control. Something like: 
```
isChecked: {
			...isChecked,
			name: "Selected",
		},
```

I believe that's doing everything the same, but it changes the control name in the UI to Selected.

components/switch/stories/switch.stories.js
```diff
@@ -50,31 +52,78 @@ WithForcedColors.parameters = {
 };
 
 // ********* DOCS ONLY ********* //
-export const Emphasized = Template.bind({});
+/**
+ * Switches can either be selected or not selected. They cannot be in an indeterminate state (unlike
+ * [checkboxes](?path=/docs/components-checkbox--docs)). When a switch represents multiple values that are not
+ * identical (mixed values), the switch should appear as not selected.
```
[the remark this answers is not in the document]
rise-erpelding: I ended up sticking it with Standalone, it felt like it flowed better to me

components/switch/stories/switch.stories.js
```diff
@@ -50,31 +55,80 @@ WithForcedColors.parameters = {
...
 };
 
-export const Checked = Template.bind({});
-Checked.tags = ["!dev"];
-Checked.args = {
-	isChecked: true
+/**
+ * When the label is too long for the horizontal space available, it wraps to form another line.
+ */
+export const WithLongerLabel = DocsSwitchGroup.bind({});
```
marissahuysentruyt, line 100: Should we do a truncation test case?

marissahuysentruyt (approved):
Looking good! Thanks for all of the updates. My only comment was about whether or not you wanted to add a wrapping/truncation test case. I'm going to approve it and leave that up to you 👍

components/switch/stories/switch.stories.js
rise-erpelding: Just added one for a wrapping label!

components/switch/stories/switch.test.js
```diff
@@ -22,9 +22,15 @@ export const SwitchGroup = Variants({
 			isChecked: true,
 		},
 		{
-			testHeading: "Checked and Disabled",
+			testHeading: "Checked and disabled",
```
marissahuysentruyt, line 25: good catch!!

components/switch/stories/switch.test.js
```diff
@@ -22,9 +22,15 @@ export const SwitchGroup = Variants({
...
 			isChecked: true,
 			isDisabled: true,
 		},
+		{
+			testHeading: "With longer label",
```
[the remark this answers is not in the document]
rise-erpelding: Ok! I see that other components like Badge, Action Button, and Radio are doing this so I'll follow suit. I didn't name it "Truncation" though because it wraps--if I should for some reason, let me know!

castastrophe (approved):

PR #3195, opened 2024-10-02
feat(storybook): modifiers table added to docs

castastrophe:
## Description

This creates a custom DocBlock for our docs that displays the boilerplate text for mods on the docs site. I called this component a `PropertiesTable` with the hope that we can eventually expand this component to document all the custom properties being used by the component.

### Validation steps

@jawinn 

- [x] Expect to see a new section added to each component between the ArgsTable and the TaggedReleases (suggest picking 5 pages at random to validate).
- [x] Expect heading "Modifiable custom properties" show a link on hover and when the icon is selected, a direct link to that section is added to the window location.
- [x] Expect pages like actionmenu to have no custom properties table (as it has no custom properties).

### Regression testing

Validate:

## Screenshots

Screenshot 2024-10-02 at 5 51 12 PM

Screenshot 2024-10-02 at 5 51 58 PM

## To-do list

### DIFF
--- a/.storybook/blocks/PropertiesTable.jsx
+++ b/.storybook/blocks/PropertiesTable.jsx
@@ -1,0 +1,66 @@
+import "@spectrum-css/table";
+
+import { DocsContext, useOf } from "@storybook/blocks";
+import { ResetWrapper } from "@storybook/components";
+import { NAVIGATE_URL } from "@storybook/core-events";
+import { styled } from "@storybook/theming";
+import React, { useContext } from 'react';
+import { Body, Code, LinkableHeading } from "./Typography.jsx";
+
+export const Table = styled.table`
+	--mod-table-cursor-row-default: auto;
+	padding-block: 10px;
+`;
+
+/**
+ * Displays the modifiable custom properties for a component based on the metadata provided in the story.
+ * The story metadata object is imported from the "metadata.json" file in the component's directory.
+ *
+ * If the metadata object does not contain a "modifiers" array, this component will not render.
+ *
+ * Usage of this doc block within MDX template(s):
+ *  <PropertiesTable />
+ */
+export const PropertiesTable = () => {
+	const context = useContext(DocsContext);
+	const storyMeta = useOf("meta");
+
+	const packageJson = storyMeta?.csfFile?.meta?.parameters?.packageJson ?? {};
+	const metadata = storyMeta?.csfFile?.meta?.parameters?.metadata ?? {};
+
+	if (!packageJson?.name) return;
+	if (!metadata?.modifiers || !metadata?.modifiers.length) return;
+
+	return (
+		<ResetWrapper>
+			<LinkableHeading id="modifiable-properties" size="m">
+				<a aria-hidden="true" href="#modifiable-properties" tabIndex="-1" target="_self" onClick={() => {
+					context.channel.emit(NAVIGATE_URL, "#modifiable-properties");
+				}}></a>
+				Modifiable custom properties
+			</LinkableHeading>
+			<Body>
+				These are empty CSS custom property hooks available in this component
+				that enable one-off customizations specific to a product implementation.
+			</Body>
+			<Table className="docblock-properties-table sb-unstyled spectrum-Table spectrum-Table--sizeL spectrum-Table--compact">
+				<thead className="spectrum-Table-head">
+					<tr>
+						<th className="spectrum-Table-headCell">Property</th>
+					</tr>
+				</thead>
+				<tbody className="spectrum-Table-body">
+					{metadata?.modifiers.map((propertyName) => (
+						<tr key={propertyName} className="spectrum-Table-row">
+							<td className="spectrum-Table-cell">
+								<Code backgroundColor={"transparent"} size="s">
+									{propertyName}
+								</Code>
+							</td>
+						</tr>
+					))}
+				</tbody>
+			</Table>
+		</ResetWrapper>
+	);
+};
--- a/.storybook/project.json
+++ b/.storybook/project.json
@@ -27,7 +27,7 @@
 					"outputs": ["{workspaceRoot}/dist/preview"]
 				}
 			},
-			"dependsOn": ["clean", "^build"],
+			"dependsOn": ["clean", "^build", { "target": "report", "tag": "component" }],
 			"executor": "nx:run-commands",
 			"inputs": ["tools", { "externalDependencies": ["storybook"] }],
 			"options": {
@@ -85,7 +85,7 @@
 		},
 		"start": {
 			"cache": true,
-			"dependsOn": ["^build"],
+			"dependsOn": [{ "target": "report", "projects": "tag:component" }, "^build"],
 			"executor": "nx:run-commands",
 			"inputs": [
 				"tools",
--- a/components/progressbar/stories/progressbar.stories.js
+++ b/components/progressbar/stories/progressbar.stories.js
@@ -1,6 +1,8 @@
 import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isIndeterminate, size, staticColor } from "@spectrum-css/preview/types";
+import metadata from "../metadata/metadata.json";
+import packageJson from "../package.json";
 import { ProgressBarGroup } from "./progressbar.test.js";
 import { IndeterminateGroup, Template } from "./template.js";
 
@@ -36,7 +38,6 @@ export default {
 			table: {
 				type: { summary: "string" },
 				category: "Component",
-
 				// todo: side labels are not supported according to S1 documentation, but will be in S2. Remove the disable: true once S2 is released
 				disable: true,
 			},
@@ -89,6 +90,10 @@ export default {
 		customWidth: 192,
 		isIndeterminate: false,
 		showValueLabel: true,
+	},
+	parameters: {
+		packageJson,
+		metadata,
 	},
 };
 
--- a/components/stepper/stories/stepper.stories.js
+++ b/components/stepper/stories/stepper.stories.js
@@ -1,9 +1,10 @@
 import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isDisabled, isFocused, isHovered, isInvalid, isKeyboardFocused, isQuiet, size } from "@spectrum-css/preview/types";
-import pkgJson from "../package.json";
+import metadata from "../metadata/metadata.json";
+import packageJson from "../package.json";
 import { StepperGroup } from "./stepper.test.js";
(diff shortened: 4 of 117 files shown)


.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,205 @@
...
+		if (container) container.innerHTML = content;
+		else return;
+	}
+
+	function toggleToast(toast, open = true) {
+		toast.classList.toggle("is-open", open);
+	}
+
+	navigator.clipboard.writeText(text).then(
+		() => {
```
jawinn, line 117: There's something going on that is causing this to be called multiple times on click. After clicking one of the mod properties, I get 5 console logs with the same info, instead of one.

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,205 @@
...
+		// Capture all the table body cells and add a click event to copy the text to the clipboard.
+		const tableBodyCells = document.querySelectorAll(
+			".docblock-properties-table tbody td",
+		);
+		tableBodyCells.forEach((cell) => {
+			cell.addEventListener("click", () => {
+				if (cell.textContent) writeTextToClipboard(cell.textContent);
+			});
+		});
+	});
```
jawinn, line 171: Instead of an event listener on each cell, this could be a single event listener on the parent, that bubbles up from the click event (then also check `event.target.tagName === "TD"`).

jawinn (changes requested):
The table showing the list of mods is a great addition.

The two bigger things I noticed when validating:
- The anchor link does not appear in the URL after clicking the "Modifiable custom properties" heading, like it does when clicking the other headings on the page
- The copy event is logging multiple times for a single click (noted below)

Personally, the copy to clipboard and toast functionality feels excessive to me, for a text list of properties. Would users be copying them that often?

It also may be a little unexpected behavior right now, as the rows show as clickable but without indication that a click would result in changing the contents of the clipboard; one might expect a click would show or link to more information, and it's only clear what will happen after you click one.

.storybook/project.json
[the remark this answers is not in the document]
castastrophe: It ensures the report is run before the storybook is built so that every component has the latest metadata.json available for the doc block.

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,205 @@
...
+			const toast = document.getElementById("clipboard-success");
+			const content = `Copied <em>${text}</em> to clipboard`;
+
+			updateMessage(toast, content);
+			toggleToast(toast);
+			setTimeout(() => {
+				toggleToast(toast, false);
+			}, 3000);
+
+			console.log(toast, content);
```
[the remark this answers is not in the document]
castastrophe: I kind of borrowed that functionality from the [SWC docs](https://opensource.adobe.com/spectrum-web-components/components/accordion/api/#attributes%20and%20properties_density) but I'm not in love with it. I'll pull it out for now and we can always revisit if it seems valuable.

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,62 @@
...
+	const packageJson = storyMeta?.csfFile?.meta?.parameters?.packageJson ?? {};
+	const metadata = storyMeta?.csfFile?.meta?.parameters?.metadata ?? {};
+
+	if (!packageJson?.name) return;
+	if (!metadata?.modifiers) return;
+
+	return (
+		<ResetWrapper>
+			<LinkableHeading id="mods" size="m">
+				<a aria-hidden="true" href="#mods" tabIndex="-1" target="_self"></a>
```
marissahuysentruyt, line 35: I know Josh mentioned this, but I'm also not getting the `#mods` added to the URL. I don't have a suggestion on how to fix that, but wanted to note it once more.

marissahuysentruyt (changes requested):
This is looking excellent! Having the mods surfaced like this will be really helpful I think. 

Most of my comments are just questions, but I did see that progress bar is missing the metadata and packageJson imports (I'm assuming you don't want to wait for the fix in the resource link PR 😆)

castastrophe:
> Expect heading "Modifiable custom properties" show a link on hover and when the icon is selected, a direct link to that section is added to the window location.

@jawinn Just pushed up a fix for this - apparently vanilla HTML anchor linking doesn't work b/c React haha so I found an example of how they're doing it in the other Storybook docs and mimic'ed that.  It seems to be working well now.

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+export const Table = styled.table`
+	padding-block: 10px;
+`;
+
+/**
+ * Displays the current version number of the component. The version is read from
+ * the component's parameters, where it was sourced from the package.json file.
+ *
+ * Also displays a component status of "deprecated" if it is set in the story's
+ * parameters.
```
marissahuysentruyt, line 19: I think this might be the `ComponentDetails` description?

castastrophe: Yes! Good catch. Fixed this in my latest update.

generator/templates/stories/{{ folderName }}.stories.js.hbs
```diff
@@ -1,9 +1,9 @@
-import { version } from "../package.json";
```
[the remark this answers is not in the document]
castastrophe: This line is deleted in the template in favor of the full packageJson import. 🥳

components/swatchgroup/stories/swatchgroup.stories.js
```diff
@@ -1,29 +1,31 @@
-import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { Sizes } from "@spectrum-css/preview/decorators";
+import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { default as Swatch } from "@spectrum-css/swatch/stories/swatch.stories.js";
-import pkgJson from "../package.json";
+import metadata from "../metadata/metadata.json";
+import packageJson from "../package.json";
 import { SwatchgroupGroup } from "./swatchgroup.test.js";
-import { Template, RoundingTemplate } from "./template";
+import { RoundingTemplate } from "./template";
```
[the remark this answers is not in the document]
castastrophe: Yes! Good catch. Fixed this in my latest update.

components/stepper/stories/stepper.stories.js
```diff
@@ -1,9 +1,11 @@
 import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isDisabled, isFocused, isHovered, isInvalid, isKeyboardFocused, isQuiet, size } from "@spectrum-css/preview/types";
-import pkgJson from "../package.json";
+import metadata from "../metadata/metadata.json";
+import packageJson from "../package.json";
 import { StepperGroup } from "./stepper.test.js";
-import { AllDefaultVariantsGroup, DisabledVariantsGroup, Template } from "./template";
+import { AllDefaultVariantsGroup, DisabledVariantsGroup } from "./template";
```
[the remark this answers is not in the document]
castastrophe: No worries! Fixed this in my latest update.

components/progressbar/stories/progressbar.stories.js
```diff
@@ -1,15 +1,15 @@
...
  * The progress bar component shows the progression of a system operation such as downloading, uploading, processing, etc. in a visual way.
- * 
+ *
  * ## Values
  * The value is the progress of a system operation (e.g., downloading, uploading, processing) within the progress bar’s range, from the min value to max value. By default, the min value starts at 0 and the max value is set to 100. These values are not applicable when a progress bar is indeterminate.
- * 
+ *
  * Progress bars can have a value label at the end of the track that gives detailed information about the progress (e.g. "60%" or "2 of 8"). This value label works alongside the label and should not be displayed if the label itself is not displayed. It should also not be displayed if the progress is indeterminate. Similar to the label, the value label is always placed above the track.
  */
 export default {
```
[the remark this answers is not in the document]
castastrophe: Yes! Good catch. Fixed this in my latest update.

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const Table = styled.table`
+	padding-block: 10px;
+`;
+
+/**
+ * Displays the modifiable custom properties for a component based on the metadata provided in the story.
+ * The story metadata object is imported from the "metadata.json" file in the component's directory.
+ *
+ * If the metadata object does not contain a "modifiers" array, this component will not render.
```
marissahuysentruyt, line 18: Thinking back to the field group (which doesn't have any modifiers in the CSS), should we remove the `"modifiers": []` from field group metadata.json? Right now, I see an empty table, but it sounds like I shouldn't.
Screenshot 2024-10-09 at 11 03 27 AM

OR if this we had `!metadata?.modifiers.length` [here](https://github.com/adobe/spectrum-css/blob/8cfec9a60a8bdbe7c6fe86a6bba59796e16fecef/.storybook/blocks/PropertiesTable.jsx#L31), it wouldn't return the mods table.

marissahuysentruyt (commented):
Thanks for making all of those updates! This is looking really good 🤩 I just had one more comment on the modifiers array for field group!

.storybook/blocks/PropertiesTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+				<a aria-hidden="true" href="#modifiable-properties" tabIndex="-1" target="_self" onClick={() => {
+					context.channel.emit(NAVIGATE_URL, "#modifiable-properties");
+				}}></a>
+				Modifiable custom properties
+			</LinkableHeading>
+			<Body>
+				These are empty CSS custom property hooks available in this component
+				that enable one-off customizations specific to a product implementation.
+			</Body>
+			<Table className="docblock-properties-table sb-unstyled spectrum-Table spectrum-Table--sizeL spectrum-Table--compact">
```
jawinn, line 45: Small suggestion: I'd add `--mod-table-cursor-row-default: auto` to the `.docblock-properties-table` so it doesn't show a pointer.

jawinn (approved):
This looks good to me; my previous comments are addressed. The only thing I see is what Marissa already called out for Field group.

.storybook/blocks/PropertiesTable.jsx
castastrophe: Good idea!  I added that and pushed up the update.

.storybook/blocks/PropertiesTable.jsx
castastrophe: Oh good idea, yes. I added that and pushed up.

marissahuysentruyt (approved):
chefs-kiss

Thanks for all of this- it looks great!

PR #3196, opened 2024-10-02
fix(rating): address inconsistent hover behavior

cdransf:
## Description

Provides more granular control over the hover behavior of child stars within the rating component to prevent hovering in space adjacent to the component from highlighting all stars.

CSS-735

Verified in local Storybook story for rating component.

### Validation steps

1. Fetch the branch and run it locally or open the Storybook URL for the PR.
2. Navigate to the rating component.
3. Verify that hovering over rating stars is now more controlled and that hovering between stars doesn't result or all/no stars being displayed as active.
4. Verify that changing the selected star works as expected with the hover behavior.

### Regression testing

Validate:

## Screenshots

Kapture 2024-10-02 at 11 07 46

## To-do list


.changeset/spotty-penguins-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/rating": minor
```
jawinn, line 2: I think this would be a major change, since it's relying on the addition and usage of a new class.

jawinn (changes requested):
When testing, I'm seeing a couple issues:

1. The focus ring should not be appearing on click. The existing `is-focused` is actually just for keyboard focus. 

    Really`.is-focused` should be named `.is-keyboardFocused` or even better the selector could be `&:has(:focus-visible), &.is-keyboardFocused { ... }`, though I'm not sure we want to do that sort of renaming now? If we're already doing a breaking change, we could, though I'm hesitant to increase the scope of this work. I'd definitely reword the docs sentence that says "When the rating receives focus" to "When the rating receives keyboard focus".
2. Clicking on any of the ratings on the Docs page doesn't update something with the current value; the underline will show in the wrong place.
3. The `is-hovered` class gets leftover on elements when Rating is no longer being hovered

components/rating/index.css
```diff
@@ -168,6 +157,16 @@
 			display: block;
 		}
 	}
+
+	&.is-hovered {
```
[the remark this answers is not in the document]
cdransf: Got it! I'll add this to the PR description and will look for the usage notes.

That is an accurate description of the issue — I first approached it using CSS but there's an issue with one of our postcss plugins when trying to style one of the stars based on the hover state of the next: https://github.com/saulhardman/postcss-hover-media-feature/issues/35.

I've updated the selector to use `:hover, .is-hoverSelection`. The latter is now removed from all child icons on `mouseout` of the parent component.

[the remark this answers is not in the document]
cdransf: I've updated `is-focused` to be `is-keyboardFocused` and changed the state in the stories and templates too.

> Clicking on any of the ratings on the Docs page doesn't update something with the current value; the underline will show in the wrong place.

I need to look at this a bit more — it doesn't look like the instance in the docs is making it into the `@change` handler that updates this (as it works in the stories for the component):

```javascript
@change=${function(e) {
  const rating = e.target.closest(`.${rootClass}`);
  if (!rating) return;

  const input = rating.closest(`.${rootClass}-input`);
  if (!input) return;
  if (!isReadOnly && !isDisabled) {
    updateArgs({ value: parseInt(input.value, 10) });
  }
}}
```

components/rating/stories/rating.stories.js
```diff
@@ -1,23 +1,24 @@
...
 /**
  * The rating component is used to display or collect a user's rating of an item as represented by a number of stars.
- * 
+ *
  * ### Usage notes
  * - All active stars have the class `is-selected` applied.
  * - The current value (the last active star) has the class `is-currentValue` applied.
- * - When the rating receives focus, the class `is-focused` should be added to the component's root element (`.spectrum-Rating`).
+ * - When the rating receives keyboard focus, the class `is-keyboardFocused` should be added to the component's root element (`.spectrum-Rating`).
+ * - When the rating is hovered, the class `is-hoverSelection` should be added to the `spectrum-Rating-icon` being hovered over.
```
marissahuysentruyt, line 14:
```suggestion
* - When the rating is hovered, the class `.is-hoverSelection` should be added to the `.spectrum-Rating-icon` being hovered over.
```

components/rating/stories/rating.stories.js
```diff
@@ -1,23 +1,24 @@
...
 
 /**
  * The rating component is used to display or collect a user's rating of an item as represented by a number of stars.
- * 
+ *
  * ### Usage notes
  * - All active stars have the class `is-selected` applied.
  * - The current value (the last active star) has the class `is-currentValue` applied.
- * - When the rating receives focus, the class `is-focused` should be added to the component's root element (`.spectrum-Rating`).
+ * - When the rating receives keyboard focus, the class `is-keyboardFocused` should be added to the component's root element (`.spectrum-Rating`).
```
marissahuysentruyt, line 13:
```suggestion
* - When the rating receives keyboard focus, the class `.is-keyboardFocused` should be added to the component's root element (`.spectrum-Rating`).
```

components/rating/stories/rating.stories.js
```diff
@@ -1,23 +1,24 @@
...
+import { isDisabled, isEmphasized, isKeyboardFocused, isReadOnly } from "@spectrum-css/preview/types";
 import pkgJson from "../package.json";
 import { RatingGroup } from "./rating.test.js";
 import { Template } from "./template.js";
 
 /**
  * The rating component is used to display or collect a user's rating of an item as represented by a number of stars.
- * 
+ *
  * ### Usage notes
```
marissahuysentruyt, line 10: Should this actually be an h2? Most of the other components that have "Usage notes" have it as an h2.

cdransf: Just updated this! ✨

components/rating/stories/rating.stories.js
```diff
@@ -48,16 +49,22 @@ export default {
...
 		max: 5,
 		value: 3,
 	},
 	parameters: {
 		packageJson: pkgJson,
+		docs: {
+			story: {
+				inline: false,
+			},
+		},
```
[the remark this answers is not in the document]
cdransf: _Huh_, so removing `inline: false;` restores the background. Adding it causes the background to be reduced to what we're seeing here and allows you to select a rating star in the docs view.

Without `inline: false;` we get the full background and selecting a star is not reflected on the docs page. _But your selection will be reflected when you then go to the `Default` view._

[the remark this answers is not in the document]
cdransf: I pushed up a change removing `inline: false;` — I'm leaning towards keeping it out and having the component be fully interactive in the `Default` view. What do you think? ✨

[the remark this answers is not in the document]
cdransf: I can't seem to reproduce the `Failed to fetch...` error in Chrome either. 🤔 

image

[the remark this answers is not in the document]
marissahuysentruyt: I wonder if I just had a weird caching issue or something from hopping between s2 branches and back. I don't get the `failed to fetch` now either!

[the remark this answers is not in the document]
cdransf: Yeah! I'm curious to see what that future fix looks like. Should be cool! 🎉

components/rating/index.css
```diff
@@ -43,8 +43,10 @@
 	--spectrum-rating-icon-count: var(--spectrum-rating-icon-count);
 }
 
+/* stylelint-disable-next-line no-duplicate-selectors */
 .spectrum-Rating {
-	&.is-focused {
```
marissahuysentruyt, line 47: In the `.spectrum-Rating--emphasized` block, there's a `&.is-focused`. Do we need to update that too, to `.is-keyboardFocused`? I feel like it might have something to do with this? Where if I have `isEmphasize: true`, the keyboard focus styles remain the dark gray.

https://github.com/user-attachments/assets/ad4f94c0-c52c-4e81-a3f3-6702d61d1b35

marissahuysentruyt: There might be more color conflicts that are definitely a separate PR actually. With `isEmphasized: true` again, if I hover juuuust right, I can see both the gray unselected icon, or the blue unselected icon. I have a feeling it's somewhere near this code? I think it's just supposed to be gray the whole time, not sure at all though hahah

https://github.com/adobe/spectrum-css/blob/c328acd63a780de14530d12f213553b1d70eb14e/components/rating/index.css#L213

cdransf: _Oof_ haha — I swapped the `.is-focused` to `.is-keyboardFocused`, thanks for catching that!

Screenshot 2024-10-10 at 3 15 38 PM

The blue unselected border color is coming from `fill: currentColor` here:

image

I suppose we can try and override it, but the icons are getting set based on the `color` set in the component. https://www.w3.org/TR/css-color-3/#currentcolor

components/rating/stories/template.js
```diff
@@ -12,21 +12,80 @@ export const Template = ({
...
+	});
 
 	return html`
 		<div
 			class=${classMap({
 				[rootClass]: true,
 				"is-disabled": isDisabled,
 				"is-readOnly": isReadOnly,
-				"is-focused": isFocused,
+				"is-keyboardFocused": isKeyboardFocused,
```
marissahuysentruyt, line 88: Just looking through the rest of the template, do we need to change any of these `isFocused` args to `isKeyboardFocused` as well? 

I think this one is down in the `repeat()`
https://github.com/adobe/spectrum-css/blob/c328acd63a780de14530d12f213553b1d70eb14e/components/rating/stories/template.js#L72

Then in the `focusin`/`focusout`
https://github.com/adobe/spectrum-css/blob/c328acd63a780de14530d12f213553b1d70eb14e/components/rating/stories/template.js#L35

I'll be honest that I don't have a good grasp on these functions- I just noticed they were still the `isFocused` arg.

components/rating/stories/template.js
```diff
@@ -12,30 +12,89 @@ export const Template = ({
...
+		icons.forEach((icon, index) => {
+			if (icon.classList.contains("is-selected")) selectedIndex = index;
+
+			icon.addEventListener("mouseover", function() {
+				hoverIndex = index;
+				updateHoverState();
+			});
+
+			icon.addEventListener("mouseleave", function(event) {
+				if (!rating.contains(event.relatedTarget)) {
```
marissahuysentruyt, line 67: I'm not sure I've seen `relatedTarget`. Could you explain what this `if` is checking for? (I'll have to look it up!)

components/rating/stories/template.js
```diff
@@ -12,30 +12,89 @@ export const Template = ({
...
+				const activeStar = icon.querySelector(".spectrum-Rating-starActive");
+				const inactiveStar = icon.querySelector(".spectrum-Rating-starInactive");
+
+				if (index <= hoverIndex ||
+					(index <= selectedIndex && hoverIndex === -1)
+				) {
+					icon.classList.add("is-hoverSelection");
+					activeStar.style.display = "block";
+					inactiveStar.style.display = "none";
+				}
```
marissahuysentruyt, line 41: Aren't `hoverIndex` and `selectedIndex` set to -1, which means that `index` will always be greater than them initially? It's working (don't get me wrong), but how is that working? I'm totally missing something here 🤦‍♀️

components/rating/index.css
```diff
@@ -98,17 +100,6 @@
...
-	/* When the entire component is hovered, show all solid icons */
-	&:hover {
-		.spectrum-Rating-starActive {
-			display: block;
-		}
-
-		.spectrum-Rating-starInactive {
-			display: none;
-		}
-	}
```
marissahuysentruyt, line 111: Did we need to double check with design or anybody else that we can completely remove this? I agree with the removal, but it certainly seems like it was intentional. Also because I'm nosy I'd love to get the context around that decision hahahaha

marissahuysentruyt (commented):
I had a brainstorm while I was reviewing this- we could totally put a little disclaimer on the docs page that the rating components "have limited functionality" or "are not fully functional." Then add a "visit the default story for extended functionality" line or something (with a link to the default story page). What do you think about that? I did that with breadcrumbs actually. 

Screenshot 2024-10-17 at 9 50 17 AM

.changeset/spotty-penguins-sort.md
```diff
@@ -0,0 +1,8 @@
+---
+"@spectrum-css/rating": major
```
castastrophe, line 2: @pfulton I'm a bit anxious about introducing such a large bump to an S1 component with the current progress on foundations work. What do you think? Should we try to scale back the scope of this work, hold it until after foundations, or move forward as-is?

pfulton: I agree on the anxiousness around S1/S2 timing. On the plus side, though, this is a component that is _not_ currently available in SWC, so the downstream impact of a big change might not be as large as it could have been if it was already implemented there. That said: it's still a _major_, and we want to respect what that means for all potential consumers.

I think we should pause on this one for now until after Foundations goes in. After that, we can work through rebasing this one and getting it merged in. The scope of changes are only within this particular component, so hopefully any potential conflicts won't be too terrible to resolve when the time comes.

pfulton:
Marking this as `blocked` for now until we get S2 Foundations merged into `main`. Once we do, we'll come back to this, get it rebased against the new `main` stuff, and then look to get it merged in.

cdransf:
Closing this — going to see if we can lift the hover logic into the S2 migration. https://github.com/adobe/spectrum-css/pull/3627

PR #3197, opened 2024-10-02
fix(button): use appropriate tokens for radii in buttons

cdransf:
## Description

Replaces all of occurrences of `--spectrum-component-pill-edge-to-text-*` (intended to be used as padding) in the button component with the appropriate `corner-radius-*` tokens.

### Validation steps

1. Fetch the branch and run Storybook locally (or access the Storybook URL for the branch).
2. Navigate to the `button` component.
3. Verify that there are no `border-radius` regressions.

### Regression testing

Validate:

## Screenshots

image

## To-do list


cdransf:
✨ We don't have 1:1 value matches between the two token types (hence the usage of calc) _but_ the resulting value should be approximately the same and behave as expected.

.changeset/soft-ways-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/button": minor
+---
+
+Replaces all of occurrences of --spectrum-component-pill-edge-to-text-_ (intended to be used as padding) in the button component with the appropriate corner-radius-_ tokens.
```
jawinn, line 5:
```suggestion
Replaces the use of `--spectrum-component-pill-edge-to-text-*` tokens within the `border-radius` of the button component, as those tokens are intended to be used as padding. The border radius is now a calculated value, using half of `--spectrum-button-height`.
```

jawinn (approved):
I think this method, using the calculated value as we're now using in S2, should work just fine. We'll just need to make sure to confirm the VRTs. And adjust the changelog a little (see suggestion).

Some thoughts that I wanted to mention, but I don't think are an issue:
- `--spectrum-button-height` not defined by default without a size, but this does not seem to be an issue since medium appears to require the medium class (and seems to be that way in web components as well); this component hasn't received the medium as default reorganization that other components have gotten yet, so the button without the medium class is missing a lot and isn't fully formed anyway.
- Since there's a `calc`, I know when this rebases with foundations we'll need to move this property to a different location. But I think it's okay where it is now.

PR #3200, opened 2024-10-03
docs(picker): migrate docs to storybook

jawinn:
## Description

Migrates docs site information to Storybook for the Picker component. Creates several docs only stories to represent the missing examples. This also updates the variants test.js data to add coverage for missing states and variants.

Some Storybook updates of note:
- Adds a control for value (`currentValue`), and separates the concept of value and placeholder, which are different things with different classes. Their text resides in the same element but the `is-placeholder` class only exists when this text is a placeholder.
- Adds a control for setting the adjacent help text that already exists in the template
- Adds a control for displaying the thumbnail variant (with a workflow icon) from the docs site. This icon has a class associated with it (`.spectrum-Picker-icon`) that was not previously represented in Storybook.
- Renames `content` to `popoverContent` and refactors some of the `iconName` code in the Picker template to clarify what they are used for.

CSS-936

### Validation steps

@marissahuysentruyt 
- [x] Relevant docs from [the docs site](https://opensource.adobe.com/spectrum-css/picker.html) have now exist in the [Storybook "Docs" page for Picker](https://pr-3200--spectrum-css.netlify.app/preview/?path=/docs/components-picker--docs)
- [x] When `currentValue` control text is set, it is used instead of the `placeholder` control text. The `is-placeholder` class only resides on the element containing the text when the placeholder text is used.
- [x] New "show workflow icon" control is functioning
- [x] New "help text" control is functioning
- [x] Updated testing preview view of Default story covers all variants and states that we should be testing

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/picker/stories/picker.stories.js
+++ b/components/picker/stories/picker.stories.js
@@ -1,11 +1,13 @@
 import { WithDividers as MenuStories } from "@spectrum-css/menu/stories/menu.stories.js";
+import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
-import { isDisabled, isInvalid, isKeyboardFocused, isLoading, isOpen, isQuiet, size } from "@spectrum-css/preview/types";
+import { isActive, isDisabled, isHovered, isInvalid, isKeyboardFocused, isLoading, isOpen, isQuiet, size } from "@spectrum-css/preview/types";
 import pkgJson from "../package.json";
 import { PickerGroup } from "./picker.test.js";
+import { ClosedAndOpenTemplate, DisabledTemplate, Template } from "./template.js";
 
 /**
- * A picker outlines a set of options for a user.
+ * The picker component (sometimes known as a "dropdown" or "select") allows users to choose from a list of options in a limited space. The list of options can change based on the context.
  */
 export default {
 	title: "Picker",
@@ -14,6 +16,7 @@ export default {
 		size: size(["s", "m", "l", "xl"]),
 		label: {
 			name: "Label",
+			description: "The text for the field label",
 			type: { name: "string" },
 			table: {
 				type: { summary: "string" },
@@ -26,23 +29,55 @@ export default {
 			type: { name: "string" },
 			table: {
 				type: { summary: "string" },
-				category: "Content",
+				category: "Component",
 			},
-			options: ["top", "left"],
+			options: ["top", "side"],
 			control: { type: "select" },
 		},
 		withSwitch: {
-			name: "Display with a switch component",
+			name: "Show switch component",
+			description: "Display a separate switch component after the picker. Helpful for testing alignment with the picker when using the side label.",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
-				category: "Component",
+				category: "Advanced",
 			},
 			control: "boolean",
-			if: { arg: "labelPosition", eq: "left" },
+			if: { arg: "labelPosition", eq: "side" },
+		},
+		showWorkflowIcon: {
+			name: "Show workflow icon",
+			description: "Display optional workflow icon before the value or placeholder",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Advanced",
+			},
+			control: "boolean",
 		},
 		placeholder: {
 			name: "Placeholder",
+			description: "The placeholder text prompts a user to select an option from the picker menu. It disappears once a user selects an option. This will not be displayed if the `value` control is set.",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "text" },
+		},
+		currentValue: {
+			name: "Value",
+			description: "The value shows the option that a user has selected.",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "text" },
+		},
+		helpText: {
+			name: "Help text",
+			description: "Optional help text that can be informational or an error message. Displays a separate help text component after the picker. For error messages, the invalid control must also be set to true.",
 			type: { name: "string" },
 			table: {
 				type: { summary: "string" },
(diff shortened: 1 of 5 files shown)


jawinn:
Note: I noticed this CSS issue with the invalid icon on the updated chromatic template that will need to be addressed separately. It looks like it may need a `flex-shrink: 0`

Screenshot 2024-10-03 at 1 35 28 PM

components/picker/stories/picker.stories.js
```diff
@@ -32,17 +35,38 @@ export default {
 			control: { type: "select" },
 		},
 		withSwitch: {
-			name: "Display with a switch component",
+			name: "Display a separate switch component after the picker. Helpful for testing alignment with the picker when using the side label.",
```
rise-erpelding, line 38: Do we want this to be a description instead? It's a bit long.

rise-erpelding (commented):
These docs are such a great improvement over what we had before! I especially appreciate the added content from the guidelines (when to use quiet, when to use side label, etc.). Also, thank you for adding a card for that invalid icon being so tiny!

Some questions I have on whether we should add extra documentation/coverage that I'd love your opinion on:
- Do we need to talk about `withSwitch` option on the Docs page at all or is that strictly for checking alignment? (seems like it is, just want to be sure)
- Is there any value to adding documentation/support in the template for the Required picker? (It might be nice, but there's also already SO MUCH going on here.)
- Pickers with and without a label are shown somewhat interchangeably, it might be nice to include that guideline about "always use a label unless it's obvious" somewhere here.
- We're not testing help text with the picker, should we? (Also disabled + invalid + helptext)
- It looks like (but tell me if I have a control on in Storybook somewhere that I shouldn't 😂) we're testing Picker pretty much exclusively with side labels, do we want to test the top label too? (Or are we and I just missed it?)
- Also, not a question about coverage, but is this a good time/place to change the "left" option in the label position controls to "side" to be more logical?

components/picker/stories/picker.stories.js
```diff
@@ -32,17 +35,38 @@ export default {
...
 			name: "Placeholder",
+			description: "The placeholder text prompts a user to select an option from the picker menu. It disappears once a user selects an option. This will not be displayed if the `value` control is set.",
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "text" },
+		},
+		currentValue: {
```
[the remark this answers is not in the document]
jawinn: I've changed the default value to be an empty string instead of undefined.

jawinn:
> * Do we need to talk about `withSwitch` option on the Docs page at all or is that strictly for checking alignment? (seems like it is, just want to be sure)
It's just for catching an edge case with alignment that was brought up a while ago. I've recategorized this under `Advanced` in the controls.

> * Is there any value to adding documentation/support in the template for the Required picker? (It might be nice, but there's also already SO MUCH going on here.)
I think this delves too much into all the options for field label. In the updated docs descriptions, I've included a link to field label.

> * Pickers with and without a label are shown somewhat interchangeably, it might be nice to include that guideline about "always use a label unless it's obvious" somewhere here.
I've included docs about this and a link to the guidelines. And to take that even further and follow the guidelines in the other examples, I swapped out the use of `Picker` with `Template` in a few cases, so the field label appears for all the examples.

> * We're not testing help text with the picker, should we? (Also disabled + invalid + helptext)
There are no picker specific help text styles that I can find (within the picker CSS), and help text is outside of picker, so it doesn't seem like this needs to be in VRTs.

> * It looks like (but tell me if I have a control on in Storybook somewhere that I shouldn't 😂) we're testing Picker pretty much exclusively with side labels, do we want to test the top label too? (Or are we and I just missed it?)
All of the others have a top label right now, so please take a look again. I did have to make some adjustments to add `display: block` because the flex properties of our utility containers were negatively affecting the display of the adjacent components.

> * Also, not a question about coverage, but is this a good time/place to change the "left" option in the label position controls to "side" to be more logical?
Seems like as good of a time as any; I've changed this to `side` and checked to make sure all uses of `Picker()` were okay.

components/picker/stories/picker.stories.js
```diff
@@ -56,39 +91,60 @@ export default {
 		isDisabled,
 		isLoading,
 		isInvalid,
-		content: { table: { disable: true } },
+		isHovered,
+		isActive,
```
marissahuysentruyt, line 95: I saw there was a section about "read-only" pickers in the guidelines: https://spectrum.adobe.com/page/picker/#Read-only

Should we add that to the controls at all? It's not on the docs site, and the S2 picker Figma designs don't look like they have a read only option.

components/picker/stories/picker.test.js
```diff
@@ -9,24 +9,88 @@ export const PickerGroup = Variants({
...
+			isInvalid: true,
+			isHovered: true,
+		},
+		{
+			testHeading: "Loading",
+			isLoading: true,
+		},
+		{
+			testHeading: "Loading + hovered",
+			isHovered: true,
```
marissahuysentruyt, line 64: Correct me if I'm wrong, but does this need the `isLoading: true`?

components/picker/stories/picker.test.js
```diff
@@ -9,24 +9,88 @@ export const PickerGroup = Variants({
...
+		{
+			testHeading: "Hovered",
+			isHovered: true,
+		},
+		{
+			testHeading: "Active",
+			isActive: true,
+		},
+		{
+			testHeading: "Keyboard focused",
```
marissahuysentruyt, line 46: Is there a focused test case at all, in addition to keyboard focused? I don't see it in the controls, and just quickly scanning the CSS, it looks like it's all been converted to using `:focus-visible`.

components/picker/stories/template.js
```diff
@@ -12,17 +12,24 @@ import { when } from "lit/directives/when.js";
 
 import "../index.css";
 
+/**
+ * Template for Picker only (no popover or help text).
```
marissahuysentruyt, line 16: Your extra comments are great- definitely clears up the `Template` and `Picker` exports. 👍

marissahuysentruyt (changes requested):
Wonderful work on this! I think it's safe to say you're our resident picker expert!

I only have one thing that I think actually needs to change (the is-loading/is-hovered test case). The rest of my comments are questions! Thanks for all your work on this!

components/picker/stories/picker.stories.js
```diff
@@ -100,3 +156,221 @@ WithForcedColors.parameters = {
...
+	onclick: () => {},
+};
+TextOverflow.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * A workflow icon can be displayed before the value or placeholder. The class `.spectrum-Picker-icon` should be used with the icon.
+ */
+export const WithWorkflowIcon = Template.bind({});
```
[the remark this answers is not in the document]
jawinn: I did exclude that one on purpose as it did look like Menu component territory and there are no Picker styles associated with that part of it. It also wasn't on the guidelines.

components/picker/stories/picker.stories.js
jawinn: There are no read-only styles in the CSS, so there isn't anything for us to show currently. It's unclear to me looking at that sparse guideline about whether anything is needed for follow-up work related to read-only; it looks like Spectrum Web Components displays read-only like the default and React Spectrum does not have this option.

components/picker/stories/picker.test.js
jawinn: There are no `:focus` styles, and all `:focus-visible` selectors are also combined with `.is-keyboardFocused`, so those styles should be covered.

components/picker/stories/picker.test.js
jawinn: Yes! Good catch. Updated.

marissahuysentruyt (approved):
chuck-norris-approve

and also I approve.

components/picker/stories/picker.stories.js
```diff
@@ -26,23 +29,55 @@ export default {
...
+			type: { name: "string" },
+			table: {
+				type: { summary: "string" },
+				category: "Content",
+			},
+			control: { type: "text" },
+		},
+		helpText: {
+			name: "Help text",
+			description: "Optional help text that can be informational or an error message. For error messages, also set the invalid control to true. Displays a separate help text component after the picker.",
```
rise-erpelding, line 80: An opinion: I read this slightly oddly at first, and I wonder if we could re-word it? The verb in the 2nd sentence is "set" and the verb in the 3rd sentence is "displays", which don't agree and that threw me off. Using "sets" instead of "set" changes the meaning, so that's not a good change. Maybe we could flip the 2nd and 3rd sentence? Like:

```suggestion
			description: "Optional help text that can be informational or an error message. Displays a separate help text component after the picker. For error messages, also set the invalid control to true.",
```

and/or rearrange the words a bit?

```suggestion
			description: "Optional help text that can be informational or an error message. Displays a separate help text component after the picker. For error messages, also the invalid control must be set to true.",
```

rise-erpelding (approved):
Looks great! Left one small comment about some wording that you can change at your discretion, but otherwise good to go!

PR #3219, opened 2024-10-07
docs(combobox): update placeholder text

rise-erpelding:
## Description

Removes the placeholder in the combobox template to eliminate concerns about accessibility (because placeholder text disappears when a value is present, it should not contain essential information or instructions). Text from the [guidelines](https://spectrum.adobe.com/page/combo-box/#Don%E2%80%99t-use-placeholder-text) explaining that placeholder text should not be used for combobox were also added to the Docs page for Combobox.

Also contains an update to the template used for the Docs page; where previously Combobox had its own .mdx page, it no longer seems necessary in comparison to other migrations that were completed after this one. Any relevant information from combobox.mdx was moved to combobox.stories.js and the comobobox.mdx file was removed. The combobox.mdx page did have the added benefit of allowing us to better control the order in which the stories appeared, but lately the norm has been that this doesn't justify a component having its own .mdx.

One thing led to another here and the Docs page was also reorganized with one story for each variant, each containing multiple states (closed, open, invalid, disabled, loading, etc.). This looks more similar to the more recently completed docs pages (for instance [Tabs](https://spectrum-css.netlify.app/preview/?path=/docs/components-tabs--docs), [Action Button](https://spectrum-css.netlify.app/preview/?path=/docs/components-action-button--docs), or [Alert Banner](https://spectrum-css.netlify.app/preview/?path=/docs/components-alert-banner--docs)).

There are two commits to make changes to the template for combobox as well:
1. Menu items were being passed into Combobox from combobox.stories.js, but weren't visible because the combobox template was displaying a Menu component with args hard-coded into the template. This was adjusted so that args from the story are now visible in Combobox's menu.
2. The value of the Textfield within Combobox was set to `selectedDay` or undefined. `selectedDay` is necessary for Date Picker, but I didn't see a reason it was needed within Combobox. I _think_ this might be a remnant from when [Combobox and Date Picker were split from the InputGroup package](https://github.com/adobe/spectrum-css/pull/1556). `selectedDay` was removed from Combobox in favor of a `value` argument.

Neither of these two commits is necessary and can be removed if there is a reason to keep the template as is.

These changes only affect documentation, so no changeset is necessary.

[CSS-41](https://jira.corp.adobe.com/browse/CSS-41)

### Validation steps

Using the deployment preview, view the combobox component [default story](https://pr-3219--spectrum-css.netlify.app/preview/?path=/story/components-combobox--default) and [docs page](https://pr-3219--spectrum-css.netlify.app/preview/?path=/docs/components-combobox--docs) and :
- [x] Confirm that placeholder has been removed
- [x] Confirm that appropriate placeholder guidelines have been added
- [x] Confirm that no information has been lost from the Docs page in the deletion of combobox.mdx (all the same stories should be displayed), compare with the Combobox Docs page in [prod](https://spectrum-css.netlify.app/preview/?path=/docs/components-combobox--docs)
- [x] Confirm that testing coverage still looks the way that we want it to--this will have an impact on VRTs because the content has changed, but should otherwise remain the same 

Other considerations in the scope of this work:
1. I didn't display Help Text with this, but could if we think it's relevant (since we are showing the field label and talking about help text)
2. I removed the fallback Menu content from the template since it was overriding any menu content coming from args, but that meant that every story (with `isOpen: true`) needs to have a content arg, thus I made `content` a default arg.
3. Do changes made to the `value` property work? Do we want to display any value text or leave it blank? (If a menu item is selected in SWC's [implementation](https://opensource.adobe.com/spectrum-web-components/components/combobox/), you'll see that as the value in the text field)
4. The Menu content in the args differs slightly from the removed fallback content in the template (same items, but adds a divider and selection).

### Regression testing

Validate:

## To-do list


pfulton (approved):
Wow, this looks fantastic and is a nice improvement over what we previously had!

components/combobox/stories/template.js
```diff
@@ -26,15 +25,9 @@ const Combobox = ({
 	isFocused = false,
 	isKeyboardFocused = false,
 	isLoading = false,
-	selectedDay,
+	value,
```
marissahuysentruyt, line 28: Does this need to be set to an empty string, like it is in `Template`? `value = ""`? Or is that `value = ""` being an empty string specific to a combobox with a popover? I'm a little fuzzy on how to use `Combobox` vs `Template` 😬

components/combobox/stories/combobox.stories.js
```diff
@@ -64,151 +90,66 @@ export default {
...
-					type: "divider",
-				},
-				{
-					label: "United States of America",
-					isDisabled: true,
-				},
-			],
-		}),
-	],
+	value: "Ballard",
```
marissahuysentruyt, line 133: Does this need to be in the controls for any reason? Even if we hid it from the controls table? I don't know, I was just curious.

components/combobox/stories/template.js
```diff
@@ -26,15 +25,9 @@ const Combobox = ({
...
 } = {}, context = {}) => {
-	const { globals = {}, updateArgs } = context;
-	const lang = globals.lang ?? "en-US";
-
-	// If selectedDay is a string, convert it to a Date object
-	if (typeof selectedDay === "string" && selectedDay.length > 0) {
-		selectedDay = new Date(selectedDay).toLocaleDateString({ language: lang });
-	}
```
marissahuysentruyt, line 37: I know we talked about this at our dev sync, and I don't totally see a reason for this to be here in combobox. As far as I can tell, this and the subsequent `value: selected Day...` lines are a good removals 🤷‍♀️

components/combobox/stories/template.js
```diff
@@ -69,11 +62,8 @@ const Combobox = ({
 				customInputClasses: [`${rootClass}-input`],
 				isLoading,
 				customProgressCircleClasses: ["spectrum-Combobox-progress-circle"],
-				placeholder: "Type here this text should truncate",
```
marissahuysentruyt, line 72: Oh ok! This was truncating, and so in the VRTs we could see the ellipsis. 

Screenshot 2024-10-09 at 4 24 27 PM

Should we show that at all in another way in our test grid now that we're using `value` instead? 

Screenshot 2024-10-09 at 4 24 01 PM

marissahuysentruyt (commented):
Looking good! I had a few small questions and thoughts for you, so I haven't approved it yet, but I could if you think any of those would be work for a separate branch! Let me know what you think.

A couple other thoughts:

- I did notice that we have `&::placeholder` styles in the combobox CSS. It looks like it's specifically for users to modify the placeholder styles, but if we remove placeholders, is it ok to leave in the CSS?

- You inspired me to make a card to go back and remove any of those MDX files where we can!

components/combobox/stories/template.js
rise-erpelding: Looks like we're not exporting `Combobox` so the only place it's used is within the template, which will get a default value if none is set. I probably didn't set it because of that, thinking it'd be fine without it, but I don't think it hurts to add it, and if some part of the code gets changed so that the value isn't being passed to `Combobox`, we're covered.

components/combobox/stories/combobox.stories.js
rise-erpelding: Yeah, we have it for Picker, I don't see a reason not to add one for Combobox!

components/combobox/stories/template.js
rise-erpelding: Ok, I'm adding a section for Truncation, and a section for Truncation quiet, let me know your thoughts.

rise-erpelding:
@marissahuysentruyt Thanks for the feedback! I made some changes.

Re this comment/question:
> I did notice that we have &::placeholder styles in the combobox CSS. It looks like it's specifically for users to modify the placeholder styles, but if we remove placeholders, is it ok to leave in the CSS?

Removing placeholders from the styles would be a separate decision, I think. Unofficially, I think we still style them to match the value text (there's a comment on this on the text field s2 migration Jira [ticket](https://jira.corp.adobe.com/browse/CSS-767)) in case they're used. That particular [use](https://github.com/adobe/spectrum-css/blob/main/components/combobox/index.css#L317) 
 in combobox though, seems like it exists in case someone wants to modify the color to be different from the value, which seems suspicious. 🤔

I'm kind of thinking that might be a follow-up and/or if there's an S2 migration card for this, a note to follow up there?

components/combobox/stories/combobox.test.js
```diff
@@ -38,6 +38,17 @@ export const ComboBoxGroup = Variants({
...
+			testHeading: "Truncation",
+			isOpen: false,
+			value: "United States of America and to the republic",
+		},
+		{
+			testHeading: "Quiet + truncation",
+			isOpen: false,
+			isQuiet: true,
+			value: "United States of America and to the republic",
+		}
```
marissahuysentruyt, line 51: I like the idea of having the quiet truncation too!

marissahuysentruyt (approved):
Thanks for the updates and explanation on the `::placeholder` styles. The truncation tests look great. I was wondering if we should add the `withStates: false` to those 2 tests since I'm not sure any of the state changes would affect the truncation, but I'll leave that up to you!

Beautiful work as always!

rise-erpelding:
Leaving all the states in, since we had an issue with truncation + invalid state on Picker (#3220), it makes sense to test truncation + invalid, and also truncation + loading, we _could_ pick some of the states out, but there doesn't seem to be a strong justification for that right now.

PR #3222, opened 2024-10-08
fix(search): addresses quiet variant clear icon positioning

cdransf:
## Description

Addresses issue where a portion of search clear button sat outside of its parent by moving the transform applied to the button to the child icon.

This repositions the icon and the enclosing button such that the icon renders in the same place in the layout, the button doesn't move outside the bounds of the container and the clickable area to clear the input remains the same size.

**Note:** the clear icon is no longer centered within the parent _but_ addressing the overflow without re-aligning the icon would result in the icon sitting further inside the field than is the case with other variants.

CSS-914

Locally in Storybook.

### Validation steps

1. Fetch branch and run locally or access the Storybook URL for the PR.
2. Navigate to the search component.
3. Enable quiet mode
4. Verify that the clear icon remains right aligned and the enclosing button does not overflow the parent container.

### Regression testing

Validate:

## Screenshots

Screenshot 2024-10-08 at 11 12 13 AM
Screenshot 2024-10-08 at 11 12 00 AM

## To-do list

### DIFF
--- a/.changeset/chatty-carrots-warn.md
+++ b/.changeset/chatty-carrots-warn.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/search": minor
+---
+
+Addresses issue where a portion of search clear button sat outside of its parent by moving the transform applied to the button to the child icon.
--- a/components/search/index.css
+++ b/components/search/index.css
@@ -224,8 +224,9 @@
 }
 
 /* Quiet Variant */
+/* stylelint-disable-next-line no-duplicate-selectors */
 .spectrum-Search--quiet {
-	.spectrum-Search-clearButton {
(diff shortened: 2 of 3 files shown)


rise-erpelding (approved):
I would agree that this seems like the best possible option given the circumstances. This sets the close button in the same position as before without its container overflowing the parent.

It _doesn't_ match Figma, which shows the clear icon in the same position regardless of variant (see screenshot), but given that [there's no specific token for the spacing](https://adobedesign.slack.com/archives/C07DK1G926S/p1725651955034969?thread_ts=1724691943.514189&cid=C07DK1G926S) it doesn't seem worthwhile to change the position now, especially if there aren't consumers calling it out.

image

Also worth mentioning that there's no quiet variant in S2! So this icon being off-centered in its container won't be a long-lived issue.

PR #3226, opened 2024-10-09
feat(storybook): remove custom styles on storybook-root

jawinn:
## Description

The width of the Storybook root element was being limited with inline-size: max-content, which was causing a few issues:
- min-widths on some components were no longer being tested
- some responsive grid layouts were no longer working as intended

After running a force-rebuild on the VRTs to check for differences, it looks like removing these should not be a problem.  This is likely due to the switch to the Storybook "centered" layout for most stories, which uses a flex layout for centering, which creates a less wide storybook-root. The only stories which were affected in the VRT diff were:
- **coachmark**: this was manually setting the "padded" layout, which I removed in this PR after testing that the centered layout looks fine
- **alertbanner**: this is manually setting the "padded" layout on purpose to see its `display: block` and max width behavior, that is lost with the "centered" layout, so I kept that as is. This particular snapshot will be a bit wider. I also adjusted this component's template slightly so we are no longer setting an inline `min-inline-size` that is outside of the component's actual behavior with its shipped styles. The expected VRT difference in the chromatic variants template will be slightly less wide alert banners for the non text wrapping examples.

CSS-959

### Validation steps

- [x] Alert banner's default story now behaves like the CSS; full width up until its max-width (832px). Shrink the browser to confirm that it shrinks as well. @castastrophe (https://pr-3226--spectrum-css.netlify.app/preview/?path=/story/components-alert-banner--default)
- [x] Only [VRT changes](https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=3519) are for alert banner, as noted in the PR description

### Regression testing

Validate:

- [x] VRTs have been run and looked at.: https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=3519 (required running `yarn test --force-rebuild` to rebuild changes)

## To-do list

### DIFF
--- a/.storybook/assets/base.css
+++ b/.storybook/assets/base.css
@@ -55,12 +55,6 @@ svg:has(symbol):not(:has(use)) {
 	border-block-end: 1px solid hsla(203deg, 50%, 30%, 15%);
 }
 
-/* This is the container Chromatic uses to determine the height and width of the story */
-#storybook-root {
-	inline-size: max-content;
-	max-inline-size: 100%;
-}
-
 /* Force the modal wrapper to be contained by the frame not the viewport */
 #root-inner {
 	.spectrum-Modal-wrapper {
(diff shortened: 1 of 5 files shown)


castastrophe (approved):
Fabulous! Thank you for tackling this.

rise-erpelding (approved):
These are nice improvements to alert banner especially! Glad this ended up being less of a problem after centering all the stories!

PR #2827, opened 2024-06-07
docs(dropzone,fieldgroup,floatingactionbutton,illustratedmessage,infieldbutton): site docs to Storybook

rise-erpelding:
## Description

This continues the migration work to move the documentation from the deprecated docs site to Storybook, covering components drop zone, field group, floating action button, illustrated message, and infield button. Variants shown on the docs site for these components have been added to Storybook if they did not already exist.

An MDX page was only created if the existing documentation template could not support displaying information in a manner consistent with the docs site. In this PR, the only component that received an MDX page was field group, because its variants encompass both checkbox groups and radio groups and organizing these stories to be displayed side-by-side felt more logical.

Since we anticipate that some components will use their own MDX docs page and others will use the existing documentation template, the documentation template was updated to make the formatting look more consistent with what is seen on the MDX page.

### Component-specific changes
- Drop zone: The `FilledAndDragged` story was updated to be able to receive custom styles, which were used to add a background image to the component as it's seen on the docs site.
- Field group: The docs site demonstrates both radio inputs and checkboxes, but Storybook was only demonstrating radio inputs. Checkboxes have now been added to the template with an arg/control to switch between these two field group types. I did some follow-up work to confirm that what I was copying from the docs site was accurate, and in the process, added a few pieces of information from the internal docs site. Some information about the invalid state was also updated (see comments).
- Floating action button - This required only minor additions to the description. I also renamed the Default story to "Default (Primary)" to make more sense alongside the "Secondary" variant.
- Illustrated message - Additional information was added to supplement the current descriptions in the JSDoc comments.

This PR is docs-only and does not require a changeset.

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps
- [ ] [Documentation template ](https://github.com/adobe/spectrum-css/pull/2827/files#diff-569b9b80de2456e5a082d20cc1f491ee06aff354c1d5d930cdfedf2d4193c576)- the changes here make sense and bring visual parity between MDX docs pages and pages using the template
- [ ] [Drop zone](https://pr-2827--spectrum-css.netlify.app/preview/?path=/docs/components-drop-zone--docs) - includes all necessary information compared to the [docs site](https://opensource.adobe.com/spectrum-css/dropzone.html)
[...]

PR #3227, opened 2024-10-09
fix(field-group): addresses regressions in field groups.

cdransf:
## Description

Addresses regressions in field groups.
- [x] Adds helper to return correct input types in `args.items`
- [x] Adds appropriate inputs to their respective stories
- [x] Updates labels to match legacy docs page
- [x] Restores helper text/labels
- [x] Adds req'd asterisks, parenthetical and label modifications to relevant stories

Verified locally in storybook and by referencing [legacy docs page](https://pr-2827--spectrum-css.netlify.app/preview/?path=/docs/components-field-group--docs).

### Validation steps

1. Fetch branch and run locally or access the Storybook URL for the PR.
2. Navigate to the field group component.
3. Verify that the rendered field groups align with the [legacy docs page](https://pr-2827--spectrum-css.netlify.app/preview/?path=/docs/components-field-group--docs).

### Regression testing

Validate:

## Screenshots

Screenshot 2024-10-10 at 8 43 44 AM

## To-do list

### DIFF
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -67,7 +67,7 @@ export default {
 		inputType: "radio",
 		labelPosition: "top",
 		layout: "vertical",
-		label: "Field Group Label",
+		label: "Field group label",
 		helpText: "Select an option.",
 		items: [
 			{
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -1,7 +1,6 @@
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isInvalid, isReadOnly, isRequired } from "@spectrum-css/preview/types";
 import { default as RadioSettings } from "@spectrum-css/radio/stories/radio.stories.js";
-import { Template as Radio } from "@spectrum-css/radio/stories/template.js";
 import metadata from "../metadata/metadata.json";
 import packageJson from "../package.json";
 import { FieldGroupSet } from "./fieldgroup.test.js";
@@ -65,24 +64,21 @@ export default {
 		layout: "vertical",
 		label: "Select one of the following options:",
 		items: [
-			(passthroughs, context) => Radio({
-				...passthroughs,
+			{
 				id: "apple",
 				label: "Apples are best",
 				customClasses: ["spectrum-FieldGroup-item"],
-			}, context),
-			(passthroughs, context) => Radio({
-				...passthroughs,
+			},
+			{
 				id: "banana",
 				label: "Bananas forever",
 				customClasses: ["spectrum-FieldGroup-item"],
-			}, context),
-			(passthroughs, context) => Radio({
-				...passthroughs,
+			},
+			{
 				id: "pear",
 				label: "Pears or bust",
 				customClasses: ["spectrum-FieldGroup-item"],
-			}, context),
+			}
 		],
 		isInvalid: false,
 		isRequired: false,
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -20,6 +20,10 @@ export default {
 	title: "Field group",
 	component: "FieldGroup",
 	argTypes: {
+		label: {
+			name: "Label",
+			type: { name: "string" },
+		},
 		inputType: {
 			name: "Input type",
 			type: { name: "string" },
@@ -62,7 +66,8 @@ export default {
 		inputType: "radio",
 		labelPosition: "top",
 		layout: "vertical",
-		label: "Select one of the following options:",
+		label: "Field Group Label",
+		helpText: "Select an option.",
 		items: [
 			{
 				id: "apple",
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -22,6 +22,7 @@ export default {
 	argTypes: {
 		label: {
 			name: "Label",
+			description: "The label for the field group component.",
 			type: { name: "string" },
 		},
 		inputType: {
--- a/components/fieldgroup/stories/template.js
+++ b/components/fieldgroup/stories/template.js
@@ -57,20 +57,21 @@ export const Template = (
 			>
 				${inputType === "radio" ?
 					items.map((item) =>
-					Radio({
-					...item,
-					isReadOnly,
-					isRequired,
-					name: "field-group-example",
-					customClasses: [`${rootClass}-item`],
-					}, context))
-					: items.map((item) =>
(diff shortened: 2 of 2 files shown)


components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -7,6 +8,48 @@ import packageJson from "../package.json";
 import { FieldGroupSet } from "./fieldgroup.test.js";
 import { Template } from "./template.js";
 
+const inputType = (type) => type === "radio" ? [
```
marissahuysentruyt, line 11: Oh nice. This is clever, and I wouldn't have thought to create a function like this here. I do have some other ideas for this, however. I'm happy to pair if this doesn't make sense!

I think we actually need to be editing `fieldgroup.template.js` instead of this stories file. If I'm understanding the card correctly, we want to be able to pass radio or checkbox to the `inputType` arg, and the story will handle the physical input changes. At one point, it was working (Josh linked to [the deploy preview](https://pr-2827--spectrum-css.netlify.app/preview/?path=/story/components-field-group--default&args=inputType:checkbox), and you can just switch the controls from radio to checkbox), and [line 44ish is I think what we want to try to do](https://github.com/adobe/spectrum-css/pull/2827/files#diff-9d5d7da10a73aa5f35dcd706223f6b2198005d038d0310251ee6d2fb9c673541): 

```
${inputType === "radio" ? 
  items.map((item) =>
	Radio({
	...item,
	isReadOnly,
	name: "field-group-example",
	customClasses: [`${rootClass}-item`],
  }, context)) 
  : items.map((item) =>
	Checkbox({
	...item,
	isReadOnly,
	customClasses: [`${rootClass}-item`],
  }, context))}
```

Or maybe we can edit the newer `renderContent` call? I had to import `Radio` and `Checkbox` into the template file instead.

```
${when(inputType, () => html`
	${renderContent(items, {
		callback: inputType == "radio" ? Radio : Checkbox,
		args: {
			label,
			isReadOnly,
			isRequired,
			customClasses,
	        }, context
	})}
	`,
)}
```

marissahuysentruyt (changes requested):
I left a big long comment, but let me know if you want to pair on it instead 😊

components/fieldgroup/stories/fieldgroup.stories.js
cdransf: Kapture 2024-10-14 at 15 42 40

Totally makes sense! I _did_ have to remove the `items` defaults in the story and set some in the template, but I'm open to another solution to that.

(Having them in the story rather than the template yielded an array of `null` objects.)

components/fieldgroup/stories/template.js
```diff
@@ -20,7 +21,23 @@ export const Template = (
...
+				id: "banana",
+				label: "Bananas forever",
+				customClasses: ["spectrum-FieldGroup-item"],
+			},
+			{
+				id: "pear",
+				label: "Pears or bust",
+				customClasses: ["spectrum-FieldGroup-item"],
+			}
+		],
```
marissahuysentruyt, line 40: If you pass _this_ as the default `items` args, does that work? So in the story file, we'd have to remove all of the `(passthroughs, context) => Radio({...passthroughs,` etc? Trying it really quickly, I think that should work still.

And then this can remain `items = []` in the template.

components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -1,7 +1,6 @@
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
 import { isInvalid, isReadOnly, isRequired } from "@spectrum-css/preview/types";
 import { default as RadioSettings } from "@spectrum-css/radio/stories/radio.stories.js";
-import { Template as Radio } from "@spectrum-css/radio/stories/template.js";
```
marissahuysentruyt, line 4: Oh yeah- this removal was probably why it was coming back as null!

components/fieldgroup/stories/template.js
```diff
@@ -53,7 +70,20 @@ export const Template = (
 					[`${rootClass}InputLayout`]: true,
 				})}
 			>
-				${renderContent(items, { args: { isReadOnly, isRequired }, context })}
+				${inputType === "radio" ?
+					items.map((item) =>
+					Radio({
+					...item,
+					isReadOnly,
```
marissahuysentruyt, line 77: We should probably add the `isRequired` arg back to the `Radio` and `Checkbox` components here.

components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -63,26 +66,24 @@ export default {
 		inputType: "radio",
 		labelPosition: "top",
 		layout: "vertical",
-		label: "Select one of the following options:",
+		label: "Field Group Label",
```
marissahuysentruyt, line 69:
```suggestion
label: "Field group label",
```

components/fieldgroup/stories/template.js
```diff
@@ -53,7 +55,22 @@ export const Template = (
...
+					name: "field-group-example",
+					customClasses: [`${rootClass}-item`],
+					}, context))
+					: items.map((item) =>
+					CheckBox({
+					...item,
+					isReadOnly,
+					isRequired,
+					customClasses: [`${rootClass}-item`],
+				}, context))}
```
marissahuysentruyt, line 73: This is a total nitpick, and doesn't really block this PR at all, but could we fix up some of the indentation? 

```
items.map((item) =>
	Radio({
		...item,
		isReadOnly,
		isRequired,
		name: "field-group-example",
		customClasses: [`${rootClass}-item`],
	}, context))
```

components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -21,6 +20,10 @@ export default {
 	title: "Field group",
 	component: "FieldGroup",
 	argTypes: {
+		label: {
+			name: "Label",
```
marissahuysentruyt, line 24: Oh nice addition. The default value for `label` is "Field group label." We could add that here to either the `name`, or we could add a `description`.

```
label: {
	name: "Label",
        description: "The field label for the entire field group",
        ...
```

And you could probably come up with a better description than me right now 😁

marissahuysentruyt (changes requested):
I had only small requests now...except for some testing. I just realized we should probably put the radio vs. checkbox `inputType` in the test file.

It looks like we could get rid of the items completely (and then remove that `Checkbox` import). And then, if we add a test case to `stateData` for checkbox, it at least covers one scenario for a checkbox field group.

```
{
	testHeading: "Checkbox",
	inputType: "checkbox",
}
```

We're missing quite a few test cases for the stories from the docs page (namely vertical), but that work should probably be a separate card.

marissahuysentruyt (approved):
Thanks for making all of those updates! It's so exciting to have that `inputType` control working again 

computer-kid-approuve

Issue #2853, opened 2024-06-19
[ActionButton, ActionGroup] examples should be accessible


astearns:
## Description

The Storybook Action button (and Action group) examples do not have an accessible name in the icon-only versions. The versions that have a name in the markup have a duplicate aria-label attribute, but the ones that do not have a name in the markup are missing that attribute.

## Link to documentation

https://opensource.adobe.com/spectrum-css/preview/index.html?path=/story/components-action-button--default

## Additional context

I assume the stories/template should be set up more like the spectrum-Button examples, which add an aria-label attribute for the versions with no markup name, and do not add the attribute for the versions with the name in the markup. I’d be happy to submit a PR with this documentation fix if this is correct.

neehhaa06:
Hello @astearns 

error: 

- The error is likely due to missing aria-label attributes on icon-only buttons, causing warnings like:
```
    Missing accessible name for button with icon
    Button lacks aria-label or aria-labelledby attribute
```

after: 

Add aria-label to icon-only buttons:

```
<button aria-label="Add item">
  <i class="icon-add"></i>
</button>
```

- No aria-label needed for buttons with text.

- Ensure icon-only buttons in Action Groups have aria-label.

PR #3230, opened 2024-10-10
chore[actionbutton, actiongroup]: fix application of aria-labels

marissahuysentruyt:
## Description

All action buttons were, by default, receiving the `aria-label` attribute. When clickable elements have visible text, typically they should not receive an `aria-label` because a screen reader will read the visible text aloud. The button component was applying `aria-label`s conditionally (when a button didn't have visible text/label), so the same approach was brought to action button. Now, any action button that has "regular" label text does not receive the `aria-label`, and if an action button is icon-only, with its `hideLabel: true` arg set, an `aria-label` will be applied.

Because action buttons are used in the action group component as well, some refactoring had to be done to make sure any arguments used to create icon-only action buttons were from the action group args, as opposed to the action button args. For instance, instead of the action buttons' `label` arg getting set to empty string to render an icon-only action button within an action group like so: 
```
ActionGroup.args = {
        content: [
		{
			iconName: "AlignTop",
			label: "",
		},
		{
			iconName: "AlignBottom",
			label: "",
		},
		{
			iconName: "AlignMiddle",
			label: "",
		},
	]
}
```

...the code for action group now relies on its own `iconOnly` control to hide the action buttons' labels. 
```
ActionGroup.args = {
	iconOnly: true,
	content: [
		{
			iconName: "AlignTop",
			label: "Align top",
		},
		{
			iconName: "AlignBottom",
			label: "Align bottom",
		},
		{
			iconName: "AlignMiddle",
			label: "Align middle",
		},
	]
}
```

This should ensure that the `label` for each action button within `content` gets passed appropriately to the action button component, and set as the value for its `aria-label`. 

White space/empty lines in the story files were removed during this work, as well as some corrections made to the sentence-case of our story names on the documentation pages.

### Jira/Specs
[CSS-853](https://jira.corp.adobe.com/browse/CSS-853)
[Github issue](https://github.com/adobe/spectrum-css/issues/2853)

### Pending Questions
1. Should we go through each instance of `ActionButton` to make sure we're using label and hideLabel correctly? For instance, the breadcrumbs action button is icon-only, and has no label text or aria-label. Is that separate work from this PR? With just a quick search, card, contextual help, pagination may be other components to look into.

### Validation steps

- [ ] Pull down the branch to run locally or [visit the deploy preview](https://pr-3230--spectrum-css.netlify.app/preview/?path=/docs/components-action-button--docs).
- [ ] Inspect [the action button docs page](https://pr-3230--spectrum-css.netlify.app/preview/?path=/docs/components-action-button--docs) in the browser.
- [ ] Verify that any of the action buttons that have visible label text do not have an `aria-label` attribute.
- [ ] Verify that any of the icon-only buttons (screenshot below) have an aria-label attribute that corresponds to the label text set in the controls. By default, you should see `aria-label="Edit"` for the icon-only buttons.

Screenshot 2024-10-10 at 2 12 42 PM

- [ ] Visit [the story page ](https://pr-3230--spectrum-css.netlify.app/preview/?path=/story/components-action-button--default)and change the label text via the "Label" control. Verify that your new label is set as the `aria-label` for the icon-only action buttons.

Screenshot 2024-10-10 at 2 16 04 PM

(I used "paradiddle" 🥁)
- [ ] Using a screen reader (like NVDA on AssistivLabs or VoiceOver with Safari), use the keyboard to tab focus onto the action buttons.
- [ ] Any icon-only action buttons show have their `aria-label` announced by the screenreader. You should hear something like "Edit, toggle button." You should also hear "submenu" for the action buttons with the "hold icon."
- [ ] Next [visit the action group docs page](https://pr-3230--spectrum-css.netlify.app/preview/?path=/docs/components-action-group--docs)
- [ ] In the inspector, verify that the "Default" action group has no `aria-label`s. The action button labels should be within `<span>`s.

Screenshot 2024-10-10 at 2 56 35 PM

- [ ] Verify that the "Icon-only" and "Quiet, icon-only" have `aria-label`s set to their label text. You should see "Edit," "Copy," and "Delete."

Screenshot 2024-10-10 at 2 57 34 PM

- [ ] Feel free to inspect a few more of the action buttons within the action group stories. The justified stories have different label text to correspond better to their icons.
- [ ] Additionally, you can inspect [the action group within the action bar component](https://pr-3230--spectrum-css.netlify.app/preview/?path=/docs/components-action-bar--docs). Those buttons should no longer have the `aria-label` attribute since they have visible text.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/actionbutton/stories/template.js
+++ b/components/actionbutton/stories/template.js
@@ -69,7 +69,7 @@ export const Template = ({
 	const { updateArgs } = context;
 	return html`
 		<button
-			aria-label=${ifDefined(label)}
+			aria-label=${ifDefined(hideLabel ? label : undefined)}
 			aria-haspopup=${ifDefined(hasPopup && hasPopup !== "false" ? hasPopup : undefined)}
 			aria-controls=${hasPopup && hasPopup !== "false" ? popupId : undefined}
 			aria-pressed=${isSelected ? "true" : "false"}
--- a/components/actiongroup/stories/actiongroup.stories.js
+++ b/components/actiongroup/stories/actiongroup.stories.js
@@ -48,7 +48,7 @@ export default {
 			control: "boolean",
 		},
 		iconOnly: {
-			name: "Icon Only",
+			name: "Icon-only",
 			type: { name: "boolean" },
 			table: {
 				type: { summary: "boolean" },
@@ -103,17 +103,14 @@ Default.tags = ["!autodocs"];
 Default.args = {};
 
 // ********* DOCS ONLY ********* //
-
 /**
  * An action group in a disabled state shows that the action buttons within the group exist, but are not available in that circumstance. This state can be used to maintain layout continuity and to communicate that an action group may become available later.
 */
-
 export const Disabled = ActionGroups.bind({});
 Disabled.tags = ["!dev"];
 Disabled.args = {
 	areDisabled: true
 };
-
 Disabled.parameters = {
 	chromatic: { disableSnapshot: true },
 };
@@ -126,7 +123,6 @@ Emphasized.tags = ["!dev"];
 Emphasized.args = {
 	areEmphasized: true
 };
-
 Emphasized.parameters = {
 	chromatic: { disableSnapshot: true },
 };
@@ -134,11 +130,9 @@ Emphasized.parameters = {
 export const Horizontal = TreatmentTemplate.bind({});
 Horizontal.tags = ["!dev"];
 Horizontal.args = {};
-
 Horizontal.parameters = {
 	chromatic: { disableSnapshot: true },
 };
-
 Horizontal.storyName = "Default";
 
 /**
@@ -149,25 +143,22 @@ HorizontalCompact.tags = ["!dev"];
 HorizontalCompact.args = {
 	compact: true
 };
-
 HorizontalCompact.parameters = {
 	chromatic: { disableSnapshot: true },
 };
+HorizontalCompact.storyName = "Horizontal compact";
 
 /**
  * The vertical option should be reserved for when horizontal space is limited.
 */
-
 export const Vertical = TreatmentTemplate.bind({});
 Vertical.tags = ["!dev"];
 Vertical.args = {
 	vertical: true
 };
-
 Vertical.parameters = {
 	chromatic: { disableSnapshot: true },
 };
-
 
 export const VerticalCompact = TreatmentTemplate.bind({});
 VerticalCompact.tags = ["!dev"];
@@ -175,11 +166,10 @@ VerticalCompact.args = {
 	compact: true,
 	vertical: true
 };
-
 VerticalCompact.parameters = {
 	chromatic: { disableSnapshot: true },
 };
-
+VerticalCompact.storyName = "Vertical compact";
 
 export const HorizontalSizing = (args, context) => Sizes({
 	Template: ActionGroups,
@@ -192,6 +182,7 @@ HorizontalSizing.tags = ["!dev"];
 HorizontalSizing.parameters = {
 	chromatic: { disableSnapshot: true },
 };
+HorizontalSizing.storyName = "Horizontal sizing";
 
 export const VerticalSizing = (args, context) => Sizes({
 	Template: ActionGroups,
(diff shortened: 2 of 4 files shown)


components/actionbutton/stories/template.js
```diff
@@ -69,7 +69,7 @@ export const Template = ({
 	const { updateArgs } = context;
 	return html`
 		<button
-			aria-label=${ifDefined(label)}
+			aria-label=${ifDefined(hideLabel ? label : undefined)}
```
marissahuysentruyt, line 72: This is the same why `Button` handles the `aria-labels`

components/actiongroup/stories/template.js
```diff
@@ -119,6 +119,7 @@ export const OverflowOption = (context) => Container({
 						},
 						{
 							iconName: "More",
+							label: "More options",
```
marissahuysentruyt, line 122: In my screenreader testing, this button was just announced as "toggle button." Do we have a specific label for a button like this, that probably triggers a popover/menu?

components/actiongroup/stories/template.js
```diff
@@ -129,59 +130,24 @@ export const OverflowOption = (context) => Container({
 	`
 }, context );
 
-export const ActionButtonOptions = (args, context ) => Container({
```
marissahuysentruyt, line 132: Correct me if I'm wrong, but I think this was basically recreating the Template with the default args. I ended up removing this template, in favor of just using the template itself in `TreatmentTemplate`.

components/actiongroup/stories/template.js
```diff
@@ -129,59 +130,24 @@ export const OverflowOption = (context) => Container({
...
+		{ iconOnly: true, heading: "Icon-only", },
+		{ iconOnly: true, areQuiet: true, heading: "Quiet, icon-only", },
+		].map(({ heading, areQuiet, iconOnly }) => Container({
+			withBorder: false,
+			heading: heading,
+			content: Template({
+				...args,
+				areQuiet, 
+				iconOnly,
+			}, context)}
```
marissahuysentruyt, line 150: Similar to the comment I made in the stories file- I opted for using the `iconOnly` arg for action group, instead of `hideLabel` from action button.

components/actiongroup/stories/template.js
```diff
@@ -65,102 +65,69 @@ export const OverflowOption = (context) => Container({
...
+						iconSet: "workflow",
+						label: "Copy",
+					},
+					{
+						iconName: "Delete",
+						iconSet: "workflow",
+						label: "Delete",
+					},
+					{
+						iconName: "More",
```
marissahuysentruyt, line 122: In my screenreader testing, this button was just announced as "toggle button." Do we have a specific label for a button like this, that probably triggers a popover/menu?

components/actiongroup/stories/template.js
```diff
@@ -65,102 +65,69 @@ export const OverflowOption = (context) => Container({
...
-						{
-							iconName: "Move",
-							iconSet: "workflow",
-							label: "Move",
-						},
-					]
-				})
-			}
-		)}
+			withBorder: false,
```
marissahuysentruyt, line 68: These changes just fixed up some indentation, removed the border from around each container, and I think passed context to the templates, instead of the parent `Container()`

components/actionbutton/stories/template.js
cdransf: Is this preferable to wrapping the label and value in the condition and omitting altogether if the condition is false?

cdransf (approved):

marissahuysentruyt:
@cdransf Because the button behavior was called out in the issue, I went straight to the button template to see how it was working there.

components/actionbutton/stories/template.js
castastrophe: It is - this is the way lit-html manages attributes

components/actiongroup/stories/actiongroup.stories.js
```diff
@@ -47,7 +47,7 @@ export default {
 			control: "boolean",
 		},
 		iconOnly: {
-			name: "Icon Only",
+			name: "Icon-only",
```
castastrophe, line 50: Nice catch

castastrophe (approved):

PR #3129, opened 2024-09-18
feat: ResourceListDetails doc block

marissahuysentruyt:
## Description

In the effort to bring as much of the docs site into Storybook, this PR creates a new `<ResourceListDetails />` doc block component. It should replicate this resource section of the docs site: 

Screenshot 2024-09-18 at 4 19 12 PM

`<ResourceListDetails />` should render each of the resource link cards (`<ResourceLinkContent />`) if the link exists. To do so, we are pulling data from each component's `package.json`. A few things should happen with that data: 
- we can get the package name, which corresponds to the package on NPM
- using that package name, we can programmatically create links to the component's package on NPM and its place in the repo (navigating to the component's README)

There was also a need for a sub-component, `<ResourceLinkContent />`, to hold the individual resource's content (the SVG icon, the "title" and "subtitle"). There are several `styled` components as well, mainly used as containers/wrappers, that made up the basis of the styles for the two main components.

### Component package.json updates

1.  Adding a `spectrum` object to each component's `package.json` (with a `guidelinesLink` and `rootClass` keys) gives us the ability to grab the link to the Spectrum Guidelines site. For the majority of components, this single addition is sufficient. 
    a. Some package.json files can generate multiple components, such as progress bar and meter. Those components have multiple objects within `spectrum`, to hold the data for the individual components. 
3. In cases where Spectrum Guidelines do not exist for a component (like accordion or dial), if documentation could be found on the Spectrum Contributions site, the `guidelinesLink` was pointing to that site instead.
4. For cases where no guidelines could be found on either of the sites mentioned above, the `spectrum` array was not added to the package.json. This allows us to just not render that specific guidelines card.

### <span id="uncovered-issues">Uncovered issues</span>
The switch and progress bar do not render some of the doc blocks. Neither renders the `ComponentDetails` or this new `ResourceListDetails` block. Although both render `TaggedReleases`, I believe there's issues with the data coming back from NPM for those two components. There is one tag found, at `0.1.0`, and [that link](https://www.npmjs.com/package/undefined/v/0.1.0) goes to a deprecated NPM package called [`undefined.js`](https://github.com/Gozala/undefined.js). 

#### Resolved!
Switch and progress bar were missing some parameters for the package metadata: https://github.com/adobe/spectrum-css/pull/3129/commits/d88aeba6fa7c35bcf13fb89db96a855f5da16039

[...]

PR #3231, opened 2024-10-10
docs(meter): let meter use its rootclass in storybook

jawinn:
## Description

An update to the meter component storybook templates to help with simplifying the comparison between `rootClass` values in #3129. The Meter component was the only component using a `rootClass` arg value that is not unique and its own (previously it was using `spectrum-ProgressBar`).

### Details

Meter's Storybook template was using the `rootClass` for progress bar, rather than its own `rootClass`. It was set up this way because of the way the two components are housed together, and meter primarily just uses ProgressBar, with some added classes.

This value not being unique made it difficult to compare and match up the value of `rootClass` in the `package.json` `spectrum` data to the actual `rootClass` defined in the story files default args.

### Validation steps

- [ ] Meter still displays correctly
- [ ] The static color background color option still works in meter; the class `spectrum-Meter--staticWhite` gets applied to the component root and the teal background still appears. 

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/progressbar/stories/meter.stories.js
+++ b/components/progressbar/stories/meter.stories.js
@@ -36,6 +36,7 @@ export default {
 	},
 	args: {
 		...ProgressBar.args,
+		rootClass: "spectrum-Meter",
 		size: "l",
 		label: "Storage space",
 	},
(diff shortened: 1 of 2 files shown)


marissahuysentruyt (approved):
Thanks for tracking this down! This definitely helps with the doc block PR!

castastrophe (approved):

PR #3232, opened 2024-10-10
feat(combobox): adding read only state to combobox

aramos-adobe:
## Description

Adding a `readOnly`  feature to the combobox component.  Where an `is-readOnly` class added to `spectrum-Combobox`. The `is-readOnly` class is also added to the `Textfield` component as well as the `readonly` attribute to the `<input>` element within it. 

## Validation Steps

- [x] Read only feature added to Storybook controls
- [x] New changeset 

### Regression testing

Validate:

## To-do list


.changeset/new-doors-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/combobox": major
+---
+
+Adding read only state to combobox
```
marissahuysentruyt, line 5: I'm not sure this is a major change, but double check with Patrick & Cassondra. A major change would result in a change in version numbers, and breaking changes. Since most of this is additive and, with the small amount of time I've spent with this branch, doesn't change existing behaviors, are they still considered "breaking changes?" 

(asking for a friend 😆)

components/combobox/index.css
```diff
@@ -318,6 +322,8 @@
 		--mod-textfield-text-color-default: var(--mod-combobox-font-color-placeholder);
 	}
 
+	
+
```
marissahuysentruyt, line 326: We can delete these empty lines.

components/combobox/stories/combobox.stories.js
```diff
@@ -165,6 +171,20 @@ QuietGroup.parameters = {
...
+
+export const ReadOnly = Template.bind({});
+ReadOnly.tags = ["!dev"];
+ReadOnly.args = {
+	isReadOnly: true,
+	value: "Ballard"
+};
+ReadOnly.parameters = {
+	chromatic: { disableSnapshot: true }
+};
```
marissahuysentruyt, line 186: Let's add `ReadOnly.storyName = "Read-only";` so the story name is in the correct sentence-case.

components/combobox/stories/combobox.stories.js
```diff
@@ -165,6 +171,20 @@ QuietGroup.parameters = {
 	chromatic: { disableSnapshot: true },
 };
 
+/**
+ * Combo boxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combo box does not have a read-only option if no selection has been made.
+*/
+
+export const ReadOnly = Template.bind({});
```
marissahuysentruyt, line 178: Would you add `isReadOnly` to the `stateData` in the test file, please? It'd probably be a good idea to capture the read-only state, especially for the open test case.

```
{
	testHeading: "Read-only",
	isReadOnly: true,
},
```

Screenshot 2024-10-15 at 2 00 55 PM

marissahuysentruyt (changes requested):
This is looking really great! I love the addition of making isOpen a conditional control when readOnly is false! 

I was wondering- are there any particular hover or focus states we should be considering for a read-only combobox?

components/combobox/stories/template.js
```diff
@@ -84,7 +86,7 @@ const Combobox = ({
 				isDisabled,
 				position: "right",
 				onclick: function () {
-					updateArgs({ isOpen: !isOpen });
+					if (!isReadOnly) updateArgs({ isOpen: !isOpen });
```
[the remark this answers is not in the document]
marissahuysentruyt: oh wait...if we don't have this, do keyboard interactions trigger `updateArgs`?

.changeset/new-doors-fold.md
aramos-adobe: You're right, it is not a breaking change. I thought since it is a feature that devs should be aware of I made it `major` .

components/combobox/index.css
```diff
@@ -213,6 +220,27 @@
 	.spectrum-Popover.is-open {
 		transform: translateY(var(--mod-combobox-spacing-edge-to-menu, var(--spectrum-combobox-spacing-edge-to-menu)));
 	}
+
+	&.is-readOnly .spectrum-Combobox-textfield {
```
marissahuysentruyt, line 224: I'm wondering if targeting the `spectrum-Combobox-textfield` is the right thing here...I tried a little bit to recreate some of the other CSS I've seen with `:not()`. I was thinking the "not quiet" part would be helpful up here too, that way we could really make sure the quiet styles are separated, like you have them. It feels to me like we're getting straight to the `spectrum-Combobox-input`, instead of aiming for its parent first, then the `input`. 

```suggestion
&.is-readOnly:not(.spectrum-Combobox--quiet) {
  .spectrum-Combobox-input:read-only {
    // probably the background-color and border and hover stuff in here?
  }
}
```

Let me know what you think of that. We'd want to make the changes across all 3 of these nested selectors- `default`, `is-invalid`, and `is-disabled`.

components/combobox/index.css
```diff
@@ -213,6 +220,27 @@
...
+			&:hover {
+				background-color: revert;
+			}
+		}
+
+		&.is-invalid input:read-only {
+			border-color: var(--highcontrast-textfield-border-color-invalid-default, var(--mod-textfield-border-color-invalid-default, var(--spectrum-combobox-readonly-border-color-invalid-default)));
+		}
+
+		&.is-disabled input:read-only {
```
marissahuysentruyt, line 238: Feel free to get a second opinion, but I definitely think we should be more specific with `.spectrum-Combobox-input:read-only`, as opposed to `input:read-only`.

components/combobox/index.css
```diff
@@ -377,6 +405,22 @@
 		&.is-invalid .spectrum-Textfield-validationIcon {
 			inset-inline-end: var(--mod-combobox-button-width, var(--spectrum-combobox-button-width));
 		}
+
+		&.spectrum-Textfield--quiet {
+			input:read-only {
```
marissahuysentruyt, line 410: Similar comment here: if it was me, I'd refactor `input:read-only` to work with `.spectrum-Combobox-input:read-only` instead.

components/combobox/index.css
```diff
@@ -377,6 +405,22 @@
 		&.is-invalid .spectrum-Textfield-validationIcon {
 			inset-inline-end: var(--mod-combobox-button-width, var(--spectrum-combobox-button-width));
 		}
+
+		&.spectrum-Textfield--quiet {
```
marissahuysentruyt, line 409: This is super close! As always, feel free to get a second opinion. I would probably do a similar approach to one I mentioned above, just without the "not quiet": 

```suggestion
&.is-readOnly {
	.spectrum-Combobox-input:read-only {
		border-block-end: var(--mod-combobox-border-width, var(--spectrum-combobox-border-width)) solid var(--mod-combobox-readonly-input-border-color, var(--spectrum-combobox-readonly-input-border-color));
	}
				
	&.is-disabled .spectrum-Combobox-input:read-only {
		color: var(--highcontrast-textfield-text-color-disabled, var(--mod-textfield-text-color-disabled, var(--spectrum-combobox-readonly-text-color-disabled)));
		border-color: var(--mod-textfield-border-color-disabled, var(--spectrum-combobox-border-color-disabled));
	}
			
	&.is-invalid > .spectrum-Combobox-input:read-only {
		border-block-end-color: var(--highcontrast-textfield-border-color-invalid-default, var(--mod-textfield-border-color-invalid-default, var(--spectrum-combobox-readonly-border-color-invalid-default)));
	}
}
```

components/combobox/index.css
```diff
@@ -90,6 +90,13 @@
 	--mod-picker-button-background-color-disabled: var(--mod-combobox-background-color-disabled);
 	--mod-picker-button-font-color-disabled: var(--mod-combobox-font-color-disabled);
 	/* @passthroughs end -- settings for nested Picker Button component */
+
+	/*** Read-only Colors ***/
+	--spectrum-combobox-readonly-input-background-color: var(--spectrum-gray-50);
+	--spectrum-combobox-readonly-input-border-color: var(--system-spectrum-textfield-border-color);
+	--spectrum-combobox-readonly-border-color-invalid-default: var(--spectrum-negative-border-color-default);
+	--spectrum-combobox-readonly-background-color-disabled: var(--spectrum-disabled-background-color);
+	--spectrum-combobox-readonly-text-color-disabled: var(--spectrum-disabled-content-color);
```
marissahuysentruyt, line 99: I know I left lots of suggestions, and in order to get some of the disabled stuff to work, I added: `--spectrum-combobox-border-color-disabled: var(--spectrum-disabled-border-color);`

marissahuysentruyt (commented):
This is getting loads closer! Let me know what you think of some of my suggestions. It was more of what we started talking about yesterday, just being a little more specific with the combobox classes & elements.

.changeset/new-doors-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/combobox": minor
+---
+
+Combobox now includes a `read-only` state. Add the `is-readOnly` class to the Combobox to enable this feature.
```
jawinn, line 5: It looks like the `readonly` attribute is also needed on the input to use the feature, and we should make it clear that it's more than adding the class.

jawinn (changes requested):

marissahuysentruyt (changes requested):
Just two more small things! I think we're missing a hover and then maybe a quick switch-a-roo with a selector block 😊

components/combobox/index.css
```diff
@@ -213,6 +221,27 @@
...
+			&:hover {
+				background-color: revert;
+			}
+		}
+
+		&.is-invalid .spectrum-Combobox-input:read-only {
+			border-color: var(--highcontrast-textfield-border-color-invalid-default, var(--mod-textfield-border-color-invalid-default, var(--spectrum-combobox-readonly-border-color-invalid-default)));
+		}
+
+		&.is-disabled .spectrum-Combobox-input:read-only {
```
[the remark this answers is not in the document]
aramos-adobe: @marissahuysentruyt I think that's why I had `unset` from the beginning so that no hover states happen at all. I will go through this again

components/combobox/stories/combobox.stories.js
```diff
@@ -171,7 +171,7 @@ QuietGroup.parameters = {
 };
 
 /**
- * Comboboxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combobox does not have a read-only option if no selection has been made.
+ * Comboboxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combobox does not have a read-only option if no selection has been made. To enable this feature, add `isReadOnly` class to the Combobox. Then within the Textfield component add `isReadOnly` class and `readonly` attribute to the `<input>`.
```
marissahuysentruyt, line 174:
```suggestion
* Comboboxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combobox does not have a read-only option if no selection has been made. To enable this feature, add the `.isReadOnly` class to the combobox. Then within the textfield component, add  the `.isReadOnly` class and `readonly` attribute to the `<input>`.
```

I'm not sure if this blocks approval, just some sentence-case corrections to the components names, and some additional punctuation.

marissahuysentruyt (approved):
Well done on this! I left a comment on fixing up some sentence-case naming, but I'm going to approve!

.changeset/new-doors-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/combobox": minor
+---
+
+Combobox now includes a `read-only` state. Add the `is-readOnly` class to the Combobox to enable this feature. By doing so, this also activates the `readonly` attribute in the `Textfield` component. To ensure the intended user experience please make sure the readonly state in the textfield is enabled by adding the `is-readOnly` class and `readonly` attribute on the input element for this component as well.
```
jawinn, line 5:
```suggestion
Combobox now includes a `read-only` state. Add the `is-readOnly` class to the Combobox to enable this feature. To ensure the intended user experience please make sure the readonly state in the textfield is enabled by adding the `is-readOnly` class and `readonly` attribute on the input element for this component as well.
```
The documentation is in context of implementations using our shipped CSS, not our Storybook.

jawinn (changes requested):

.changeset/new-doors-fold.md
jawinn: I also just inspected the component markup and saw that `is-readOnly` is added to `spectrum-Textfield`, and the `readonly` attribute is added to the input within it. So both this and the docs need a little adjustment to clarify; right now it sounds like both are added to the input.

jawinn (changes requested):
For keyboard focus, I'm seeing the focus indicator on quiet + readonly, but not on default + readonly. If this element is still focusable, should it still show the focus indicator?

components/combobox/stories/combobox.stories.js
```diff
@@ -165,6 +170,21 @@ QuietGroup.parameters = {
 	chromatic: { disableSnapshot: true },
 };
 
+/**
+ * Comboboxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combobox does not have a read-only option if no selection has been made. To enable this feature, add the `.isReadOnly` class to the combobox. To enable this feature, add the .isReadOnly class to the combobox. Then within the nested textfield component, add the .isReadOnly class and readonly attribute to the <input> element.
```
jawinn, line 174:
```suggestion
 * Comboboxes have a read-only option for when content in the disabled state still needs to be shown. This allows for content to be copied, but not interacted with or changed. A combobox does not have a read-only option if no selection has been made. To enable this feature, add the `.isReadOnly` class to the combobox. To enable this feature, add the .isReadOnly class to the combobox. Then within the nested textfield component, add the .isReadOnly class and readonly attribute to the `<input>` element.
```
One last update and this should be good to go! This was rendering an input on the Docs.

jawinn (changes requested):

components/combobox/stories/combobox.stories.js
aramos-adobe: @jawinn Good catch!

jawinn (approved):

PR #3256, opened 2024-10-15
fix(actionbutton): prevent diacritic clipping in thai

marissahuysentruyt:
## Description

Recently, another bug was captured in relation to the diacritics and tone marks getting cut off in the action button when the text is Thai: https://jira.corp.adobe.com/browse/CCEX-131368. 

This PR attempts to fix that bug by adding padding to the action button's label, allowing for the diacritics to overflow into the padding space, and no longer get cut off. A similar solution was implemented recently for picker (https://github.com/adobe/spectrum-css/pull/2914) 

If we decide that adding padding is causing VRT diffs or other unexpected issues we want to avoid, we could explore using line height as well (a PR for the same picker issue mentioned above was closed in favor of the padding fix instead: https://github.com/adobe/spectrum-css/pull/2913)

The 2 commits related to adding and then removing the Thai text were meant to be temporary and serve as validation purposes. These could be removed/dropped before merging (or we could just clean up the squash commit message)

BEFORE 🚫

action button
Screenshot 2024-10-15 at 12 15 58 PM

action group
Screenshot 2024-10-18 at 1 41 39 PM

AFTER ✅ 

action button
Screenshot 2024-10-15 at 12 16 46 PM

action group
Screenshot 2024-10-18 at 1 43 31 PM

I tested the deploy preview in AssistivLabs, for WHCM in Chrome, Edge, Firefox. 

### Validation steps

- [ ] Pull down the branch to run locally or visit [the deploy preview](https://pr-3256--spectrum-css.netlify.app/preview/?path=/docs/guides-contributing--docs)
- [ ] [Visit the action button story page](https://pr-3256--spectrum-css.netlify.app/preview/?path=/story/components-action-button--default)
- [ ] Change the action button text in the controls to Thai: "ล้างทั้งหมด"
    - [ ] Alternatively, if you're viewing the branch locally, you can checkout the second commit (`git checkout 35c9809de3fd760ff8f3848f2c031a4164d849a0`) to see the Thai text
- [ ] Verify that the tone marks and diacritics are no longer getting cut off visually (especially over the fourth character)
    - [ ] When viewing [the default action button story in production](https://opensource.adobe.com/spectrum-css/preview/index.html?path=/story/components-action-button--default), changing the label text to "ล้างทั้งหมด" causes the topmost diacritic to get cut off over the fourth character.
- [ ] When viewing [the action button testing preview](https://pr-3256--spectrum-css.netlify.app/preview/?path=/story/components-action-button--default&globals=testingPreview:!true), a new Internationalization test case appears, with the Thai characters (with no clipping of tone marks)
- [ ] Visit [the action group testing preview](https://pr-3256--spectrum-css.netlify.app/preview/?path=/story/components-action-group--default&globals=testingPreview:!true)
- [ ] Verify the action buttons in the vertical-justified action group are now the same height (32px) as all other buttons (this is an expected VRT)

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/actionbutton/index.css
+++ b/components/actionbutton/index.css
@@ -234,6 +234,7 @@ a.spectrum-ActionButton {
 .spectrum-ActionButton-label {
 	@extend %spectrum-ButtonLabel;
 	pointer-events: none;
+	line-height: var(--spectrum-actionbutton-height);
 
 	font-size: var(--mod-actionbutton-font-size, var(--spectrum-actionbutton-font-size));
 	white-space: nowrap;
--- a/components/actiongroup/index.css
+++ b/components/actiongroup/index.css
@@ -154,5 +154,5 @@
 }
 
 .spectrum-ActionGroup--justified .spectrum-ActionGroup-item {
-	flex: 1;
+	flex-grow: 1;
 }
--- a/.changeset/calm-ears-think.md
+++ b/.changeset/calm-ears-think.md
@@ -1,0 +1,7 @@
+---
+"@spectrum-css/actiongroup": patch
+---
+
+Action group
+
+`flex: 1` has now changed to `flex-grow: 1` in the `.spectrum-ActionGroup--justified .spectrum-ActionGroup-item` selector to specify the desired behavior of justified action groups.
--- a/.changeset/fair-badgers-repeat.md
+++ b/.changeset/fair-badgers-repeat.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/actionbutton": patch
+---
+
+Adds line-height on `.spectrum-ActionButton-label` in order to accommodate text with diacritics that may be cut off vertically.
--- a/components/actionbutton/stories/actionbutton.test.js
+++ b/components/actionbutton/stories/actionbutton.test.js
@@ -75,6 +75,10 @@ export const ActionButtonGroup = Variants({
 			testHeading: "Static white",
 			staticColor: "white",
 		},
+		{
+			testHeading: "Internationalization (Thai)",
+			label: "ล้างทั้งหมด",
+		},
 	],
 	stateData: [{
 		testHeading: "Disabled",


marissahuysentruyt:
> This addresses the core issue of preventing the clipping, although it's very tight for XS per current design (non-blocking issue).

Oh yes, great catch. In comparison, what if we went with a line-height fix instead? It looks like maybe it's a little better. A solution with line-height was another option in some recent fixes for picker (but I can't remember why we didn't end up going with it). Maybe that's a better solution here- what do you think? It also looks like maybe that span no longer overflows with line-height.

Padding:
Screenshot 2024-10-15 at 12 44 17 PM
Screenshot 2024-10-15 at 12 48 13 PM


Line-height:
Screenshot 2024-10-15 at 12 43 51 PM
Screenshot 2024-10-15 at 12 48 02 PM

5t3ph:
@marissahuysentruyt since we do not allow line-wrapping, line-height does alleviate the issue in a slightly more "correct" way since it changes the way the text-box is evaluated by increasing the room for the characters.

However, if we allowed line-wrapping, then the larger line-height, especially pixel-based, would not be ideal, so padding may be a preferred solution in other cases.

For future reference back on this PR, it's probably also worth noting that the clipping is only happening due to `overflow: hidden` which is a requirement for `text-overflow: ellipsis`

5t3ph (approved):

components/actionbutton/index.css
```diff
@@ -232,6 +232,8 @@ a.spectrum-ActionButton {
 }
 
 .spectrum-ActionButton-label {
+	line-height: var(--spectrum-actionbutton-height);
+
```
castastrophe, line 236: Great addition! Can we add this below the extend? I think ideally the extends appear at the top of the style block.

castastrophe:
Can we add a test case to the actionbutton with an example of this issue so we can prevent regressions in future?

components/actionbutton/index.css
marissahuysentruyt: ahhh yes absolutely. And I also added the test case as well.

components/actiongroup/index.css
```diff
@@ -57,6 +57,12 @@
 	gap: var(--mod-actiongroup-vertical-spacing-regular, var(--spectrum-actiongroup-vertical-spacing-regular));
 	display: inline-flex;
 	flex-direction: column;
+
+	&.spectrum-ActionGroup--justified .spectrum-ActionGroup-item {
+		/* flex: unset removes the flex: 1 set on `.spectrum-ActionGroup--justified .spectrum-ActionGroup-item`,
+		in order to keep the action buttons from shrinking */
+		flex: unset;
+	}
```
marissahuysentruyt, line 65: @5t3ph I know you suggested that maybe we should use `flex-grow: 1` as opposed to the `flex: 1` (which would also set flex-shrink which was probably causing our squish-ness), however, I was struggling to not undo the horizontal justified behavior.

I tried `flex-grow: 1` in a couple of places:
```
.spectrum-ActionGroup--justified .spectrum-ActionGroup-item {
	flex-grow: 1;
}

// OR
.spectrum-ActionGroup--vertical {
	gap: var(--mod-actiongroup-vertical-spacing-regular, var(--spectrum-actiongroup-vertical-spacing-regular));
	display: inline-flex;
	flex-direction: column;

	&.spectrum-ActionGroup--justified .spectrum-ActionGroup-item {
		flex-grow: 1;
	}
}
```
The first selector fixed the vertical justified button heights (from 36px to the correct 32px), but then unjustified the horizontal justified action group:
Screenshot 2024-10-18 at 3 11 53 PM

The second selector....didn't do anything? The action buttons were still the incorrect 36px, (and the icon-only aren't tall enough). 
Screenshot 2024-10-18 at 3 28 58 PM

With the new selector here with `flex: unset`, the horizontal justified remains the same, AND the vertical justified now is not squished and the proper 32px.

Screenshot 2024-10-18 at 3 29 57 PM

Am I missing anything with `flex-grow`? I know you mentioned `min-block-size` might be redundant, but would you rather that than the `flex: unset`? I haven't tried anything with that yet.

marissahuysentruyt: After our pairing, we realized I just didn't understand what _should_ be happening in a justified actiongroup! `flex-grow: 1` is now up: https://github.com/adobe/spectrum-css/pull/3256/commits/85656a0cf06f3f61bbd052cb35fbfa1b378c9701

castastrophe:
Passed VRT: https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=3495

aramos-adobe (approved):

PR #3258, opened 2024-10-15
fix(clear button): align background color and hover state for disabled variant

cdransf:
## Description

More closely aligns the clear button styling with the close button styling by adopting the pattern of classes targeting the disabled state. Scopes hover indicators to components that are not disabled and aligns the disabled colors used by the two components.

Locally in storybook.

### Validation steps

1. Fetch the branch and run it locally or access the Storybook URL for the branch.
2. Navigate to the clear button component and toggle on the disabled state.
3. Do the same for the close button.
4. Verify that the two are visually consistent.

### Regression testing

Validate:

## Screenshots

Screenshot 2024-10-15 at 10 39 50 AM

## To-do list


components/clearbutton/index.css
```diff
@@ -64,13 +64,24 @@
 		--mod-clear-button-icon-color-down: var(--spectrum-clear-button-icon-color-down-disabled, var(--spectrum-disabled-content-color));
```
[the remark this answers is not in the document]
cdransf: Ignore

rise-erpelding (changes requested):
These look like great changes! I’m a little on the fence about taking out the `--spectrum-clear-button-icon-color-hover-disabled` and `--spectrum-clear-button-icon-color-down-disabled` props (more so wondering if they should turn into mods, but on the other hand maybe it’s just better not to provide any modification for these) and want to know your thoughts.

Either way though this will need a changeset and we’d probably want to note that these properties have changed. It seems doubtful that this has many effects anywhere (not sure I’ve seen a disabled clear button in the wild) but just to be on the safe side!

components/clearbutton/index.css
```diff
@@ -60,17 +60,29 @@
 	&:disabled,
 	&.is-disabled {
 		--mod-clear-button-icon-color: var(--mod-clear-button-icon-color-disabled, var(--spectrum-disabled-content-color));
-		--mod-clear-button-icon-color-hover: var(--spectrum-clear-button-icon-color-hover-disabled, var(--spectrum-disabled-content-color));
-		--mod-clear-button-icon-color-down: var(--spectrum-clear-button-icon-color-down-disabled, var(--spectrum-disabled-content-color));
```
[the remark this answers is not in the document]
castastrophe: I think you're correct that they were meant to be --mods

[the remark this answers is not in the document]
cdransf: I spoke with @castastrophe and they may have been intended as mods initially, but aren't used now and can be safely removed. 🙂

components/clearbutton/index.css
```diff
@@ -65,13 +65,25 @@
 		--mod-clear-button-icon-color-down: var(--spectrum-disabled-content-color);
```
rise-erpelding, line 65: Are we keeping these `--mod-clear-button-icon-color-hover` and `--mod-clear-button-icon-color-down` after all?

cdransf: Ah, yep! I'm leaving those in place as they are referenced elsewhere. ✨

castastrophe (changes requested):
This is look good - it will need a changeset and a note to the SWC team when it merges as it will likely require config mapping updates on their end.

rise-erpelding (commented):
Looks good! I just left a question about static black.

components/clearbutton/index.css
```diff
@@ -65,13 +65,25 @@
...
 		--mod-clear-button-background-color: var(--mod-clear-button-background-color-disabled, transparent);
 	}
+
+	&.spectrum-ClearButton--staticWhite {
+		--spectrum-clear-button-icon-color-disabled: var(--spectrum-disabled-static-white-content-color);
+	}
+
+	&.spectrum-ClearButton--staticBlack {
```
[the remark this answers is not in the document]
cdransf: Good catch! I don't think it should be necessary — I'm going to drop it and let VRT run.

rise-erpelding (approved):

components/clearbutton/index.css
```diff
@@ -65,13 +65,20 @@
 		--mod-clear-button-icon-color-down: var(--spectrum-disabled-content-color);
 		--mod-clear-button-background-color: var(--mod-clear-button-background-color-disabled, transparent);
 	}
+
+	&.spectrum-ClearButton--staticWhite {
```
marissahuysentruyt, line 69: So this clear button also gets an [`--overBackground` modifier class](https://github.com/adobe/spectrum-css/blob/a3c858577dcb8c73a18dd5f876ee7029cf80f460/components/clearbutton/stories/template.js#L26) added to it, as opposed to just a `--staticWhite`. I believe we have intentions of deprecating the `--overBackground` modifier in favor of `--staticWhite` to match more of our other static color component variants. Which is great.

My question is if `staticWhite` is sort of the preferred naming, what do we do with the `--overBackground` selector blocks?

Should we add the `&.spectrum-ClearButton--staticWhite` to the [`&:focus-visible` block](https://github.com/adobe/spectrum-css/blob/a3c858577dcb8c73a18dd5f876ee7029cf80f460/components/clearbutton/index.css#L140), now that we have it? Should we add the static white class to the[ block with the token definitions](https://github.com/adobe/spectrum-css/blob/a3c858577dcb8c73a18dd5f876ee7029cf80f460/components/clearbutton/index.css#L49) for `--overBackground`?

components/clearbutton/index.css
```diff
@@ -112,6 +119,12 @@
 	}
 }
 
+.spectrum-ClearButton:disabled {
```
marissahuysentruyt, line 122: I know there's a few other components that double up the selector here like this: 

```
.spectrum-ClearButton:disabled,
.spectrum-ClearButton.is-disabled
```

I'm curious if you think we need that here or not. 🤷‍♀️ I'm not sure I have a strong preference either way, just thinking.

marissahuysentruyt (commented):
YOU CAN SEE THE DISABLED BUTTON! 🎉 Left a couple of questions for you, but it looks really great!

components/clearbutton/index.css
cdransf: I was debating whether to add that since it behaves without it, but I dropped it in for the sake of consistency. ✨

components/clearbutton/index.css
cdransf: That makes sense to me! I added `.spectrum-ClearButton--staticWhite` to both blocks that currently have `--overBackground` so we can remove `--overBackground` once it's been deprecated. ✨

marissahuysentruyt (approved):
Thanks for digging into this one!

.changeset/silly-donkeys-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/clearbutton": minor
+---
+
+Improves the visibility of the clear button disabled variant on static color backgrounds by more closely aligning styles with close button.
```
castastrophe, line 5: Can you update this as much as possible to detail the selector and property updates so that it's easier for SWC to understand the impact of the changes?

castastrophe (approved):

components/clearbutton/index.css
```diff
@@ -46,6 +46,7 @@
 		--mod-clear-button-background-color-key-focus: transparent;
 	}
 
+	&.spectrum-ClearButton--staticWhite,
```
castastrophe, line 49:
```suggestion
	/* @deprecated .spectrum-ClearButton--overBackground */
	&.spectrum-ClearButton--staticWhite,
```

castastrophe: Sorry missed that overBackground was being deprecated

cdransf: No problem! I'll get that incorporated. ✨

PR #3296, opened 2024-10-22
chore(sidenav): move declaration, add description and apply token

cdransf:
## Description

Resolve stylelint violations by moving declaration and adding comment as stylelint disable description.

## To-do list

### DIFF
--- a/components/sidenav/index.css
+++ b/components/sidenav/index.css
@@ -221,6 +221,7 @@
 	margin-block-end: var(--mod-sidenav-heading-bottom-margin, var(--spectrum-sidenav-heading-bottom-margin));
 	padding-inline: var(--mod-sidenav-inline-padding, var(--spectrum-sidenav-inline-padding));
 	color: var(--mod-sidenav-header-color, var(--spectrum-sidenav-header-color));
+	font-family: var(--spectrum-sidenav-header-font-family);
 	font-size: var(--mod-sidenav-header-font-size, var(--spectrum-sidenav-header-font-size));
 	font-weight: var(--mod-sidenav-header-font-weight, var(--spectrum-sidenav-header-font-weight));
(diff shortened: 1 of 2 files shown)


components/sidenav/index.css
```diff
@@ -92,7 +92,6 @@
 		--spectrum-sidenav-top-level-line-height: var(--spectrum-cjk-line-height-100);
 	}
 
-	--spectrum-sidenav-header-font-family: var(--spectrum-sans-font-family-stack);
```
castastrophe, line 95: Oh, I wonder if we should be using this variable and aren't. This might be one where we want to add the font-family property rather than remove the variable.

castastrophe (changes requested):

components/sidenav/index.css
cdransf: Good catch! I applied it alongside the other tokens in that block — that class was missing a `font-family` declaration.

components/sidenav/index.css
```diff
@@ -221,6 +221,7 @@
 	margin-block-end: var(--mod-sidenav-heading-bottom-margin, var(--spectrum-sidenav-heading-bottom-margin));
 	padding-inline: var(--mod-sidenav-inline-padding, var(--spectrum-sidenav-inline-padding));
 	color: var(--mod-sidenav-header-color, var(--spectrum-sidenav-header-color));
+	font-family: var(--mod-sidenav-header-font-family, var(--spectrum-sidenav-header-font-family));
```
castastrophe, line 224: @cdransf No need to add a new --mod here since that will add to our supported API. If a customer requires it down the road, we can address the need then.

castastrophe (changes requested):

components/sidenav/index.css
cdransf: Sounds good! Just dropped it. 🙂

castastrophe (approved):

PR #3297, opened 2024-10-22
fix(search): update icon sizing calculation in the quiet variant [CSS-1012]

castastrophe:
## Description

A bug fix for the Search component's quiet variant. The `--spectrum-search-quiet-button-offset` was being calculated using a static workflow icon size of 100 but the search icon sizing is set using t-shirt sizing and varies. This PR updates the calculation to use the variable `--spectrum-search-icon-size` which updates based on the determined t-shirt size.

This also adds a deprecation notice for the `--mod-workflow-icon-size-100` and suggests the use of the `--mod-search-icon-size` instead.

I didn't add a changeset for the update to clearbutton because it just combines the duplicate selectors for `.spectrum-ClearButton`.

### Validation steps

- [x] Expect VRTs to reflect the updated offset spacing between the search icon and the Search text in the quiet variant.
- [x] The space between clear icon and the end of the input field should vary based on which t-shirt size is used for the clear icon.

Screenshot 2024-10-22 at 4 56 02 PM

Screenshot 2024-10-22 at 4 56 13 PM

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/search/index.css
+++ b/components/search/index.css
@@ -80,6 +80,15 @@
 	--mod-textfield-background-color: var(--mod-search-background-color, var(--spectrum-search-background-color));
 	--mod-textfield-background-color-disabled: var(--mod-search-background-color-disabled, var(--spectrum-search-background-color-disabled));
 	/* @passthrough end */
+
+	display: inline-block;
+	position: relative;
+	inline-size: var(--mod-search-inline-size, var(--spectrum-search-inline-size));
+	min-inline-size: var(--mod-search-min-inline-size, var(--spectrum-search-min-inline-size));
+
+	.spectrum-HelpText {
+		margin-block-start: var(--mod-search-to-help-text, var(--spectrum-search-to-help-text));
+	}
 }
 
 .spectrum-Search--sizeS {
@@ -100,19 +109,6 @@
 	--spectrum-search-text-to-icon: var(--spectrum-text-to-visual-300);
 }
 
-.spectrum-Search--quiet {
-	--spectrum-search-quiet-button-offset: calc((var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2) - (var(--mod-workflow-icon-size-100, var(--spectrum-workflow-icon-size-100)) / 2));
-	--spectrum-search-background-color: transparent;
-	--spectrum-search-background-color-disabled: transparent;
-	--spectrum-search-border-color-disabled: var(--spectrum-disabled-border-color);
-
-	/* Added specificity, otherwise they are overridden by system specific themes. */
-	&.spectrum-Search {
-		--spectrum-search-border-radius: 0;
-		--spectrum-search-edge-to-visual: var(--spectrum-field-edge-to-visual-quiet);
-	}
-}
-
 @media (forced-colors: active) {
 	.spectrum-Search .spectrum-Search-textfield,
 	.spectrum-Search .spectrum-Search-textfield .spectrum-Search-input {
@@ -128,18 +124,6 @@
 			forced-color-adjust: none;
 			background-color: transparent;
 		}
-	}
-}
-
-/* Standard / Default */
-.spectrum-Search {
-	display: inline-block;
-	position: relative;
-	inline-size: var(--mod-search-inline-size, var(--spectrum-search-inline-size));
-	min-inline-size: var(--mod-search-min-inline-size, var(--spectrum-search-min-inline-size));
-
-	.spectrum-HelpText {
-		margin-block-start: var(--mod-search-to-help-text, var(--spectrum-search-to-help-text));
 	}
 }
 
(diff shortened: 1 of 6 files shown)


castastrophe:
VRT run: https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=3532

components/search/index.css
```diff
@@ -130,18 +127,6 @@
 	}
 }
 
-/* Standard / Default */
-.spectrum-Search {
```
castastrophe, line 134: Moved up to line 84

components/search/index.css
```diff
@@ -99,19 +109,6 @@
 	--spectrum-search-text-to-icon: var(--spectrum-text-to-visual-300);
 }
 
-.spectrum-Search--quiet {
```
castastrophe, line 102: Moved down to line 213

rise-erpelding (approved):
Looks good! This matches the spec better, the clear button sits nicely in the middle of its container, and the padding between the text input and clear button makes more sense, and supports anyone who was using the previous mods.

components/search/index.css
```diff
@@ -226,14 +210,26 @@
...
-		padding-inline-end: calc(var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size)) - var(--mod-search-quiet-button-offset, var(--spectrum-search-quiet-button-offset)));
+		padding-inline-end: calc(var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size)) - var(--mod-search-quiet-button-offset, calc(var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2 - var(--mod-search-icon-size, var(--spectrum-search-icon-size)) / 2)));
 		padding-block-start: var(--mod-search-top-to-text, var(--spectrum-search-top-to-text));
 	}
+
+	.spectrum-Search-clearButton .spectrum-ClearButton-icon {
+		/* @deprecated --mod-workflow-icon-size-100 -- use --mod-clear-button-width instead */
+		/* stylelint-disable-next-line spectrum-tools/no-unknown-custom-properties -- variable --spectrum-clear-button-width will resolve via the .spectrum-ClearButton class */
+		transform: translateX(var(--mod-search-quiet-button-offset, calc(var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2 - var(--mod-clear-button-width, var(--mod-workflow-icon-size-100, var(--spectrum-clear-button-width))) / 2)));
+	}
```
jawinn, line 234: Is this `transform` still needed? I disabled this style in the inspector for all t-shirt sizes and I'm not seeing any visual changes. It looks like the calc() is calculating to 0.  I see the icon is centered within the button with the button fill element's flex properties.

components/search/index.css
```diff
@@ -226,14 +210,26 @@
...
+	&.spectrum-Search {
+		--spectrum-search-border-radius: 0;
+		--spectrum-search-edge-to-visual: var(--spectrum-field-edge-to-visual-quiet);
 	}
 
 	&.spectrum-Search .spectrum-Search-input {
 		border-radius: var(--mod-search-border-radius, var(--spectrum-search-border-radius));
 		padding-inline-start: calc(var(--mod-search-edge-to-visual, var(--spectrum-search-edge-to-visual)) + var(--mod-search-icon-size, var(--spectrum-search-icon-size)) + var(--mod-search-text-to-icon, var(--spectrum-search-text-to-icon)));
-		padding-inline-end: calc(var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size)) - var(--mod-search-quiet-button-offset, var(--spectrum-search-quiet-button-offset)));
+		padding-inline-end: calc(var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size)) - var(--mod-search-quiet-button-offset, calc(var(--mod-search-block-size, var(--spectrum-search-block-size)) / 2 - var(--mod-search-icon-size, var(--spectrum-search-icon-size)) / 2)));
```
jawinn, line 226: Now that the clear button is no longer sitting outside the bounds of the quiet search, can this be simplified to just 
`padding-inline-end: var(--mod-search-button-inline-size, var(--spectrum-search-button-inline-size));`? 
I noticed that the padding does not account for the full size of the button:
search-field-padding

components/search/index.css
castastrophe: Great point, I've pushed up this update.

components/search/index.css
castastrophe: I think it's meant to be there as a guardrail to prevent text from impeding on the close icon, even if the icon sizing is altered.

jawinn: It seems odd to me from an interaction standpoint for the text to go under the clickable button. If longer text goes all the way to the end of the input (of the blue in that screenshot), then when the user may be thinking they're clicking at the end of the text to edit, they'd actually be clicking the clear button by mistake.

AFAIK design did not specify that increased padding for quiet, and React doesn't have it either. I think it may have just been always a little wrong and tied to the horizontal mis-positioning that you fixed in this PR.

components/search/index.css
castastrophe: That mod is also being used for the inline-end which is more relevant to placement so it should still do what customers are expecting.

components/search/index.css
jawinn: 👍 This is written up in **CSS-1052**.

PR #3301, opened 2024-10-22
feat(colorloupe): migrate to s2 drop-shadow tokens

rise-erpelding:
## Description

Migrates color loupe to Spectrum 2 tokens. According to the [S2 token spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9596-3359&t=o1lmG0VUo1R6zDuL-4) for color loupe, the only updated tokens concern the drop shadow. Color loupe now uses the new drop shadow tokens `drop-shadow-elevated-x`, `drop-shadow-elevated-y`, `drop-shadow-elevated-blur`, and `drop-shadow-elevated-color` rather than specific color loupe tokens.

[CSS-707](https://jira.corp.adobe.com/browse/CSS-707)

### Validation steps
- Confirm that the color loupe component for S2 is using the tokens listed in the S2 token spec

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/colorloupe/index.css
+++ b/components/colorloupe/index.css
@@ -9,87 +9,84 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 .spectrum-ColorLoupe {
-  --spectrum-colorloupe-width: var(--spectrum-color-loupe-width);
-  --spectrum-colorloupe-height: var(--spectrum-color-loupe-height);
+	--spectrum-colorloupe-width: var(--spectrum-color-loupe-width);
+	--spectrum-colorloupe-height: var(--spectrum-color-loupe-height);
 
-  --spectrum-colorloupe-offset: var(--spectrum-color-loupe-bottom-to-color-handle);
-  --spectrum-colorloupe-animation-distance: 8px; /* TODO: replace with forthcoming animation token */
+	--spectrum-colorloupe-offset: var(--spectrum-color-loupe-bottom-to-color-handle);
+	--spectrum-colorloupe-animation-distance: 8px; /* TODO: replace with forthcoming animation token */
 
-  --spectrum-colorloupe-drop-shadow-x: var(--spectrum-drop-shadow-x);
-  --spectrum-colorloupe-drop-shadow-y: var(--spectrum-color-loupe-drop-shadow-y);
-  --spectrum-colorloupe-drop-shadow-blur: var(--spectrum-color-loupe-drop-shadow-blur);
-  --spectrum-colorloupe-drop-shadow-color: var(--spectrum-color-loupe-drop-shadow-color);
+	--spectrum-colorloupe-drop-shadow-x: var(--spectrum-drop-shadow-elevated-x);
+	--spectrum-colorloupe-drop-shadow-y: var(--spectrum-drop-shadow-elevated-y);
+	--spectrum-colorloupe-drop-shadow-blur: var(--spectrum-drop-shadow-elevated-blur);
+	--spectrum-colorloupe-drop-shadow-color: var(--spectrum-drop-shadow-elevated-color);
 
-  --spectrum-colorloupe-outer-border-width: var(--spectrum-color-loupe-outer-border-width);
-  --spectrum-colorloupe-inner-border-width: var(--spectrum-color-loupe-inner-border-width);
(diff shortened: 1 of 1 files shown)


components/colorloupe/index.css
```diff
@@ -9,87 +9,84 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+	--spectrum-colorloupe-drop-shadow-y: var(--spectrum-drop-shadow-elevated-y);
+	--spectrum-colorloupe-drop-shadow-blur: var(--spectrum-drop-shadow-elevated-blur);
+	--spectrum-colorloupe-drop-shadow-color: var(--spectrum-drop-shadow-elevated-color);
 
-  --spectrum-colorloupe-outer-border-width: var(--spectrum-color-loupe-outer-border-width);
-  --spectrum-colorloupe-inner-border-width: var(--spectrum-color-loupe-inner-border-width);
-  --spectrum-colorloupe-outer-border-color: var(--spectrum-color-loupe-outer-border);
-  --spectrum-colorloupe-inner-border-color: var(--spectrum-color-loupe-inner-border);
+	--spectrum-colorloupe-outer-border-width: var(--spectrum-color-loupe-outer-border-width);
+	--spectrum-colorloupe-inner-border-width: var(--spectrum-color-loupe-inner-border-width);
```
marissahuysentruyt, line 24: The `color-loupe-outer-border-width` token is set to `border-width-200`, which is 2px. `color-loupe-inner-border-width` is just set to 1px. But there is a `border-width-100` that's 1px I think. I wonder why we didn't use that token.

This is in the tokens/dist. It's been a while since I've been in some of these files, so maybe I'm looking at things wrong. 
```
--spectrum-color-loupe-outer-border-width: var(--spectrum-border-width-200);
--spectrum-color-loupe-inner-border-width:1px;
```

Nothing's wrong with your code, I was just curious!

marissahuysentruyt, line 24: Should we let design know that there's just a typo in the Border section of the tokens file? I think the second `color-loupe-outer-border` (which is supposed to be transparent black 200) is meant to say `color-loupe-inner-border`. Do they want to update and correct that?

marissahuysentruyt (approved):
This looks great! Seems like it was a really simple migration- we probably won't get many of those! 😆

Maybe it's just a weird, rendering thing on my end, but does the docs color loupe _not_ have the opacity checkerboard, while the default story does? The markup certainly looks identical to me (I ran it through diffchecker), but I don't know why the checkerboard isn't rendering on the docs instance. 

Screenshot 2024-10-23 at 2 41 05 PM
Screenshot 2024-10-23 at 2 40 57 PM

The docs color loupe just seems to behave a little differently to me, even when I change the stroke/stroke width on the svgs: 

Screenshot 2024-10-23 at 3 11 40 PM
Screenshot 2024-10-23 at 3 11 17 PM

I see both of these locally and in the PR preview. Any ideas for that? I'm not sure if that's something that would block the approval. 🤷‍♀️ I don't think I see either one of these issues on production so...maybe it's just an old bug. I'm approving anyways!

rise-erpelding:
@marissahuysentruyt This is a known bug that I just noticed in Jira earlier today when I was looking for the design verification ticket! https://jira.corp.adobe.com/browse/CSS-775 It looks like it got closed out because it couldn't be reproduced, but I'm seeing it too in Chrome and Firefox, might need to be resurrected!

> Maybe it's just a weird, rendering thing on my end, but does the docs color loupe _not_ have the opacity checkerboard, while the default story does? The markup certainly looks identical to me (I ran it through diffchecker), but I don't know why the checkerboard isn't rendering on the docs instance.
> 
> Screenshot 2024-10-23 at 2 41 05 PM Screenshot 2024-10-23 at 2 40 57 PM
> The docs color loupe just seems to behave a little differently to me, even when I change the stroke/stroke width on the svgs:
> 
> Screenshot 2024-10-23 at 3 11 40 PM Screenshot 2024-10-23 at 3 11 17 PM
> I see both of these locally and in the PR preview. Any ideas for that? I'm not sure if that's something that would block the approval. 🤷‍♀️ I don't think I see either one of these issues on production so...maybe it's just an old bug. I'm approving anyways!

components/colorloupe/index.css
rise-erpelding: Yeah, I'm seeing that too! Looks like it is indeed coming from the tokens: 
image

PR #3325, opened 2024-10-24
chore(vscode-settings): remove deprecated property; disable postcss validate

cdransf:
## Description

- Removes deprecated `aslant.packageManager` setting
- Adds `postcss.validate: false` as `postcss.validate` conflicts with our `stylelint` and `posts`

Verified locally. Removing the following:

```
"at-rule-no-unknown": [
  true,
  {
    ignoreAtRules: ["extend", "container", "each", "include", "mixin"],
  },
],
```

Raises a lint violation against `at-rule-no-unknown`, establishing that `ignoreAtRules` is being applied (as no exceptions are now being defined in the rule definition).

Restoring the rule then returns our `unknownAtRules` violation which is attributed to `postcss`.

Adding `postcss.validate: false` to the workspace settings resolves the issue. Additionally — removing `"extend"` from `ignoreAtRules` and restarting the stylelint server in VS Code causes stylelint to surface a violation — we can then rely on stylelint to provide coverage for syntax violations like this.

## To-do list

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -119,7 +119,6 @@
 		"postcss": "css"
 	},
 	"eslint.format.enable": true,
-	"eslint.packageManager": "yarn",
 	"eslint.useESLintClass": true,
 	"files.associations": {
 		"*.css": "postcss"
@@ -167,6 +166,7 @@
 		}
 	],
 	"js/ts.implicitProjectConfig.experimentalDecorators": true,
+	"postcss.validate": false,
 	"prettier.configPath": ".prettierrc",
 	"prettier.ignorePath": ".prettierignore",
 	"prettier.prettierPath": "node_modules/prettier",


5t3ph (approved):

.vscode/settings.json
```diff
@@ -119,7 +119,6 @@
 		"postcss": "css"
 	},
 	"eslint.format.enable": true,
-	"eslint.packageManager": "yarn",
```
castastrophe, line 122: Why remove this?

.vscode/settings.json
```diff
@@ -167,6 +166,7 @@
 		}
 	],
 	"js/ts.implicitProjectConfig.experimentalDecorators": true,
+	"postcss.validate": false,
```
castastrophe, line 169: This looks good

.vscode/settings.json
```diff
@@ -188,5 +188,5 @@
 		".storybook/deprecated/*/*.mdx",
 		"components/*/stories/*.stories.js",
 		"components/*/stories/*.mdx"
-	]
+	],
```
castastrophe, line 191: I don't think hanging commas are allowed in vanilla json.

castastrophe (changes requested):

.vscode/settings.json
cdransf: We can keep it — but VS Code indicated that it's now handled by default and the setting has been deprecated.

cdransf: https://github.com/microsoft/vscode-eslint/blob/fa396e45eb25087950c81c9a9f9eb6f7d1f2818f/README.md?plain=1#L218

castastrophe (approved):

PR #3328, opened 2024-10-25
fix: correct the display of checkbox read-only state

jawinn:
## Description

As noted in CSS-833, the CSS, SWC, and RSP implementations of read-only checkboxes/checkbox groups, and the docs surrounding them, conflicted with each other. After discussing with the design team, it was decided that the first step would be for CSS to bring our visual styles in line with React's implementation.

**Field group:**
- This PR updates field group to remove the additional checkbox read-only styles that were causing the box to not display and for commas to be appended.
- Template: added `aria-labeledby` connected to the group label, so this is announced as a group by the screen reader.

**Checkbox:**
- Removes some unnecessary read-only CSS. Read-only just needs to override disabled styles. It was confirmed with PJ on the design team that otherwise it uses default styles (for both default and emphasized; React also displays this).
- Template: The checkboxes are made sure to be shown as both focusable and immutable. Some JS was added to prevent the checkbox value from changing.
- Template: `disabled` attribute removed when read-only. Read-only still needs to be focusable / in the tab order.
- Template: Added `aria-disabled="true"` when read-only, so screen readers do not announce as a normal interactive checkbox. For VoiceOver, this will add "dimmed" now and won't announce the shortcut to check or uncheck. Note that I've not used `aria-readonly` like as seen in React Spectrum as it is not announced by the screen reader (any differently than a normal checkbox) and MDN states that it is ["not relevant" for input with type checkbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-readonly).
- Adds additional VRT tests to cover some missing combination of states

This also updates all of the docs surrounding these things and adds a new storybook Docs example for Checkbox. The field group usage notes have been cleaned up as well.

CSS-1004

### Validation steps

- [x] Field group docs page: read-only example shows full checkbox with boxes and no added commas, and read-only colors
- [x] Checkbox docs page: read-only example shows full checkbox with boxes and no added commas, and read-only colors
- [x] Checkbox docs page: read-only default shows default colors and read-only emphasized shows emphasized colors (should include the blue, which is different than it was previously) @marissahuysentruyt 
- [x] Checkbox testing preview: additional test cases appear and cover the usage of read-only @marissahuysentruyt 
- [x] Review and proofread docs for both examples
- [x] Read-only checkbox is immutable in both docs and stories (clicking does not change checkedness)
- [x] Read-only checkboxes are focusable
- [x] Field group of checkboxes is announced as a group by screen reader (VoiceOver) @marissahuysentruyt 
- [x] Read only checkboxes are announced as dimmed/disabled by screen reader @marissahuysentruyt 

### Regression testing

Expected changes:
- Additional tests added
- Read-only checked for emphasized now displays with blue
- Field group read-only no longer shows comma display (see screenshots)

Validate:

## Screenshots

Before:
Screenshot 2024-10-25 at 4 24 36 PM

After:
Screenshot 2024-10-25 at 4 26 27 PM

## To-do list

### DIFF
--- a/components/checkbox/stories/checkbox.test.js
+++ b/components/checkbox/stories/checkbox.test.js
@@ -28,6 +28,10 @@ export const CheckboxGroup = Variants({
 			isChecked: true,
 		},
 		{
+			testHeading: "Indeterminate",
+			isIndeterminate: true,
+		},
+		{
 			testHeading: "Invalid",
 			isInvalid: true,
 		},
@@ -35,6 +39,11 @@ export const CheckboxGroup = Variants({
 			testHeading: "Invalid, checked",
 			isInvalid: true,
 			isChecked: true,
+		},
+		{
+			testHeading: "Invalid, indeterminate",
+			isInvalid: true,
+			isIndeterminate: true,
 		},
 		{
 			testHeading: "Disabled",
@@ -46,7 +55,8 @@ export const CheckboxGroup = Variants({
 			isChecked: true,
 		},
 		{
-			testHeading: "Indeterminate",
+			testHeading: "Disabled, indeterminate",
+			isDisabled: true,
 			isIndeterminate: true,
 		},
 		{
@@ -55,6 +65,16 @@ export const CheckboxGroup = Variants({
 		},
 		{
 			testHeading: "Read-only, checked",
+			isReadOnly: true,
+			isChecked: true,
+		},
+		{
+			testHeading: "Read-only, indeterminate",
+			isReadOnly: true,
+			isIndeterminate: true,
+		},
+		{
+			testHeading: "Read-only, checked, disabled",
 			isReadOnly: true,
 			isChecked: true,
 		},
--- a/components/checkbox/stories/template.js
+++ b/components/checkbox/stories/template.js
@@ -50,7 +50,7 @@ export const Template = ({
 					typeof size !== "undefined",
 				[`${rootClass}--emphasized`]: isEmphasized,
 				["is-indeterminate"]: isIndeterminate,
-				["is-disabled"]: isDisabled|| isReadOnly,
+				["is-disabled"]: isDisabled,
 				["is-invalid"]: isInvalid,
 				["is-readOnly"]: isReadOnly,
 				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
@@ -62,13 +62,19 @@ export const Template = ({
 				type="checkbox"
 				class="${rootClass}-input"
 				aria-labelledby=${ifDefined(ariaLabelledby)}
+				aria-disabled=${ifDefined(isReadOnly ? "true" : undefined)}
 				?checked=${isChecked}
-				?disabled=${isDisabled || isReadOnly}
+				?disabled=${isDisabled}
 				title=${ifDefined(title)}
 				value=${ifDefined(value)}
-				@change=${function() {
-					if (isDisabled) return;
-					updateArgs({ isChecked: !isChecked });
+				@change=${(e) => {
+					if (isReadOnly) {
+						// Make checked value immutable for read-only.
+						e.preventDefault();
+						e.target.checked = !e.target.checked;
+					}
+					if (isDisabled || isReadOnly) return;
+					updateArgs?.({ isChecked: e.target.checked });
 				}}
 				id=${ifDefined(id ? `${id}-input` : undefined)}
 			/>
@@ -103,21 +109,19 @@ export const DocsCheckboxGroup = (args, context) => Container({
 			...args,
 			context,
 			iconName: undefined,
+			label: "Unchecked",
 		})}
 		${Template({
 			...args,
 			context,
 			isChecked: true,
+			label: "Checked",
 		})}
 		${Template({
 			...args,
 			context,
 			isIndeterminate: true,
-		})}
-		${Template({
-			...args,
-			context,
-			isDisabled: true,
+			label: "Indeterminate",
 		})}
 		${Template({
 			...args,
--- a/components/fieldgroup/stories/fieldgroup.stories.js
+++ b/components/fieldgroup/stories/fieldgroup.stories.js
@@ -271,13 +271,16 @@ HorizontalSideLabelCheckbox.parameters = {
 };
 
 /**
- * A group of read-only checkboxes that have been checked. In U.S. English, use commas to delineate items within read-only checkbox groups. In other languages, use the locale-specific formatting.
+ * Implementations should include the following behavior for read-only checkboxes:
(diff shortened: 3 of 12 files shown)


marissahuysentruyt (approved):
Love this- the read-only checkboxes are focusable, but cannot be checked/unchecked. 👍 I love that you added a bunch of the documentation from the guidelines site- thank you!

I did have a question about the expectations for what would be announced by a screenreader in read-only cases. Maybe that's something that's part of the research on read-only that's coming up. When using voice over, a read-only checkbox is announced like any of the other non-disabled checkboxes, so how would a user know that they cannot change the selection? I would think this would probably be a bigger discussion involving all of our team and some accessibility folks, so I'm not going to let it block my approval right now. Just wondered if you had any knowledge I don't!

components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -268,13 +271,16 @@ HorizontalSideLabelCheckbox.parameters = {
 };
 
 /**
- * A group of read-only checkboxes that have been checked. In U.S. English, use commas to delineate items within read-only checkbox groups. In other languages, use the locale-specific formatting.
+ * Implementations should include the following behavior for read-only checkboxes:
+ * - Read-only checkboxes are immutable, i.e. their checked state cannot be changed.
+ * - Unlike disabled checkbox groups, the normally focusable elements of a checkbox group should remain focusable.
  */
 export const ReadOnly = Template.bind({});
```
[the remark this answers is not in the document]
jawinn: That'd be a good idea for CSS-1005 that looks at this issue for radios.

jawinn:
> When using voice over, a read-only checkbox is announced like any of the other non-disabled checkboxes, so how would a user know that they cannot change the selection?

Good call-out. VoiceOver reads it the same way for React Spectrum's version that uses `aria-readonly`. I've updated this to use `aria-disabled`, which now announces it as "dimmed" and does not announce the keyboard shortcut to check/uncheck. See the updated PR notes about that.

components/fieldgroup/stories/template.js
```diff
@@ -46,6 +50,7 @@ export const Template = (
 						label,
 						isRequired,
 						alignment: labelPosition === "side" ? "right" : "top",
+						id: groupLabelId,
 					},
 					context,
 				),
```
5t3ph, line 56: For the input loop, consider checking one of them to be able to test for sure the immutability. [Here's how I did it](https://github.com/adobe/spectrum-css/pull/3350/files#diff-9d5d7da10a73aa5f35dcd706223f6b2198005d038d0310251ee6d2fb9c673541R72) for the radio update as one option.

components/fieldgroup/stories/fieldgroup.stories.js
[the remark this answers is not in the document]
5t3ph: Added in #3350

components/checkbox/stories/template.js
```diff
@@ -50,7 +50,7 @@ export const Template = ({
 					typeof size !== "undefined",
 				[`${rootClass}--emphasized`]: isEmphasized,
 				["is-indeterminate"]: isIndeterminate,
-				["is-disabled"]: isDisabled|| isReadOnly,
+				["is-disabled"]: isDisabled,
```
marissahuysentruyt, line 53: I like that we were able to remove the `isReadOnly` arg from this. It immediately makes sense now to any newcomer to the codebase.

components/checkbox/stories/template.js
```diff
@@ -62,13 +62,19 @@ export const Template = ({
 				type="checkbox"
 				class="${rootClass}-input"
 				aria-labelledby=${ifDefined(ariaLabelledby)}
+				aria-disabled=${ifDefined(isReadOnly ? "true" : undefined)}
```
marissahuysentruyt, line 65: I can look it up on MDN, but I wonder why `aria-disabled` adds the "dimmed" part of the screen reader announcement, as opposed to just saying something like "read-only." 🤔

components/checkbox/stories/template.js
```diff
@@ -62,13 +62,19 @@ export const Template = ({
...
-				@change=${function() {
-					if (isDisabled) return;
-					updateArgs({ isChecked: !isChecked });
+				@change=${(e) => {
+					if (isReadOnly) {
+						// Make checked value immutable for read-only.
+						e.preventDefault();
+						e.target.checked = !e.target.checked;
+					}
+					if (isDisabled || isReadOnly) return;
```
marissahuysentruyt, line 76: Would we ever get to this "or" portion of the code, the `|| isReadOnly`?

The code would run through the first `if(isReadOnly)`, and if it's also disabled, it'll return anyways, but I guess my question is do we need the "or read only" in this `if(isDisabled || isReadOnly)` block. Could it run through _both_ if blocks, if a read-only checkbox isn't disabled? Should the return just go in the `if(isReadOnly)` block?

components/checkbox/stories/checkbox.test.js
```diff
@@ -27,6 +27,10 @@ export const CheckboxGroup = Variants({
 			testHeading: "Checked",
 			isChecked: true,
 		},
+		{
+			testHeading: "Indeterminate",
```
marissahuysentruyt, line 31: Great catch on all of the missing indeterminate tests!

marissahuysentruyt (approved):
I think this is looking good. The read-only behavior works as you described, VoiceOver is announcing the read-only checkboxes in a way that is distinct from other editable checkboxes, and field group no longer has those commas! Just a few things from me! 


- love all of the new tests you added
- thanks for going so deep into the `aria-labelled` / `aria-readonly` / `aria-disabled` attributes. I might ping you for resources so I can keep those straight and know when to use them.

components/checkbox/stories/template.js
jawinn: Yes, since readonly is not disabled. This is as intended. If readonly, it keeps the value from changing on click. Then it returns if it's disabled or readonly, and only runs updateArgs to changed the isChecked control value for checkboxes that aren't disabled or readonly.

components/fieldgroup/stories/template.js
jawinn: That's a good idea to show one checked in the example. I added a commit for this. I adjusted it slightly so it won't affect any `isChecked` value coming from the `items` data (in case that changes to show any checked by default): `...(isReadOnly ? {isChecked: i === 1} : {}),`

5t3ph (approved):

pfulton:
@jawinn Just need you to talk a look at the Chromatic changes here. Once those are good to go, we can merge!

PR #3350, opened 2024-10-29
fix(radio): correct display of read-only state

5t3ph:
## Description

Same motivation as CSS-1004 for read-only checkbox, as being updated in PR #3328 

**Field group**:
- modifications to Storybook templates to support custom names for radio groups for enabling screen reader testing
- added story for read-only radios

**Radio**:
- Modifies read-only CSS to enable a visible UI input
- Template (similar modifications as for checkbox): 
  - JS added to make radio choice immutable as a group to prevent changing via pointer or keyboard
  - `disabled` attribute removed when read-only. Read-only still needs to be focusable / in the tab order
  - added `aria-disabled="true"` to not announce as interactive by screen readers (ex. "dimmed" in VO)
      - confirmed that `aria-readonly` had no effect on radio buttons, or even when set on `role="radiogroup"` at least in VO, but left in case other AT make use of it

### Validation steps

@jawinn 
- [x] field group docs page: read-only radio shows full input UI
- [x] radio docs page: read-only radio shows full input UI
- [x] review docs updates for both examples
- [x] read-only radio groups are immutable via both pointer (click) and keyboard access
- [x] read-only radios are keyboard focusable
- [x] field group of radios are announced as a group by screen reader (VO)
- [x] read-only radios are announced as dimmed/disabled by screen reader

### Regression testing

Expected changes:
- read-only radio now displays the input UI

Validate:

## Screenshots

**Before**:
image

**After**:
image

## To-do list


components/fieldgroup/stories/fieldgroup.stories.js
```diff
@@ -71,17 +71,14 @@ export default {
 		helpText: "Select an option.",
 		items: [
 			{
-				id: "apple",
```
5t3ph, line 74: these were causing issues with the full docs page when testing with VO since they weren't unique, but the radio template handles for assigning a unique ID

components/radio/stories/radio.stories.js
```diff
@@ -130,11 +130,15 @@ Disabled.parameters = {
...
+ * This allows for label content to be copied, but prevents the input from being interacted with or changed.
  *
- * - Read-only radio buttons are disabled, but not all disabled radio buttons are read-only.
- * - Read-only radio buttons do not have a focus ring, but the button should be focusable.
+ * Read-only radio buttons:
+ * - prevent interaction like disabled, but not all disabled radio buttons are read-only
+ * - are immutable, i.e. their checked state cannot be changed
+ * - are keyboard focusable and communicate state to assistive technology
+ * - use `aria-disabled` since the `readonly` attribute [is not valid](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#overview)
+ * - use `preventDefault()` on click to prevent changing the selection via pointer or keyboard interaction
```
jawinn, line 141:
```suggestion
 * Read-only radio buttons:
 * - prevent interaction like disabled, but not all disabled radio buttons are read-only
 * - are immutable, i.e. their checked state cannot be changed
 * - are keyboard focusable and communicate state to assistive technology
 * - use `aria-disabled` since the `readonly` attribute [is not valid](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly#overview) and `aria-readonly` is not currently announced by the majority of screen readers.
```
What do you think about leaving out the last sentence since we don't necessarily need to prescribe the how, and that is covered under the "immutable" bullet point?

I could also see dropping the first bullet point that originates from the original docs, as it feels a little confusing and not totally necessary? I'm on the fence about that one and don't have a problem with keeping it.

jawinn (changes requested):
This is looking really good! I just had a few suggestions after wrapping up with the related checkbox work:

components/radio/stories/radio.stories.js
5t3ph: I guess I was trying to emphasize that it required scripting to accomplish and wasn't an inherent feature, which is different behavior than applying `readonly` to other input types, but if WC pulls in the change we can confirm.

.changeset/hot-timers-count.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/radio": patch
+---
+
+Corrects the styles of the read-only state to show radio the inputs and allow visible focus. Also adds `aria-disabled` since `aria-readonly` isn't well supported, and story demonstrates scripting to make selection for read-only radios immutable.
```
cdransf, line 5: Should this be `show the radio inputs`? ✨

cdransf (approved):

jawinn (approved):

PR #2786, opened 2024-05-24
feat!: 🎉 Spectrum 2 foundations

pfulton:
## Description

This pull request brings in the Spectrum 2 design foundations tokens to start the migration from our current design system to the latest. To accommodate these changes, this pull request makes a few updates to our build to facilitate more robust theme switching.

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

## Remaining TODOs

- [x] Need a decision about how we're going to handle the QuickAction deprecation in the Card documentation
- [x] Resolve the remaining S1 & Express VRTs
- [x] Write out additional tests cases for the postcss-splitinator plugin to cover the expanded pseudo and combinator support
- [x] Storybook: clean up theme switching to load only the required assets per context and check for bugs or regressions
- [x] After the modes branch is merged, resolve conflicts in rebase and add new context mode; check vrt baselines for regressions in express and legacy

### Validation steps

**Validating the compiled output**

- [x] Expect to see only the 1 top-level class `.spectrum--express` in the `dist/themes/express` file; this class should contain --system prefixed variables
- [x] Expect to see only the 1 top-level class `.spectrum--legacy` in the `dist/themes/legacy.css` file; this class should contain --system prefixed variables
- [x] Expect to see only the 1 top-level class `.spectrum` in the `dist/themes/spectrum.css` file; this class should contain --system prefixed variables
- [x] Expect to see the component class names in the `dist/index-theme.css` file but no top-level classes (i.e., no `.spectrum` or `.spectrum--legacy`)

**Validating the compiled token output**

- [x] Expect to see a `tokens/components` and `tokens/dist/css/components` folder with `bridge`, `express`, `legacy`, and `spectrum` folders inside
- [x] Expect the `tokens/components/bridge/actionbutton.css` file to match the `components/actionbutton/dist/index-theme.css` output
- [x] Expect the `tokens/components/spectrum/actionbutton.css` file to match the `components/actionbutton/dist/themes/spectrum.css` output
- [x] Expect the `tokens/components/legacy/actionbutton.css` file to match the `components/actionbutton/dist/themes/legacy.css` output
- [x] Expect the `tokens/components/express/actionbutton.css` file to match the `components/actionbutton/dist/themes/express.css` output

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [x] The pages render correctly, are accessible, and are responsive.

2. If components have been modified, VRTs have been run on this branch:

[...]

PR #3365, opened 2024-11-04
feat(popover): migrate s2 popover

marissahuysentruyt:
## Description

Welcome to the new and improved S2 popover! 🥳 

This PR updates tokens used in the popover component. The work includes new S2 tokens specifically for colors, padding and the border radius for popover. There is also a transparent border in light mode and a visible border in dark mode.

There are several components that have popover as a subcomponent. Those were updated with just enough additional code to make them "work." Coachmark and picker did have some additional CSS needs, particularly relating to the new popover corner rounding, and the new `bottom-start` positioning.

### Popover Mods

Some mods have been renamed to match their new token names, or better reflect their purposes:

| Old mod name | New mod name |
| ----- | ------ |
| `--mod-popover-content-area-spacing-vertical` | `--mod-popover-content-area-spacing` |
| `--mod-popover-shadow-blur`                   | `--mod-popover-drop-shadow-blur`           |
| `--mod-popover-shadow-color`                  | `--mod-popover-drop-shadow-color`          |
| `--mod-popover-shadow-horizontal`             | `--mod-popover-drop-shadow-x`              |
| `--mod-popover-shadow-vertical`   | `--mod-popover-drop-shadow-y`              |

### Notes
- the background color and the border radius of the menu items is a known bug. It will be addressed in a separate branch.
- there's a few coach mark questions that are still outstanding as well, including the needed drop shadow on the action menu popover ([css-1066](https://jira.corp.adobe.com/browse/CSS-1066)), and the padding (as noted in a few comments).
- the picker 's popover styling is a known issue. I attempted to implement something like: 
```
    /* popover passthroughs */
    --mod-popover-animation-distance: var(--spectrum-picker-spacing-picker-to-popover);
```
but the cascade doesn't quite work because of the markup differences. [CSS-1065](https://jira.corp.adobe.com/browse/CSS-1065) captures some of the thoughts and issues, and Josh brought up some good questions/context in [his comment below](https://github.com/adobe/spectrum-css/pull/3365#discussion_r1852943133) in regards to nesting elements and SWC.

### Designs
[S2 Popover Token specs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=11533-696)
[S2 / Desktop Popover](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?node-id=42086-5684)

### Jira
[CSS-615](https://jira.corp.adobe.com/browse/CSS-615)

### Validation steps

- [ ] Pull down the branch or use the deploy preview
- [ ] Visit the [popover storybook](https://pr-3365--spectrum-css.netlify.app/preview/?path=/docs/components-popover--docs). Verify [all stories from `main`](https://opensource.adobe.com/spectrum-css/?path=/docs/components-popover--docs) appear on the docs page. No changes should have been made to popover positions or tip placements.
- [ ] The default popover matches the [s2 popover specs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=27417-7447).
- [ ] Verify the action button trigger and popover menu items are accessible via keyboard.
- [ ] Create various combinations of the popover to verify combos are considered in the CSS:
    - [ ] mobile vs desktop 
    - [ ] light vs dark mode (the border color/transparency should change in dark mode)
    - [ ] popover positioning and tip placement
- [ ] Verify new tokens are used: 
    - [ ] `popover-border-color` (see the comment in index.css)
    - [ ] `popover-border-opacity`
    - [ ] `drop-shadow-elevated-color`
    - [ ] `drop-shadow-elevated-x`
    - [ ] `drop-shadow-elevated-y`
    - [ ] `drop-shadow-elevated-blur`
- [ ] Verify updated tokens are used: 
    - [ ] `corner-radius-large-default`
    - [ ] `popover-edge-to-content-area` (`popover-top-to-content-area` was removed from tokens)
    - [ ] `border-width-100`
    - [ ] `background-layer-2-color` (it may not be pointing to `gray-25` as intended, but should once we make the full change to S2 tokens)
- [ ] Chromatic coverage continues to maintain test templates, and does not have additional variations/test scenarios.
- [ ] Action bar, action menu, coach mark, combobox, contextual help, date picker, menu (specifically the "submenu in popover" story), and picker (which all call for a popover component) look acceptable. There may be additional style issues to handle during those components' migrations, as opposed to this PR. 

### Regression testing

Validate:

## Screenshots

Before: 
Screenshot 2024-11-06 at 5 16 11 PM

After: 
Screenshot 2024-11-06 at 5 16 28 PM

## To-do list


components/popover/metadata/mods.md
```diff
@@ -1,18 +1,18 @@
...
+| `--mod-popover-animation-distance`        |
+| `--mod-popover-background-color`          |
+| `--mod-popover-border-color`              |
+| `--mod-popover-border-width`              |
+| `--mod-popover-content-area-spacing`      |
+| `--mod-popover-corner-radius`             |
+| `--mod-popover-dropshadow-blur`           |
+| `--mod-popover-dropshadow-color`          |
+| `--mod-popover-dropshadow-x`              |
+| `--mod-popover-dropshadow-y`              |
```
marissahuysentruyt, line 14: Is this the correct way to write these properties? Should it be "dropshadow" or "drop-shadow"?

components/popover/stories/template.js
```diff
@@ -16,146 +20,284 @@ export const Template = ({
...
+		}, 500);
+	});
 
-					if (x) transforms.push(`translateX(calc(var(--flow-direction) * calc(${parseInt(x, 10)}px ${xOffset})))`);
-					if (y) transforms.push(`translateY(calc(${y}px ${yOffset}))`);
+	if (!skipAlignment) {
+		switch (position) {
+			case "top":
+				popoverWrapperStyles["inline-size"] = "var(--spectrum-popover-width)";
+				popoverAlignment["inset-block-end"] = "100%";
```
marissahuysentruyt, line 80: I removed 2 lines from these test cases about "inset-inline-start" being set to "0". Something about those 2 lines was uncentering the top and bottom positioning of the popover. I was wondering if maybe it had to do with the additional horizontal padding popover has now? 


https://github.com/adobe/spectrum-css/blob/196ef4c0dd64f9c6c3803197453cd8dc28cc4ff4/components/popover/stories/template.js#L70
https://github.com/adobe/spectrum-css/blob/196ef4c0dd64f9c6c3803197453cd8dc28cc4ff4/components/popover/stories/template.js#L95

components/actionbar/stories/actionbar.stories.js
```diff
@@ -68,7 +68,7 @@ export default {
 	parameters: {
 		actions: {
 			handles: [
-				...Popover.parameters.actions.handles,
+				...(Popover?.parameters?.actions?.handles ?? []),
```
marissahuysentruyt, line 71: This is pulled over from `main`. I tried to pull the least amount of code changes in order to get them all to render.

components/actionbar/stories/template.js
```diff
@@ -68,6 +68,8 @@ export const Template = ({
 					],
 				}),
 			],
+			popoverHeight: 42,
```
marissahuysentruyt, line 71: Same here: this is pulled over from main.

components/coachmark/index.css
```diff
@@ -120,7 +119,11 @@ governing permissions and limitations under the License.
 .spectrum-CoachMark-action-menu {
 	white-space: nowrap;
 	z-index: 1;
-	margin-inline-start: var(--mod-coachmark-heading-to-action-button, var(--spectrum-coachmark-heading-to-action-button))
+	margin-inline-start: var(--mod-coachmark-heading-to-action-button, var(--spectrum-coachmark-heading-to-action-button));
+
+	& .spectrum-Popover {
+		--mod-popover-filter: drop-shadow(var(--mod-popover-drop-shadow-x, var(--spectrum-popover-drop-shadow-x)) var(--mod-popover-drop-shadow-y, var(--spectrum-popover-drop-shadow-y)) var(--mod-popover-drop-shadow-blur, var(--spectrum-popover-drop-shadow-blur)) var(--mod-popover-drop-shadow-color, var(--spectrum-popover-drop-shadow-color)));
+	}
```
marissahuysentruyt, line 126: I think this needs some help. For the record, none of these variables are defined. I just copied them verbatim from the popover CSS. I was trying to add the drop shadow back onto this popover, since it's lying on top of the rest of the coach mark's popover. I assume that's what is supposed to happen given this note: "Containers that appear on top of content, such as menus and tooltips, have drop shadows to show elevation" in Figma, but then also the CSS comment to "prevent nested popovers from affecting each others drop shadow filters" (found here)

components/popover/index.css
```diff
@@ -14,94 +14,110 @@ governing permissions and limitations under the License.
...
+	@extend %spectrum-overlay--open;
 }
 
 /* prevent nested popovers from affecting each others drop shadow filters */
 .spectrum-Popover * {
-  --mod-popover-filter: none;
+	--mod-popover-filter: none;
+
+	& > .spectrum-Popover {
+		margin: var(--spectrum-popover-nested-container-spacing);
```
marissahuysentruyt, line 119: This was my attempt at trying to incorporate the "container padding" that's outlined in Figma. I'm still unsure if this component (the popover itself) is really the right place for that "container padding." Should it go on that popover wrapper we have instead?

https://github.com/adobe/spectrum-css/blob/61af0743f874d799b8fbb789aff8ad2580c5994a/components/popover/stories/template.js#L165

marissahuysentruyt: /* & > .spectrum-Popover {
		margin: var(--spectrum-popover-nested-container-spacing);
	} */

components/coachmark/index.css
marissahuysentruyt: Doing this is also introducing extra mods to coach mark and I'm not sure we _need_ to do that.

components/popover/stories/popover.stories.js
```diff
@@ -38,147 +47,140 @@ export default {
...
-			triggerId: "trigger-nested-2",
+			...passthroughs,
+		}, context),
+		(passthroughs, context) => Template({
+			position: "end-top",
 			isOpen: true,
 			customStyles: {
-				"margin-inline-start": "136px",
-				"margin-block-start": "32px"
+				"margin": "var(--spectrum-popover-nested-container-spacing)",
```
marissahuysentruyt, line 176: This was my attempt at adding the "container padding" that is outlined in Figma. It felt weird to put those styles on the popover itself, i.e.: 
```
& > .spectrum-Popover {
  margin: var(--spectrum-popover-nested-container-spacing);
}
```

And then users can add those styles if they need. Any thoughts on where these styles are supposed to go?

...since it certainly feels like they belong on the containing element. So that's what this is doing- putting that container padding on the popover wrapper. 

https://github.com/adobe/spectrum-css/blob/61af0743f874d799b8fbb789aff8ad2580c5994a/components/popover/stories/template.js#L165

components/popover/stories/popover.stories.js
```diff
@@ -197,8 +199,169 @@ Nested.args = {
...
+		disableSnapshot: true,
+	},
+};
+
+/**
+ * Popovers can display different elements within their content area. This example uses the
+ * [dialog](?path=/docs/components-dialog--docs) component within the popover.
+ */
+// @see https://opensource.adobe.com/spectrum-web-components/components/popover/#dialog-popovers
+export const DialogStyle = Template.bind({});
```
marissahuysentruyt, line 248: This isn't displaying correctly on the docs page. 

maybe we need to remove `display: inline-flex;` for this story?

components/popover/index.css
```diff
@@ -14,94 +14,106 @@ governing permissions and limitations under the License.
...
 
-  /* animation distance is equal to and responsible for popover offset */
-  --spectrum-popover-animation-distance: var(--spectrum-spacing-100);
+	/* animation distance is equal to and responsible for popover offset */
+	--spectrum-popover-animation-distance: var(--spectrum-spacing-100);
 
-  --spectrum-popover-background-color: var(--spectrum-background-layer-2-color);
-  --spectrum-popover-border-color: var(--spectrum-gray-400);
+	--spectrum-popover-background-color: var(--spectrum-background-layer-2-color);
+	--spectrum-popover-border-thickness: var(--spectrum-border-width-100);
```
rise-erpelding, line 28: It looks like there are still several references to `--spectrum-popover-border-width` which is defined in the `themes/spectrum.css`, this custom prop either needs to be updated to the same name or the references to the old one maybe should be removed and replaced... 🤔 I'm honestly not sure what the best move is here, it might be worth looking over the [Foundations PR](https://github.com/adobe/spectrum-css/pull/2786) to get a feel for how we anticipate the different themes will work together in `main`.

components/popover/index.css
```diff
@@ -14,94 +14,106 @@ governing permissions and limitations under the License.
...
+	padding: var(--mod-popover-content-area-spacing, var(--spectrum-popover-content-area-spacing));
+
+	display: inline-flex;
+	flex-direction: column;
+
+	border-style: solid;
+	border-radius: var(--mod-popover-corner-radius, var(--spectrum-popover-corner-radius));
+	border-color: var(--highcontrast-popover-border-color, var(--mod-popover-border-color, var(--spectrum-popover-border-color-with-transparency)));
+
+	/* border-color: var(--spectrum-popover-border-color-default); */
```
rise-erpelding, line 87: A comment here about why this is commented out would be nice, but feel free not to add one if you disagree.

components/picker/stories/template.js
```diff
@@ -147,11 +149,14 @@ export const Template = ({
 	const popoverMarkup = content.length !== 0 ? Popover({
 		isOpen: isOpen && !isDisabled && !isLoading,
 		withTip: false,
-		position: "bottom",
+		position: "bottom-start",
```
rise-erpelding, line 152: Something is looking a little funny with this position in Picker, although changing it back to `bottom` wouldn't fix it here:

image

rise-erpelding (changes requested):
This is looking great! I gave this a (relatively) quick once-over and all the tokens look like they're being used. I called out one issue I noticed in the CSS with border-width and noted in Picker that the spacing is looking a little weird.

I'll plan on circling back to this and looking more closely at your questions on the next round!

components/popover/stories/popover.stories.js
rise-erpelding: Great question! I don't really have an answer, but I did look into it and will share what I've found so far:

Strangely enough, you [asked](https://github.com/adobe/spectrum-css/pull/2743#discussion_r1672743196) about this exact thing when we did S2 for Tooltip, which has the same container padding token in its [spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=9706-2117&m=dev), and there I think we decided that it was sort of out of our hands to ensure that that token was used in the implementation.

It's also worth mentioning that, if I dig through the [wiki](https://wiki.corp.adobe.com/display/AdobeDesign/Spectrum+Design%3A+Token+Specs) for the S1 specs, I can find the Popover S1 token specs in the XD file from the [Sharepoint](https://adobe.sharepoint.com/sites/Spectrum/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FSpectrum%2FShared%20Documents%2FSpectrum%20Design%2FToken%20specs), and the container padding doesn't change from S1 to S2, and it doesn't seem like it was implemented anywhere in S1 (it's the same case for tooltip, if you're curious).

I think it's fine if we want to add some margin to ensure that there's spacing here, but I have a couple of follow-up questions about that:
1. Is this `--spectrum-popover-nested-container-spacing` custom prop meant to be used for all container spacing, or is there a reason it's only used for the nested container/only named for nested container spacing?
2. Was the intention to put this margin on the wrapper? It looks like this margin is ending up on the popover (on the same element as `.spectrum-Popover`), not the wrapper, but applying them to `popoverWrapperStyles` instead puts the margin on the action button, which isn't what we want either. It seems like we'd probably want some space here anyway (maybe?) but I'm not sure we need to use the token (but if I'm missing something let me know!).
    image
    image
3. As far as I can tell, the `--spectrum-popover-nested-container-spacing` is defined in index.css then not used in index.css, is that going to cause some linting issues in `main`? (I would have thought there would be some in this branch too, but I didn't see any.)

Does any of that help here?

rise-erpelding:
> 2. Should popover handle the nested menu overlap margin? My guess is no...

Does this apply in this situation? It feels like it's slightly different than that example because this one has an action button? This one for [submenu in popover](https://spectrum-css.netlify.app/?path=/docs/components-menu--docs#submenu%20in%20popover) applies the custom styles to Popover to move it, although I'm sure there are probably other ways we could handle it.

components/popover/index.css
```diff
@@ -14,94 +14,106 @@ governing permissions and limitations under the License.
...
+
+	border-width: var(--mod-popover-border-width, var(--spectrum-popover-border-thickness));
```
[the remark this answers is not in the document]
marissahuysentruyt: Sure! Good catch 😄

components/popover/index.css
marissahuysentruyt: It was an extraneous comment from earlier edits, trying to popover-border-color-default, which is one of those rgba() tokens.

components/picker/stories/template.js
marissahuysentruyt: I think I have that worked out now.

https://github.com/adobe/spectrum-css/pull/3365/commits/b76e4ac82a4def48c3c50ee792ca24a2ea8f8d96
https://github.com/adobe/spectrum-css/pull/3365/commits/5e12534072872bf1f5abd6ba5b4638544dcb24ea

Would this be considered "fragile?" Picker on `main` has some significant updates that this picker doesn't, so the new CSS in these commits won't really apply with the new picker markup. Maybe I should leave some comments? Very happy for feedback on the fixes either way.

marissahuysentruyt: also, the side label shouldn't be broken now!

components/popover/stories/popover.stories.js
marissahuysentruyt: I am totally fine with _not_ implementing the container padding, but I suppose the question is the same for this specific, nested popover story.

> Is this --spectrum-popover-nested-container-spacing custom prop meant to be used for all container spacing, or is there a reason it's only used for the nested container/only named for nested container spacing?

I don't believe this should be used for all container spacing. I'm using it only for the nested popover because those 2 popovers "touch" now. The spacing between the trigger and the popover itself is dictated by the `animation-distance`. That animation distance if from the trigger, which in this case is 8px from the action buttons. But now with the new, additional padding popover has (which is also 8px), the `.spectrum-Popover` divs "touch." 

I suppose instead of `spacing-100`, I could use the `animation-distance` token?

without `--spectrum-popover-nested-container-spacing`:
Screenshot 2024-11-18 at 1 39 27 PM

with `--spectrum-popover-nested-container-spacing`:
Screenshot 2024-11-18 at 1 40 31 PM

> As far as I can tell, the --spectrum-popover-nested-container-spacing is defined in index.css then not used in index.css, is that going to cause some linting issues in main? (I would have thought there would be some in this branch too, but I didn't see any.)

Correct, I didn't actually use it in the CSS. I only used it in the stories file, so I moved it and just defined it in the stories file too. Good call.

Let me know if this solution also addresses your second question about not needing to use that token! https://github.com/adobe/spectrum-css/pull/3365/commits/3ee1873a2fd21f2f69fc93c0e6f055bb4116151c

rise-erpelding: This looks great with the animation distance token!

components/popover/stories/template.js
rise-erpelding: This is a good callout/observation! After looking into it, I concur that it looks like it's because the horizontal padding is added, but I think we might be able to keep these two lines as they were if we make an adjustment to the `popoverWidth` and `popoverHeight` that are being set in the template args and the stories.js. Those dimensions were being used to match the calculated popover size but don't look like they are anymore, so either way we probably want to update them to account for that padding.

(Adding some screenshots, in case it makes more sense with illustrations:)

Here's the old popover (on main) with 89x142 dimensions seen in the custom props in the wrapper and also for .spectrum-Popover:
image

And here's the branch, but the dimensions don't match the custom props:
image

components/popover/index.css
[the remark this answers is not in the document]
rise-erpelding: Oh no I just spotted 2 more `--mod-popover-border-width`s in coachmark 😬

components/popover/index.css
rise-erpelding: Oh! I thought it went with this one, and if we were able to get to a point where we could use that, we'd be able to use the border-color that was commented out:
```
/* TODO: Cannot use popover-border-color token until the value no longer has the rgba() function alpha/opacity */
/* --spectrum-popover-border-color-default: var(--spectrum-popover-border-color); */
```

rise-erpelding (commented):
Still looking over this, but want to drop the current batch of comments!

components/popover/index.css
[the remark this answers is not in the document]
marissahuysentruyt: Eagle eyes! Thanks https://github.com/adobe/spectrum-css/pull/3365/commits/11481406b846997060144910ff84da99c916c07a

components/popover/index.css
marissahuysentruyt: Yes, it was related to that `TODO` comment. `popover-border-color` doesn't have the transparency built in though, which is why I went with the rgb token instead. I added a little clarity in that comment- does that make more sense? https://github.com/adobe/spectrum-css/pull/3365/commits/05fc517e22ee3c0b93ebcb41f3d3e931a37d808c

components/picker/stories/template.js
marissahuysentruyt: ah yes! `:has()`! That feels better to me at least. https://github.com/adobe/spectrum-css/pull/3365/commits/cb2ac7ad427fc9f64394d06879fb49cf858382cf

jack-black-yes

I would have to reach out to Josh on why popover has the wrapper. This top-most div is the wrapper of the popover in production, so I _think_ it is in `main`.

Screenshot 2024-11-21 at 2 49 11 PM

components/popover/stories/template.js
marissahuysentruyt: Ok. I added the additional padding (16px I guess, on both sides) to the args, and like you said, that fixes the off-centering. 🥳 The `inset-inline-start` lines were added back in! Those args are now set to 105px and 158px.

I'll try to get with @jawinn so I can understand more about the `popoverWidth` and `popoverHeight`. Are they supposed to be like min-width & height values? I'm definitely seeing smaller popovers, so those must be coming from the event listener. I'm also curious if i need to adjust the new combobox [popoverHeight](https://github.com/adobe/spectrum-css/blob/8382527c9d1aa0889829ef9868b609f3b018b9b3/components/combobox/stories/template.js#L110)/[popoverWidth](https://github.com/adobe/spectrum-css/blob/8382527c9d1aa0889829ef9868b609f3b018b9b3/components/combobox/stories/template.js#L143) I introduced (which were on main). Do each of those sizes need to be adjusted for the extra padding?

Screenshot 2024-11-21 at 3 59 35 PM
Screenshot 2024-11-21 at 3 40 37 PM

jawinn: From what I can tell about `popoverWidth` and `popoverHeight`, they are to define the actual fixed dimensions of the popover wrapper, I think to help with our example positioning. Either passed as args or the event listener tries to determine it dynamically. It appears that `popoverWrapperStyles["inline-size"] = "var(--spectrum-popover-width)";` is used in the alignment code of the template, and perhaps the height custom property is not being used. Perhaps we can look at this if there are still questions here.

jawinn (changes requested):

components/coachmark/index.css
jawinn: As far as I can tell after some research, the `filter(drop-shadow)` being used will be affected by the children and there isn't a way to exclude the nested popover from affecting the coachmark popover edge.

We use `drop-shadow` instead of `box-shadow` on popover in order for the shadow to also follow the triangle tip. So one thought I had is that in popover we could use `box-shadow` instead only when there is no tip displayed, like for coachmark. I think that would solve this situation, since coachmark doesn't appear to have a tip option. I'd like to get others thoughts on that idea.

If there are other situations where the popover _with a tip_ has an issue with these nested shadows, the only other workaround I saw was applying the filter only to a separate element like a pseudo element---that doesn't sound great, since it'd be a messy markup change, likely involving an empty element and moving where the background and tip psuedo element live.

components/picker/stories/template.js
marissahuysentruyt: > I have some concerns with this picker CSS change being discussed here.

I agree- the picker on main has some significant updates that this picker doesn't, so I'm not even sure these CSS changes will really apply with the new picker markup. Should I hold off on more changes? Picker has already been migrated to S2, but without the changes from `main`, so I'm not sure where the line is in stopping work.

jawinn (changes requested):
One other thing I noticed is a display issue with the corners overlapping when I hover over the coachmark action button's popover menu items. Is this missing the padding? Feel free to leave this for the coachmark migration if it's outside the scope of this PR:

PR:
Screenshot 2024-11-22 at 9 43 18 AM

S1:
Screenshot 2024-11-22 at 10 24 10 AM

.changeset/twenty-ravens-divide.md
```diff
@@ -0,0 +1,8 @@
+---
+"@spectrum-css/coachmark": minor
```
marissahuysentruyt, line 2: Just making sure here: the changes in the coachmark CSS are actually a result of the breaking changes from popover. Would that still be considered a minor change in the coachmark CSS?

components/popover/stories/popover.stories.js
marissahuysentruyt: How's this? https://github.com/adobe/spectrum-css/pull/3365/commits/a0ff0b26c9651db3d7f7d7cd2e2339eb228e854c

marissahuysentruyt:
> One other thing I noticed is a display issue with the corners overlapping when I hover over the coachmark action button's popover menu items. Is this missing the padding? Feel free to leave this for the coachmark migration if it's outside the scope of this PR:

I was going to leave this for the coachmark. I have a card that captures this right now: https://jira.corp.adobe.com/browse/CSS-1067

marissahuysentruyt:
Some follow-up has been created: 
[CSS-1065](https://jira.corp.adobe.com/browse/CSS-1065)(in response to picker questions found in this branch)
[CSS-1066](https://jira.corp.adobe.com/browse/CSS-1066)(in response to the coachmark dropshadow/filter questions)
[CSS-1067](https://jira.corp.adobe.com/browse/CSS-1067)(in response to the coachmark menu padding questions)

components/picker/stories/template.js
```diff
@@ -134,19 +134,22 @@ export const Template = ({
...
 		labelPosition,
 		ariaLabeledBy: fieldLabelId,
 	}, context);
 
-	const popoverMarkup = content.length !== 0 ? Popover({
+	const popoverMarkup = popoverContent.length !== 0 ? Popover({
 		isOpen: isOpen && !isDisabled && !isLoading,
 		withTip: false,
-		position: "bottom",
+		position: "bottom-start",
```
marissahuysentruyt, line 145: `bottom-start` is from `main` & `s2-foundations-redux`. There's not explicit direction on the popover's positioning within picker that I found, but I don't think I see an issue with this change. Combobox, action menu, contextual help, date picker all use `bottom-start`. Happy to revert if there are any concerns.

components/coachmark/index.css
```diff
@@ -18,8 +18,8 @@
 	/* @passthrough start */
 	--mod-buttongroup-justify-content: flex-end;
 	--mod-popover-border-width: var(--spectrum-border-width-100);
-	--mod-popover-corner-radius: var(--spectrum-corner-radius-100);
-	--mod-popover-content-area-spacing-vertical: 0;
+	--mod-popover-corner-radius: var(--spectrum-corner-radius-large-default)
+	--mod-popover-content-area-spacing: 0;
```
marissahuysentruyt, line 22: I updated this mod name because I changed popover-content-area-spacing in the popover CSS- there's now padding on all sides of popover instead of just on the top & bottom. However, I am unsure if this is actually the correct styling for the popovers in coachmark. Is this something to address in this PR, or in the coachmark PR #3412?

components/coachmark/index.css
marissahuysentruyt: [CSS-1066](https://jira.corp.adobe.com/browse/CSS-1066) captures this comment.

aramos-adobe (approved):

rise-erpelding (approved):
I think this is good to go! Thanks for all your hard work on this, and for all the follow-up cards that you've made to explore some of the other complexities that came about with this migration - those cover a whole lot of the things that I noticed reviewing this.

I made a tiny callout about maybe linking Down state docs if you want, but I have a couple of other non-blocking questions/maybe concerns:
* I'm honestly a little puzzled as to why the left/right-positioned popovers with tip are so far away from the source compared to the top/bottom-positioned popovers with tip. This isn't anything you introduced, because I saw it on main too! To me, this seems like it could be a bug - the spacing is set based on `popover-pointer-width` if it's side-positioned, and based on `popover-pointer-height` if it's vertically positioned from its source, but `popover-pointer-width` and `popover-pointer-height` don't ever change, if I'm not mistaken - it's just that for side-positioned popovers inline-size is set to the height and otherwise it's set to the width.
    image
* It also totally feels weird that we get popovers that look like this in rtl where the tip is way away from the source, but that would be a reason to use the logical position, I guess, so we probably don't need to take any further action there.
    image
* Lastly, I know you have a card for the drop-shadows as pertaining to coachmark already, but I can't help but notice how much bigger the shadows look compared to Figma. I know I've done my fair share of digging and I think the mechanisms used to render the shadows in Figma are completely different than what is used in CSS for `filter: drop-shadow()` but that would be something that I'd love for us to standardize further at some point in the future.

That's all I've got! Nice work, once again!

components/popover/stories/popover.stories.js
```diff
@@ -101,10 +102,10 @@ export default {
 };
 
 /**
- * By default, popovers do not have a tip. Popovers without a tip should be used when the source has a
+ * By default, popovers do not have a tip and is positioned at the top. Popovers without a tip should be used when the source has a
  * visually distinct down state, in order to show the connection between the popover and its source.
```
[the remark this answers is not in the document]
marissahuysentruyt: christrageryes

components/popover/index.css
```diff
@@ -144,7 +144,7 @@
 	&.spectrum-Popover--withTip {
 		/* tip size minus where it overlaps with popover border */
 		/* stylelint-disable-next-line csstools/use-logical -- intentional, right stays on the same side even for RTL */
-		margin-left: calc(var(--mod-popover-pointer-width, var(--spectrum-popover-pointer-width)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
+		margin-left: calc(var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance)) - var(--mod-popover-border-width, var(--spectrum-popover-border-width)));
```
rise-erpelding, line 147: This change looks good! I'm wondering if `popover-animation-distance` is the token we want but some of the context may have left my brain since yesterday--the top/bottom positions are still using `popover-pointer-height` while the side positions use `popover-animation-distance`, right? It all computes to the same value in the end, but I'm wondering if the tokens ever change their values if that leaves room for inconsistency down the road?

marissahuysentruyt: Good points! So, I've used the `popover-pointer-height` token in place of the popover-animatino-distance`, which makes much more sense I think. 

To address your other comments from yesterday: 
- There are little documentation blurbs on the docs page for the positions that don't adapt to RTL. "* Text direction does not affect the position" and "* Changes side with text direction (like a logical property)". Is that offer enough clarity on the expected behaviors for users when they use particular positions? 
- I agree on the drop shadow. I thought maybe that our popover shadows were just looking "bigger" than the shadows in Figma because our background color is white, while Figma's background colors are light gray but...it's not light gray in all the files so that argument is nothing 😆 We have that card to document that we know it looks maybe a little funny, so I don't think I'm going to address anything further in this PR. Does that sound ok?

components/popover/stories/popover.stories.js
```diff
@@ -101,10 +102,10 @@ export default {
 };
 
 /**
- * By default, popovers do not have a tip. Popovers without a tip should be used when the source has a
- * visually distinct down state, in order to show the connection between the popover and its source.
+ * By default, popovers do not have a tip and is positioned at the top. Popovers without a tip should be used when the source has a
```
jawinn, line 105:
```suggestion
 * By default, popovers do not have a tip and are positioned at the top. Popovers without a tip should be used when the source has a
```

components/popover/stories/popover.stories.js
```diff
@@ -101,10 +102,10 @@ export default {
...
+ * visually distinct [down state](/docs/foundations-down-state--docs), in order to show the connection between the popover and its source.
  *
- * This example uses the [menu](?path=/docs/components-menu--docs) component within the popover, and a button as the source.
+ * This example uses the [menu](/docs/components-menu--docs) component within the popover, its position is `bottom`, and uses a button as the source.
```
jawinn, line 108:
```suggestion
 * This example uses the [menu](/docs/components-menu--docs) component within the popover, has a position of `bottom`, and uses a button as the source.
```

jawinn (approved):
Nice work working through this one. Approved!
I didn't see any issues. Just two tiny docs wording nitpicks here.

PR #3369, opened 2024-11-05
fix: remove storybook assets from shipped output

castastrophe:
## Description

Remove the storybook assets from the shipped output for components. These assets can cause issues downstream due to their dependencies on other Storybook packages in the project. For projects using automated dependency tooling, they might flag these dependencies as missing from the component.

## To-do list


pfulton (changes requested):
It looks like this may have had a negative side effect of making the publish check fail due to imports not being resolved.

castastrophe:
> It looks like this may have had a negative side effect of making the publish check fail due to imports not being resolved.

Good point!  I pushed up an alias'ing to Vite to allow internal resolution for storybook assets inside the monorepo.

pfulton (approved):
Looks good to me!

PR #3380, opened 2024-11-10
feat(progress-circle): migrating s2 tokens, refactor markup to use SVG element

aramos-adobe:
### S2 migration for progress circle
Changed from using border styles to `stroke-*` styles and attributes. The svg element allows for `stroke-linecap` to enable to the rounded path. 

#3430 Infield progress circle will be sharing this. This component is used in button, picker, and combobox

### Animation

`@keyframes spectrum-dashoffset-animation`:
This animation was created by React Spectrum and used to keep the same speed and bezier curve.

### Added tokens
`--spectrum-in-field-progress-circle-edge-to-fill`
`--spectrum-in-field-progress-circle-size-75`
`--spectrum-in-field-progress-circle-size-100`
`--spectrum-in-field-progress-circle-size-200`
`--spectrum-in-field-progress-circle-size-300`

### Removed styles & mods

`--spectrum-progress-circle-track-border-color-over-background`
`--spectrum-progress-circle-fill-border-color-over-background`
`--spectrum-progress-circle-track-border-style`
`--spectrum-progress-circle-track-border-style`
`--highcontrast-progress-circle-track-border-style`
`--mod-progress-circle-track-border-style`
`--spectrum-progress-circle-track-border-style`

### Validation steps
  1. Open the storybook doc for the progress circle component:
    - [x] Choose static color options to see correct tokens applied
    - [x] Toggle indeterminate option see the animation
    - [x] Range slider to see the flow of progress track fill

### Regression testing

Validate:

## To-do list


marissahuysentruyt (changes requested):
Woo- love the updated styles. I think I wrapped my head around the animation stuff, but if anything is confusing, let's pair! 🍐 

A couple of other thoughts I wasn't sure where to put: 
- Cassondra separated some of nested components in the `s2-foundations-redux` branch. For instance, progress bar and meter _were_ nested together, but now they're totally separate (https://github.com/adobe/spectrum-css/tree/s2-foundations-redux/components/meter). Same thing with form being separated from field label. You may have already know, but is there a way in this branch we can prep for infield progress circle being separate from progress circle? Or maybe we leave our comment/note/Jira ticket for ourselves? Should these 2 components be nested like they are now, or should they be separated into different component directories? 

- Let's just fix the typo in the file names. (infield-progress_cirlce_ > progresscircle)

components/progresscircle/index.css
```diff
@@ -13,173 +13,123 @@ governing permissions and limitations under the License.
...
-                      var(--mod-progress-circle-fill-border-color-over-background,
-                          var(--spectrum-progress-circle-fill-border-color-over-background)));
-  }
+/* animation masks - circle indeterminate with continuous animation */
+.spectrum-ProgressCircle--indeterminate {
+	.spectrum-ProgressCircle-fill {
+		will-change: transform;
+		animation:
+			1s cubic-bezier(0.25, 0.85, 0.48, 0.89) 0.2s infinite spectrum-fills-rotate,
+			1s cubic-bezier(0.25, 0.1, 0.25, 1.3) infinite 0s spectrum-dashoffset-animation;
```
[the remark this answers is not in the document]
aramos-adobe: Mmmmm!! Good call out on the bezier curve. I removed it and it moves the way React Spectrum's does now.

components/progresscircle/animation.css
```diff
@@ -10,500 +10,512 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-.spectrum-ProgressCircle--indeterminate-fill-submask-2 {
-  animation: 1s infinite linear spectrum-fill-mask-2;
+.spectrum-ProgressCircle--indeterminate {
+	animation: 1s infinite linear spectrum-fill-mask-1;
```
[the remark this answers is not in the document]
aramos-adobe: It does absolutely nothing so I'm removing them both

components/progressbar/stories/progressbar.stories.js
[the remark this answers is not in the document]
marissahuysentruyt: There's actually a bug in the `spectrum-two` branch: https://github.com/adobe/spectrum-css/blob/dd96363745d39e200cfea84092903fcd53c340b3/components/progressbar/stories/template.js#L15

The progress bar's arg `isStaticWhite` should actually be `staticColor`. Regardless, I think these changes with the indeterminate group and template should be removed.

components/progresscircle/stories/progresscircle.stories.js
```diff
@@ -1,6 +1,6 @@
 import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
-import { isIndeterminate, size, staticColor } from "@spectrum-css/preview/types";
+import { isIndeterminate, size } from "@spectrum-css/preview/types";
```
marissahuysentruyt, line 3: Let's get the `staticColor` import back! Yay!

`import { isIndeterminate, size, staticColor } from "@spectrum-css/preview/types";`

That means we can basically revert these `staticColor` control changes as well. In the controls, we should just be able to do `staticColor`, since we need both the black and white options now too. 👍 

Screenshot 2025-01-08 at 4 14 42 PM

components/progresscircle/stories/progresscircle.stories.js
```diff
@@ -102,3 +118,25 @@ StaticWhiteIndeterminate.args = {
...
+export const StaticBlackIndeterminate = Template.bind({});
+StaticBlackIndeterminate.tags = ["!dev"];
+StaticBlackIndeterminate.storyName = "Static black, indeterminate";
+StaticBlackIndeterminate.args = {
+	staticColor: "black",
+	isIndeterminate: true,
+};
+StaticBlackIndeterminate.parameters = {
+	chromatic: { disableSnapshot: true },
+};
```
marissahuysentruyt, line 142: Now that we do have the `Sizing()` function, what do you tihnk about using it for both of these static black stories? Instead of binding to `Template()`, bind to `Sizing()` like the static white stories? 

```suggestion
export const StaticBlackDeterminate = Sizing.bind({});
StaticBlackDeterminate.tags = ["!dev"];
StaticBlackDeterminate.storyName = "Static black, default";
StaticBlackDeterminate.args = {
	staticColor: "black",
	isIndeterminate: false,
};
StaticBlackDeterminate.parameters = {
	chromatic: { disableSnapshot: true },
};

export const StaticBlackIndeterminate = Sizing.bind({});
StaticBlackIndeterminate.tags = ["!dev"];
StaticBlackIndeterminate.storyName = "Static black, indeterminate";
StaticBlackIndeterminate.args = {
	staticColor: "black",
	isIndeterminate: true,
};
StaticBlackIndeterminate.parameters = {
	chromatic: { disableSnapshot: true },
};
```

There's nothing wrong with what you have, it's just the odd ones out of the bunch on the docs page.

Screenshot 2025-01-08 at 4 17 36 PM

components/tooltip/stories/tooltip.stories.js
marissahuysentruyt, line 1: Similar to my progress bar comment, I think we should revert these changes completely. As far as I can tell, they shouldn't be related to your progress circle work, and there's a separate bug with tooltip (not having the `variant = "neutral",` arg in it's template).

components/progresscircle/stories/template.js
```diff
@@ -15,45 +17,55 @@ export const Template = ({
...
 };
+
+export const ProgressCircleGroup = (args) => html`
+	${window.isChromatic() ? html`
+		${Template(args)}
+		${Template({
+			...args,
+			isIndeterminate: true,
+		})}
+	` : Template(args)}
```
marissahuysentruyt, line 70: We have the test files after the `spectrum-two` update, so we don't even need this anymore! 🥳🥳🥳🥳

marissahuysentruyt (changes requested):
Your rebase cleaned this up so nice! It's looking good, but I do have some questions for you: 

Just wanted to be sure- the `corner-radius-full` token for the fill isn't used correct? Because we're using the `stroke-linecap` attribute? I saw it in the Figma file, but didn't see it used in a quick search of the CSS file. But I'm pretty sure that's why.

There's the changes in `progress-bar.stories.js` and `tooltip.stories.js`. I vote we revert those and I'll get a bug branch going to fix those pages. 

The test files are here! Would you add the static black test case? 
```
// progresscircle.test.js
...
{
			testHeading: "Static white",
			staticColor: "white",
		},
{
			testHeading: "Static black",
			staticColor: "black",
		},
...
```

components/progresscircle/index.css
```diff
@@ -1,31 +1,16 @@
...
+*/
 
-		/* fill over background is bright color */
-		--highcontrast-progress-circle-fill-border-color-over-background: Highlight;
-	}
-
-	/* track double for visibility */
-	.spectrum-ProgressCircle-track {
-		--spectrum-progress-circle-track-border-style: double;
-	}
```
marissahuysentruyt, line 27: I think we need to revisit the high contrast mode. The only reason I started looking into this was because I saw no track for the staticBlack stories. So I pulled it up in AssistivLabs and noticed that there's just some discrepancies where the the staticBlack stories still has its transparent-black track (so you can't really visually see it), the default/sizing/indeterminate stories have a brighter white track (which I think is the gray-200), and then staticWhite has a different track color (the transparent-white token). You can see all 3 different track colors in this screenshot:

Screenshot 2025-01-09 at 9 47 59 AM

I don't really have an answer for this besides asking in the channel.

marissahuysentruyt: I was thinking maybe we could emulate what progress bar is doing with a `--highcontrast-progresscircle-track-color: ButtonFace;`, but I think that would require additional SVG work in order to get a "border" on the track? That `ButtonFace` color is coming through, but it's just really low contrast.

Screenshot 2025-01-09 at 10 12 14 AM
Screenshot 2025-01-09 at 10 11 55 AM

components/progresscircle/index.css
```diff
@@ -1,31 +1,16 @@
...
-}
+@import "animation.css";
 
 .spectrum-ProgressCircle {
 	/* circle unfilled border color */
```
marissahuysentruyt, line 16: Should `--spectrum-progress-circle-track-border-color` be aliased to `--spectrum-track-color` instead of the global, gray-200? When this goes to design review, are they going to be looking for `track-color` since that's in Figma? It also looks like `track-color` is actually gray-300 anyways.

Screenshot 2025-01-09 at 10 19 39 AM

components/progresscircle/index.css
marissahuysentruyt: I looked back at the static white progress circle, in WHCM, and this is what it looks like on `spectrum-two`. I feel like that's probably what we want to try to recreate, particularly with the borders. What do you think?

Screenshot 2025-01-09 at 10 24 33 AM

marissahuysentruyt (changes requested):
I went through the CSS this morning and things are looking really good! I found we were missing the `track-color` token, but overall, I really love all the animation code we were able to remove! Nice work 😄 

I did have a question about the animation (and this is a real nit-pick, so feel free to disregard if you are happy with the fill animation!). Is it "stopping" at the top of the progress circle just before the animation loops? I can't tell if that's my eyes playing tricks on me or if there's really a pause. 

https://github.com/user-attachments/assets/15812014-8b4e-4c9c-b879-4ad97f7d2081

Other than that, my final questions are really regarding high contrast mode.

components/progresscircle/index.css
aramos-adobe: @marissahuysentruyt I've added the borders for WHCM. For indeterminate mode, the borders are not there. Speaking with design, as of right now the track is not needed for indeterminate just to show the loading state. But for progress the track is necessary.

components/progresscircle/index.css
```diff
@@ -68,591 +68,56 @@
...
-.spectrum-ProgressCircle-fillSubMask1,
-.spectrum-ProgressCircle-fillSubMask2 {
-	inline-size: 100%;
-	block-size: 100%;
-	transform-origin: 100% center;
-	overflow: hidden;
-	transform: rotate(-180deg);
+.spectrum-ProgressCircle-track {
+	stroke-width: var(--mod-progress-circle-thickness, var(--spectrum-progress-circle-thickness));
+	stroke: var(--highcontrast-progressbar-track-color, var(--mod-progress-circle-track-border-color, var(--spectrum-progress-circle-track-border-color)));
```
marissahuysentruyt, line 92: I think we need to delete this progress bar custom property in favor of a `--highcontrast-progress-circle-track-border-color` instead.

marissahuysentruyt (approved):
Let's get that second approval and get this thing merrggedd! It looks awesome. Thanks for wading through all of my comments and ramblings!

michael-scott-disco-coffee

PR #3388, opened 2024-11-11
feat(colorarea): S2 migration

cdransf:
## Description

`CSS-1032`

This change migrates the `colorarea` component to S2. It updates the `--spectrum-colorarea-border-radius` token to `--spectrum-colorarea-border-rounding`. Additionally, it leverages updated tokens and the `rgba(...)` color function, replacing the existing `TODO` and hardcoded values.

### Validation steps

1. Run Storybook locally or open the link for the PR.
2. Navigate to the `colorarea` component.
3. Verify that no regressions are visible.

## To-do list

### DIFF
--- a/.changeset/spotty-onions-study.md
+++ b/.changeset/spotty-onions-study.md
@@ -4,4 +4,4 @@
 
 # colorarea S2 migration
 
-This change migrates the `colorarea` component to S2. It updates the `--spectrum-colorarea-border-radius` token to `--spectrum-colorarea-border-rounding`. Additionally, it leverages updated tokens and the `rgba(...)` color function, replacing the existing `TODO` and hardcoded values.
+This change migrates the `colorarea` component to S2. It leverages updated tokens and the `rgba(...)` color function, replacing the existing `TODO` and hardcoded values.
--- a/.changeset/spotty-onions-study.md
+++ b/.changeset/spotty-onions-study.md
@@ -1,0 +1,7 @@
+---
+"@spectrum-css/colorarea": major
+---
+
+# colorarea S2 migration
+
+This change migrates the `colorarea` component to S2. It updates the `--spectrum-colorarea-border-radius` token to `--spectrum-colorarea-border-rounding`. Additionally, it leverages updated tokens and the `rgba(...)` color function, replacing the existing `TODO` and hardcoded values.
--- a/components/colorarea/index.css
+++ b/components/colorarea/index.css
@@ -22,26 +22,6 @@ governing permissions and limitations under the License.
 	--spectrum-colorarea-min-height: var(--spectrum-color-area-minimum-height);
 }
 
-/* Windows High Contrast Mode */
-@media (forced-colors: active) {
-	.spectrum-ColorArea {
-		--highcontrast-colorarea-border-color-disabled: GrayText;
-		--highcontrast-colorarea-border-color: Canvas;
-		--highcontrast-colorarea-fill-color-disabled: Canvas;
-	}
-
-	.spectrum-ColorArea {
-		&.is-disabled {
-			forced-color-adjust: none;
-		}
-	}
-
-	.spectrum-ColorArea-gradient,
-	.spectrum-ColorHandle-color {
-		forced-color-adjust: none;
-	}
-}
-
 .spectrum-ColorArea {
 	position: relative;
 	display: inline-block;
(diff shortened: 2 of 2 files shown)


components/colorarea/index.css
```diff
@@ -9,87 +9,89 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-  --spectrum-colorarea-border-radius: var(--spectrum-color-area-border-rounding);
-  --spectrum-colorarea-border-color: rgba(0, 0, 0, 10%); /* TODO replace with token --spectrum-color-area-border-color and --spectrum-color-area-border-opacity using RGBA function */
-  --spectrum-colorarea-disabled-background-color: var(--spectrum-disabled-background-color);
-  --spectrum-colorarea-border-width: var(--spectrum-color-area-border-width);
-  --spectrum-colorarea-height: var(--spectrum-color-area-height);
-  --spectrum-colorarea-width: var(--spectrum-color-area-width);
-  --spectrum-colorarea-min-width: var(--spectrum-color-area-minimum-width);
-  --spectrum-colorarea-min-height: var(--spectrum-color-area-minimum-height);
+	--spectrum-colorarea-border-radius: var(--spectrum-color-area-border-rounding);
+	--spectrum-colorarea-border-color-rgb: var(--spectrum-gray-1000-rgb);
```
marissahuysentruyt, line 13: Should we keep the TODO comment? Technically, the designs are asking for `gray-1000`, not `gray-1000-rgb`. Once `gray-1000` gets updated without that rgba() color function, we could remove the use of that custom property `--spectrum-colorarea-border-color-rgb` to "properly" use the `gray-1000`.

marissahuysentruyt (commented):
This one seems pretty straightforward. I did come away with some questions after reviewing.

- The designs make a point to include the color loupe in the list of components for color area. Should we add some documentation regarding that? I'm not sure we need to add the color loupe to this template, but I was thinking at least a blurb about "implementations can use the color loupe to..." I don't think there's anything on the `main` about color loupe with this color area component, but it is mentioned in the Spectrum guidelines: https://spectrum.adobe.com/page/color-area/#Color-loupe-on-down/touch-state
- Something I just thought of- I wonder if we should move the `forced-colors` query. I would have to check with Rise and Josh (or dig through old PRs), but I think when we were doing these components before, we were trying to move the forced-colors block to the bottom of the CSS file to work with the cascade.

components/colorarea/index.css
```diff
@@ -9,87 +9,89 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+	--spectrum-colorarea-border-color: rgba(var(--spectrum-colorarea-border-color-rgb), var(--spectrum-color-area-border-opacity));
+	--spectrum-colorarea-border-rounding: var(--spectrum-color-area-border-rounding);
```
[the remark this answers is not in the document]
cdransf: We don't! Just pushed up a fix ✨

components/colorarea/index.css
cdransf: That makes sense to me — I dropped in the same component that I used in the `colorwheel` since it's the same pattern. ✨

cdransf:
> Looks like everything is connecting to the right tokens, there are some duplicated custom properties being used for the border-radii that we'd probably want to sort out though.
> 
> Comparing this branch with `main`, I'm noticing that the component sizes are different, which is fine, but zooming in on the smaller size in this branch I noticed that there's a gap in the corners between the border and the gradient: image
> 
> This happens in `main` too, but I'm wondering if that's a thing we can fix for S2. We'd probably want to set the `border-radius` in `.spectrum-ColorArea-gradient` to something like `calc(var(--spectrum-colorarea-border-radius) - var(--spectrum-colorarea-border-width))` and maybe leave a comment explaining it (I didn't include the mods here but you'd want to add them in too).

I'd noticed that too but I wasn't sure how best to approach it — thanks so much for clarifying! Just pushed up a fix for that.

Screenshot 2024-11-13 at 4 29 17 PM

components/colorarea/index.css
cdransf: > Something I just thought of- I wonder if we should move the forced-colors query. I would have to check with Rise and Josh (or dig through old PRs), but I think when we were doing these components before, we were trying to move the forced-colors block to the bottom of the CSS file to work with the cascade.

I went ahead and moved this down — it's a fairly consistent pattern and was the case when I did WHC work recently. Happy to revert if @jawinn or @rise-erpelding would like though. 🙂

rise-erpelding (approved):
Looks great! I think the changeset needs updating to match the changes that have been made here, but otherwise I think this is good to go!

marissahuysentruyt (approved):
Thanks for making all of the changes. Looks like everything was take care of from my perspective and things are looking great. I think the only thing we should do is update that changeset like Rise mentioned, but I'm going to approve. Nice work!

jawinn (approved):

.changeset/spotty-onions-study.md
```diff
@@ -0,0 +1,7 @@
+---
+"@spectrum-css/colorarea": major
+---
+
+# colorarea S2 migration
+
+This change migrates the `colorarea` component to S2. It updates the `--spectrum-colorarea-border-radius` token to `--spectrum-colorarea-border-rounding`. Additionally, it leverages updated tokens and the `rgba(...)` color function, replacing the existing `TODO` and hardcoded values.
```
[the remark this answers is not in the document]
cdransf: Just updated this — thanks so much for catching it! ✨

PR #2818, opened 2024-06-05
feat(statuslight): s2 migration

marissahuysentruyt:
## Description

This migrates the `Status light` to S2. 🎉 New color tokens and text-to-visual tokens were added to match the design specs. The neutral variant label uses `default-font-family` instead of italic, and the semantic accent variant has been removed, both according to design specs.

This PR also increases Chromatic coverage for Status light variants.

### Designs
[S2 Tokens specs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?m=auto&node-id=10725-1075&t=ZkVK6F8ShI1pizHB-1)
[S2 / Desktop](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?m=auto&node-id=36797%3A2020&t=Lbp6Z9NHARTCBSr1-1)

- New mods:
    - `--mod-statuslight-font-family`
    - `--mod-statuslight-font-style`
    - `--mod-statuslight-nonsemantic-brown-color`
    - `--mod-statuslight-nonsemantic-cinnamon-color`
    - `--mod-statuslight-nonsemantic-pink-color`
    - `--mod-statuslight-nonsemantic-silver-color`
    - `--mod-statuslight-nonsemantic-turquoise-color`
       
- Removed mods:
    - `--mod-statuslight-semantic-accent-color`

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [x] Design review and approval (@marissahuysentruyt and Amun)
- [x] Status light matches [S2 designs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?m=auto&node-id=10725-1075&t=ZkVK6F8ShI1pizHB-1)
- [x] Removed elements include: 
    - no `accent` variant 
    - the `neutral` variant's label no longer uses italics 
- [x] New tokens are used:
    -  `--spectrum-status-light-text-to-visual-75`
    -  `--spectrum-status-light-text-to-visual-100`
    - `--spectrum-status-light-text-to-visual-200`
    - `--spectrum-status-light-text-to-visual-300`
    - `--spectrum-pink-visual-color`
    - `--spectrum-turquoise-visual-color`
    - `--spectrum-cinnamon-visual-color`
    - `--spectrum-brown-visual-color`
    - `--spectrum-silver-visual-color`
- [x] Updated tokens are used:
[...]

PR #3424, opened 2024-12-02
feat(colorslider): S2 migration

cdransf:
## Description

`CSS-1028`

# S2 colorslider migration

This migrates the `colorslider` component to S2. Custom properties have been remapped per the design spec.

| Before                    | After                      |
| ------------------------- | -------------------------- |
| `--spectrum-gray-900-rgb` | `--spectrum-gray-1000-rgb` |

### Validation steps

1. Open the URL for the PR.
2. Navigate to the `colorslider` component and verify no regressions have occurred.

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/colorslider/index.css
+++ b/components/colorslider/index.css
@@ -21,7 +21,7 @@
 	/* @passthrough -- settings for nested color handle */
 	--mod-colorhandle-hitarea-border-radius: var(--mod-color-slider-handle-hitarea-border-radius, 0px);
 
-	--spectrum-colorslider-fill-color-disabled: var(--spectrum-disabled-background-color);
+	--spectrum-color-slider-fill-color-disabled: var(--spectrum-disabled-background-color);
 
 	position: relative;
 	display: block;
@@ -76,8 +76,8 @@
 	}
 
 	.spectrum-ColorSlider.is-disabled & {
-		--spectrum-color-slider-border-color-local: var(--highcontrast-color-slider-border-color-disabled, var(--mod-color-slider-border-color-disabled, var(--spectrum-colorslider-fill-color-disabled)));
-		background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-colorslider-fill-color-disabled)));
+		--spectrum-color-slider-border-color-local: var(--highcontrast-color-slider-border-color-disabled, var(--mod-color-slider-border-color-disabled, var(--spectrum-color-slider-fill-color-disabled)));
+		background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-color-slider-fill-color-disabled)));
 	}
 }
 
--- a/components/colorslider/index.css
+++ b/components/colorslider/index.css
@@ -11,28 +11,24 @@
  * governing permissions and limitations under the License.
  */
 
-@media (forced-colors: active) {
-	.spectrum-ColorSlider {
-		--highcontrast-color-slider-border-color: CanvasText;
-		--highcontrast-color-slider-border-color-disabled: GrayText;
-		--highcontrast-color-slider-background-color-disabled: Canvas;
-		forced-color-adjust: none;
-	}
-}
-
 .spectrum-ColorSlider {
+	/* Size and Spacing */
+	--spectrum-color-slider-control-track-width: var(--spectrum-color-control-track-width);
+
 	/* @todo Refactor with --spectrum-color-slider-border-color once gray rgb token is no longer necessary to workaround nested rgb color token value using rgba(). */
-	--spectrum-color-slider-border-color-default: rgba(var(--spectrum-gray-900-rgb), var(--spectrum-color-slider-border-opacity));
+	--spectrum-color-slider-border-color-default: rgba(var(--spectrum-gray-1000-rgb), var(--spectrum-color-slider-border-opacity));
 
 	/* @passthrough -- settings for nested color handle */
 	--mod-colorhandle-hitarea-border-radius: var(--mod-color-slider-handle-hitarea-border-radius, 0px);
+
+	--spectrum-colorslider-fill-color-disabled: var(--spectrum-disabled-background-color);
 
 	position: relative;
 	display: block;
 
 	min-inline-size: var(--mod-color-slider-minimum-length, var(--spectrum-color-slider-minimum-length));
 	inline-size: var(--mod-color-slider-length, var(--spectrum-color-slider-length));
-	block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-control-track-width));
+	block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-slider-control-track-width));
 
 	/* Otherwise we randomly drag a file icon */
 	user-select: none;
@@ -59,11 +55,6 @@
 	block-size: var(--mod-color-slider-vertical-height, var(--mod-color-slider-length, var(--spectrum-color-slider-length)));
 	min-inline-size: 0;
 	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-control-track-width)));
-
-	.spectrum-ColorSlider-handle {
-		inset-inline-start: 50%;
-		inset-block-start: 0;
-	}
 }
 
 .spectrum-ColorSlider-handle {
@@ -86,7 +77,7 @@
 
 	.spectrum-ColorSlider.is-disabled & {
 		--spectrum-color-slider-border-color-local: var(--highcontrast-color-slider-border-color-disabled, var(--mod-color-slider-border-color-disabled, var(--spectrum-disabled-background-color)));
-		background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-disabled-background-color)));
+		background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-colorslider-fill-color-disabled)));
 	}
 }
 
@@ -113,4 +104,13 @@
 	z-index: 0;
 	margin: 0;
 	pointer-events: none;
+}
+
+@media (forced-colors: active) {
+	.spectrum-ColorSlider {
+		--highcontrast-color-slider-border-color: CanvasText;
+		--highcontrast-color-slider-border-color-disabled: GrayText;
+		--highcontrast-color-slider-background-color-disabled: Canvas;
+		forced-color-adjust: none;
+	}
 }
--- a/components/colorslider/index.css
+++ b/components/colorslider/index.css
@@ -57,6 +57,7 @@
 	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-slider-control-track-width)));
 
 	.spectrum-ColorSlider-handle {
+		inset-inline-start: 50%;
 		inset-block-start: 0;
 	}
 }
--- a/components/colorslider/index.css
+++ b/components/colorslider/index.css
@@ -55,6 +55,10 @@
 	block-size: var(--mod-color-slider-vertical-height, var(--mod-color-slider-length, var(--spectrum-color-slider-length)));
 	min-inline-size: 0;
 	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-slider-control-track-width)));
+
+	.spectrum-ColorSlider-handle {
+		inset-block-start: 0;
+	}
 }
 
 .spectrum-ColorSlider-handle {
--- a/components/colorslider/index.css
+++ b/components/colorslider/index.css
@@ -54,7 +54,7 @@
 	min-block-size: var(--mod-color-slider-vertical-minimum-height, var(--mod-color-slider-minimum-length, var(--spectrum-color-slider-minimum-length)));
 	block-size: var(--mod-color-slider-vertical-height, var(--mod-color-slider-length, var(--spectrum-color-slider-length)));
 	min-inline-size: 0;
-	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-control-track-width)));
+	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-slider-control-track-width)));
 }
 
 .spectrum-ColorSlider-handle {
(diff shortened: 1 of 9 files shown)


cdransf:
Per design we can ignore the spacing tokens as they're not required for the implementation (alignment of the handle within the color slide is handled by this rule:

```css
.spectrum-ColorSlider-handle {
  inset-inline-start: 0;
  inset-block-start: 50%;
}
```

rise-erpelding (changes requested):
Migration-wise, this looks fine, since not a lot needed to change here! I have a few other recommendations though:

**CSS**
I have a few recommendations for the CSS that would make this file more consistent with some of the other migration work that's been done:
- Generally, we've been moving forced-colors media query blocks to the bottom in recent migrations (see #3388, #2743, or #2818 for instance), and I see that Color Slider's is still further up in the CSS file, that could probably be moved down.
- We have this line:
    ```css
    block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-control-track-width));
    ```
    I'm really not sure why that `color-control-track-width` token has that name, but it seems like a good idea to create an alias custom property for this, especially given that its token name doesn't match its mod name, something like this, we can even take the word "control" out and change the mod name completely, up to you:
    ```css
     /* Up at the topmost .spectrum-ColorSlider block */
     --spectrum-color-slider-control-track-width: var(--spectrum-color-control-track-width);

    /* further down in the code, same place it is now */
    block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-slider-control-track-width));
    ```
- In that same vein, we could also alias the global token for the disabled background color that we have here, similar to how it's done in [Color Wheel](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/colorwheel/index.css#L18), [Picker](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/picker/index.css#L100), or [Search](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/search/index.css#L48), for instance:
    ```css
    --spectrum-color-slider-border-color-local: var(--highcontrast-color-slider-border-color-disabled, var(--mod-color-slider-border-color-disabled, var(--spectrum-disabled-background-color)));
    background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-disabled-background-color)));
    ```
- There's also an unused token for the spacing between the field label and the color slider, but this isn't new. I don't know if there's some specific reason we're not adding the label to this implementation (Slider has a label that we display), but it feels like something we ought to be doing. Feel free to get more opinions on this one.
image

**Storybook**
Otherwise, this component could probably benefit from a few Storybook enhancements similar to what we added for several of our other components back when we were doing docs to storybook migrations (this component got migrated early on, before we decided that we would be doing that kind of thing):
- Being able to change the colors in the slider from the Storybook controls or add an image (Swatch on main added this with the docs to storybook migration https://github.com/adobe/spectrum-css/pull/2925, so we might be able to copy that over)
- A control to be able to view the vertical variant from the default story in Storybook. In this branch, all the variants are showing, but on `main`, on the default variant is visible and the others are used just for display purposes on the Docs page, that's probably where we'd like to end up one day when `spectrum-two` has some of those docs changes from main.

Last thing! I can't use the keyboard to focus on the slider handle (I can in `main`, but not on this branch or `spectrum-two`), this may not be a big deal, but I'm curious about why it's happening.

cdransf:
> Migration-wise, this looks fine, since not a lot needed to change here! I have a few other recommendations though:
> 
> **CSS** I have a few recommendations for the CSS that would make this file more consistent with some of the other migration work that's been done:
> 
> * Generally, we've been moving forced-colors media query blocks to the bottom in recent migrations (see [feat(colorarea): S2 migration #3388](https://github.com/adobe/spectrum-css/pull/3388), [feat(tooltip): S2 migration #2743](https://github.com/adobe/spectrum-css/pull/2743), or [feat(statuslight): s2 migration #2818](https://github.com/adobe/spectrum-css/pull/2818) for instance), and I see that Color Slider's is still further up in the CSS file, that could probably be moved down.

Alrighty! `forced-colors` has been moved to the bottom of the file.

> * We have this line:
>   ```css
>   block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-control-track-width));
>   ```
>     
>   I'm really not sure why that `color-control-track-width` token has that name, but it seems like a good idea to create an alias custom property for this, especially given that its token name doesn't match its mod name, something like this, we can even take the word "control" out and change the mod name completely, up to you:
>   ```css
>    /* Up at the topmost .spectrum-ColorSlider block */
>    --spectrum-color-slider-control-track-width: var(--spectrum-color-control-track-width);
>   
>   /* further down in the code, same place it is now */
>   block-size: var(--mod-color-slider-control-track-width, var(--spectrum-color-slider-control-track-width));
>   ```

`--spectrum-color-control-track-width` has been aliased to `--spectrum-color-slider-control-track-width`.

> * In that same vein, we could also alias the global token for the disabled background color that we have here, similar to how it's done in [Color Wheel](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/colorwheel/index.css#L18), [Picker](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/picker/index.css#L100), or [Search](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/search/index.css#L48), for instance:
>   ```css
>   --spectrum-color-slider-border-color-local: var(--highcontrast-color-slider-border-color-disabled, var(--mod-color-slider-border-color-disabled, var(--spectrum-disabled-background-color)));
>   background: var(--highcontrast-color-slider-background-color-disabled, var(--mod-color-slider-background-color-disabled, var(--spectrum-disabled-background-color)));
>   ```

`--spectrum-disabled-background-color` has been aliased to `--spectrum-colorslider-fill-color-disabled`.

> * There's also an unused token for the spacing between the field label and the color slider, but this isn't new. I don't know if there's some specific reason we're not adding the label to this implementation (Slider has a label that we display), but it feels like something we ought to be doing. Feel free to get more opinions on this one.

@jawinn and I discussed the label visible in this component and (apparently) SWC doesn't have the label as part of the component and the consensus was that we didn't need to include it at this point.

> **Storybook** Otherwise, this component could probably benefit from a few Storybook enhancements similar to what we added for several of our other components back when we were doing docs to storybook migrations (this component got migrated early on, before we decided that we would be doing that kind of thing):
> 
> * Being able to change the colors in the slider from the Storybook controls or add an image (Swatch on main added this with the docs to storybook migration [docs(swatch,swatchgroup,table,tabs): docs migrations to storybook #2925](https://github.com/adobe/spectrum-css/pull/2925), so we might be able to copy that over)
> * A control to be able to view the vertical variant from the default story in Storybook. In this branch, all the variants are showing, but on `main`, on the default variant is visible and the others are used just for display purposes on the Docs page, that's probably where we'd like to end up one day when `spectrum-two` has some of those docs changes from main.

I added a `colors` control to the story for the component that defaults to the previously hardcoded linear gradient and a `vertical` boolean control that defaults to false and allows the orientation to be changed.

> Last thing! I can't use the keyboard to focus on the slider handle (I can in `main`, but not on this branch or `spectrum-two`), this may not be a big deal, but I'm curious about why it's happening.

I can reproduce this issue too, but haven't been able to sort out why it's happening yet. 🤔

cdransf:
I am able to focus on the handle in this branch (testing in Chrome) — I can't in Safari, but I often get odd keyboard focus behavior in Safari.

image

rise-erpelding (commented):
Looking great! I love the improvements to Storybook, and for whatever reason the keyboard focus issue has resolved itself! 🎉 

One minor nitpick about the `--spectrum-colorslider-fill-color-disabled` custom prop.

components/colorslider/metadata/metadata.json
```diff
@@ -37,13 +36,15 @@
     "--spectrum-color-slider-border-opacity",
     "--spectrum-color-slider-border-rounding",
     "--spectrum-color-slider-border-width",
+    "--spectrum-color-slider-control-track-width",
     "--spectrum-color-slider-length",
-    "--spectrum-color-slider-minimum-length"
+    "--spectrum-color-slider-minimum-length",
+    "--spectrum-colorslider-fill-color-disabled"
```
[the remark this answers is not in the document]
cdransf: Makes sense to me! Just updated it. ✨

jawinn (changes requested):

components/colorslider/stories/colorslider.stories.js
jawinn, line 1: ### Alpha 
It looks like the Alpha story is still using the `gradientStops` arg that renamed, so it's not displaying correctly:
Screenshot 2025-01-17 at 3 53 30 PM
Prod:
Screenshot 2025-01-17 at 3 58 38 PM

### Vertical 
The Vertical story also still needs an update for its story in our docs, and the VRT test. The gradient should be in the `to bottom` direction for these instances where we set the arg value, though we talked about not needing any complicated logic if someone is manually flipping the vertical control on the Default story:

Screenshot 2025-01-17 at 3 53 16 PM
Screenshot 2025-01-17 at 3 54 08 PM

jawinn (changes requested):
I'm still seeing a few issues with the Alpha and Vertical stories being displayed on the Docs.

This'll also need some rebasing after the large rebase in `spectrum-two` that recently happened, which is causing the large number of changed files to be shown. It may be easier to cherry-pick your commits than to handle the conflicts (at least that's what I found in my other branch).

components/colorslider/stories/colorslider.stories.js
cdransf: Got it! I renamed and updated that arg and updated the gradient direction in the tests you highlighted. ✨

rise-erpelding (changes requested):
A couple of other details on the docs page that look like they still need some attention:

* This Alpha story doesn't look the same as in Foundations or main, it looks like the gradient in the args is using the same rainbow gradient rather than the previous one that included rgba values.
    image
* The Vertical variant is improved! But it looks like it needs its handle adjusted, the color doesn't match the color next to it in the gradient and it's off-center. Previous versions showed the handle at the top which probably isn't strictly necessary, but it seems like maybe that's where the red color is coming from. Feels like a bit of a silly ask since it's not a working implementation, but would help it look the part anyway!
    image

cdransf:
> A couple of other details on the docs page that look like they still need some attention:
> 
> * This Alpha story doesn't look the same as in Foundations or main, it looks like the gradient in the args is using the same rainbow gradient rather than the previous one that included rgba values.

Totally missed that! Thank you! That should be all fixed (updated with the value from the `redux` branch). ✨

> * The Vertical variant is improved! But it looks like it needs its handle adjusted, the color doesn't match the color next to it in the gradient and it's off-center. Previous versions showed the handle at the top which probably isn't strictly necessary, but it seems like maybe that's where the red color is coming from. Feels like a bit of a silly ask since it's not a working implementation, but would help it look the part anyway!

I adjusted the `inset-block-start` for the `.spectrum-ColorSlider-handle` when the slider is oriented vertically. The default class hasn't been updated in ~2-5 years.

```
.spectrum-ColorSlider-handle {
  inset-inline-start: 0;
  inset-block-start: 50%;
}
```

Which is why we're seeing the handle 50% of the way through the slider. I added:

```
.spectrum-ColorSlider--vertical {
  ...
  .spectrum-ColorSlider-handle {
    inset-block-start: 0;
  }
}
```

Which I'm not sure is the right path — would love some feedback on that. 😄 

image

rise-erpelding:
Hey! The docs page Alpha story looks good, I think it's still going to need an update for the testing preview:
image

Re: the vertical story, it looks like maybe that `--vertical` variant is just missing the `inset-inline-start: 50%` like it had in [Foundations](https://github.com/adobe/spectrum-css/blob/s2-foundations-redux/components/colorslider/index.css#L64)?

components/colorslider/stories/colorslider.stories.js
```diff
@@ -40,12 +42,19 @@ export default {
...
+			options: ["color", "image"],
 			control: { type: "select" },
 			table: { disable: true },
 		},
+		vertical: {
+			name: "Vertical",
+			description: "The orientation of the component.",
+			type: { name: "boolean" },
+			control: "boolean",
+		},
```
rise-erpelding, line 57: Ooof, I'm still seeing that funny gradient when the vertical control is used:
image

I don't have ideas off the top of my head on how we handle this, since it would need to be one gradient for horizontal and another for vertical... although I'm sure we've run into a similar issue before... any thoughts?

rise-erpelding (changes requested):
Looks like the Alpha gradient in the tests and the handle for the vertical variant still need to be addressed, I also left another comment about some other weirdness with the vertical gradient.

components/colorslider/stories/colorslider.stories.js
cdransf: I talked to @jawinn about this and the thinking at the time was that, since we've moved the color control to a string that the onus would then be on the user to change that direction. We could couple the two, I can investigate that a bit more and see what we'd need to do it. 😃

components/colorslider/index.css
```diff
@@ -55,6 +55,10 @@
 	block-size: var(--mod-color-slider-vertical-height, var(--mod-color-slider-length, var(--spectrum-color-slider-length)));
 	min-inline-size: 0;
 	inline-size: var(--mod-color-slider-vertical-control-track-width, var(--mod-color-slider-control-track-height, var(--spectrum-color-slider-control-track-width)));
+
+	.spectrum-ColorSlider-handle {
+		inset-block-start: 0;
+	}
```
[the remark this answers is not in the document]
cdransf: Just restored this! ✨

components/colorslider/stories/colorslider.test.js
```diff
@@ -14,14 +14,15 @@ export const ColorSliderGroup = Variants({
...
+			color: "linear-gradient(to bottom, rgb(255, 0, 0) 0%, rgb(255, 255, 0) 17%, rgb(0, 255, 0) 33%, rgb(0, 255, 255) 50%, rgb(0, 0, 255) 67%, rgb(255, 0, 255) 83%, rgb(255, 0, 0) 100%)",
 			wrapperStyles: {
 				// Adjust for the indicator
 				"padding-block": "20px",
 			},
 		},
 		{
 			testHeading: "Alpha",
-			gradientStops: ["rgba(0, 0, 0, 1) 0%", "rgba(0, 0, 0, 0) 100%"],
+			color: "linear-gradient(to right, rgb(255, 0, 0) 0%, rgb(255, 255, 0) 17%, rgb(0, 255, 0) 33%, rgb(0, 255, 255) 50%, rgb(0, 0, 255) 67%, rgb(255, 0, 255) 83%, rgb(255, 0, 0) 100%)",
```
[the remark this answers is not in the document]
cdransf: Updated this as well. ✨

components/colorslider/stories/colorslider.stories.js
rise-erpelding: I think that's fine. We could add a little blurb in the Color control description about the gradient direction, maybe saying that the direction would need to be changed for the vertical variant. Not sure I feel strongly about that though, so I'll leave that up to you!

rise-erpelding (approved):
Looks great, I think everything I had has been addressed now! Thanks Cory!

jawinn (approved):

PR #3440, opened 2024-12-10
chore(storybook): fix code loading on docs pages [CSS-1070]

castastrophe:
## Description

The "Show code" feature is causing Storybook to freeze, which means that a user cannot view the reference markup on the documentation (without using the inspector).

Upon investigation, I determined this was caused by the testing grid injecting so much markup onto docs pages that it was overloading the code previewer. The testing code now returns only the template on docs pages when the Variant() function is used.

### Validation steps

Open the [storybook docs page](https://pr-3440--spectrum-css.netlify.app/?path=/docs/components-action-bar--docs) for any component:
- [x] Click the "show code" button at the bottom of one of the examples; expect the code to display in a reasonable amount of time
- [x] Repeat for at least 2 other snippets on the page
- [x] Validate on Safari [@castastrophe]
- [x] Validate on Chrome
- [x] Validate on Firefox

### Regression testing

Validate:

## To-do list


.storybook/package.json
```diff
@@ -40,7 +40,6 @@
 		"@etchteam/storybook-addon-status": "^5.0.0",
 		"@storybook/addon-a11y": "^8.4.6",
 		"@storybook/addon-actions": "^8.4.6",
-		"@storybook/addon-console": "^3.0.0",
```
castastrophe, line 43: This plugin adds little value and is pretty slow as well so I included here to ensure reasonable load times for pages.

.storybook/package.json
```diff
@@ -51,7 +50,6 @@
 		"@storybook/core-events": "^8.4.6",
 		"@storybook/manager-api": "^8.4.6",
 		"@storybook/preview-api": "^8.4.6",
-		"@storybook/testing-library": "^0.2.2",
```
castastrophe, line 54: Also unused so removed this to prevent any conflicting code loaded.

.storybook/preview.js
```diff
@@ -94,11 +81,6 @@ export const parameters = {
 	},
 	status: {
 		statuses: {
-			migrated: {
```
castastrophe, line 97: Unused code

cdransf (approved):
I see a brief delay before the code loads after hitting `Show code` but it no longer freezes or locks up the application.

rubencarvalho:
These changes reduced the amount of code in the "Show code" snippets. However, for components with large amounts of markup, they still render very slowly and can cause the page to become unresponsive.
If I try to:
1. Visit [https://pr-3440–spectrum-css.netlify.app/?path=/docs/components-table–docs](https://pr-3440--spectrum-css.netlify.app/?path=/docs/components-table--docs).
2. Click `Show code` under the first “Default variant” and wait for the snippet to render.
3. Then click `Show code` under the “Collapsible” variant. At this point, the page becomes unresponsive.

Chrome 131.0.6778.86; Mac (M3 Max)
Screenshot 2024-12-10 at 16 41 35

jawinn (commented):
As Ruben just noted, it's still giving "Page Unresponsive" for many components (Chrome, Mac). Some ones I tested of note:
- Accordion: will show the unresponsive message, but the code actually loads a few seconds later after that initial delay
- Checkbox: unresponsive message but the code never loads and page fully locks up after waiting
- Card and Button: code does load but there is still a bit of a delay

.storybook/project.json
```diff
@@ -18,7 +18,7 @@
 		"build": {
 			"configurations": {
 				"ci": {
-					"commands": ["cross-env NODE_OPTIONS=\"--no-warnings\" storybook build --config-dir ."]
+					"commands": ["cross-env NODE_OPTIONS=\"--no-warnings\" storybook build --test --config-dir ."]
```
jawinn, line 21: Should this `--test` change be here? Was that for local testing?

.storybook/preview-head.html
```diff
@@ -2,27 +2,43 @@
...
+						document.dispatchEvent(new CustomEvent("typekit-loaded", { detail: { kitId: this.kitId } }));
+					},
+				});
+		} catch (e) {
+			console.log(e);
+		}
+	};
+
+	const s = document.getElementsByTagName("script")[0];
+	s.parentNode.insertBefore(tk, s);
```
jawinn, line 43: Could you give more descriptive names to some of the variables here? 
`d`, `tk`, `rs`, `s`

.storybook/main.js
```diff
@@ -72,10 +80,20 @@ export default {
...
+					"@storybook/blocks",
+					"@storybook/theming",
+					"@storybook/components",
+				],
+			},
 			build: {
 				sourcemap: configType === "DEVELOPMENT",
 				manifest: true,
 				minify: configType === "PRODUCTION",
+				cssCodeSplit: false,
```
jawinn, line 96: What is this change and is there any way to validate?

aramos-adobe (approved):

.changeset/famous-comics-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/preview": patch
+---
+
+Fix an issue where "show code" was blocking loading in Storybook docs pages [CSS-1070]
```
jawinn, line 5: Is it worth noting what changes are being made here?
- Changes to the way Typekit is loaded (+ some more info about what these changes are)
- Icon now uses ref by default
- Removing some packages
- Excluding Variants decorator from rendering on the Docs pages
- anything else?

jawinn (changes requested):
This appears to be loading a lot faster now. I'm now able to load the Checkbox and Accordion code. Was it related to the icon loading and/or Typekit loading?

One significant question below about whether `--test` should be in there, otherwise I think it's good to go.

.storybook/project.json
castastrophe: It was recommended in the Storybook blog to optimize the performance of the build. It has to do with how files are broken up.

jawinn: https://storybook.js.org/docs/api/main-config/main-config-build#test
It looks like by default this disables a lot of things for the purpose of testing performance. Would we want all of those things disabled? I see it removes the blocks package used for docs blocks, mdx documentation, and autodocs?

.storybook/main.js
castastrophe: I can write up a test case, it's an optimization suggestion I thought would reduce the amount of fetches being done on load.
https://vite.dev/config/build-options#build-csscodesplit

.storybook/project.json
castastrophe: Right but it's just being added to the CI command which is used for VRT. It depends if we start using chromatic's build as a full example of the docs site. Currently, we used a netlify build for that which builds using a different command.

.storybook/main.js
castastrophe: When writing out the test case, I didn't see any differences in the Network tab for how many files were loaded per component so I removed this update until we find we need it.

jawinn (approved):

PR #3444, opened 2024-12-13
fix(slider): adjust label width

5t3ph:
## Description

[Per a report](https://adobedesign.slack.com/archives/C5N154FEY/p1732612802178059), allowing the slider label to take up full width leads to unexpected activation and precision issues on mobile.

Proposed fix was to remove `flex-grow: 1` on the label and switch to `justify-content: space-between` on the label container.

Visually in Storybook

### Validation steps

- [x] Open the storybook for the slider component and inspect whether any variant does not behave well with the change to how space is applied between the label and the current value output for the top label examples [@castastrophe]

### Regression testing

Validate:

**Note**: Confirmed that no VRT changes occurred from this change.

## To-do list


castastrophe (approved):
This looks great! We'll need a changeset added with details of the update and I'm wondering if there's a test case we're missing to prevent future regressions on this issue?

5t3ph:
> This looks great! We'll need a changeset added with details of the update and I'm wondering if there's a test case we're missing to prevent future regressions on this issue?

Since there was no visual change, I'm not sure there's a way to add a visual-only test for this. Did you have something in mind? Or am I missing a test method we have available? The slider also isn't functional on our side and the issue was preventing clicks basically outside the label boundary from activating the slider, so even a functional test wouldn't quite work on our Storybook I don't think?

PR #3347, opened 2024-10-29
feat(dialog): s2 takeover dialog migration

marissahuysentruyt:
## Description

This implements the design updates and new tokens for the fullscreen and fullscreen takeover dialogs in S2. 🎉  The majority of the migration has been done in the default, standard dialog (from #2860). The effected stories are called Takeover Dialog in Figma. New tokens were utilized to match the design specs. Fullscreen and fullscreenTakeover dialogs do not support additional footer content, checkboxes, or hero images. They are also not dismissible.

Documentation has been updated with info on the takeover dialog variants and the divider story was removed since dividers are no longer supported in dialog components in S2. 

Takeover dialogs in Figma support swapping out text for other components in the header content and body areas. The template for dialogs changed to reflect that, utilizing `renderContent`. Additionally, some class names were corrected to follow our typical naming convention (`.spectrum-Dialog-header-content > .spectrum-Dialog-headerContent` and
`.spectrum-Dialog-footer-content > .spectrum-Dialog-footerContent`).

New mod properties should reflect the "takeover dialog" language:

_New Mods_
`--mod-takeover-dialog-grid-spacing`
`--mod-takeover-dialog-spacing-header-content-gap`
`--mod-takeover-dialog-title-font-size`

_New modal custom property_
`--spectrum-modal-takeover-window-to-edge`

New aliased tokens were created for mobile value changes: 
For `.spectrum--large`:
--spectrum-takeover-dialog-spacing-grid-padding:var(--spectrum-spacing-400);
--spectrum-takeover-dialog-spacing-header-gap:var(--spectrum-spacing-300);
--spectrum-takeover-dialog-spacing-title-to-body:var(--spectrum-spacing-400);

For `.spectrum--medium`: 
--spectrum-takeover-dialog-spacing-grid-padding:var(--spectrum-spacing-500);
--spectrum-takeover-dialog-spacing-header-gap:var(--spectrum-spacing-400);
--spectrum-takeover-dialog-spacing-title-to-body:var(--spectrum-spacing-500);

### Designs
[S2 Takeover Dialog Token specs](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=27417-7447)
[S2 / Desktop Takeover dialog](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?node-id=61935-5399)

### Jira
[...]

PR #3476, opened 2025-01-03
fix(*): define missing custom properties from theme directory

cdransf:
## Description

`CSS-1084`

- [x] fixes missing custom properties by transferring them from `s2-foundations-redux`

image
image

## To-do list


components/actionbutton/index.css
```diff
@@ -11,65 +11,18 @@
...
 	--spectrum-actionbutton-background-color-hover: var(--highcontrast-actionbutton-background-color-hover, var(--mod-actionbutton-background-color-hover, var(--spectrum-gray-100)));
 	--spectrum-actionbutton-background-color-down: var(--highcontrast-actionbutton-background-color-down, var(--mod-actionbutton-background-color-down, var(--spectrum-gray-200)));
 	--spectrum-actionbutton-background-color-focus: var(--highcontrast-actionbutton-background-color-focus, var(--mod-actionbutton-background-color-focus, var(--spectrum-gray-100)));
-	--spectrum-actionbutton-background-color-selected-disabled: var(--spectrum-transparent-black-100);
-
-	--spectrum-actionbutton-background-color-selected: var(--spectrum-neutral-background-color-selected-default);
-	--spectrum-actionbutton-background-color-selected-hover: var(--spectrum-neutral-background-color-selected-hover);
-	--spectrum-actionbutton-background-color-selected-down: var(--spectrum-neutral-background-color-selected-down);
-	--spectrum-actionbutton-background-color-selected-focus: var(--spectrum-neutral-background-color-selected-key-focus);
```
marissahuysentruyt, line 64: Looks like these were all just "Unused custom properties," which I get why we would remove. But are these the values we should be using? https://github.com/adobe/spectrum-css/blob/9c3a6253d4dd7208bcb9c2cab544c204e9cc9fcf/components/actionbutton/themes/spectrum-two.css#L20

How does that work is the `spectrum-two.css` in the foundations branch? It's late on a Friday so my brain might not be working hahaha

marissahuysentruyt (commented):
Nice work on this! I think we just to rebuild the metadata and it'll be good to merge! I just wanted to double check with you on some of those action button "unused custom properties" (I left a comment). Otherwise, I didn't get any linter warnings/errors for each of the components ✅

components/actionbutton/index.css
```diff
@@ -11,65 +11,18 @@
...
 
 	--spectrum-actionbutton-border-color-default: var(--highcontrast-actionbutton-border-color-default, var(--mod-actionbutton-border-color-default, var(--spectrum-gray-400)));
 	--spectrum-actionbutton-border-color-hover: var(--highcontrast-actionbutton-border-color-hover, var(--mod-actionbutton-border-color-hover, var(--spectrum-gray-500)));
 	--spectrum-actionbutton-border-color-down: var(--highcontrast-actionbutton-border-color-down, var(--mod-actionbutton-border-color-down, var(--spectrum-gray-600)));
 	--spectrum-actionbutton-border-color-focus: var(--highcontrast-actionbutton-border-color-focus, var(--mod-actionbutton-border-color-focus, var(--spectrum-gray-500)));
 	--spectrum-actionbutton-border-color-disabled: var(--spectrum-disabled-border-color);
-
-	--spectrum-actionbutton-content-color-selected: var(--spectrum-gray-50);
+	--spectrum-actionbutton-content-color-disabled: var(--spectrum-disabled-content-color);
```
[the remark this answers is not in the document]
cdransf: It is — it's not in `s2-foundations-redux` (the custom property isn't present in `index.css` or `spectrum-two.css` and it's not used in `index.css`). `--spectrum-disabled-content-color` is widely used though, so I'm comfortable with this alias (but I'm I don't have strong opinions either way).

components/actionbutton/index.css
cdransf: Ah, good catch! It looks like those are unused because we replaced fallback definitions for quite a few mods with more generic Spectrum tokens. Those custom properties had been used here:

```
--mod-actionbutton-background-color-default: var(--mod-actionbutton-background-color-default-selected, var(--spectrum-neutral-background-color-selected-default));
--mod-actionbutton-background-color-hover: var(--mod-actionbutton-background-color-hover-selected, var(--spectrum-neutral-background-color-selected-hover));
--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-neutral-background-color-selected-down));
--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-neutral-background-color-selected-key-focus));
```

I know — in a lot of cases — we'll create component-specific aliases for tokens, but the implementation on `spectrum-two` _isn't_ doing that. For example:

**s2-foundations-redux**

```css
--mod-actionbutton-background-color-default: var(--mod-actionbutton-background-color-default-selected, var(--spectrum-actionbutton-background-color-selected));
--mod-actionbutton-background-color-hover: var(--mod-actionbutton-background-color-hover-selected, var(--spectrum-actionbutton-background-color-selected-hover));
--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-actionbutton-background-color-selected-down));
--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-actionbutton-background-color-selected-focus));
--mod-actionbutton-background-color-disabled: var(--spectrum-actionbutton-background-color-selected-disabled);

--mod-actionbutton-border-color-default: transparent;
--mod-actionbutton-border-color-hover: transparent;
--mod-actionbutton-border-color-down: transparent;
--mod-actionbutton-border-color-focus: transparent;
--mod-actionbutton-border-color-disabled: transparent;

--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-actionbutton-content-color-selected));
--mod-actionbutton-content-color-hover: var(--mod-actionbutton-content-color-hover-selected, var(--spectrum-actionbutton-content-color-selected));
--mod-actionbutton-content-color-down: var(--mod-actionbutton-content-color-down-selected, var(--spectrum-actionbutton-content-color-selected));
--mod-actionbutton-content-color-focus: var(--mod-actionbutton-content-color-focus-selected, var(--spectrum-actionbutton-content-color-selected));
```

**spectrum-two**

```css
--mod-actionbutton-background-color-default: var(--mod-actionbutton-background-color-default-selected, var(--spectrum-neutral-background-color-selected-default));
--mod-actionbutton-background-color-hover: var(--mod-actionbutton-background-color-hover-selected, var(--spectrum-neutral-background-color-selected-hover));
--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-neutral-background-color-selected-down));
--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-neutral-background-color-selected-key-focus));

--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-gray-25));
--mod-actionbutton-content-color-hover: var(--mod-actionbutton-content-color-hover-selected, var(--spectrum-gray-25));
--mod-actionbutton-content-color-down: var(--mod-actionbutton-content-color-down-selected, var(--spectrum-gray-25));
--mod-actionbutton-content-color-focus: var(--mod-actionbutton-content-color-focus-selected, var(--spectrum-gray-25));
```

Redux used the aliasing pattern, but the spectrum-two branch doesn't, so I'm not sure which pattern or direction to take. 😅 Does the fact that they're mods factor in too?

marissahuysentruyt: Happy Monday 👋  Just wanted to clarify this comment. It looks like in `s2-foundations-redux` that at least `--spectrum-actionbutton-background-color-selected-disabled` is in fact used: https://github.com/adobe/spectrum-css/blob/9c3a6253d4dd7208bcb9c2cab544c204e9cc9fcf/components/actionbutton/index.css#L136

I guess my question is the same, how are we reconciling what's going on in the foundations branch or main? I feel like I'm missing something 😵‍💫 maybe I'm thinking too hard, or not even looking in the right files.

cdransf: Ahhh, that makes sense! I pushed up [08087d](https://github.com/adobe/spectrum-css/pull/3476/commits/08087dcf8e9c6cc1ac86c16d2c7b45243e09095f) which restores those and references that went missing from `s2-foundations-redux`.

components/dialog/index.css
```diff
@@ -414,6 +414,6 @@
 
 @media (forced-colors: active) {
 	.spectrum-Dialog {
-		border: 1px solid var(--highcontrast-standard-dialog-border-color, rgba(var(--spectrum-black-rgb), var(--spectrum-overlay-opacity)));
+		border: 1px solid rgba(var(--spectrum-black-rgb), var(--spectrum-overlay-opacity));
```
rise-erpelding, line 417: Judging by what Marissa had done in her dialog migration PR #2860, this is probably the right call, but I'm not able to verify what it looks like because Dialog is looking like this for me:
image
Is anyone else having the same experience? If so, maybe we can make a follow-up card to look into this issue since it seems unrelated to the current fixes in this PR (if one doesn't already exist)?

Unless @marissahuysentruyt remembers the migration well enough and can spot something missing that would fix it here.

components/actionbutton/index.css
```diff
@@ -144,15 +102,17 @@
 	}
 
 	&.is-selected {
-		--mod-actionbutton-background-color-default: var(--mod-actionbutton-background-color-default-selected, var(--spectrum-neutral-background-color-selected-default));
-		--mod-actionbutton-background-color-hover: var(--mod-actionbutton-background-color-hover-selected, var(--spectrum-neutral-background-color-selected-hover));
-		--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-neutral-background-color-selected-down));
-		--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-neutral-background-color-selected-key-focus));
-
-		--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-gray-25));
```
rise-erpelding, line 152: Do we still need this (and maybe the hover, down, focus colors here)? I'm seeing selected content color coming through like this now, looking black:
image

rise-erpelding (changes requested):
Wow, this ended up being quite the collection of components, especially with actionbutton and dialog in the mix! It looks like you took care of all the no undefined custom props warnings, so that's great!

I think there are a few remaining oddities particularly with actionbutton, like the selected state content color (and also the default/disabled state border, which I didn't comment on within the code because I wasn't able to figure out if that was because of a missing/undefined custom prop or something else). Hopefully we can resolve them here, or if not, in the S2 migration. I'd also noticed that when I was running `yarn build` on this branch that there were some lingering metadata changes that needed to be committed, but that may or may not still be an issue if your branch has been updated since then.

components/dialog/index.css
cdransf: Got it! I'm seeing the same but had gone ahead and lifted it from the S2 theme file on the redux branch. I'll wait for @marissahuysentruyt to chime in,

components/actionbutton/index.css
cdransf: Ohh nice catch! The first block of mods had been restored, but not `--mod-actionbutton-content-color-default`. Restoring that fixes the selected button state:

image

cdransf:
> Wow, this ended up being quite the collection of components, especially with actionbutton and dialog in the mix! It looks like you took care of all the no undefined custom props warnings, so that's great!
> 
> I think there are a few remaining oddities particularly with actionbutton, like the selected state content color (and also the default/disabled state border, which I didn't comment on within the code because I wasn't able to figure out if that was because of a missing/undefined custom prop or something else). Hopefully we can resolve them here, or if not, in the S2 migration. I'd also noticed that when I was running `yarn build` on this branch that there were some lingering metadata changes that needed to be committed, but that may or may not still be an issue if your branch has been updated since then.

I left a few comments, fixed the `actionbutton` selected state and updated the metadata ✨

components/dialog/index.css
marissahuysentruyt: @rise-erpelding @cdransf I do have a fix for this in the takeover dialog PR! https://github.com/adobe/spectrum-css/pull/3347/files#r1878562052 `layout: fullscreen` 

And yes, Cory and I talked about the removal of that highcontrast variable. I think it makes sense. We never defined it- I added it as if it were a mod, but that was the incorrect usage.

components/actionbutton/index.css
```diff
@@ -144,15 +102,19 @@
...
+		--mod-actionbutton-background-color-hover: var(--mod-actionbutton-background-color-hover-selected, var(--spectrum-actionbutton-background-color-selected-hover));
+		--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-actionbutton-background-color-selected-down));
+		--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-actionbutton-background-color-selected-focus));
+		--mod-actionbutton-background-color-disabled: var(--spectrum-actionbutton-background-color-selected-disabled);
 
 		--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-hover: var(--mod-actionbutton-content-color-hover-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-down: var(--mod-actionbutton-content-color-down-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-focus: var(--mod-actionbutton-content-color-focus-selected, var(--spectrum-gray-25));
```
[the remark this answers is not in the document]
cdransf: Thank you! Just fixed that,

components/clearbutton/index.css
```diff
@@ -22,6 +21,11 @@
 	--spectrum-clear-button-icon-color-down: var(--spectrum-neutral-content-color-down);
 	--spectrum-clear-button-icon-color-key-focus: var(--spectrum-neutral-content-color-key-focus);
 
+	--spectrum-clear-button-background-color: transparent;
+	--spectrum-clear-button-background-color-hover: transparent;
+	--spectrum-clear-button-background-color-down: transparent;
+	--spectrum-clear-button-background-color-key-focus: transparent;
+
```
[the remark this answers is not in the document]
cdransf: Ah yes! I believe so. Just added those as well. ✨

rise-erpelding (dismissed):
Looks good to me! I know @marissahuysentruyt was planning to take another look at this, I'd be curious to know if her eagle eyes can spot anything here.

Documenting that I still think actionbutton is missing the basebutton styles and that's what's causing the two-colored outset border style, but we weren't able to get those base button styles to work on action button, so there may need to be a follow-up card for that (or we'll address it in the migration).
image

components/actionbutton/index.css
```diff
@@ -132,7 +92,6 @@
 		--mod-actionbutton-border-color-focus: var(--spectrum-transparent-white-600);
 
 		--spectrum-actionbutton-background-color-disabled: transparent;
-		--spectrum-actionbutton-background-color-selected-disabled: var(--spectrum-transparent-white-200);
```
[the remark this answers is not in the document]
marissahuysentruyt: Here's what I was seeing before, without the background-color-selected-disabled and content-color-selected variables:

Screenshot 2025-01-08 at 10 09 12 AM

But I think we want something more like this, with the background-color-selected-disabled? I'm trying not to just fix styles and save that for the migration (hence why the content color is still incorrect), but I think the background-color-selected-disabled should be defined here.

Screenshot 2025-01-08 at 10 29 01 AM

marissahuysentruyt (commented):
I believe I found one or two more variables that _should_ be defined here- let me know what you think!

components/actionbutton/index.css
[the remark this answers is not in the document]
cdransf: Just updated this!

components/actionbutton/index.css
```diff
@@ -114,7 +74,6 @@
 		--mod-actionbutton-border-color-focus: var(--spectrum-transparent-black-600);
 
 		--spectrum-actionbutton-background-color-disabled: transparent;
-		--spectrum-actionbutton-background-color-selected-disabled: var(--spectrum-transparent-black-200);
```
marissahuysentruyt, line 117: Do we need this line still? It looks like the static white has both a disabled and a selected-disabled custom property, so maybe we still need both properties here. What do you think?

rise-erpelding (commented):
Wanted to drop a quick comment since I know this is still in review!

Since actionbutton particularly has been causing problems, I dug into this a bit more to try to figure out what's going on with it. Comparing how it's changed from `s2-foundations-redux`, I noticed that the styles that previously applied with the import of `commons/basebutton.css` are now hardcoded into the component, like `cursor: pointer;`, `user-select: none;` etc.

We lost the `border-style: solid;` in commons/basebutton.css in this [commit](https://github.com/adobe/spectrum-css/commit/cc3b7d7645b9e4cf500e95fa6e07d37f069cf92b#diff-715275ed27a56c42d8cf769b25514cbd54a4fd44bf5349430f0acd9ba1a268f2) so it didn't make it in when the commons styles were pasted in in this later [commit](https://github.com/adobe/spectrum-css/pull/2352/commits/9444cdbb3a6d8c2d880b53c691c0c067e04e7e1f#diff-b86e2a5a1675710bcd76aa3ceec37bdabcfcbf4082593cdc602f41712a6165a0). That was the style we needed for the button outlines and importing basebutton in this branch wasn't doing us any good because that style isn't in basebutton anymore.

Anyway, a couple of takeaways from that knowledge:
1. We should probably paste that `border-style: solid;`, probably below this section:

```
/* Remove the inheritance of text transform on button in Edge, Firefox, and IE. */
	text-transform: none;

	/* stylelint-disable-next-line property-no-vendor-prefix -- Correct the inability to style clickable types in iOS and Safari (normalize). */
	-webkit-appearance: button;
```

2. If things are getting really hairy with fixes, it might help to refer back to the way the code looked on the redux branch and how it changed between there and here. We know we have to add in any missing themes custom properties and ideally that would be it, but hopefully we can uncover any issues outside of that by looking back. 🤞

cdransf:
> Wanted to drop a quick comment since I know this is still in review!
> 
> Since actionbutton particularly has been causing problems, I dug into this a bit more to try to figure out what's going on with it. Comparing how it's changed from `s2-foundations-redux`, I noticed that the styles that previously applied with the import of `commons/basebutton.css` are now hardcoded into the component, like `cursor: pointer;`, `user-select: none;` etc.
> 
> We lost the `border-style: solid;` in commons/basebutton.css in this [commit](https://github.com/adobe/spectrum-css/commit/cc3b7d7645b9e4cf500e95fa6e07d37f069cf92b#diff-715275ed27a56c42d8cf769b25514cbd54a4fd44bf5349430f0acd9ba1a268f2) so it didn't make it in when the commons styles were pasted in in this later [commit](https://github.com/adobe/spectrum-css/pull/2352/commits/9444cdbb3a6d8c2d880b53c691c0c067e04e7e1f#diff-b86e2a5a1675710bcd76aa3ceec37bdabcfcbf4082593cdc602f41712a6165a0). That was the style we needed for the button outlines and importing basebutton in this branch wasn't doing us any good because that style isn't in basebutton anymore.
> 
> Anyway, a couple of takeaways from that knowledge:
> 
> 1. We should probably paste that `border-style: solid;`, probably below this section:
> 
> ```
> /* Remove the inheritance of text transform on button in Edge, Firefox, and IE. */
> 	text-transform: none;
> 
> 	/* stylelint-disable-next-line property-no-vendor-prefix -- Correct the inability to style clickable types in iOS and Safari (normalize). */
> 	-webkit-appearance: button;
> ```
> 
> 2. If things are getting really hairy with fixes, it might help to refer back to the way the code looked on the redux branch and how it changed between there and here. We know we have to add in any missing themes custom properties and ideally that would be it, but hopefully we can uncover any issues outside of that by looking back. 🤞

I just reverted 

> Wanted to drop a quick comment since I know this is still in review!
> 
> Since actionbutton particularly has been causing problems, I dug into this a bit more to try to figure out what's going on with it. Comparing how it's changed from `s2-foundations-redux`, I noticed that the styles that previously applied with the import of `commons/basebutton.css` are now hardcoded into the component, like `cursor: pointer;`, `user-select: none;` etc.
> 
> We lost the `border-style: solid;` in commons/basebutton.css in this [commit](https://github.com/adobe/spectrum-css/commit/cc3b7d7645b9e4cf500e95fa6e07d37f069cf92b#diff-715275ed27a56c42d8cf769b25514cbd54a4fd44bf5349430f0acd9ba1a268f2) so it didn't make it in when the commons styles were pasted in in this later [commit](https://github.com/adobe/spectrum-css/pull/2352/commits/9444cdbb3a6d8c2d880b53c691c0c067e04e7e1f#diff-b86e2a5a1675710bcd76aa3ceec37bdabcfcbf4082593cdc602f41712a6165a0). That was the style we needed for the button outlines and importing basebutton in this branch wasn't doing us any good because that style isn't in basebutton anymore.
> 
> Anyway, a couple of takeaways from that knowledge:
> 
> 1. We should probably paste that `border-style: solid;`, probably below this section:
> 
> ```
> /* Remove the inheritance of text transform on button in Edge, Firefox, and IE. */
> 	text-transform: none;
> 
> 	/* stylelint-disable-next-line property-no-vendor-prefix -- Correct the inability to style clickable types in iOS and Safari (normalize). */
> 	-webkit-appearance: button;
> ```
> 
> 2. If things are getting really hairy with fixes, it might help to refer back to the way the code looked on the redux branch and how it changed between there and here. We know we have to add in any missing themes custom properties and ideally that would be it, but hopefully we can uncover any issues outside of that by looking back. 🤞

Alrighty! I just added that missing rule. I had tried reintroducing the action button updates by stepping through the changes again, but things still weren't quite right.

components/actionbutton/index.css
marissahuysentruyt: @cdransf I thought we had added this back in for the static black, but we must have reverted it. That way both static white & static black has their own, specific `--spectrum-actionbutton-background-color-selected-disabled` value.

components/actionbutton/index.css
```diff
@@ -144,15 +105,16 @@
...
+		--mod-actionbutton-background-color-down: var(--mod-actionbutton-background-color-down-selected, var(--spectrum-actionbutton-background-color-selected-down));
+		--mod-actionbutton-background-color-focus: var(--mod-actionbutton-background-color-focus-selected, var(--spectrum-actionbutton-background-color-selected-focus));
+		--mod-actionbutton-background-color-disabled: var(--spectrum-actionbutton-background-color-selected-disabled);
 
-		--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-hover: var(--mod-actionbutton-content-color-hover-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-down: var(--mod-actionbutton-content-color-down-selected, var(--spectrum-gray-25));
-		--mod-actionbutton-content-color-focus: var(--mod-actionbutton-content-color-focus-selected, var(--spectrum-gray-25));
+		--mod-actionbutton-content-color-default: var(--mod-actionbutton-content-color-default-selected, var(--spectrum-actionbutton-content-color-selected));
```
rise-erpelding, line 114: Are we missing this [chunk](https://github.com/adobe/spectrum-css/blob/s2-foundations-redux/components/actionbutton/index.css#L138-L142) of border-color mods that s2-foundations has about here?

components/actionbutton/index.css
rise-erpelding, line 1: There's this selector in [s2 foundations](https://github.com/adobe/spectrum-css/blob/s2-foundations-redux/components/actionbutton/index.css#L68) that sets the border color to transparent that looks like its missing here, after `--spectrum-logical-rotation`. You have this selector here but I think would maybe need to add the border-color?

```css
&.spectrum-ActionButton--quiet {
		--spectrum-actionbutton-border-color: transparent;
	}
```

Except that I don't think that works because now we're not setting `--spectrum-actionbutton-border-color` anywhere, so maybe try something like
```css
                 --mod-actionbutton-border-color-default: transparent;
		--mod-actionbutton-border-color-hover: transparent;
		--mod-actionbutton-border-color-down: transparent;
		--mod-actionbutton-border-color-focus: transparent;
		--mod-actionbutton-border-color-disabled: transparent;
```

----------

We're also missing [setting](https://github.com/adobe/spectrum-css/blob/s2-foundations-redux/components/actionbutton/index.css#L104C3-L104C96) the static-black-focus-indicator-color:

```css
--mod-actionbutton-focus-indicator-color: var(--spectrum-static-black-focus-indicator-color);
```

And [static white](https://github.com/adobe/spectrum-css/blob/s2-foundations-redux/components/actionbutton/index.css#L128C3-L128C96):
```css
--mod-actionbutton-focus-indicator-color: var(--spectrum-static-white-focus-indicator-color);
```

rise-erpelding (changes requested):
Noting a couple of diffs I saw between this branch and s2-foundations that are causing some regressions!

components/actionbutton/index.css
cdransf: Just added those back — thank you! ✨

components/actionbutton/index.css
cdransf: I believe the `spectrum-action-border-color` change(s) should be handled in this commit: [fef056](https://github.com/adobe/spectrum-css/pull/3476/commits/fef056c39b9d42d1da83b80a5d76d4091c67aa0e). I stepped through the states on this branch/PR and matched them to a recent deploy with the previous, working state of the button(s).

image
image

rise-erpelding:
Hey! I think we've come a long way on action button and while we're still getting unintended borders on several quiet variants, I think a lot of this is due to some of the structural differences between the branches, which isn't really what we intended to fix here, and is proving to be a lot more difficult than originally anticipated. Overall, it looks fine and there are no glaring contrast or background issues that I'm seeing. If there are no other improvements that you or @marissahuysentruyt see to make on action button, I think that one's good to go.

When scanning the rest of the components fixed here, I did still notice some possible inconsistencies with button that I'd love some opinions on:
image
In this image, I think the second set of accent/negative buttons is supposed to be the outline variant? But also, I think button has already been migrated to S2 in this branch and it looks as if maybe those variants don't exist anymore. Could you confirm? It seems reasonable to think that we might need to remove test cases or maybe make some adjustments to the template after all the updates to `spectrum-two`, and if that's the case, it'd be ok to make another card for that.

As far as I can tell I think the other components look good!

tokens/dist/index.css
rise-erpelding, line 1: Is this file deletion intentional?

cdransf: It wasn't! Should be restored now. ✨

marissahuysentruyt (commented):
Running `yarn linter` against each component doesn't return any errors 🥳 With the project running, I don't see anything that's absolutely broken (wrong styles - yes, but nothing totally broken or unusable). I was a little concerned at the number of mods that look like they've been deleted in action button, but you mentioned in standup that they seemed self-referential, so maybe it's ok for the `spectrum-two` branch. 

Like @rise-erpelding said, there's lots of work that probably could be done with action button. We'll have to readdress those things when the action button migration gets restarted. I also see the missing outlines in the test cases for outline buttons (she left that screenshot).

components/actionbutton/index.css
```diff
@@ -69,7 +29,9 @@
 	--spectrum-actionbutton-border-color-focus: var(--highcontrast-actionbutton-border-color-focus, var(--mod-actionbutton-border-color-focus, var(--spectrum-gray-500)));
 	--spectrum-actionbutton-border-color-disabled: var(--spectrum-disabled-border-color);
 
-	--spectrum-actionbutton-content-color-selected: var(--spectrum-gray-50);
+	--spectrum-actionbutton-content-color-selected: var(--spectrum-gray-25);
+	--spectrum-actionbutton-content-color-disabled: var(--mod-disabled-content-color, var(--spectrum-disabled-content-color));
+	--spectrum-actionbutton-background-color-disabled: var(--mod-disabled-background-color, var(--spectrum-disabled-background-color));
```
marissahuysentruyt, line 34: For these 2 new mods (mod-disabeld-content-color and mod-disabled-background-color), what do you think of naming them more in the format we have for the other mods? 

```
--mod-actionbutton-content-color-disabled
--mod-actionbutton-background-color-disabled
```

Just a thought to match the other properties.

marissahuysentruyt (approved):
Thanks for all your hard work on this!

rise-erpelding (approved):
I think this is good to go!

Re-summarizing my notes:
* Action button has several instances of visible borders where they should be transparent, but this doesn't seem to be because of missing custom properties so hopefully this can be taken care of in the migration.
* Button - some follow up should be done to remove unneeded outline variants from tests since it's already been migrated and those variants don't exist anymore.

Clear button, Avatar, Dialog, and Infield button look good! Thanks @cdransf!

PR #3488, opened 2025-01-09
feat(tokens): remove components from token build

castastrophe:
### Breaking change

This update removes the compiled `components` directory from the token output. Instead, customers are asked to source the theming data from the component's `theme/(express,spectrum,spectrum-two).css` and `index-theme.css`. For example:

```
import "@spectrum-css/actionbutton/dist/index-base.css";
import "@spectrum-css/actionbutton/dist/index-theme.css";
import "@spectrum-css/actionbutton/dist/themes/express.css";
```

**Why?** Pre-compiling and publishing the components directory inside the tokens package can lead to mismatched component styles and theme assets. Because these assets are generated by the component's CSS and then moved into the tokens package, if you pull down a version of the tokens package that is misaligned with the release of the component, there is a risk that they will not be compatible. It is safer to use the theming layers provided by the components directly.

In addition, the `custom-*-vars.css` files previously shipped in the `spectrum` and `express` folders will no longer be shipped separately. This data already exists in the `*-vars.css` file with a matching name. For example, `spectrum/custom-large-vars.css` already exists within `spectrum/large-vars.css`, concatenated with the token-generated output.

### Minor

Whitespace has been cleaned up in the exported content for readability.

### Validation steps

- [x] `yarn builder tokens`: expect no `dist/css/components` directory to be generated

### Regression testing

Validate:

## To-do list

- [n/a] I have updated relevant storybook stories and templates.
- [n/a] I have tested these changes in Windows High Contrast mode.


pfulton (approved):
Pulled the branch down and ran the cleaner and then the builder on the `tokens` dir, and everything worked as expected. The deploy URL also works as expected!

PR #2697, opened 2024-04-25
feat(picker): migrate to spectrum 2

jawinn:
## Description

Migrates the Picker component to use Spectrum 2 specifications. 

Includes a total overhaul of the Chromatic-only Storybook template in order to increase coverage for this component's many options. This format was based off of the stories for Action button and the grid of Picker options in the S2 / Desktop Figma. See the "Show testing preview" within Storybook. A control for the optional workflow icon that appears to the left of the text within the Picker was added; this was previously on the YML docs but missing from Storybook.

This also fixes:
- The positioning of Popover in Storybook for the side label variant (this previously was incorrectly appearing to the right instead of below). The Popover width will also now match the width of the Picker in the Storybook examples.

**Mod custom property changes**
- The mod custom property `--mod-picker-spacing-text-to-alert-icon-inline-start` that was previously marked for deprecation has been removed. This is replaced by `--mod-picker-spacing-text-to-icon-inline-end`.
- Some of the icon to text and text to icon spacing mods were not working as intended. This has been cleared up and documented. 
    - `--mod-picker-spacing-starting-icon-to-text` added. This affects the optional starting workflow icon.
    - `--mod-picker-spacing-text-to-icon-inline-end` This affects text spacing to visual (indicator or disclosure).
    - `--mod-picker-spacing-text-to-icon` removed. This had been applied to different things and was affecting the inline end icon. 
- Additional mod custom properties were added for some quiet values.

CSS-617

## How and where has this been tested?

Please tag yourself on the tests you've marked complete to confirm the tests have been run by someone other than the author.

### Validation steps

- [x] Reviewed and approved by design team (@jawinn) -- good except for followup work to use "inline progress circle" and possible changes to down state for quiet; followup captured in CSS-777
- [x] Test the default Picker and its controls @marissahuysentruyt 
- [x] Picker is using the new tokens for field width and rounding, as defined on the token specs design file @rise-erpelding 
- [x] Examples in the Picker Chromatic template look correct and like the S2 / Desktop Figma + token specs @rise-erpelding 
- [x] Background color and color do not change when hovering over any disabled state @marissahuysentruyt 
- [x] Fix: Popover appears beneath the Picker instead of weirdly to the side when viewing the side label positioning @marissahuysentruyt 
- [x] "Icon" content field in Storybook is working @marissahuysentruyt 

### Regression testing

Validate:

1. The documentation pages for at least two other components are still loading, including:

- [x] The pages render correctly, are accessible, and are responsive.

[...]

PR #3514, opened 2025-01-21
chore: spectrum-two fixes

marissahuysentruyt:
## Description

This PR reimplements some fixes for the `spectrum-two` branch after a rebase. 

1. Adds a missing `variant = "neutral"` default arg to the tooltip template. This should alleviate the errors on the docs page.

Before 🚫 
Screenshot 2025-01-08 at 4 51 59 PM

After ✅ 
Screenshot 2025-01-08 at 4 51 36 PM

2. An additional fix on the steplist docs page occurred since the `WithTooltip` story calls for the tooltip component. 🥳 

Before 🚫 
Screenshot 2025-01-22 at 10 20 01 AM

After ✅ 
Screenshot 2025-01-22 at 10 22 00 AM

3. The calls to `window.isChromatic` have been removed.  It also corrects the arg names for picker, in order for them to be properly passed to the popover used in the template and render the content. This should alleviate all errors and most of the missing content on the docs page. (additional styling and content changes will need to be addressed in a separate PR).

Before 🚫 
Screenshot 2025-01-22 at 9 53 23 AM

After ✅ 
Screenshot 2025-01-22 at 10 04 48 AM

4. An additional fix on the form docs page occurred since the examples we use contain the picker component. 🥳 

Before 🚫 
Screenshot 2025-01-22 at 9 53 35 AM

After ✅ 
Screenshot 2025-01-22 at 10 04 55 AM

5. An additional fix on the tabs docs page occurred since the Overflow story calls for the picker component. 🥳 

Before 🚫
Screenshot 2025-01-22 at 10 11 01 AM

After ✅ 
Screenshot 2025-01-22 at 10 11 33 AM

### Validation steps

- Pull down the branch to run locally or use the deploy preview.
- Visit the picker docs page. 
- Verify no errors referencing the `variant is not defined` appear in place of the `WithTooltip` story canvas. Each story step in the steplist should have a rendered tooltip with step number content.
- Visit the form docs page. 
- Similarly to the picker docs, now errors should occur regarding `window.isChromatic()`. All form components should render with a closed picker component that displays the placeholder text.
- Visit the tooltip docs page.
- Verify there are no errors referencing `variant is not defined`. All tooltips should now display as expected.
- Visit the tabs docs page. 
- Similarly to the picker docs, no errors should occur regarding `window.isChromatic()`. The overflow tabs components should render with a closed picker component that displays the first tab item's text & chevron.
- Visit the steplist docs page.
- Verify 

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/picker/stories/picker.stories.js
+++ b/components/picker/stories/picker.stories.js
@@ -97,7 +97,7 @@ export default {
 		},
 		isHovered,
 		isActive,
-		content: { table: { disable: true } },
+		popoverContent: { table: { disable: true } },
 	},
 	args: {
 		rootClass: "spectrum-Picker",
--- a/components/picker/stories/template.js
+++ b/components/picker/stories/template.js
@@ -64,8 +64,7 @@ export const Picker = ({
 			id=${id}
 			style=${styleMap(customStyles)}
 			type="button"
-			@click=${() => {
-				if (window.isChromatic()) return;
+			@click=${function() {
 				updateArgs({ isOpen: !isOpen });
 			}}
 			aria-labelledby=${ifDefined(ariaLabeledBy)}
@@ -195,12 +194,6 @@ export const Template = ({
 				}, context)
 			)}
 		</div>`;
-
-	// Make sure there is a wrapper around sibling components when using the Chromatic
-	// template, so their layout is not affected by the flex and grid layouts used.
-	if (window.isChromatic()) {
-		return html`<div style="position: relative;">${markup}</div>`;
-	}
 	return markup;
 };
 
--- a/components/picker/stories/template.js
+++ b/components/picker/stories/template.js
@@ -109,6 +109,7 @@ export const Template = ({
 	size = "m",
 	label,
 	labelPosition = "top",
+	placeholder,
 	helpText,
 	isQuiet = false,
 	isOpen = false,
@@ -125,7 +126,7 @@ export const Template = ({
 		// Helps ensure that Popover appears below the Picker, with side labels layout.
 		display: "block",
 	},
-	content = [],
+	popoverContent = [],
 } = {}, context = {}) => {
 	const pickerMarkup = Picker({
 		size,
@@ -134,17 +135,18 @@ export const Template = ({
 		isInvalid,
 		isDisabled,
 		isLoading,
-		content,
+		placeholder,
+		popoverContent,
 		labelPosition,
 		ariaLabeledBy: fieldLabelId,
 	}, context);
 
-	const popoverMarkup = content.length !== 0 ? Popover({
+	const popoverMarkup = popoverContent.length !== 0 ? Popover({
 		isOpen: isOpen && !isDisabled && !isLoading,
 		withTip: false,
 		position: "bottom",
 		isQuiet,
-		content,
+		content: popoverContent,
 		size,
 		customStyles: customPopoverStyles,
 	}, context) : "";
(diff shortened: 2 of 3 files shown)


components/picker/stories/picker.stories.js
```diff
@@ -97,7 +97,7 @@ export default {
 		},
 		isHovered,
 		isActive,
-		content: { table: { disable: true } },
+		popoverContent: { table: { disable: true } },
```
marissahuysentruyt, line 100: During the picker migration, we tried to clarify this picker control by calling it `popoverContent`. This is the content that gets passed to the popover component in the template. 

Popover has its own `content` arg, so that's why you'll see the `content: popoverContent` assignment.

components/tooltip/stories/template.js
```diff
@@ -12,6 +12,7 @@ export const Template = ({
 	rootClass = "spectrum-Tooltip",
 	label,
 	placement,
+	variant = "neutral",
```
marissahuysentruyt, line 15: Looks like the informative & negative tooltips got added back to Figma 11/18/24 (approximately). We'll have to create a card to implement the CSS styles and stories (for the docs page & controls)

```
variant: {
			name: "Variant",
			type: { name: "string" },
			table: {
				type: { summary: "string" },
				category: "Component",
			},
			options: ["neutral", "info", "negative"],
			control: "inline-radio",
```

tooltip migration PR: https://github.com/adobe/spectrum-css/pull/2743/files#diff-14d4b487bc2de8da9dccaef67bce6cccc92cbce28e2ac16912884edbce436cbc

components/picker/stories/picker.stories.js
marissahuysentruyt: Separately, there might be more that we can pull from the docs migration PR listed above. It looks like some of those additional controls/args (like `currentValue`, `isHovered`, `showWorkflowIcon`, and maybe some others) are missing currently. That could probably be figured out in a separate PR.

components/picker/stories/template.js
```diff
@@ -64,8 +64,7 @@ export const Picker = ({
 			id=${id}
 			style=${styleMap(customStyles)}
 			type="button"
-			@click=${() => {
-				if (window.isChromatic()) return;
+			@click=${function() {
 				updateArgs({ isOpen: !isOpen });
 			}}
```
marissahuysentruyt, line 69: On `main`, [this line](https://github.com/adobe/spectrum-css/blob/a5a8d7e9d23445bbfe42378e2b7f92845371882a/components/picker/stories/template.js#L57C4-L57C21) is `@click=${onclick}`, and the `onclick` function was [defined in the `Popover` call](https://github.com/adobe/spectrum-css/blob/a5a8d7e9d23445bbfe42378e2b7f92845371882a/components/picker/stories/template.js#L166). This template is pretty different from `main`, so I plopped the function here for now. 

Any thoughts or reservations on this?

rise-erpelding: I'm not sure I know enough about why this part of the code is different between the branches/how Chromatic normally handles Picker, but it does work as is, and would work as an arrow function as well 🤷‍♀️

components/tooltip/stories/template.js
rise-erpelding: 🙀 Good catch! Who knew those variants would come back again?

rise-erpelding (approved):
This works and fixes many of the other issues I was noticing in other components too!

This does a nice job preserving the changes from the picker migration in #2697 but keeps what we need to make things work! 🎉

PR #3531, opened 2025-01-30
feat(checkbox/group): s2 token migration, control spacing, subtle color changes

aramos-adobe:
## Checkbox/Group S2 Migrations

Checkbox has some new tokens for control and color. Most of the tokens have been updated in the global vars. 

### New tokens

`--spectrum-component-size-width-ratio-down`
`--spectrum-checkbox-bottom-to-text(S,M,L,XL)`
`--spectrum-checkbox-top-to-control (S,M,L,XL)`
`--spectrum-accent-content-color-default`
`--spectrum-accent-content-color-hover`
`--spectrum-accent-content-color-down`
`--spectrum-accent-content-color-key-focus`

### Modified tokens

`--spectrum-checkbox-checkmark-color`
`--spectrum-checkbox-invalid-color-down`
`--spectrum-checkbox-control-color-default`
`--spectrum-checkbox-control-color-hover`
`--spectrum-checkbox-control-color-down`
`--spectrum-checkbox-control-color-focus`

### Validation steps
1. Look at the checkbox component in storybook
2. Check light and dark modes in Testing Mode

### Regression testing

Validate:

3. If components have been modified, VRTs have been run on this branch:

## To-do list


components/checkbox/index.css
rise-erpelding, line 1: Going through the Figma specs here are some things I noticed on the CSS:
- [ ] Checkbox layout - all of the tokens look good here except for **Component size (down)**:
    - [ ] Awhile ago we wrote up some docs for down state in #2520, at that time the down state for checkbox was migrated to S2 as well so that it could be used on that docs page. There's also a comment in that PR from SWC about the approach there not working, that's something that we probably need to take some action on and find out more about to follow up. As it in this PR, it's changed from the checkbox being just a bit smaller when active, to being really small when active, so this will probably need another look.
- [X] Checkbox group layout - Checking the fieldgroup component, this looks good!
- [ ] Checkbox colors - this needs some work:
    - [ ] The invalid/error default checked hover (screenshot below is hovered) color is the `neutral-content-color-hover` black and not the `negative-color-1000` red. I think this is a bug that exists on `main` too, not something you introduced.
    image
    - [ ] The down/active color on this same invalid default checked checkbox is also a black color and not a red color, so this would need to be updated as well.
    It's a little strange that there's not a visual for the error, I wonder what's going on with that, but I think that's outside of the work of this PR 🤷‍♀️
- [ ] Checkbox typography
    - [ ] We're not using `default-font-family` in the component, should we be?
    - [ ] Same for `regular-font-weight` and
    - [ ] Also `default-font-style`, but I see TODO comments in here around L370, it might be time to start setting those styles? I'm not sure of the historical reason why they weren't before.
    - [ ] CJK - this looks good and works just fine! But could you revise it so that the custom properties are redefined instead of resetting the style (probably moving the whole selector up to the top with the custom property definitions)? Similar to how you did it in [Illustrated Message](https://github.com/adobe/spectrum-css/blob/aramos-adobe/css781-illustrated-message-s2-tokens/components/illustratedmessage/index.css#L46) but with line height. Nothing is wrong with the way you have it, but after analyzing our components and realizing we're doing it 3 different ways across all of them, we want to try to standardize how we do this wherever we can, I just documented this in our style guide [draft](https://adobe.enterprise.slack.com/docs/T024FSURM/F07RKLELX6V?focus_section_id=temp:C:KTa8d082b1ae0014ae6975dc7a3d).

components/checkbox/dist/index.css.map
rise-erpelding, line 1: Should we have this `components/checkbox/dist/index.css.map` and `components/checkbox/dist/index.css` committed?

components/progresscircle/dist/index.css
rise-erpelding, line 1: Wondering if these dist files for progresscircle also need to be removed.

rise-erpelding (dismissed):
This is a great start! I pointed out some of the changes that I thought were still needed to the CSS within the CSS file, but here's a quick rundown of what I think needs another look:

- [ ] Down state
- [ ] Some of the invalid default colors (emphasized colors seem to be working fine for whatever reason) - this is a (hopefully small) bugfix that I also noticed on main
- [ ] Typography use of tokens
- [ ] Updating CJK to match other components
- [ ] Probably removing some stray dist files
- [ ] There are SO many test cases covered already, but for some reason we don't have any focus, hover, or active states. I don't think there's any particular reason we've left them out as far as I know, so I'm wondering how you would feel about adding them? I think it would make catching bugs (like the bug I just mentioned above) a little easier!

As far as I can tell, I think the template.js and stories.js files for this component are fine, so probably no changes needed there!

components/checkbox/index.css
aramos-adobe: @rise-erpelding the default-font-family doesn't render correctly for some reason. For component size (down) there is something in documentation describing the usage.

jawinn, line 1: I'm noticing a couple things that need adjustment in high contrast mode.

With the addition of `--spectrum-checkbox-background-color`, this needs a highcontrast variable added where it is used. To contrast with the existing CanvasText border and content color, this should be set to Canvas:
Screenshot 2025-02-12 at 10 43 45 AM

I'm seeing a few discrepancies between different combinations on hover and focus. These actually exist in Prod as well:
- Invalid + Unchecked and Invalid + Indeterminate should also be showing the highlight color on hover and focus, similar to the default. Just the Invalid + Checked shows it on hover currently.
- Read-only + Unchecked + Hover and Read-only + Unchecked + Keyboard Focus shouldn't be showing the highlight color

When the background of the checkbox is the Highlight color, it should be using the matching foreground pair HighlightText within it to ensure contrast in different color schemes; right now the checkmark and the indeterminate dash are using the wrong keyword "Canvas" via `--spectrum-checkbox-checkmark-color`.

components/checkbox/index.css
```diff
@@ -20,15 +20,16 @@
 	--spectrum-checkbox-checkmark-color: var(--highcontrast-checkbox-background-color-default, var(--mod-checkbox-checkmark-color, var(--spectrum-gray-50)));
```
jawinn, line 20: The pattern used on this line doesn't seem to be consistent with where we typically define the mods and highcontrast variables. This seems like a good time to clean this up.

components/checkbox/index.css
```diff
@@ -408,6 +418,8 @@
 
 	cursor: pointer;
 
+	margin-block-start: var(--spectrum-checkbox-top-to-control);
```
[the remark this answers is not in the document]
jawinn: I was thinking that we would replace the value of the existing margin property. So instead of:

```
margin: calc(var(--mod-checkbox-spacing, var(--spectrum-checkbox-spacing)) / 2) 0;
margin-block-start: var(--spectrum-checkbox-top-to-control);
```
it'd be something like
```
margin: var(--mod-checkbox-margin-block, var(--spectrum-checkbox-top-to-control)) 0;
```
Now that we have this top to control token value, it looks like we can remove the previously used `--spectrum-checkbox-spacing`?

We want to make sure this value can still be customized with a mod; the previous mod `--mod-checkbox-spacing` wouldn't work anymore on the top margin when both `margin` and `margin-block-start` are used.

components/checkbox/stories/checkbox.test.js
```diff
@@ -13,6 +13,14 @@ export const CheckboxGroup = Variants({
...
 			isEmphasized: true,
 		},
+		{
+			testHeading: "Focused",
+			isFocused: true,
+		},
+		{
+			testHeading: "Hovered",
+			isHovered: true,
+		},
```
jawinn, line 23: This is a nice addition. I'd suggest also adding an emphasized + hovered test.

.changeset/six-donuts-march.md
```diff
@@ -0,0 +1,26 @@
+---
+"@spectrum-css/checkbox": major
+---
+
+# Checkbox S2 Migration
+
+Checkbox has some new control and color tokens.
```
jawinn, line 7: Could you add some elaboration here about the changes? Such as the removal of `.spectrum-Checkbox--sizeM` as medium is the default, high contrast updates, maybe font changes, and anything else that would be helpful to know about what has changed? I also see the thickness mod renamed, fixing its typo.

.changeset/six-donuts-march.md
```diff
@@ -0,0 +1,26 @@
...
+
+## New tokens
+
+`--spectrum-component-size-width-ratio-down`
+`--spectrum-checkbox-bottom-to-text(S,M,L,XL)`
+`--spectrum-checkbox-top-to-control (S,M,L,XL)`
```
jawinn, line 13: Could you list these full token names out individually?

jawinn (changes requested):

jawinn (changes requested):

.changeset/six-donuts-march.md
aramos-adobe: @jawinn Oh it's because I didn't actually use the actual name. They all share the same token name but swapping the values each shirt size.

components/checkbox/index.css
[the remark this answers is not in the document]
aramos-adobe: giphy

components/checkbox/index.css
```diff
@@ -511,12 +490,13 @@
 
 /* Disabled */
 .spectrum-Checkbox .spectrum-Checkbox-input,
-.spectrum-Checkbox .spectrum-Checkbox-input:checked {
+.spectrum-Checkbox .spectrum-Checkbox-input:checked,
+&.is-disabled {
```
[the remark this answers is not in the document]
aramos-adobe: @jawinn I'm not sure where that addition came from honestly. I think that came in from a merge or something.

jawinn:
Could you look at the following?
1. https://github.com/adobe/spectrum-css/pull/3531#discussion_r1964116979 I've been following up with this thread but am not seeing any changes or replies yet
2. https://github.com/adobe/spectrum-css/pull/3531#discussion_r1990103048 A change is needed here
3. Did you look at this comment for high contrast? "The focus outline border radius doesn't seem to match and the gap doesn't look like the same as the regular focus indicator. I'm assuming that this should look just like the non high contrast focus indicator, just with a different (system) color."
4. The colors are looking better. One thing I'm seeing in high contrast is a change of the background color on hover for default readonly checked and indeterminate (does not occur for emphasized)

aramos-adobe:
> Could you look at the following?
> 
> 1. [feat(checkbox/group): s2 token migration, control spacing, subtle color changes #3531 (comment)](https://github.com/adobe/spectrum-css/pull/3531#discussion_r1964116979) I've been following up with this thread but am not seeing any changes or replies yet
> 2. [feat(checkbox/group): s2 token migration, control spacing, subtle color changes #3531 (comment)](https://github.com/adobe/spectrum-css/pull/3531#discussion_r1990103048) A change is needed here
> 3. Did you look at this comment for high contrast? "The focus outline border radius doesn't seem to match and the gap doesn't look like the same as the regular focus indicator. I'm assuming that this should look just like the non high contrast focus indicator, just with a different (system) color." Of note; this looks and behaves even stranger on keyboard focus if I look at with a light theme in high contrast: Screenshot 2025-03-11 at 5 05 39 PM
> 4. The colors are looking better. One thing I'm seeing in high contrast is a change of the background color on hover for default readonly checked and indeterminate (does not occur for emphasized)

@jawinn I missed these comments!! I'll address them ASAP

components/checkbox/index.css
```diff
@@ -272,50 +441,34 @@
...
 			&::before {
-				border-color: var(--highcontrast-checkbox-color-hover, var(--mod-checkbox-invalid-color-hover, var(--spectrum-checkbox-invalid-color-hover)));
+				border-color: var(--highcontrast-checkbox-highlight-color-focus, var(--mod-checkbox-emphasized-color-focus, var(--spectrum-checkbox-emphasized-color-focus)));
 			}
 		}
 	}
 
-	/* Selected and Indeterminate Hover States */
-	&:hover {
+	/* TODO: Because this selector was moved to the default variant's styles, this selector block can be deleted when it is safe to make changes to selectors. */
```
jawinn, line 464: This comment was recently added to main and came in with the rebase. Can you see whether we can remove the selector and/or the comment? I'd check against main to make sure about what the comment was referring to is the same.

jawinn (changes requested):
I'm seeing a couple things left:

Disabled styling looks to have changed. There's also a stylelint warning on the PR that might be related?
Screenshot 2025-03-24 at 4 53 17 PM

On default focus, the checked state isn't showing the darker blue background. Same for the read-only checked. The darker blue does show on focus for indeterminate; see the difference in blues here:
Screenshot 2025-03-24 at 4 56 37 PM

The high contrast focus indicator definitely looks better than before, thanks!
When inspecting I happened to notice some styles in the forced colors media query that are overridden and don't appear to be used. Can this line be removed? It also seems to have an extra "0" on the box-shadow as compared to the other styles: https://github.com/adobe/spectrum-css/blob/a737bf0598df96ed1153319882a7eacdf7a2d2a4/components/checkbox/index.css#L514-L516

jawinn (changes requested):
It looks like the comments from the previous review are all addressed. I'm seeing this one new difference now. The unchecked emphasized focus shouldn't be showing a blue border:
Screenshot 2025-04-01 at 2 30 05 PM

components/checkbox/index.css
```diff
@@ -188,16 +358,15 @@
...
 		}
 
-		/* Overrides disabled styles */
-		.spectrum-Checkbox-input,
-		.spectrum-Checkbox-input:checked {
-			&:disabled + .spectrum-Checkbox-box::before {
-				border-color: var(--highcontrast-checkbox-color-default, var(--mod-checkbox-control-selected-color-default, var(--spectrum-checkbox-control-selected-color-default)));
-				background-color: var(--spectrum-checkbox-checkmark-color);
+		&:hover,
+		.is-focus-visible {
```
jawinn, line 362:
```suggestion
		&:focus-visible {
```
We shouldn't be using the `.is-focus-visible` class in our CSS, as this class is only added within our Storybook by the build process so we can test pseudo classes on the VRTs. That's done by the `postcss-pseudo-classes` plugin.

This is leading to an incorrect color appearing on the Docs page when I tab focus on the emphasized indeterminate checkbox, which is different than what appears on the testing preview grid (that looks correct):

Screenshot 2025-04-03 at 2 20 19 PM

components/checkbox/index.css
```diff
@@ -271,51 +433,37 @@
...
+	&:not(.is-disabled).spectrum-Checkbox-input:checked:focus-visible + .spectrum-Checkbox-box {
 		&::before {
-			border-color: var(--highcontrast-checkbox-highlight-color-focus, var(--mod-checkbox-emphasized-color-focus, var(--spectrum-checkbox-emphasized-color-focus)));
+			border-color: var(--highcontrast-checkbox-highlight-color-default, var(--mod-checkbox-emphasized-color-down, var(--spectrum-checkbox-emphasized-color-down)));
 		}
 	}
 
-	/* Selected Invalid Focus State */
-	&.is-invalid .spectrum-Checkbox-input:focus-visible:checked + .spectrum-Checkbox-box {
+	&.is-indeterminate .spectrum-Checkbox-input.is-focus-visible + .spectrum-Checkbox-box {
```
jawinn, line 442:
```suggestion
	&.is-indeterminate .spectrum-Checkbox-input:focus-visible + .spectrum-Checkbox-box {
```
This one references the Storybook-only testing class as well.

jawinn (changes requested):

components/checkbox/dist/metadata.json
```diff
@@ -2,59 +2,49 @@
...
-    ".spectrum-Checkbox--emphasized.is-invalid.is-indeterminate:hover .spectrum-Checkbox-box:before",
-    ".spectrum-Checkbox--emphasized.is-invalid.is-indeterminate:hover .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box:before",
-    ".spectrum-Checkbox--emphasized.is-invalid:active .spectrum-Checkbox-box:before",
-    ".spectrum-Checkbox--emphasized.is-invalid:active .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box:before",
-    ".spectrum-Checkbox--emphasized.is-invalid:hover .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box:before",
+    ".spectrum-Checkbox--emphasized.is-indeterminate:not(.is-disabled):focus-visible .spectrum-Checkbox-box:before",
+    ".spectrum-Checkbox--emphasized.is-indeterminate:not(.is-disabled):focus-visible .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box:before",
+    ".spectrum-Checkbox--emphasized.is-readOnly:hover .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box:before",
+    ".spectrum-Checkbox--emphasized.spectrum-Checkbox-input:not(.is-disabled):checked:focus-visible + .spectrum-Checkbox-box:before",
```
jawinn, line 33: Correct me if I'm wrong, but it looks like `.is-disabled` may be a similar situation to `.is-focus-visible` mentioned in the previous comments. I'm not seeing any mention of this class in the original metadata.json, or in the docs about implementations needing to apply it. I think any use of this class needs to be adjusted to use the existing `:disabled` selector with the input.

jawinn (changes requested):

components/checkbox/dist/metadata.json
jawinn, line 1: This is almost there.

**Some pseudo selectors that need to apply to `.spectrum-Checkbox-input`**
For the updated uses of `:focus-visible` and `:disabled`, those need to be applied to the input and not the root class. Reviewing the metadata.json I saw some selectors starting like these where they are being applied to the root classes:
`.spectrum-Checkbox.is-readOnly:focus-visible`...
`.spectrum-Checkbox--emphasized:not(:disabled):focus-visible`...
`.spectrum-Checkbox--emphasized.is-indeterminate:not(:disabled):focus-visible`...

**One leftover disabled class**
There is also still one instance of `.is-disabled`.

**Down state when selecting with keyboard**
I noticed this when testing the pseudo states and considering our `:active` selectors. The down state is not shown when tabbing to a checkbox and selecting it with space, which I'm guessing it should? I think this could be resolved by moving the `:active` to also apply to the input and not the root element---that'd need some testing.

jawinn (changes requested):

components/checkbox/index.css
```diff
@@ -158,26 +169,185 @@
...
+	 * The input is visually hidden (with opacity), and remains accessible using assistive tech.
+	 * It is still sized and positioned where it would naturally be present so that touch users can find it with haptics.
+	 */
+	.spectrum-Checkbox-input {
+		color: var(--mod-checkbox-control-color-default, var(--spectrum-checkbox-control-color-default));
+
+		/* Change the font styles in all browsers for input. */
+		font-family: inherit;
+		font-size: 100%;
+		line-height: 1.15;
```
rise-erpelding, line 212: Really curious about these styles and why exactly we need them, the comments don't really explain it for me really well and/or I don't have a great grasp of how this would behave with various types of assistive technology. But I know these were here before so this may not be an answerable question.

components/checkbox/stories/checkbox.test.js
rise-erpelding, line 1: The added test coverage is so welcome!!! ❤️ 

I'm wondering if we can pare this back just a little though to exclude some of the repetitive cases? For instance, the default section has "hovered" and then also the "Hovered" section has "hovered", are they both necessary?

I'm personally a fan of using [`include`/`ignore`](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/textfield/stories/textfield.test.js#L49) to hone in on what needs to be tested and take out what doesn't, but that definitely hasn't been a common/popular approach among the team and may not even be necessary here.

I also see that we have a "focused" section and that shows a focus ring around checkboxes that don't normally get one, like disabled checkboxes, which might be misleading.

Last thing on testing: should we test active/down state? Is there a good reason not to that I don't know of?

aramos-adobe: @rise-erpelding I tried adding active state in the testing environment but couldn't figure out the best place to put it 😅

rise-erpelding: How would you feel about adjusting these a bit more, that might make it easier to incorporate active as well?

The original intention of `testData` vs. `stateData`, I think, was for `testData` to cover all the variants we might have in a component (like quiet, emphasized, static black, etc.), and `stateData` to cover all the states that those variants might experience (hovered, focused, invalid, disabled, read-only, combinations would also be fine, like how you have invalid+checked+hovered). Then, by default, each item in the `testData` array gets combined with each item in the `stateData` array to make the testing grid, but we have options (like `withStates`, or `ignore`, or `include`) to further refine and include or exclude cases from a particular variant or state, if for instance, 

Using that as a model to guide tests here, we could narrow down what we have in `testData` to Default, Emphasized, and Truncation, and move Hovered and Focused down to states. We could also add Active here, once we set the template and Storybook up to support it!

components/checkbox/index.css
```diff
@@ -153,7 +153,8 @@
 		}
 	}
 
-	&:active {
+	&:active,
+	.is-active {
```
rise-erpelding, line 157: Was this `.is-active` added for Storybook? The active state may work just fine in Storybook without it, when the dist/index.css is built I think PostCSS processes the state and adds the `.is-active` class 🙂 

Although it's fair to say that we have plenty of instances of `.is-focused` in CSS files and I'm not sure whether we need those for a different reason.

rise-erpelding (commented):
Leaving a question/comment about `:active`, but the testing grid looks great!

Since I was already in selector land and thinking about the `:active` pseudo-class, I have one other question/thought that I couldn't put in the code, some of @jawinn's previous feedback was:

> I noticed this when testing the pseudo states and considering our :active selectors. The down state is not shown when tabbing to a checkbox and selecting it with space, which I'm guessing it should? I think this could be resolved by moving the :active to also apply to the input and not the root element---that'd need some testing.

That seems valid and hasn't been done yet, would something like this work? (Down around L170, inside of `.spectrum-Checkbox` but after `:hover` and `:after`--this isn't new code in this PR so GitHub won't let me comment inline.)

```css
	&:not(.is-readOnly) {
		.spectrum-Checkbox-input:active:not(:disabled) + .spectrum-Checkbox-box,
		&:active .spectrum-Checkbox-input:not(:disabled) + .spectrum-Checkbox-box {
			transform: perspective(var(--spectrum-component-size-minimum-perspective-down)) translateZ(var(--spectrum-component-size-difference-down));
		}
	}
```

rise-erpelding (approved):
All of my feedback items have been addressed and this is looking pretty good to me! Thanks for all your work on this!!

components/checkbox/index.css
```diff
@@ -152,395 +167,251 @@
...
-.spectrum-Checkbox--emphasized {
-	/* Checked and Indeterminate Default States */
-	.spectrum-Checkbox-input:checked + .spectrum-Checkbox-box,
-	&.is-indeterminate .spectrum-Checkbox-input:checked + .spectrum-Checkbox-box,
-	&.is-indeterminate .spectrum-Checkbox-box {
 		&::before {
-			border-color: var(--highcontrast-checkbox-highlight-color-default, var(--mod-checkbox-emphasized-color-default, var(--spectrum-checkbox-emphasized-color-default)));
-		}
-	}
+			forced-color-adjust: none;
```
jawinn, line 287: Ideally all uses of `forced-color-adjust` should be organized within the forced-colors media query. I think that makes it a lot easier to reason about the forced-color styles.

jawinn (changes requested):
Really close on this one. Thanks for addressing my last set of feedback.

1. On the PR link I'm seeing one background color difference on the Docs page for invalid indeterminate. It's not showing the darker red on focus like the regular checked: 
Screenshot 2025-04-17 at 3 03 55 PM

2. In the testing preview grid refactoring, it looks like we lost some states that were important to test, such as the one that would catch number 1 above:
Focused + checked, focused + indeterminate, Focused + checked + invalid, focused + indeterminate + invalid, Hover + checked, hovered + indeterminate, hover + checked + invalid, hover + indeterminate + invalid, the same for read-only states, hover on disabled states

rise-erpelding:
> Really close on this one. Thanks for addressing my last set of feedback.
> 
> 1. On the PR link I'm seeing one background color difference on the Docs page for invalid indeterminate. It's not showing the darker red on focus like the regular checked:
>    Screenshot 2025-04-17 at 3 03 55 PM
> 2. In the testing preview grid refactoring, it looks like we lost some states that were important to test, such as the one that would catch number 1 above:
>    Focused + checked, focused + indeterminate, Focused + checked + invalid, focused + indeterminate + invalid, Hover + checked, hovered + indeterminate, hover + checked + invalid, hover + indeterminate + invalid, the same for read-only states, hover on disabled states

😱 This is my bad. Yeah, I think the way we have the testing grid set up is a little too simple. Since some of these states (disabled, read only, indeterminate) need to cross with each other, we'd have to set up test cases with them as `testData` as well as being `stateData`, perhaps?

What if `testData` looked something like:
```js
testData: [
		{
			testHeading: "Default",
		},
		{
			testHeading: "Emphasized",
			isEmphasized: true,
		},
		{
			testHeading: "Invalid with states",
			isInvalid: true,
		},
		{
			testHeading: "Checked with states",
			isChecked: true,
		},
		{
			testHeading: "Indeterminate with states",
			isIndeterminate: true,
		},
		{
			testHeading: "Disabled with states",
			isDisabled: true,
		},
		{
			testHeading: "Read-only with states",
			isReadOnly: true,
		},
		{
			testHeading: "Truncation",
			withStates: false,
			label: "Checkbox with an extraordinarily long label. Please don't do this but if you did, it should wrap text when it gets longer than the container that houses the checkbox with the unacceptably long label",
			customStyles: {
				"max-width": "200px",
			}
		}
	],
```

and `stateData`:
```js
stateData: [
		{
			testHeading: "Checked",
			isChecked: true,
			ignore: ["Checked with states"],
		},
		{
			testHeading: "Indeterminate",
			isIndeterminate: true,
			ignore: ["Indeterminate with states"],
		},
		{
			testHeading: "Hovered",
			isHovered: true,
		},
		{
			testHeading: "Active",
			isActive: true,
		},
		{
			testHeading: "Focused",
			isFocused: true,
		},
		{
			testHeading: "Invalid",
			isInvalid: true,
			ignore: ["Invalid with states"],
		},
		{
			testHeading: "Invalid, checked",
			isInvalid: true,
			isChecked: true,
			ignore: ["Invalid with states"],
		},
		{
			testHeading: "Invalid, indeterminate",
			isInvalid: true,
			isIndeterminate: true,
			ignore: ["Invalid with states", "Indeterminate with states"],
		},
		{
			testHeading: "Disabled",
			isDisabled: true,
			ignore: ["Disabled with states"],
		},
		{
			testHeading: "Disabled, checked",
			isDisabled: true,
			isChecked: true,
			ignore: ["Disabled with states"],
		},
		{
			testHeading: "Disabled, indeterminate",
			isDisabled: true,
			isIndeterminate: true,
			ignore: ["Disabled with states", "Indeterminate with states"],
		},
		{
			testHeading: "Read-only",
			isReadOnly: true,
			ignore: ["Read-only with states"],
		},
		{
			testHeading: "Read-only, checked",
			isReadOnly: true,
			isChecked: true,
			ignore: ["Read-only with states", "Checked with states"],
		},
		{
			testHeading: "Read-only, indeterminate",
			isReadOnly: true,
			isIndeterminate: true,
			ignore: ["Read-only with states", "Indeterminate with states"],
		},
		{
			testHeading: "Read-only, checked, disabled",
			isReadOnly: true,
			isChecked: true,
			isDisabled: true,
			ignore: ["Read-only with states", "Checked with states", "Disabled with states"],
		},
	],
```

I still don't think that's perfect, since it leaves out Focused + checked + invalid, focused + indeterminate + invalid, and possibly others. Open to other ideas for sure but that might be a good first step if nothing else.

aramos-adobe:
@rise-erpelding + @jawinn, just pushed some changes to the test grid before seeing Rise's suggestion. What do y'all think of this? I'm going to try Rise's suggestion as well to see what it looks like 😁

`testData: [
		{
			testHeading: "Default",
		},
		{
			testHeading: "Emphasized",
			isEmphasized: true,
		},
		{
			testHeading: "Truncation",
			withStates: false,
			label: "Checkbox with an extraordinarily long label. Please don't do this but if you did, it should wrap text when it gets longer than the container that houses the checkbox with the unacceptably long label",
			customStyles: {
				"max-width": "200px",
			}
		}
	],`
	
	`stateData: [
		{
			testHeading: "Hovered",
			isHovered: true,
		},
		{
			testHeading: "Checked",
			isChecked: true,
		},
		{
			testHeading: "Indeterminate",
			isIndeterminate: true,
		},
		{
			testHeading: "Hovered + checked",
			isHovered: true,
			isChecked: true,
		},
		{
			testHeading: "Hovered + indeterminate",
			isHovered: true,
			isChecked: true,
		},
		{
			testHeading: "Hovered + indeterminate + invalid",
			isHovered: true,
			isChecked: true,
			isIndeterminate: true,
		},
		{
			testHeading: "Active",
			isActive: true,
		},
		{
			testHeading: "Focused",
			isFocused: true,
		},
		{
			testHeading: "Focused + checked",
			isFocused: true,
			isChecked: true,
		},
		{
			testHeading: "Focused + indeterminate",
			isFocused: true,
			isIndeterminate: true,
		},
		{
			testHeading: "Invalid",
			isInvalid: true,
		},
		{
			testHeading: "Invalid, checked",
			isInvalid: true,
			isChecked: true,
		},
		{
			testHeading: "Invalid, indeterminate",
			isInvalid: true,
			isIndeterminate: true,
		},
		{
			testHeading: "Hovered + checked + invalid",
			isHovered: true,
			isChecked: true,
			isInvalid: true,
		},
		{
			testHeading: "Focused + checked + invalid",
			isFocused: true,
			isChecked: true,
			isInvalid: true,
		},
		{
			testHeading: "Focused + indeterminate + invalid",
			isFocused: true,
			isIndeterminate: true,
			isInvalid: true,
		},
		{
			testHeading: "Read-only, checked + invalid",
			isReadOnly: true,
			isChecked: true,
			isInvalid: true,
		},
		{
			testHeading: "Read-only, indeterminate + invalid",
			isReadOnly: true,
			isIndeterminate: true,
			isInvalid: true,
		},
		{
			testHeading: "Disabled",
			isDisabled: true,
		},
		{
			testHeading: "Disabled, checked",
			isDisabled: true,
			isChecked: true,
		},
		{
			testHeading: "Disabled, indeterminate",
			isDisabled: true,
			isIndeterminate: true,
		},
		{
			testHeading: "Read-only",
			isReadOnly: true,
		},
		{
			testHeading: "Read-only, checked",
			isReadOnly: true,
			isChecked: true,
		},
		{
			testHeading: "Read-only, checked + hovered",
			isReadOnly: true,
			isChecked: true,
			isHovered: true,
		},
		{
			testHeading: "Read-only, indeterminate",
			isReadOnly: true,
			isIndeterminate: true,
		},
		{
			testHeading: "Read-only, indeterminate + hovered",
			isReadOnly: true,
			isIndeterminate: true,
			isHovered: true,
		},
		{
			testHeading: "Read-only, checked, disabled",
			isReadOnly: true,
			isChecked: true,
			isDisabled: true,
		},
	],`
	
	

https://github.com/user-attachments/assets/b84e781c-5a30-47ab-a69d-f180c852fd45

aramos-adobe:
Just added "hovered and invalid"

jawinn (approved):

PR #3538, opened 2025-02-05
chore(search): add missing size-specific CSS properties

rubencarvalho:
## Description

Added missing size-specific CSS custom properties to the Search component in both Spectrum 2 and Express themes. These properties are required by downstream consumers (like Spectrum Web Components) to properly render the component across different sizes.

Even though this change is not reflected in this project (no VRT changes), it will cascade and fix undefined variables in Spectrum Web Components

You can see the properties not defined on the initial runs of https://github.com/adobe/spectrum-web-components/pull/5085, resulting in a broken component design. This change makes it so the CSS properties are defined in the SWC's `#textfield` element which will then render it correctly (tested manually by overriding the local CSS in SWC).

### Validation steps
  1. Check out the [SWC PR](https://github.com/adobe/spectrum-web-components/pull/5085).
  2. Manually copy the produced CSS artifacts from this PR to respective SWC's `node_modules`
  3. Run `yarn start` and verify the http://localhost:8080/storybook/index.html?path=/story/search-sizes--xl renders correctly.

### Regression testing

Validate:

## Screenshots

Broken component in SWC:
XL

## To-do list


rubencarvalho:
I am getting a Typography regression in VRT which seems flaky, not sure if this is known / how to proceed here! 

https://www.chromatic.com/build?appId=64762974a45b8bc5ca1705a2&number=4191

components/search/themes/express.css
rise-erpelding, line 1: The addition of `--spectrum-search-block-size`, `--spectrum-search-icon-size`, and `--spectrum-search-text-to-icon` seem a little repetitive here in Express, on the CSS side of things as far as I can tell, those custom properties look like they're being inherited in Express from where they're set in `spectrum-two.css`, would things still work as you'd want them in SWC if we removed the extra code you added here in `express.css`?

rise-erpelding (commented):
Looks good - just leaving a quick question here for you!

components/search/themes/express.css
Rajdeepc: `express.css` inherits `spectrum.css` if I am not wrong! Can you please check @rise-erpelding

Rajdeepc: Yes this make sense. Thanks

rise-erpelding (approved):

jawinn (commented):
I'm wondering why this is needed in SWC, when those same values for S/L/XL are already being set in the `index.css`, which should apply to all themes?
http://github.com/adobe/spectrum-css/blob/main/components/search/index.css#L93-L109

If an increase in specificity is needed, would that be done in the linked section above instead?

Do you have a testing link where the screenshotted issue is present? When I go to the Storybook at the linked PR 5085, I do not see that issue appearing.

rubencarvalho:
Sorry for the late reply, I'm only now picking this back up. Yes - I can confirm this is unnecessary 😃  I am therefore closing it. Thank you for everyone chiming in!

PR #3541, opened 2025-02-06
chore: add peer dep alignment to constraints

castastrophe:
## Description

Updated the constraints `yarn.config.cjs`:
- Update the local package alignment to run on workspaces whether they have peerDependencies or not
- Include peerDependencies, devDependencies, and dependencies in the package validation
- Add logic to update any existing peerDependencies to reflect the local major version
- Add a check that all local peerDependencies are listed as optional (reduce errors in downstream projects when packages installed or served separately)
- Add logic to maintain consistent versions for external dependencies across workspaces

### Validation steps

- [x] `yarn constraints`, expect this to return warnings about misaligned shared dependencies, recommending update to the latest version required in the project

```
─ spectrum-css-monorepo@workspace:.
│  └─ ⚙ Invalid field devDependencies["postcss"]; expected '^8.5.0', found '^8.4.49'
```

- [x] `yarn constraints`, expect peerDependencies to match local major version, all local peerDependencies should be listed as optional in the peerDependenciesMeta

```
├─ @spectrum-css/tag@workspace:components/tag
│  ├─ ⚙ Invalid field peerDependencies["@spectrum-css/avatar"]; expected '>=9.0.0 <10.0.0', found '>=7'
│  ├─ ⚙ Invalid field peerDependencies["@spectrum-css/clearbutton"]; expected '>=7.0.0 <8.0.0', found '>=6'
│  ├─ ⚙ Invalid field peerDependencies["@spectrum-css/icon"]; expected '>=9.0.0 <10.0.0', found '>=7'
│  ├─ ⚙ Invalid field peerDependencies["@spectrum-css/tokens"]; expected '>=16.0.0 <17.0.0', found '>=14 || >=15'
│  └─ ⚙ Invalid field peerDependenciesMeta; expected { '@spectrum-css/avatar': { optional: true }, '@spectrum-css/clearbutton': { optional: true }, '@spectrum-css/icon': { optional: true }, '@spectrum-css/tokens': { optional: true } }, found { '@spectrum-css/avatar': { optional: true }, '@spectrum-css/icon': { optional: true } }
```

## To-do list


components/alertbanner/package.json
```diff
@@ -25,27 +25,35 @@
...
+			"optional": true
+		},
+		"@spectrum-css/divider": {
+			"optional": true
+		},
 		"@spectrum-css/icon": {
 			"optional": true
+		},
+		"@spectrum-css/tokens": {
+			"optional": true
```
jawinn, line 48:
```suggestion
		"@spectrum-css/tokens": {
			"optional": true
```
Tokens shouldn't be appearing as optional should it? (appearing on multiple components)

components/picker/package.json
```diff
@@ -25,33 +25,41 @@
...
 		},
+		"@spectrum-css/icon": {
+			"optional": true
+		},
+		"@spectrum-css/menu": {
+			"optional": true
+		},
+		"@spectrum-css/popover": {
+			"optional": true
+		},
```
jawinn, line 47: These aren't optional for a functioning Picker. I'm seeing a lot of components with peer dependencies added as optional that aren't really optional, or is there some other criteria here for what is optional that I'm not aware of?

components/accordion/package.json
```diff
@@ -25,16 +25,12 @@
 	},
 	"main": "dist/index.css",
 	"peerDependencies": {
-		"@spectrum-css/icon": ">=7",
-		"@spectrum-css/tokens": ">=14 || >=15"
+		"@spectrum-css/icon": ">=9.0.0 <10.0.0",
+		"@spectrum-css/tokens": ">=16.0.0 <17.0.0"
```
jawinn, line 29: For a lot of these semver range changes appearing with both `>=` and `<`, is there any reason not to use the caret range instead? For example, this line could just be `"@spectrum-css/tokens": "^16.0.0`?

components/alertbanner/package.json
castastrophe: So it's not optional to appear in the DOM with the component but it is optional whether downstream consumers choose to install it as a dependency in the same package. It's problematic for us to list it as a required peerDep because SWC ships those styles from different packages. I definitely don't want to loose the peerDependency information but thought this would be a good way to reduce errors in the install that just end up ignored.

components/accordion/package.json
castastrophe: They are equivalent, that's true. I just like the readability of this syntax in peerDeps; it's subtly more readable. Any reason to use caret instead?

PR #3566, opened 2025-02-19
feat(popover): overflow bug on Safari

aramos-adobe:
## Description

Drop shadow effect disappears when overflow is applied to the popover component on Safari browser. The error occurs on the contextual menu and the rewrite popover.
Reported by Bianca on this [PR](https://git.corp.adobe.com/Horizon/hz/pull/113285)

[Jira ticket](https://jira.corp.adobe.com/browse/CSS-1095)

## Solution
Adding `translateZ(0)` or `translate3d(0,0,0)` prevents the clipping of filter drop-shadow on the popover. The element needs to be accelerated to Safari's GPU layer. If anyone applies `overflow` to popover at minimum there's a fallback in one of these two that'll prevent any weird changes. 

[Source](https://blog.teamtreehouse.com/increase-your-sites-performance-with-hardware-accelerated-css)
[Source 2](https://stackoverflow.com/questions/18529381/what-does-webkit-transform-translate3d0-0-0-exactly-do-apply-to-body#:~:text=Using%20%2Dwebkit%2Dtransform%3A%20translate3d,to%20force%20the%20hardware%20acceleration.)

- [x] Safari browser
- [x] Chrome
- [x] `overflow: auto` added to popover without tip and `overflow: visible` to CSS `--withTip`

### Validation steps

- [x] Validate popover [@castastrophe, @cdransf]
- [x] Validate action bar [@castastrophe, @cdransf]
- [x] Validate action menu [@castastrophe, @cdransf]
- [x] Validate coachmark [@castastrophe, @cdransf]
- [x] Validate combobox [@castastrophe, @cdransf]
- [x] Validate context help [@castastrophe, @cdransf]
- [x] Validate menu [@castastrophe, @cdransf]

### Regression testing

None visual changes have been made. 

## To-do list

### DIFF
--- a/components/popover/index.css
+++ b/components/popover/index.css
@@ -72,9 +72,12 @@
 
 	background-color: var(--mod-popover-background-color, var(--spectrum-popover-background-color));
 	filter: var(--mod-popover-filter, var(--spectrum-popover-filter));
+	overflow: auto;
 
 	/* has tip triangle */
 	&.spectrum-Popover--withTip {
+		overflow: visible;
+
 		.spectrum-Popover-tip {
 			/* triangle polygon */
 			.spectrum-Popover-tip-triangle {
@@ -115,7 +118,7 @@
 
 	/* popover animates upward ⬆ */
 	&.is-open {
-		transform: translateY(calc(-1 * var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance))));
+		transform: translateY(calc(-1 * var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance)))) translate3d(0, 0, 0);
 	}
 }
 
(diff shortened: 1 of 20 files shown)


components/popover/index.css
```diff
@@ -115,7 +118,7 @@
 
 	/* popover animates upward ⬆ */
 	&.is-open {
-		transform: translateY(calc(-1 * var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance))));
+		transform: translateY(calc(-1 * var(--mod-popover-animation-distance, var(--spectrum-popover-animation-distance)))) translate3d(0, 0, 0);
```
castastrophe, line 121: Can you add a comment to this line about what the translate function is doing here so it doesn't get accidentally removed in future? If it's specific to the Safari bug, can you include a note about that too?

castastrophe (approved):

castastrophe:
@aramos-adobe Is there any test case that can be added to cover this customer's use-case?

castastrophe (changes requested):
Something appears to be up with nested popovers now in the snapshots. I think the overflow settings need to be adjusted.

components/popover/index.css
aramos-adobe: Good call out! I'll add that to the next commit

castastrophe (approved):

cdransf (approved):
Ran through all validation cases and didn't encounter any issues. ✨

PR #3570, opened 2025-02-20
feat(link): new typography tokens, focus color and design, WHCM focus state

aramos-adobe:
# Link S2 migration

### Color

`spectrum-link-focus-indicator-color`
`spectrum-link-focus-indicator-thickness`
`spectrum-link-focus-indicator-gap`
`spectrum-link-corner-radius`

### Typography

`spectrum-link-line-height`
`spectrum-link-line-height-cjk-100`
`spectrum-link-font-size`
`spectrum-link-font-style`
`spectrum-link-font-weight`
`spectrum-link-text-underline-thickness`
`spectrum-link-text-underline-gap`

### Regression testing

Validate:

## Screenshots

Screenshot 2025-02-20 at 2 16 30 PM

Screenshot 2025-02-20 at 2 16 50 PM

## To-do list


rise-erpelding (changes requested):
This was such a nicer-sized migration to review compared to some of the ones we've had lately! 😸 

The primary, secondary, and static colors are on point and looking good, same for standalone default and standalone quiet! The main recommendation that I'd make here is to ensure that we can see the inline variant with Storybook controls, tests, and with a story of its own on the Docs page. It looks like the only real change there is maybe just the font weight? It seems like there's a clear differentiation between standalone and inline links in S2, so it's probably worth adding something more in the documentation. For instance, we probably want to demonstrate it without filler text, and even if we don't name the variant standalone in CSS (IMO it's fine if the default is standalone), we should probably note somewhere on the docs that standalone is the default.

The [changelog](https://www.figma.com/design/M9P2s2IDhGWZuNyvJtXrx3/S2-foundations-and-components-changelog?node-id=1751-3221&t=LeXgNYB8ZYce39yp-4) says that inline quiet links were removed for accessibility reasons, that's probably a good thing we can integrate into our updated docs page for link. Ideally, we can add this into the JSDoc comments somewhere, and hopefully we'll have Storybook picking those up and putting them back into story descriptions again soon!

Last thing: I'm also wondering if it's possible to stack variants within a story? Like put primary and primary (quiet) in a single frame? I don't know if that was a deliberate decision made when we did the docs site to storybook migration for Link or whether the migration happened before we started doing that, but it might be nice to make the docs page a little bit neater. Up to you.

components/link/stories/link.stories.js
```diff
@@ -89,105 +98,81 @@ Default.tags = ["!autodocs"];
...
+Primary.args = {};
 Primary.tags = ["!dev"];
 Primary.parameters = {
 	chromatic: { disableSnapshot: true },
 };
+Primary.storyName = "Standalone, primary";
 
 /**
  * The secondary variant is the same gray color as the paragraph text. Its subdued appearance is optimal for when the primary variant is too overwhelming, such as in blocks of text with several references linked throughout.
+ * Link is using `spectrum-Link--secondary`
```
rise-erpelding, line 113: Can we make this statement and the ones you have above the static colors stories in the comments into sentences? And maybe throw in the word "class" somewhere like "This link is using the `spectrum-Link--secondary` class." or "Link is using the class `spectrum-Link--secondary`." I think it makes it a little clearer and it's also searchable in case someone is looking for information on classes on the page.

.changeset/tasty-eagles-draw.md
rise-erpelding, line 1: This will eventually get added automatically to the changelog in the component, so we'll want to think about how this reads in that context, could we make this a little more explanatory/documentation-y?

Some examples I saw that I liked were for [Breadcrumbs](https://github.com/adobe/spectrum-css/blob/spectrum-two/.changeset/fresh-crabs-deliver.md) and [Tooltip](https://github.com/adobe/spectrum-css/blob/spectrum-two/.changeset/light-years-speak.md), although certainly there are others in that `.changeset/` directory.

rise-erpelding (changes requested):
I think this is almost there!

There are a few, mostly documentation things that I think could make this even more polished - some of these might also be in the individual comments, but I'm duplicating them here because occasionally those comments get lost:

* Docs page:
    * We could say something about the standalone link being the default variant vs. the inline variant.
    * Rename the Multiline story to include something about focus state in the title (I don't feel super strongly about this one since you talk about it in the JSDoc comment, so I'll leave it to your discretion).
    * Consider changing the wrapping element for  TemplateWithFillerText and MultilineText to a `<p>` instead of `<div>`
    * Mention the classes being used for variants in sentences.
* Storybook controls - Remove the ability to turn on both inline and quiet controls.
* CSS - We should change the font-family to the font-stack.
* Changeset - Add a bit more about what's changing from S1 to S2 so it reads nicely in the changelog.

rise-erpelding (approved):

.changeset/tasty-eagles-draw.md
```diff
@@ -0,0 +1,30 @@
+---
+"@spectrum-css/link": major
+---
+
+# Link S2 migration
```
castastrophe, line 5:
```suggestion
### Link S2 migration
```

Super minor knit but the only first-level heading in the changelogs is the top-level one. All the change sets should be h3s or lower

.changeset/tasty-eagles-draw.md
```diff
@@ -0,0 +1,30 @@
...
+
+The link component is now updated with S2 specs, colors, and typography.There is a new inline variant which uses the `.spectrum-Link--inline` class and the default variant is the standalone variant.
```
castastrophe, line 7:
```suggestion
The link component is updated with S2 specifications, colors, and typography. There is a new inline variant which uses the `.spectrum-Link--inline` modifier class and the default is the standalone variant.
```

A couple minor wording updates, can you say more about the "standalone" variant. Will most readers of the changelog know what that means or is there a way to describe it that might be more clear? Or can we link to the design documentation?

.changeset/tasty-eagles-draw.md
```diff
@@ -0,0 +1,30 @@
...
+
+**Notes**
```
castastrophe, line 9:
```suggestion
#### Note
```

The linter typically suggests using headings over bold text for a stronger navigability when rendered.

.changeset/tasty-eagles-draw.md
```diff
@@ -0,0 +1,30 @@
...
+- Quiet styling does not apply to the inline variant for accessiblity reasons.
```
castastrophe, line 10:
```suggestion
- Quiet styling does not apply to the inline variant for accessiblity reasons.
```

Any chance we can expand on those reasons a bit just to make sure we have them documented here for the changelog?

.changeset/tasty-eagles-draw.md
```diff
@@ -0,0 +1,30 @@
...
+- `--mod-spectrum-link-font-weight` can be used to adjust inline variant to match surrounding text. Apply font weights to this mod `inherit` does not apply.
+
+## New tokens
```
castastrophe, line 13: Are these new custom properties on the component-level or are these tokens from the token system that we're now using in the component styles?

components/link/index.css
```diff
@@ -36,29 +43,45 @@
...
+	font-size: var(--spectrum-link-font-size);
+	font-style: var(--spectrum-link-font-style);
+	line-height: var(--spectrum-link-line-height);
+	text-decoration-thickness: var(--spectrum-link-text-underline-thickness);
+	text-underline-offset: var(--spectrum-link-text-underline-gap);
 
 	&:hover {
-		/* @deprecated --mod-link-text-color-primary-hover in favor of --mod-link-text-color-hover */
-		--mod-link-text-color: var(--mod-link-text-color-hover, var(--mod-link-text-color-primary-hover, var(--spectrum-accent-content-color-hover)));
+		--mod-link-text-color: var(--mod-link-text-color-hover, var(--spectrum-accent-content-color-hover));
```
castastrophe, line 56: This is really cool and a nice efficiency! Love the consistency you implemented it with throughout as well.

components/link/stories/link.stories.js
```diff
@@ -54,8 +54,25 @@ export default {
...
-		isQuiet,
+		isQuiet: {
+			name: "Quiet styling",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Component",
+			},
+			if: { arg: "isInline", neq: true },
+		},
```
castastrophe, line 75:
```suggestion
		isQuiet: {
			...isQuiet,
			if: { arg: "isInline", neq: true },
		},
```

Just a note, you can keep it aligned with the shared isQuiet variants by spreading the original object and adding your if statement.

components/link/stories/link.test.js
```diff
@@ -48,6 +48,15 @@ export const LinkGroup = Variants({
 			staticColor: "white",
 			isQuiet: true,
 		},
+		{
```
castastrophe, line 51: I like that you added these at the bottom because they're new but for S2, I'm wondering what you think of listing the inline variant right under the default because it's a more central style setting than say, the static color tests.

components/link/stories/template.js
```diff
@@ -43,9 +45,52 @@ export const Template = ({
...
+		withBorder: false,
+		direction: "column",
+		heading: "Quiet",
+		content: html`
+			${Template({
+				...args,
+				text: "Learn more",
+				isQuiet: true,
+			}, context)}
+		`,
```
castastrophe, line 93:
```suggestion
		content: Template({
			...args,
			text: "Learn more",
			isQuiet: true,
		}, context)
```

This can simplify it a bit for you since the Template returns an html object already - you don't have to wrap it again if you don't want to.

components/link/stories/template.js
```diff
@@ -43,9 +45,52 @@ export const Template = ({
...
+		direction: "column",
+		heading: "Default",
+		content: html`
+			${Template({
+				...args,
+				text: "Learn more",
+			}, context)}
+		`,
+	}, context)}
+	${Container({
```
castastrophe, line 83:
```suggestion
	content: [
		Container({
			withBorder: false,
			direction: "column",
			heading: "Default",
			content: Template({
				...args,
				text: "Learn more",
			}, context),
		}, context),
		Container({
```

One cool feature of all the content fields is that they support adding elements in an array and then joins them for you when rendering. If it simplifies things a bit, you can call them this way to leverage that and might not even need to add the html dependency for this.

castastrophe (approved):
Looks great! I left some questions, some suggestions, some ideas, but all of this could go in as-is if needed!

.changeset/tasty-eagles-draw.md
aramos-adobe: These are tokens from the token system now being used in the component.

PR #3606, opened 2025-03-05
feat(actionbutton): use closer to s2 colors in spectrum-two theme

jawinn:
#### SWC-497 action button color updates

This includes requested colors update for action button, aligning the colors closer to the S2 design, post-foundations.

In the spectrum-two theme:
- Removes the border
- Changes the default background colors to match s2 specs
- Updates the background colors used for static black and static white

Screenshot 2025-03-06 at 3 54 14 PM
Screenshot 2025-03-06 at 3 54 29 PM
Screenshot 2025-03-06 at 3 54 23 PM

#### High contrast fix

This also includes a bug fix for selected + disabled in forced-colors / high contrast mode. The disabled colors were not being used for this combination in both themes, and previously had a mix of disabled border and selected background in S1.
- S2 and S1, Before:
Screenshot 2025-03-06 at 10 33 22 AM
- S1 + Quiet, Before
Screenshot 2025-03-06 at 3 36 45 PM
- S2 and S1, After:
Screenshot 2025-03-06 at 3 18 32 PM

### Validation steps

- [x] Requested updates in Jira issue SWC-497 have been made. spectrum-two version of action button in storybook does not have borders, has the darker grey background, and the updated static color variants look correct by default and on hover and focus. [@rise-erpelding]
- [x] No regressions in spectrum 1 theme VRT. There is one expected change in the spectrum 1 VRTs for selected disabled in forced-colors / high contrast. [@rise-erpelding]
- [x] VRT diff shows expected changes for S2 theme [@rise-erpelding]

### Regression testing

Validate:

## To-do list


rise-erpelding (approved):
These look great!

I had a few small, non-blocking comments though:
- Corner radius: no changes were made here and this follows the S2 spec beautifully. However the ticket says "Ensure that border radius is pointing to / inheriting the token value 'corner-radius-100'", is it possible (or necessary) to confirm that we do in fact want the corner radius to scale at different sizes?
- I raised the question about the down/active color being different than hover and focus, even though the [S2 spec](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-token-specs?node-id=13572-5954&m=dev) has them as all the same color, but this color difference could be due to the fact that we're not implementing the perspective-based down state in Foundations, and I see you've already [asked](https://adobedesign.slack.com/archives/G01P2PHJMC6/p1741619165730499) about whether this is accurate. No questions from me on this, I'm sure we'll end up with the colors that everyone's happy with here.
- I only checked the Chrome emulation for forced colors, and noticed for dark mode that the quiet selected variants have a border. There's still good contrast here so I'm not all that concerned, but it does feel a little unintentional.
Screenshot 2025-03-10 at 8 43 18 AM

components/actionbutton/index.css
```diff
@@ -45,6 +38,13 @@
 				forced-color-adjust: none;
 			}
 		}
+
+		&:disabled,
```
castastrophe, line 42: This is a great move to ensure disabled styles always take precedence over other states. I can see that this is why the disabled+selected variant is now showing disabled colors in high contrast mode. Great job here.

castastrophe (approved):
These updates are great; thanks for tackling this. I can see in the baselines how much closer the S2 visuals are now. I approved the VRT changes and I like the update to the disabled+selected in the forced colors mode. I think this is good to merge.

castastrophe:
@jawinn Oh and this is ready for you to add a changeset once you feel confident in the changes.

jawinn:
I've updated this with the revised down state color after confirmation from design that the Jira requested value was incorrect. It's now matching the S2 spec.

rise-erpelding (approved):
Took one more quick glance at it, this is still looking great! Thanks for making those updates!

PR #3611, opened 2025-03-10
fix(slider): offset variant border radius bug fix

aramos-adobe:
## Slider: offset variant track fix

The border radius styles were not being applied to the second instance of the `spectrum-Slider-track` when the offset variant is activated. The reason for this bug is because when the `offset` is selected, the template structure changes as `spectrum-Slider-fill` gets added to the slider. 

Adding a sibling combinator `&~.spectrum-Slider-track` to `spectrum-Slider-track` when offset is activated resolved the issue.

### Regression testing

Validate:

## Screenshots
Screenshot 2025-03-10 at 2 47 15 PM

## To-do list

### DIFF
--- a/.changeset/rude-jokes-judge.md
+++ b/.changeset/rude-jokes-judge.md
@@ -4,36 +4,6 @@
 
 # Slider: offset variant track fix
 
-The border radius styles were not being applied to the second instance of the `spectrum-Slider-track` when the offset variant is activated. The reason for this bug is because when the `offset` is selected, the template structure changes as `spectrum-Slider-fill` gets added to the slider. This likely negated the initial declaration of the border radius styles written below:
-
-```css
-.spectrum-Slider-track {
-	&:first-of-type {
-		&::before {
-			border-start-start-radius: var(
-				--mod-slider-track-corner-radius,
-				var(--spectrum-slider-track-corner-radius)
-			);
-			border-end-start-radius: var(
-				--mod-slider-track-corner-radius,
-				var(--spectrum-slider-track-corner-radius)
-			);
-		}
-	}
-
-	&:last-of-type {
-		&::before {
-			border-start-end-radius: var(
-				--mod-slider-track-corner-radius,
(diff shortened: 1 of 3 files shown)


cdransf (approved):
Nice! Looks good when pulling down the branch and stepping through the styles in the inspector. ✨

marissahuysentruyt (changes requested):
Glad this was a nice quick fix! Looks like this just needs a rebase now.

This is really nitpicky, but I think when this merges, we should change this commit from `feat` to `fix`- it is a bug/regression fix, as opposed to something new for slider. In my mind, `fix` also corresponds to the minor version bump a bit better than `feat` does.

rise-erpelding (changes requested):
I'm so glad we're looking into this one! There are a few things here that I noticed that I thought were really odd, which is holding back my approval:
* First, this isn't a problem in [SWC](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/slider--fill-start), and after taking a quick glance, I really can't figure out why. The `&:last-of-type::before` selector seems to be working great there. I would think that any changes we make here would carry over fine to SWC and not cause problems, but I'm wondering if it's worth more of an investigation on how what appears to be the same CSS is working differently between the two environments.
* Second, and maybe more importantly, this is throwing some regressions into the range variant, here it has a border-radius on the right side of the middle segment:
image
In comparison, this is how it looked before, with the border-radius only being applied to the end segments:
image
~~* Lastly, can you see the color diffs between the top and bottom image? For some reason the build preview and your local branch are running slightly different colors (honestly higher contrast in your branch tbh) than main... but that seems unexpected unless I missed something? That might end up being completely unrelated to your PR specifically, but noting it just in case?~~ Ok it looks like there were some slider contrast things added here, so that doesn't seem like it's relevant here!

marissahuysentruyt (approved):
Thanks for looking into this one!

rise-erpelding (approved):

PR #3614, opened 2025-03-11
fix(slider): corrects contrast bug caused by template default arg

cdransf:
## Description

Corrects the contrast issue observed between the fill element and slider track that caused the fill element to be difficult to see.

- Adjusts colors per design feedback.
- Removes default control/template value that caused contrast issue in dark modew.

### Validation steps

1. Open Storybook URL
2. Navigate to slider
3. Change theme from light to dark
4. Enable testing preview
5. Scroll to filled offset and verify offset and track have sufficient contrast

## Screenshots

image

## To-do list

### DIFF
--- a/components/slider/stories/template.js
+++ b/components/slider/stories/template.js
@@ -21,7 +21,7 @@ export const Template = ({
 	values = [],
 	variant,
 	labelPosition,
-	fillColor = "rgb(213, 213, 213)",
+	fillColor,
 	showTicks = false,
 	showTickLabels = false,
 	isDisabled = false,
@@ -127,8 +127,8 @@ export const Template = ({
 			})}
 			id=${ifDefined(id)}
 			style=${styleMap({
+				"--spectrum-slider-track-color": fillColor ? fillColor : undefined,
 				"inline-size": "240px",
-				["--spectrum-slider-track-color"]: fillColor,
 				...customStyles,
 			})}
 			role=${ifDefined(values.length > 1 ? "group" : undefined)}


components/slider/stories/template.js
```diff
@@ -18,7 +18,7 @@ export const Template = ({
 	values = [],
 	variant,
 	labelPosition,
-	fillColor = "rgb(213, 213, 213)",
+	fillColor = "",
```
marissahuysentruyt, line 21: As I was looking at this, I remembered that [swatch](https://github.com/adobe/spectrum-css/blob/8637d396dcf15b0b3310b65419512a25b65a2181/components/swatch/stories/swatch.stories.js#L96) also has a similar default color situation. The color is defined in the args in the stories file, so [it is different](https://github.com/adobe/spectrum-css/blob/8637d396dcf15b0b3310b65419512a25b65a2181/components/swatch/stories/template.js#L27), but would any of that approach apply here at all? 

I do like the idea of the ternary in the styleMap like in swatch, but what do you think? 

`"--spectrum-slider-track-color": fillColor ? fillColor : undefined,`

marissahuysentruyt (commented):
Thanks for digging into the gray values- glad we got them corrected! 

Looking [at prod](https://opensource.adobe.com/spectrum-css/?path=/story/components-slider--default&args=variant:offset&globals=color:dark), is this PR supposed to address `main`? I think I see the same issue, so if we fix it on `main`, could we just cherry pick over to `spectrum-two`? 

Screenshot 2025-03-11 at 11 49 54 AM

components/slider/stories/template.js
cdransf: Sort of! I left the default value out, but I did remove the empty string assignment and change the syntax to a ternary. Adding the default color arg reintroduces the contrast issue, unfortunately. :(

.changeset/few-glasses-bathe.md
```diff
@@ -0,0 +1,10 @@
+---
+"@spectrum-css/slider": minor
+---
+
+# Slider contrast fix
+
+Corrects the contrast issue observed between the fill element and slider track that caused the fill element to be difficult to see.
+
+- Adjusts colors per design feedback.
+- Removes default control/template value that caused contrast issue in dark modew.
```
marissahuysentruyt, line 10: There's a typo in `mode` here. 

(it almost looks like an abbreviation for mountain dew 😆)

cdransf: Oh no! (And I don't even drink soda.) All fixed! ✨

cdransf:
> Thanks for digging into the gray values- glad we got them corrected!
> 
> Looking [at prod](https://opensource.adobe.com/spectrum-css/?path=/story/components-slider--default&args=variant:offset&globals=color:dark), is this PR supposed to address `main`? I think I see the same issue, so if we fix it on `main`, could we just cherry pick over to `spectrum-two`?

Absolutely! I lifted everything over to main and squashed it down so we can cherry-pick in a follow up PR. ✨

.changeset/few-glasses-bathe.md
```diff
@@ -0,0 +1,10 @@
+---
+"@spectrum-css/slider": minor
+---
+
+# Slider contrast fix
+
+Corrects the contrast issue observed between the fill element and slider track that caused the fill element to be difficult to see.
+
+- Adjusts colors per design feedback.
```
marissahuysentruyt, line 9: I think you had these changes at one point, but I don't see them in the file diff now. Do we no longer want or need to adjust the gray-200's to gray-300s?

cdransf: Good call! We don't, strictly speaking — that's what they will be in S2, but the actual fix was removing that arg. I'll drop that right now.

castastrophe (approved):

PR #3627, opened 2025-03-17
feat(rating): S2 migration

cdransf:
## Description

`CSS-1141`

This migrates the `rating` component to S2. Custom properties have been remapped and added per the design specification.

##### Additions

The medium component variation is the default and a t-shirt size medium variation has been added.

A tooltip may be displayed to a user indicating whether interacting with the component will clear or edit the rating. A control and documentation example have been added.

A rating may have a partially filled star. The width and fill of this star is controlled by adding `.is-partial` to the parent `span` with classes of `spectrum-Rating-icon` and `is-selected` and then setting `--mod-rating-icon-fill` to the necessary fill percentage (e.g. `50%`).

##### New mods

`--mod-rating-width`
`--mod-rating-height`
`--mod-rating-bottom-to-content-area`
`--mod-rating-edge-to-content-area`
`--mod-rating-top-to-content-area`
`--mod-rating-top-to-content-area`
`--mod-rating-icon-fill`

##### Removed mods

`--mod-rating-icon-spacing-vertical-top`
`--mod-rating-icon-count`
`--mod-rating-indicator-border-radius`
`--mod-rating-indicator-height`

##### New custom properties

`--spectrum-rating-width`
`--spectrum-component-size-difference-down`
`--spectrum-component-size-minimum-perspective-down`
`--spectrum-corner-radius-medium-size-medium`
`--spectrum-corner-radius-medium-size-small`
`--spectrum-neutral-content-color-default`
`--spectrum-neutral-content-color-down`
`--spectrum-neutral-content-color-hover`
`--spectrum-spacing-75`
`--spectrum-workflow-icon-size-100`
`--spectrum-workflow-icon-size-75`

##### Removed custom properties

`--spectrum-rating-indicator-width`
`--spectrum-rating-icon-color-key-focus`
`--spectrum-rating-icon-spacing-vertical-top`
`--spectrum-rating-focus-indicator-gap`
`--spectrum-rating-indicator-height`
`--spectrum-rating-indicator-border-radius`
`--spectrum-accent-content-color-key-focus`
`--spectrum-border-width-200`
`--spectrum-component-top-to-workflow-icon-100`
`--spectrum-corner-radius-100`
`--spectrum-neutral-subdued-content-color-default`
`--spectrum-neutral-subdued-content-color-down`
`--spectrum-neutral-subdued-content-color-hover`
`--spectrum-neutral-subdued-content-color-key-focus`
`--spectrum-workflow-icon-size-100`

### Validation steps

1. Open the Storybook URL for the PR.
2. Navigate to the rating component.
3. Confirm that the new sizing stories render in the docs view.
4. Verify that all new tokens have been adopted and properly implemented.
  - Small rating size is the new default, medium is available with a new t-shirt size class.
  - `isFocused` has been changed to `isKeyboardFocused`
  - Sizing story control behaves as expected
  - No regressions are visible

## To-do list


components/rating/stories/rating.test.js
```diff
@@ -18,12 +18,24 @@ export const RatingGroup = Variants({
...
+			withTooltip: true,
+		},
+		{
+			testHeading: "Small size",
+			size: "s",
+		},
+		{
+			testHeading: "Medium size",
+			size: "m",
+		},
```
castastrophe, line 39: We don't need to test sizing separately - this is handled automatically by the testing grid.

components/rating/index.css
```diff
@@ -56,17 +44,19 @@
 	position: relative;
 	display: inline-flex;
 	flex: 0 0 auto;
+	justify-content: space-between;
```
castastrophe, line 47: Nice addition!

components/rating/index.css
```diff
@@ -56,17 +44,19 @@
...
 
 	gap: var(--mod-rating-icon-spacing, var(--spectrum-rating-icon-spacing));
-
-	inline-size: calc((var(--mod-rating-icon-width, var(--spectrum-rating-icon-width)) + var(--mod-rating-icon-spacing, var(--spectrum-rating-icon-spacing))) * var(--mod-rating-icon-count, var(--spectrum-rating-icon-count)));
 	border-radius: var(--mod-rating-border-radius, var(--spectrum-rating-border-radius));
 
-	padding-inline: calc((var(--mod-rating-icon-spacing, var(--spectrum-rating-icon-spacing)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap))) - 1.5px);
+	padding-inline: var(--mod-rating-icon-spacing, var(--spectrum-rating-icon-spacing));
```
castastrophe, line 52: Love that you simplified this so much

components/rating/index.css
```diff
@@ -116,6 +100,17 @@
 	}
 }
 
+.spectrum-Rating--sizeM {
```
castastrophe, line 103: Just checking, small is the default here right? Should we push back a bit on design for that? From an API perspective, it seems confusing for users. Could we alter this to s -> m and m -> l?

components/rating/stories/template.js
```diff
@@ -88,3 +100,21 @@ export const Template = ({
 		</div>
 	`;
 };
+
+export const RatingSizeGroup = (args, context) => Container({
```
castastrophe, line 104: The Sizes utility does the same thing but scales for you if the arg options change. Check out the Accordion story for an example:

```
export const Sizing = (args, context) => Sizes({
	Template,
	withBorder: false,
	withHeading: false,
	items: content,
	...args,
}, context);
```

tokens/custom/large-vars.css
```diff
@@ -99,8 +99,6 @@
 	--spectrum-pagination-textfield-width: 60px;
 	--spectrum-pagination-item-inline-spacing: 6px;
 
-	--spectrum-rating-icon-spacing: var(--spectrum-spacing-100);
```
castastrophe, line 102: Love seeing these removed 🕺

castastrophe (changes requested):
Left a few suggestions for optimizing the tests and storybook. Otherwise approved but I do think we should push back on design designating a default size other than medium if medium exists. It's really confusing for developers and it's not the norm in our system right now.

components/rating/index.css
cdransf: Digging back through the specs and it looks like. I misread that! I'll get it swapped out. 🙈

components/rating/index.css
```diff
@@ -12,35 +12,23 @@
...
 	/* Icon */
 	--spectrum-rating-icon-height: var(--spectrum-workflow-icon-size-100);
 	--spectrum-rating-icon-width: var(--spectrum-workflow-icon-size-100);
-	--spectrum-rating-icon-color-default: var(--spectrum-neutral-subdued-content-color-default);
-	--spectrum-rating-icon-color-hover: var(--spectrum-neutral-subdued-content-color-hover);
-	--spectrum-rating-icon-color-down: var(--spectrum-neutral-subdued-content-color-down);
+	--spectrum-rating-icon-color-default: var(--spectrum-neutral-content-color-default);
+	--spectrum-rating-icon-color-hover: var(--spectrum-neutral-content-color-hover);
+	--spectrum-rating-icon-color-down: var(--spectrum-neutral-content-color-down);
 	--spectrum-rating-icon-color-key-focus: var(--spectrum-neutral-subdued-content-color-key-focus);
```
rise-erpelding, line 23: I didn't see a `key-focus` color in the spec, should we have one?

rise-erpelding (changes requested):
Oh, it's so cool to see this migration with the cute pudgy stars! It looks really nice, and the tooltip is now part of it, awesome! The basics are there but there are some trickier changes that I think need to be made and also some questions I have:

Changes needed:
* The spacing/padding tokens for the top/bottom/edges, for instance `rating-top-to-content-area-small`, aren't being used, and the padding is being set with what looks like `spacing-75`/4px all around. The small tokens (`rating-top-to-content-area-small`, `rating-bottom-to-content-area-small`, `rating-edge-to-content-area-small`) all alias to `spacing-75`, but we should probably be using the rating-specific tokens in case the value ever changes. The medium tokens (`rating-top-to-content-area-medium`, `rating-bottom-to-content-area-medium`, `rating-edge-to-content-area-medium`) all evaluate to 6px, so we'd be wanting to see that padding for the medium size rating as well.
    * Speaking of the padding, I'm wondering how this plays with the tooltip. In the spec, it looks like the tooltip gets placed 4px (spacing-75) from the edge of the star, does that mean for medium where the edge to content padding is supposed to be 6px that it would encroach on the padding there? I'm wondering if that means we might need to use a calc to get this in the right spot?
* I also didn't see the height tokens (`rating-height-small`/30px, `rating-height-medium`/40px) being used, and the computed height I see for small is 24px and for medium 28px, we'll probably want to adjust that to match the spec.
* Down state - looks like this gets the new S2 transform that will make the star a little smaller when active, I can't say I have a great understanding on how this is implemented but we do have a Storybook story for it with some notes in the top Foundations section and the way it's done in #3531 for checkbox might be similar, since the star and checkbox feel similarly sized.

Questions/maybe changes needed:
* I see that the Figma [changelog](https://www.figma.com/design/M9P2s2IDhGWZuNyvJtXrx3/S2-foundations-and-components-changelog?node-id=26-2&p=f&t=LQCML4EVl40PfmkY-0) has sections for "Must have", "Should have", and "Could have", everything is covered in this migration, except the "Could have" that talks about being able to show fractional/half-star. That feels like a CSS change that needs to be shown for S2, is that something that needs to be done here? Is there a separate ticket for it somewhere else? I'd be fine to implement that feature later.
* The `focus-indicator-gap` isn't used; why? There's no space between the padding on `.spectrum-Rating` and the blue focus outline. This is in the S1 spec too and was never implemented there, I'm wondering if there's a specific reason or whether it was just something that was overlooked since there's already some padding between the stars and the focus outline. It feels like maybe it was just overlooked? I think this could probably be implemented by using the same `::after` technique we use in action button, button, and other components.
* There's no key-focus icon color in the spec, but we have one in the CSS, should we? (Noted this one inline in the code as well).

components/rating/index.css
cdransf: This was a holdover from the rating behavior changes I tried to carry over and then reverted, so I've gone ahead and removed it.

I've applied the spacing, padding and left the tooltip as is. I've been leaving the tooltip spacing untouched and just included it as is.

I added the height tokens from the spec and applied those.

For the partial star display, I _did_ have an example, but it was specific to 50% and the story display was pretty unreliable (even with the updated behavior with the JS carried over). I'm leaning in the direction of letting consumers control this as they could dynamically change the masked percentage without us providing something fixed in CSS.

I applied the S2 down state and updated the changelog too. ✨

components/rating/index.css
```diff
@@ -12,61 +12,65 @@
...
 
 .spectrum-Rating {
-	--spectrum-rating-icon-color-default: var(--spectrum-neutral-subdued-content-color-default);
-	--spectrum-rating-icon-color-hover: var(--spectrum-neutral-subdued-content-color-hover);
-	--spectrum-rating-icon-color-down: var(--spectrum-neutral-subdued-content-color-down);
-	--spectrum-rating-icon-color-key-focus: var(--spectrum-neutral-subdued-content-color-key-focus);
-
-	--spectrum-rating-indicator-border-radius: 2px;
+	--spectrum-rating-height: var(--spectrum-rating-height-medium);
+	--spectrum-rating-width: var(--spectrum-rating-width-medium);
```
rise-erpelding, line 16: These height/width tokens should account for the sum of all the padding, gaps, and icons in the rating, but we're getting larger dimensions than spec'd (medium is 128px wide for instance), would it fix that issue to add a `box-sizing: border-box;`? The blue boxes that the stars are sitting in here also seem a little tall, I'm unsure if that'd be a separate issue or if it's related.
image

components/rating/index.css
```diff
@@ -116,6 +114,21 @@
 	}
 }
 
+.spectrum-Rating--sizeS {
+	--mod-rating-height: var(--spectrum-rating-height-medium);
```
rise-erpelding, line 118: Do we want this `--mod-rating-height` to be small here?

rise-erpelding (changes requested):
The down state and padding differentiation for small and medium look great!

I think the rest of it is mostly spacing things and the focus indicator. We'll want that focus gap to not take over our existing padding, ideally, and then I'm wondering about the space that the component takes up since it's larger than spec'd with the padding, I'm also thinking that the tooltip seems a little high? But maybe that would be fixed once we adjust spacing.

Last thing: I don't think we need to have the controls doing partial stars, but if you were able to make it work for say, a docs page story, that would be great and really useful! Is that possible to do with the work that you previously did on this?

components/rating/index.css
```diff
@@ -12,61 +12,65 @@
...
 	/* Colors */
 	/* selected + emphasized */
 	--spectrum-rating-emphasized-icon-color-default: var(--spectrum-accent-content-color-default);
 	--spectrum-rating-emphasized-icon-color-hover: var(--spectrum-accent-content-color-hover);
 	--spectrum-rating-emphasized-icon-color-down: var(--spectrum-accent-content-color-down);
-	--spectrum-rating-emphasized-icon-color-key-focus: var(--spectrum-accent-content-color-key-focus);
 
 	/* Disabled */
 	--spectrum-rating-icon-color-disabled: var(--spectrum-disabled-content-color);
 	--spectrum-rating-icon-count: var(--spectrum-rating-icon-count);
```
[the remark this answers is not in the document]
cdransf: We don't need it! Good catch — GitHub is pretty good at catching those unused tokens but nothing changed close enough to it. 👀

components/rating/index.css
cdransf: We do! That's all fixed. 🙈

components/rating/index.css
```diff
@@ -12,61 +12,65 @@
...
+
+	&:not(.is-readOnly) {
+		.spectrum-Rating-input:not(:disabled) ~ .spectrum-Rating-icon:active {
+			transform: perspective(var(--spectrum-component-size-minimum-perspective-down)) translateZ(var(--spectrum-component-size-difference-down));
+		}
+	}
+
 	&.is-focused {
+		padding-block: var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap));
+		padding-inline: var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap));
```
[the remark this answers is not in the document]
cdransf: Ahhh yeah — I had assumed we'd be swapping one for the other. I updated lines `72-73` to use calcs that combine the two. It preserves the mods, but they end up being pretty verbose.

We could add make the values in the calcs into their own custom properties, but then we're just trading verbosity for abstraction. 🤔

(Totally open to ideas on this. ✨)

components/rating/index.css
cdransf: It looks like `box-sizing: border-box;` did the trick 🎉

Screenshot 2025-04-02 at 4 40 47 PM

cdransf:
> Last thing: I don't think we need to have the controls doing partial stars, but if you were able to make it work for say, a docs page story, that would be great and really useful! Is that possible to do with the work that you previously did on this?

Hmm, let me see if I can get it to work using inline styles. I'd added a class to mask it at 50% and rebased it away, but an example that lives in the docs sounds good. 👍🏻

cdransf:
> > Last thing: I don't think we need to have the controls doing partial stars, but if you were able to make it work for say, a docs page story, that would be great and really useful! Is that possible to do with the work that you previously did on this?
> 
> Hmm, let me see if I can get it to work using inline styles. I'd added a class to mask it at 50% and rebased it away, but an example that lives in the docs sounds good. 👍🏻

@rise-erpelding alrighty — I added support for this and an example in the docs, but I'd love feedback: https://github.com/adobe/spectrum-css/pull/3627/commits/1d70d377c82c4dd378a529021e6fe5a7bf0463fb

I added an `.is-partial` class with a custom property that defaults to `100%` fill. There's a mod to adjust that percentage that consumers can use.

components/rating/index.css
```diff
@@ -12,61 +12,65 @@
...
+	--spectrum-rating-top-to-content-area: var(--spectrum-rating-top-to-content-area-medium);
+	--spectrum-rating-bottom-to-content-area: var(--spectrum-rating-bottom-to-content-area-medium);
+	--spectrum-rating-edge-to-content-area: var(--spectrum-rating-edge-to-content-area-medium);
 
 	/* Colors */
 	/* selected + emphasized */
 	--spectrum-rating-emphasized-icon-color-default: var(--spectrum-accent-content-color-default);
 	--spectrum-rating-emphasized-icon-color-hover: var(--spectrum-accent-content-color-hover);
 	--spectrum-rating-emphasized-icon-color-down: var(--spectrum-accent-content-color-down);
-	--spectrum-rating-emphasized-icon-color-key-focus: var(--spectrum-accent-content-color-key-focus);
```
aramos-adobe, line 42:
```suggestion
	--spectrum-rating-emphasized-icon-color-key-focus: var(--spectrum-accent-content-color-default);
```

I'd add the focus color token here. The emphasized focused state (LINE 222) should still appear as the accent color.

components/rating/index.css
```diff
@@ -12,61 +12,65 @@
...
+			transform: perspective(var(--spectrum-component-size-minimum-perspective-down)) translateZ(var(--spectrum-component-size-difference-down));
+		}
+	}
+
 	&.is-focused {
+		padding-block: calc(var(--mod-rating-top-to-content-area, var(--spectrum-rating-top-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap))) calc(var(--mod-rating-bottom-to-content-area, var(--spectrum-rating-bottom-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		padding-inline: calc(var(--mod-rating-edge-to-content-area, var(--spectrum-rating-edge-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
 		box-shadow: 0 0 0 var(--mod-rating-focus-indicator-thickness, var(--spectrum-rating-focus-indicator-thickness)) var(--mod-rating-focus-indicator-color, var(--spectrum-rating-focus-indicator-color));
 
 		.spectrum-Rating-icon {
```
aramos-adobe, line 76:
```css
.spectrum-Rating-icon {
			color: var(--highcontrast-rating-icon-color-key-focus, var(--mod-rating-icon-color-key-focus, var(--spectrum-rating-icon-color-key-focus)));
		}

		.spectrum-Rating-icon.is-selected {
			color: var(--highcontrast-rating-icon-color-key-focus, var(--mod-rating-icon-color-key-focus, var(--spectrum-rating-icon-color-key-focus)));
		}
```

Do we need these for the default focus state? The colors don't seem to change when in this state.

cdransf: Good catch! Just dropped those. ✨

components/rating/index.css
```diff
@@ -247,7 +259,6 @@
 
 @media (forced-colors: active) {
```
rise-erpelding, line 260: I did open up Assistiv Labs to check WHCM this time around. There is an S1 WHCM spec linked in the guidelines, and I'm wondering after looking at that and the guidelines if we've got this correct. The colors are off, and I would guess we would want the outline to be 2px (or focus-indicator width) instead of 1px.

That said, I wouldn't consider that a blocker here! I noted on the ticket for WHCM specs that we'd probably want some updated specs for Rating, and we could address that later. For now, I don't see anything low-contrast that seems like it would be an issue.

components/rating/stories/rating.stories.js
```diff
@@ -82,6 +97,20 @@ WithForcedColors.parameters = {
 
 // ********* DOCS ONLY ********* //
 
+/**
+ * Small and medium rating
+ */
```
rise-erpelding, line 102: Some of these comments are pretty self-explanatory given the title, we could maybe take some of them out. That's a nit though.

components/rating/stories/rating.stories.js
```diff
@@ -123,3 +161,44 @@ Disabled.args = {
...
+withTooltip.args = {
+	withTooltip: true,
+};
+withTooltip.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+/**
+ * A rating with a partially filled icon
+ */
```
rise-erpelding, line 192: I'd love to see some additional usage details here, particularly around what class and custom prop to use to get this implemented.

components/rating/stories/rating.test.js
```diff
@@ -25,5 +25,9 @@ export const RatingGroup = Variants({
 			testHeading: "Read-only",
 			isReadOnly: true,
 		},
+		{
+			testHeading: "With tooltip",
+			withTooltip: true,
+		},
```
rise-erpelding, line 31: Would it be possible to add some extra padding or something here? The tooltip is getting in the way of the test heading.

image

components/rating/stories/rating.test.js
rise-erpelding, line 1: It looks like the hover color of the stars is slightly different, while I'm skeptical that it's different enough that VRTs would even pick it up if it changed, the hover state might be a nice thing to add to our testing grid here.

components/rating/stories/rating.stories.js
```diff
@@ -123,3 +161,44 @@ Disabled.args = {
...
+withPartial.args = {
+	isPartial: true,
+	value: 5,
+	customStyles: {
+		"--mod-rating-icon-fill": "50%"
+	}
+};
+withPartial.parameters = {
+	chromatic: { disableSnapshot: true },
+};
```
rise-erpelding, line 204: 🤩 Oh this looks so good!

rise-erpelding (changes requested):
This is getting closer! Thanks for all your work on this!

.changeset/eighty-terms-lead.md
```diff
@@ -0,0 +1,72 @@
...
+"@spectrum-css/rating": major
+---
+
+#### S2 migration rating
+
+This migrates the `rating` component to S2. Custom properties have been remapped and added per the design specification.
+
+##### Additions
+
+The small component variation is now the default and a t-shirt size medium variation has been added.
```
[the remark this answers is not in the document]
cdransf: Good catch! I missed updating that when I swapped the defaults. ✨

components/rating/stories/rating.stories.js
```diff
@@ -123,3 +161,44 @@ Disabled.args = {
...
+ * A rating with a tooltip displayed.
+ */
+export const withTooltip = Template.bind({});
+withTooltip.tags = ["!dev"];
+withTooltip.args = {
+	withTooltip: true,
+};
+withTooltip.parameters = {
+	chromatic: { disableSnapshot: true },
+};
```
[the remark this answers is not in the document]
cdransf: Thank you! This is updated. ✨

components/rating/stories/rating.stories.js
cdransf: 🙌🏻

components/rating/stories/rating.stories.js
cdransf: Can do!

```
* - To implement a partially filled rating star, the star must have the `is-selected` class applied to its parent and the `is-partial` and `spectrum-Rating-starActive` classes applied. The fill can then be adjusted by setting `--mod-rating-icon-fill` to a valid percentage (e.g. 50%).
```

components/rating/stories/rating.stories.js
cdransf: Good call! I removed those comments and left the comment for the partial fill with usage notes. 🎶

.changeset/eighty-terms-lead.md
```diff
@@ -0,0 +1,72 @@
...
+`--mod-rating-icon-spacing`
+`--mod-rating-width`
+`--mod-rating-bottom-to-content-area`
+`--mod-rating-edge-to-content-area`
+`--mod-rating-height`
+`--mod-rating-top-to-content-area`
+`--mod-rating-width`
+`--mod-rating-top-to-content-area`
+`--mod-rating-width`
+`--mod-rating-icon-fill`
```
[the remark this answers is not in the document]
cdransf: I had `--mod-rating-width` in there 3 times too 😵‍💫

Cleaned that up!

components/rating/stories/rating.test.js
cdransf: Can do! 

```
{
  testHeading: "With tooltip",
  withTooltip: true,
  wrapperStyles: {
    "padding-top": "35px"
  }
},
```

image

components/rating/stories/rating.test.js
cdransf: Added in https://github.com/adobe/spectrum-css/pull/3627/commits/8a6092e1c443aa8bc16314a5ccfc4ce17a6c160f — let me know if you think this approach is too brittle or contrived. I added an `isHovered` arg and an `is-hovered` class that can be used in conjunction with it. I couldn't come up with a better way to force the hover state (as it depends on the `:hover` selector).

components/rating/index.css
```diff
@@ -12,70 +12,67 @@
...
 		}
+	}
 
-		.spectrum-Rating-icon.is-selected {
-			color: var(--highcontrast-rating-icon-color-key-focus, var(--mod-rating-icon-color-key-focus, var(--spectrum-rating-icon-color-key-focus)));
-		}
+	&.is-focused {
+		padding-block: calc(var(--mod-rating-top-to-content-area, var(--spectrum-rating-top-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap))) calc(var(--mod-rating-bottom-to-content-area, var(--spectrum-rating-bottom-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		padding-inline: calc(var(--mod-rating-edge-to-content-area, var(--spectrum-rating-edge-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		box-shadow: 0 0 0 var(--mod-rating-focus-indicator-thickness, var(--spectrum-rating-focus-indicator-thickness)) var(--mod-rating-focus-indicator-color, var(--spectrum-rating-focus-indicator-color));
```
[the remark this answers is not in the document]
rise-erpelding: I'm still seeing the stars hopping around here, here's a different view to try to explain what I think might be going on, initially `.spectrum-Rating` has padding on it like this, 6px on each side:

Screenshot 2025-04-16 at 4 59 22 PM

Then, when the focus state is applied, it has padding like this, 8px on each side, plus the inner dimensions of `.spectrum-Rating` have changed from above:

Screenshot 2025-04-16 at 4 59 52 PM

So I think that explains why, if I apply the focus state, I'll see the stars move slightly to the right, then back to the left if the focus state is removed.

I do also see the stars change position slightly if the focus state is applied and I hover over each of the individual stars. I'm not totally sure if that's related to this issue or if it's a separate issue though.

components/rating/stories/rating.stories.js
```diff
@@ -156,6 +156,20 @@ Focused.parameters = {
 	chromatic: { disableSnapshot: true },
 };
 
+/**
+ * For the emphasized variant, when a selected star is hovered, the color of the star is slightly darker.
+ */
```
rise-erpelding, line 161: So I mentioned this already on the .test.js file comment, but let's make sure we update this as well, since both variants do get a subtle color change on hover. Thanks for adding documentation for this!

components/rating/stories/rating.test.js
rise-erpelding: It does work, kudos to you on that! Adding comments in the css file, I think we can leverage PostCSS for this instead.

components/rating/index.css
```diff
@@ -227,10 +227,12 @@
 		}
 	}
 
-	&:hover .spectrum-Rating-icon {
+	&:hover .spectrum-Rating-icon,
+	&.is-hovered .spectrum-Rating-icon {
 		/* Make all stars selected when the component is hovered */
 		color: var(--highcontrast-rating-emphasized-icon-color-default, var(--mod-rating-emphasized-icon-color-default, var(--spectrum-rating-emphasized-icon-color-default)));
 
+		&.is-hovered,
```
rise-erpelding, line 235: This is something I definitely want to ask and clarify about with our team as a whole because I ran into this same issue with `:active` on a different PR review, but I believe that the PostCSS pseudo-classes [plugin](https://www.npmjs.com/package/postcss-pseudo-classes) takes care of this, you can see if you `yarn build` and look at the dist/index.css for Rating, it adds the companion classes automatically so we don't have to, in this case, the `postcss.config.js` has it set to prefix states with "is-", so you should see it adding `.is-hover` anywhere where `:hover` is used.

(Now, for me, the question is, why do we see this so often in CSS files if PostCSS takes care of it for us? But I think in this case we'll be ok reverting this back to what it was before.)

With that, we can also change the setup in `template.js` to set that `.is-hover` class based on the arg, and it _should_ work pretty much the same as what you have already, but will leverage PostCSS. I'll stick a comment down there also!

components/rating/stories/template.js
```diff
@@ -33,6 +34,7 @@ export const Template = ({
 				"is-disabled": isDisabled,
 				"is-readOnly": isReadOnly,
 				"is-focused": isFocused,
+				"is-hovered": isHovered,
```
rise-erpelding, line 37:
```suggestion
                                "is-hover": isHovered,
```

components/rating/stories/template.js
```diff
@@ -82,6 +84,7 @@ export const Template = ({
 						class=${classMap({
 							[`${rootClass}-icon`]: true,
 							"is-selected": idx <= value - 1,
+							"is-hovered": idx === 4 && isHovered,
```
rise-erpelding, line 87:
```suggestion
                                                        "is-hovered": idx === 4 && isHovered,
```

rise-erpelding (changes requested):
Thanks for the documentation additions here, particularly for Tooltip and Partial Star! The rating with tooltip instance in the testing grid also looks great!

I left a few thoughts on showing `:hover` in the testing grid, plus I still think the focus state needs a little bit of work, and we need to make sure our changeset is freshly updated.

.changeset/eighty-terms-lead.md
```diff
@@ -0,0 +1,72 @@
...
+`--spectrum-spacing-75`
+`--spectrum-workflow-icon-size-100`
+`--spectrum-workflow-icon-size-75`
+
+##### Removed custom properties
+
+`--spectrum-rating-indicator-width`
+`--spectrum-rating-icon-color-key-focus`
+`--spectrum-rating-icon-spacing-vertical-top`
+`--spectrum-rating-focus-indicator-gap`
```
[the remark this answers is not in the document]
cdransf: Updated! I compared the additions against `spectrum-two` to verify that they were added here and the deletions to the current state of the metadata in this branch to verify that they've been removed.

components/rating/index.css
cdransf: Oh awesome! Yay postcss! I reverted those classes and update the template and the testing grid looks good/in line with what I'd expect. ✨

castastrophe: Oh yes, so PostCSS adds the is-prefix to states only in the Storybook version of the assets. We _could_ ship it but in the past I think we discussed it and decided not to.

components/rating/index.css
```diff
@@ -70,8 +70,9 @@
 	}
 
 	&.is-focused {
-		padding-block: calc(var(--mod-rating-top-to-content-area, var(--spectrum-rating-top-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap))) calc(var(--mod-rating-bottom-to-content-area, var(--spectrum-rating-bottom-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
-		padding-inline: calc(var(--mod-rating-edge-to-content-area, var(--spectrum-rating-edge-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		padding-block-start: calc(var(--mod-rating-top-to-content-area, var(--spectrum-rating-top-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		padding-block-end: calc(var(--mod-rating-bottom-to-content-area, var(--spectrum-rating-bottom-to-content-area)) + var(--mod-rating-focus-indicator-gap, var(--spectrum-rating-focus-indicator-gap)));
+		padding-inline: var(--mod-rating-edge-to-content-area, var(--spectrum-rating-edge-to-content-area));
```
rise-erpelding, line 75: This does keep our stars from jumping around, but effectively now the `padding-inline` doesn't change between focused and not focused, and we're adding extra padding to account for that focus indicator gap only on the top and bottom and missing it on the sides.

Typically the components here don't use padding to create the focus indicator gap and I'm wondering if that's part of the reason why, if you search the repo for `focus-indicator-gap` a lot of the components will set a focus indicator with the `::after` pseudo-element and the gap is a margin on that `::after`.

It's probably not the only way though, you could definitely explore an approach using outline and outline-offset, it looks like that's what switch does!

rise-erpelding (commented):
Leaving a quick note on the focus state and the emphasized hovered story!

components/rating/stories/rating.stories.js
```diff
@@ -157,7 +157,7 @@ Focused.parameters = {
 };
 
 /**
- * For the emphasized variant, when a selected star is hovered, the color of the star is slightly darker.
+ * When a selected star is hovered, the color of the star is slightly darker.
  */
 export const EmphasizedHovered = Template.bind({});
```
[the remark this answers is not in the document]
cdransf: Thanks for catching this! I updated it to just `Hovered` — I think that makes more sense since it _does_ apply to both states, visibility concerns aside.

components/rating/index.css
cdransf: Just pushed up https://github.com/adobe/spectrum-css/pull/3627/commits/9b25e3a3a0c3f13374aa80454ee6c602308b575c to address this using `outline` + `outline-offset` like we discussed. ✨

rise-erpelding (approved):
⭐  ⭐  ⭐  ⭐  ⭐

PR #3628, opened 2025-03-17
feat(helptext): S2 migration

aramos-adobe:
## Help text S2 migration

The helptext S2 component has some updated spacing and typography tokens. The error message validation icon has also been updated.

### Validation steps

- Open the storybook for the helptext component
- Check the icon look correct

### Regression testing

Validate:

## To-do list


components/helptext/index.css
```diff
@@ -11,95 +11,75 @@
...
-		--spectrum-helptext-content-color-default: var(--spectrum-neutral-subdued-content-color-default);
-		--spectrum-helptext-icon-color-default: var(--spectrum-neutral-subdued-content-color-default);
-	}
+.spectrum-HelpText--sizeS {
+	--spectrum-helptext-min-height: var(--spectrum-component-height-75);
+	--spectrum-helptext-icon-size: var(--spectrum-workflow-icon-size-75);
+	--spectrum-helptext-font-size: var(--spectrum-font-size-75);
+	--spectrum-helptext-text-to-visual: var(--spectrum-text-to-visual-75);
+	--spectrum-helptext-top-to-workflow-icon: var(--spectrum-component-top-to-workflow-icon-75);
+	--spectrum-helptext-bottom-to-workflow-icon: var(--spectrum-helptext-top-to-workflow-icon);
```
rise-erpelding, line 47: Hmmm, we're repeating this `--spectrum-helptext-bottom-to-workflow-icon: var(--spectrum-helptext-top-to-workflow-icon);` a few times and I think we might not need to? It looks like it reference the top for every size without having to redeclare it.

But even so, I'm wondering if it might be clearer if we name `--spectrum-helptext-top-to-workflow-icon` something else entirely instead of using it and then defining `--spectrum-helptext-bottom-to-workflow-icon` to reference it? What about something like `--spectrum-helptext-workflow-icon-padding-block`, but keep the same mods where you have `padding-block-start` and `padding-block-end`? Or does that feel too inconsistent? I'm not opposed to either approach, just wondering which one feels the cleanest.

All of that said, I know it was like that before and I don't feel really strongly about keeping it the same way here, it just feels a little unusual to me.

components/helptext/index.css
```diff
@@ -112,36 +92,63 @@
...
-	&.is-disabled {
-		.spectrum-HelpText-text {
-			color: var(--highcontrast-helptext-content-color-default, var(--mod-helptext-content-color-default, var(--spectrum-helptext-content-color-default)));
-		}
+	.spectrum-HelpText-validationIcon {
+		color: var(--highcontrast-helptext-icon-color-default, var(--mod-helptext-icon-color-default, var(--spectrum-helptext-icon-color-default)));
+	}
+}
+
+.is-disabled {
```
rise-erpelding, line 129: Is it possible to keep the `&` here? So the selector is `&.is-disabled` (`.spectrum-HelpText.is-disabled`)? It looks like the metadata.json file is showing some new/added `.is-disabled` selectors not scoped to `.spectrum-HelpText` and it looks like this might be the cause of it.

components/helptext/index.css
```diff
@@ -11,95 +11,75 @@
...
+	--spectrum-helptext-icon-size: var(--spectrum-workflow-icon-size-100);
+	--spectrum-helptext-text-to-visual: var(--spectrum-text-to-visual-75);
+	--spectrum-helptext-top-to-workflow-icon: var(--spectrum-component-top-to-workflow-icon-100);
+	--spectrum-helptext-bottom-to-workflow-icon: var(--spectrum-helptext-top-to-workflow-icon);
+	--spectrum-helptext-top-to-text: var(--spectrum-component-top-to-text-75);
+	--spectrum-helptext-bottom-to-text: var(--spectrum-component-bottom-to-text-75);
+
+	/* Typpography */
+	--spectrum-helptext-font-size: var(--spectrum-font-size-75);
+	--spectrum-helptext-font-family: var(--spectrum-default-font-family);
```
rise-erpelding, line 29: Would using the `--spectrum-sans-font-family-stack` work here? I noticed using Assistiv Labs the font's not rendering (for Windows) and I'm wondering if that would help.

components/helptext/index.css
```diff
@@ -112,36 +92,63 @@
...
-		.spectrum-HelpText-text {
-			color: var(--highcontrast-helptext-content-color-default, var(--mod-helptext-content-color-default, var(--spectrum-helptext-content-color-default)));
-		}
+	.spectrum-HelpText-validationIcon {
+		color: var(--highcontrast-helptext-icon-color-default, var(--mod-helptext-icon-color-default, var(--spectrum-helptext-icon-color-default)));
+	}
+}
+
+.is-disabled {
+	--spectrum-helptext-content-color-default: var(--spectrum-disabled-content-color);
```
rise-erpelding, line 130: Figma is wrong here, right? Because there is no `disabled-content-color-default` token?

components/helptext/index.css
```diff
@@ -112,36 +92,63 @@
...
+		color: var(--highcontrast-helptext-icon-color-default, var(--mod-helptext-icon-color-default, var(--spectrum-helptext-icon-color-default)));
+	}
+}
 
-		.spectrum-HelpText-validationIcon {
-			color: var(--highcontrast-helptext-icon-color-default, var(--mod-helptext-icon-color-default, var(--spectrum-helptext-icon-color-default)));
-		}
+.spectrum-HelpText--negative {
+	--spectrum-helptext-content-color-default: var(--spectrum-negative-color-900);
+	--spectrum-helptext-icon-color-default: var(--spectrum-negative-color-900);
```
rise-erpelding, line 118: Is there a reason these are using `--spectrum-negative-color-900` instead of `--spectrum-negative-content-color-default`?

rise-erpelding (changes requested):
This one is almost there! There are some adjustments needed to the sizing, font, and text color mostly.

I don't think we need to do much to the tests/template/story file, since this is a pretty small migration, but it might be a nice addition to test multiline as well as single line text with the negative icon next to it.

Nice work!

components/helptext/index.css
marissahuysentruyt: @aramos-adobe I agree with Rise- I think we need `--spectrum-sans-font-family-stack` instead of `default-font-family`. Even in Chrome sometimes, this font isn't rendering properly: 

Screenshot 2025-03-26 at 7 02 32 AM

I think it's got something to do with `adobe clean` vs `adobe-clean`...I'd have to hunt for the thread!

components/helptext/index.css
```diff
@@ -11,95 +11,75 @@
...
-		--spectrum-helptext-icon-size: var(--spectrum-workflow-icon-size-200);
-		--spectrum-helptext-font-size: var(--spectrum-font-size-100);
-		--spectrum-helptext-text-to-visual: var(--spectrum-text-to-visual-100);
-		--spectrum-helptext-top-to-workflow-icon: var(--spectrum-help-text-top-to-workflow-icon-large);
-		--spectrum-helptext-bottom-to-workflow-icon: var(--spectrum-helptext-top-to-workflow-icon);
-		--spectrum-helptext-top-to-text: var(--spectrum-component-top-to-text-100);
-		--spectrum-helptext-bottom-to-text: var(--spectrum-component-bottom-to-text-100);
-	}
+	/* Spacing */
+	--spectrum-helptext-min-height: var(--spectrum-component-height-75);
```
marissahuysentruyt, line 19: I believe this medium help text needs the `component-height-100` token.

marissahuysentruyt: To extend this comment, I noticed the large and extra large need some corrected min height tokens as well:
```
`&.spectrum-HelpText--sizeL {
  --spectrum-helptext-min-height: var(--spectrum-component-height-200); // instead of 100
}

&.spectrum-HelpText--sizeXL {
  --spectrum-helptext-min-height: var(--spectrum-component-height-300); // instead of 200
}
```

components/helptext/index.css
marissahuysentruyt: I agree with this ☝️ - that's probably what I would do.

components/helptext/index.css
```diff
@@ -112,36 +92,63 @@
...
-			color: var(--highcontrast-helptext-content-color-default, var(--mod-helptext-content-color-default, var(--spectrum-helptext-content-color-default)));
-		}
+	.spectrum-HelpText-validationIcon {
+		color: var(--highcontrast-helptext-icon-color-default, var(--mod-helptext-icon-color-default, var(--spectrum-helptext-icon-color-default)));
+	}
+}
+
+.is-disabled {
+	--spectrum-helptext-content-color-default: var(--spectrum-disabled-content-color);
+	--spectrum-helptext-icon-color-default: var(--spectrum-disabled-content-color);
```
marissahuysentruyt, line 131: I think we've landed on the preference being to define and redefine all of the custom properties at the top of the file. So we might need to nest some small selector blocks within the first `.spectrum-HelpText` block with the rest of the tokens and custom properties. 

There's nothing wrong with the way this is, I just think we were trying to organize any custom property definitions at the top of the file instead.

marissahuysentruyt (changes requested):
It's close! The tokens need some cleaning up in a few places for the t-shirt sizes. I didn't want to litter your PR with similar comments over and over, but I think we need to double check the following for each t-shirt size: 

`font-size-*`
`component-height-*` (I did leave a comment or two regarding these tokens)
`text-to-visual-*`
`top-to-text-*
`bottom-to-text-*`

For each of our class sizes, the size/number for the tokens should all be the same according to the token specs, and they're a little mixed up right now, that's all. 

Thanks for getting this component back up and ready!

components/helptext/index.css
```diff
@@ -11,99 +11,87 @@
...
-	&:lang(ko) {
-		--mod-helptext-line-height: var(--mod-helptext-line-height-cjk, var(--spectrum-cjk-line-height-100));
-	}
-
 	.spectrum-HelpText-validationIcon {
 		margin-inline-end: var(--mod-helptext-text-to-visual, var(--spectrum-helptext-text-to-visual));
-		padding-block-start: var(--mod-helptext-top-to-workflow-icon, var(--spectrum-helptext-top-to-workflow-icon));
-		padding-block-end: var(--mod-helptext-bottom-to-workflow-icon, var(--spectrum-helptext-bottom-to-workflow-icon));
+		padding-block-start: var(--mod-helptext-edge-to-workflow-icon, var(--spectrum-helptext-edge-to-workflow-icon));
+		padding-block-end: var(--mod-helptext-edge-to-workflow-icon, var(--spectrum-helptext-edge-to-workflow-icon));
```
rise-erpelding, line 94: Personally, I think I like this better than keeping top and bottom separate variables! Pointing out though that we do lose the ability to modify different spacings for top vs. bottom. Given that design specs often point out when it's really important that consumers be able to modify something, and it's not here, I think using a single variable is ok, but pointing this out anyway in case someone has a stronger opinion about it.

rise-erpelding (approved):
Nothing blocking for me here! I see that the sizing has been adjusted, our fonts look good with the font stack, and otherwise the tokens appear to be used as we'd expect!

Left a few comments with small suggestions, but again, I don't consider those blocking. Nice work!

.changeset/wild-lemons-spend.md
```diff
@@ -0,0 +1,14 @@
+---
+"@spectrum-css/helptext": major
+---
+
+### Helptext S2 Migration
+
+The helptext S2 component has some updated spacing and typography tokens. The error message validation icon has also been updated.
```
marissahuysentruyt, line 7: Only because we just talked about detailed changesets with SWC, should we list out any of the new or renamed mods created during this work? We were doing that back in the fall, but I think we got away from it after we had been focused on foundations for so long! Looks like it's probably only 3 new mods: 

"--mod-helptext-font-family",
"--mod-helptext-font-style",
"--mod-helptext-font-weight",

and then one renamed: 
"--mod-helptext-bottom-to-workflow-icon" >> "--mod-helptext-edge-to-workflow-icon",

You'd know better than me on those! It's up to you, I won't let it block my approval.

marissahuysentruyt (approved):
amazing

PR #3641, opened 2025-03-25
fix(tooltip): restore negative and informative semantic variants

cdransf:
## Description

This restores the negative and informative tooltip semantic variant styles, story controls and test cases. Icons have been removed for all variants as they are not present in the specifications provided by design.

##### Restored mods

`--mod-tooltip-background-color-informative`
`--mod-tooltip-background-color-negative"`

### Validation steps

1. Open the Storybook link for this PR.
2. Navigate to the `Default` story for the tooltip component.
3. Select the informative and negative tooltip styles in controls and verify that they match the Figma design(s).

## Screenshots

Screenshot 2025-03-25 at 10 39 32 AM

## To-do list

### DIFF
--- a/components/tooltip/index.css
+++ b/components/tooltip/index.css
@@ -14,8 +14,6 @@
 @import "@spectrum-css/commons/overlay.css";
 
 .spectrum-Tooltip {
-	--spectrum-tooltip-backgound-color-default-neutral: var(--spectrum-neutral-subdued-background-color-default);
-
 	--spectrum-tooltip-animation-duration: var(--spectrum-animation-duration-100);
 	--spectrum-tooltip-animation-distance: var(--spectrum-spacing-75);
 
@@ -43,6 +41,10 @@
 	--spectrum-tooltip-content-color: var(--spectrum-gray-25);
 	--spectrum-tooltip-background-color-default: var(--spectrum-neutral-background-color-default);
 
+	/* colors */
+	--spectrum-tooltip-background-color-informative: var(--spectrum-informative-background-color-default);
+	--spectrum-tooltip-background-color-negative: var(--spectrum-negative-background-color-default);
+
 	/* tip */
 	--spectrum-tooltip-tip-block-size: var(--spectrum-tooltip-tip-height);
 
@@ -57,6 +59,7 @@
 
 	--highcontrast-tooltip-border-width: 0px;
 }
+
 .spectrum-Tooltip {
 	@extend %spectrum-overlay;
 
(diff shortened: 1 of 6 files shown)


components/tooltip/index.css
```diff
@@ -576,11 +596,13 @@
...
 		--highcontrast-tooltip-border-width: 1px;
-        
+
 		border: var(--highcontrast-tooltip-border-width) solid CanvasText;
 	}
 
 	.spectrum-Tooltip-tip {
 		--highcontrast-tooltip-background-color-default: CanvasText;
+		--highcontrast-tooltip-background-color-informative: CanvasText;
+		--highcontrast-tooltip-background-color-negative: CanvasText;
```
marissahuysentruyt, line 606: I wonder if these are necessary. They're all set to the same `CanvasText`, so instead of creating a separate variable to use in each modifier class, could we still just use the default custom property? 

I'm not sure if I'm explaining that ☝️ very well, so here's what I'm thinking. Instead of: 
```
.spectrum-Tooltip--negative & {
	background-color: var(--highcontrast-tooltip-background-color-negative, var(--mod-tooltip-background-color-negative, var(--spectrum-tooltip-background-color-negative)));
}
```
...where we use `--highcontrast-tooltip-background-color-negative`, we could just do...

```
.spectrum-Tooltip--negative & {
	background-color: var(--highcontrast-tooltip-background-color-default, var(--mod-tooltip-background-color-negative, var(--spectrum-tooltip-background-color-negative)));
}
```
...continuing to use `--highcontrast-tooltip-background-color-default`, which is already set to `CanvasText`.

Any thoughts on that? My thought is they're all the same anyways, but you may have more knowledge around the decision to add/use them!
Screenshot 2025-04-01 at 2 46 36 PM

components/tooltip/index.css
```diff
@@ -14,8 +14,6 @@
 @import "@spectrum-css/commons/overlay.css";
 
 .spectrum-Tooltip {
-	--spectrum-tooltip-backgound-color-default-neutral: var(--spectrum-neutral-subdued-background-color-default);
```
marissahuysentruyt, line 17: wwe-i-sentence-you-to-deletion

marissahuysentruyt (approved):
I think this looks great! 

Just a note that the commit when we squash this down, should probably be a `fix`, since we're changing CSS, but not introducing a breaking change. [Cassondra mentioned once](https://adobedesign.slack.com/archives/C062WL2L5E0/p1723559949388659) that if there's CSS changes, we should label our commits as `fix` or `feat`.

PR #3655, opened 2025-04-03
fix: reduce excess stylelint warnings from bundle module

jawinn:
## Description

Fixes large number of stylelint warnings appearing after yarn start. These were the result of stylelint running on the
`tools/bundle/dist/index.module.css` file.

- Adds stylelint override to better lint module files that have a different class format.
- Add ignoreFiles for the tools/bundle/dist, as stylelint should not be running on the dist folders during this step; it seems to be disregarding the .stylelintignore file.

CSS-1155

### Validation steps

Pull down the branch locally and:
- [x] `yarn install` and `yarn start`. After Storybook starts up (after loader and pages are functional), you should not see the 9k stylelint warnings related to "._" classes expecting to match pattern. No `[stylelint]` labeled warnings should appear for the bundle dist file.
- [x] To test the rule override, delete the `ignoreFiles` property in the root `stylelint.config.js` and run `yarn start` again. There should only be a few dozen stylelint warnings for the module file, instead of 9k.

## Screenshots

Original problem:

Screenshot 2025-03-13 at 4 56 52 PM

## To-do list

### DIFF
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -224,7 +224,7 @@ module.exports = {
 		},
 		{
 			/* Validate that the legacy themes don't introduce any new selectors or custom properties */
-			files: ["components/*/themes/express.css", "components/*/themes/spectrum-two.css"],
+			files: ["components/*/themes/express.css", "components/*/themes/spectrum.css"],
 			rules: {
 				"spectrum-tools/no-unused-custom-properties": null,
 				"selector-class-pattern": [
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -224,7 +224,7 @@ module.exports = {
 		},
 		{
 			/* Validate that the legacy themes don't introduce any new selectors or custom properties */
-			files: ["components/*/themes/*.css", "!components/*/themes/spectrum.css"],
+			files: ["components/*/themes/express.css", "components/*/themes/spectrum-two.css"],
 			rules: {
 				"spectrum-tools/no-unused-custom-properties": null,
 				"selector-class-pattern": [
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -14,6 +14,9 @@ module.exports = {
 		"@spectrum-tools/stylelint-theme-alignment",
 		// "stylelint-high-performance-animation",
 	],
+	ignoreFiles: [
+		"tools/bundle/dist/*.css",
+	],
 	rules: {
 		/** --------------------------------------------------------------
 		 * Disabled rules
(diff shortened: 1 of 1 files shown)


stylelint.config.js
```diff
@@ -14,6 +14,9 @@ module.exports = {
 		"@spectrum-tools/stylelint-theme-alignment",
 		// "stylelint-high-performance-animation",
 	],
+	ignoreFiles: [
+		"tools/bundle/dist/*.css",
+	],
```
castastrophe, line 19: If I'm reading this correctly, I think we should be using _either_ the .stylelintignore file _or_ the `ignoreFiles` property in the config. I'm flexible with whichever approach we use but I'd like to see all the same data in it. Would you mind either migrating the content from the stylelintignore to here or moving this out to the stylelintignore file?

https://stylelint.io/user-guide/ignore-code#files-entirely

stylelint.config.js
```diff
@@ -235,5 +238,17 @@ module.exports = {
 				],
 			}
 		},
+		{
+			/* Module CSS file classes have an underscore before classes, and use underscores instead of a dash as separators (e.g. _spectrum_well) */
+			files: ["tools/bundle/**/*.module.css"],
```
castastrophe, line 243: Do you know if this works if we simplify it to just `files:["*.module.css"],` - I'm just thinking all module files would have these same linting issues so if we add more, it'd be lovely to be able to inherit these automatically

stylelint.config.js
```diff
@@ -235,5 +238,17 @@ module.exports = {
 				],
 			}
 		},
+		{
+			/* Module CSS file classes have an underscore before classes, and use underscores instead of a dash as separators (e.g. _spectrum_well) */
```
castastrophe, line 242: This is a great clarification here, thanks for adding it!

castastrophe (changes requested):
One note about where we list our ignored files, otherwise this works for me when I pull it down locally.

stylelint.config.js
jawinn: That should work. I can test and make that change.

stylelint.config.js
castastrophe: I'm happy to include it but I feel like the ignoreFiles list should match what's in .stylelintignore and/or we just remove .stylelintignore and let this be the source of truth

jawinn: I've added to the `ignoreFiles` list to mirror what's in the .stylelintignore. The syntax had to change a bit to use the glob patterns and also apply for some of the linting results that were one directory up.

jawinn:
@castastrophe When you get a chance, can you look at this again, including the new commit "fix a stylelint config entry for theme-alignment". I found that the negation in this setting was causing some issues, and wanted to check with you that the theme files being specifically included now are correct for this setting (express.css and spectrum-two.css).

castastrophe (approved):
Looks great! Thank you!

stylelint.config.js
```diff
@@ -210,7 +224,7 @@ module.exports = {
 		},
 		{
 			/* Validate that the legacy themes don't introduce any new selectors or custom properties */
-			files: ["components/*/themes/*.css", "!components/*/themes/spectrum.css"],
+			files: ["components/*/themes/express.css", "components/*/themes/spectrum-two.css"],
```
castastrophe, line 227:
```suggestion
			files: ["components/*/themes/express.css", "components/*/themes/spectrum.css"],
```

Ooops this was probably supposed to be !spectrum-two.css. 🙃 #bugfix

rise-erpelding (approved):

PR #3656, opened 2025-04-03
feat(touch-action): adding touch action none to draggable components

aramos-adobe:
CSS1078 | [Jira Ticket](https://jira.corp.adobe.com/browse/CSS-1078)

## Including the touch action rule for draggable content

The slider, color slider, color area, color wheel, color handle all deserve to have their touch-action property manages so that trying to set the value of those interfaces doesn't cause page scrolling in touch context.

Adding `touch-action: none` to a slider or any draggable component is necessary to prevent the browser's default touch behaviors—such as scrolling, pinch-zooming, or double-tap zooming—from interfering with the component's intended interaction.

SWC has been adding this override to these components. They asked if CSS team can add it to the components so that they don't have to do it all the time.

### Validation steps

- [ ] Check to see `touch-action: none` has been added to listed draggable components' CSS

## Next steps

- Remove `touch-action: none` in listed components in SWC 
- Test to see if touch action still works as expected without any zooming or scrolling


.changeset/thick-ears-win.md
```diff
@@ -0,0 +1,12 @@
+---
+"@spectrum-css/colorhandle": minor
+"@spectrum-css/colorslider": minor
+"@spectrum-css/colorarea": minor
+"@spectrum-css/slider": minor
+---
+
+### Including the touch action rule for draggable content
+
+The slider, color slider, color area, color wheel, color handle all deserve to have their touch-action property manages so that trying to set the value of those interfaces doesn't cause page scrolling in touch context.
```
marissahuysentruyt, line 10: I think "manages" should be past tense, "managed."

.changeset/thick-ears-win.md
```diff
@@ -0,0 +1,12 @@
...
+
+Adding `touch-action: none` to a slider or any draggable component is necessary to prevent the browser's default touch behaviors—such as scrolling, pinch-zooming, or double-tap zooming—from interfering with the component's intended interaction.
```
marissahuysentruyt, line 12: I also think we can remove the hyphen between "zooming" and "from."

components/slider/index.css
```diff
@@ -207,6 +207,9 @@
 	padding-inline-end: var(--mod-slider-handle-gap, var(--spectrum-slider-handle-gap));
 	margin-inline-start: var(--mod-slider-track-margin-offset, var(--spectrum-slider-track-margin-offset));
 
+	touch-action: none;
+	user-select: none;
```
marissahuysentruyt, line 211: You left really great comments in color handle CSS regarding these two properties. Did you want to add them here as well? 

Not a blocker if you don't feel like it- I just really liked the comments! 😆

.changeset/thick-ears-win.md
```diff
@@ -0,0 +1,12 @@
+---
+"@spectrum-css/colorhandle": minor
+"@spectrum-css/colorslider": minor
+"@spectrum-css/colorarea": minor
+"@spectrum-css/slider": minor
```
marissahuysentruyt, line 5: Color wheel should be added to this list, too!

marissahuysentruyt (changes requested):
I know I just pinged you about further validation instructions, but I did find a few things to request changes on. 

The big thing I'm wondering is if we are supposed to be committing all of the `tokens/dist/css` files. I wasn't under the impression that we were committing those to the repo. Has that changed, or can we remove the `tokens/dist/css` commits from this branch?

.changeset/thick-ears-win.md
aramos-adobe: @marissahuysentruyt already had it before these additions

marissahuysentruyt: It looks like the color wheel index.css file was changed, so it should be added to this list of components, right? So that it gets the proper minor patch as well?

components/slider/index.css
marissahuysentruyt: @aramos-adobe I was trying to test this, and struggling with the slider. I wonder if it's because `.spectrum-Slider-track` and `.spectrum-Slider-fill` are only like, 2px high. This is the only component I feel like I can't really verify.

I also ended up with another question -do you think touch-action should go on the slider handle, also? What about the `.spectrum-Slider-track ~ .spectrum-Slider-track` selector?

marissahuysentruyt (commented):
Besides just looking at the styles (which check out according to your validation instructions), I tried the following: 

- I opened the PR preview on my phone.
- I went to all of the components that had changes: color area, color handle, color slider, color wheel, slider. 
- I try to zoom and scroll as I was touching that component, and most were in fact disabled (like you mentioned in our Slack convo). I left one comment about slider that I'd like your help with.

I wondered about whether or not color loupe needed this touch-action property. It's not really an interactive component, so I can understand if it's not needed! 

Once we get some clarity on if the tokens/dist files should all be committed, and color wheel is add to the version bumps in the changeset, I think I'll be good to approve. Just let me know what you think about slider. 👍

marissahuysentruyt (approved):
Thanks @aramos-adobe! I tested all of the components once again on my phone and they're working as far as I can tell. Inspecting each component shows that the `touch-action` and `user-select` properties are showing up. I still have the same problem with slider (I left a comment about it here: https://github.com/adobe/spectrum-css/pull/3656#discussion_r2060676978), but who knows- maybe it's me!! 

What's the next step for these? Do we need to open a PR in SWC to take in all of these new versions and remove their `touch-action` overrides? We should at least make a story card for it so we don't forget!

Nice work!

rise-erpelding (approved):
Looks great, works on my phone as far as I can tell (a bit tricky since our draggable components don't really drag)! Left a couple of tiny non-blocking suggestions.

aramos-adobe:
> Thanks @aramos-adobe! I tested all of the components once again on my phone and they're working as far as I can tell. Inspecting each component shows that the `touch-action` and `user-select` properties are showing up. I still have the same problem with slider (I left a comment about it here: [#3656 (comment)](https://github.com/adobe/spectrum-css/pull/3656#discussion_r2060676978)), but who knows- maybe it's me!!
> 
> What's the next step for these? Do we need to open a PR in SWC to take in all of these new versions and remove their `touch-action` overrides? We should at least make a story card for it so we don't forget!
> 
> Nice work!

I did bring this up initially to SWC, but they are not prioritizing it right now hence why we're putting it in spectrum-two branch for now until they are ready to remove those overrides.

PR #3659, opened 2025-04-08
feat(in-line-alert): S2 migration

cdransf:
## To-do list


.changeset/big-jars-pump.md
```diff
@@ -0,0 +1,56 @@
...
+`.spectrum-InLineAlert--neutral--subtle`
+`.spectrum-InLineAlert--neutral--subtle .spectrum-InLineAlert-icon`
+`.spectrum-InLineAlert--notice--bold`
+`.spectrum-InLineAlert--notice--bold .spectrum-InLineAlert-icon`
+`.spectrum-InLineAlert--notice--subtle`
+`.spectrum-InLineAlert--notice--subtle .spectrum-InLineAlert-icon`
+`.spectrum-InLineAlert--positive--bold`
+`.spectrum-InLineAlert--positive--bold .spectrum-InLineAlert-icon`
+`.spectrum-InLineAlert--positive--subtle`
+`.spectrum-InLineAlert--positive--subtle .spectrum-InLineAlert-icon`
```
castastrophe, line 46: These are updated selectors - did you want them to be in here with the updated properties or in a separate section?

components/inlinealert/index.css
```diff
@@ -136,6 +151,28 @@
 	}
 }
 
+.spectrum-InLineAlert--info--subtle:not(.spectrum-InLineAlert--info--bold) {
+	--mod-inlinealert-border-width: 0;
+	--mod-inlinealert-background-color: var(--spectrum-inlinealert-subtle-background-color-informative);
+	--mod-inlinealert-border-and-icon-color-info: var(--spectrum-inlinealert-subtle-color-informative);
```
castastrophe, line 157: When using these modifier hooks, is the goal that customers should not be able to overwrite the subtle variant? I think if we overwrite the background color, for example, we'd want to add a new, more specific hook:

```
--mod-inlinealert-background-color: var(--mod-inlinealert-background-color-info-subtle, var(--spectrum-inlinealert-subtle-background-color-informative));
```

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -107,8 +125,6 @@ Negative.tags = ["!dev"];
 
 /**
  * The positive variant uses the positive semantic color (green) and has a "checkmark" icon to help those with color vision deficiency discern the message tone. This variant should be used to inform someone of a successful function or result of an action they took.
- *
- * _Spectrum for Adobe Express uses a different icon. Use the SX_CheckmarkCircle_18_S.svg icon in the Express workflow icon set._
```
castastrophe, line 111: Woo!! 🎉

components/inlinealert/stories/template.js
```diff
@@ -57,6 +61,8 @@ export const Template = ({
 			class=${classMap({
 				[rootClass]: true,
 				[`${rootClass}--${variant}`]: typeof variant !== "undefined",
+				[`${rootClass}--${variant}--${"subtle"}`]: typeof variant !== "undefined" && isSubtle,
+				[`${rootClass}--${variant}--${"bold"}`]: typeof variant !== "undefined" && isBold,
```
castastrophe, line 65:
```suggestion
				[`${rootClass}--${variant}--subtle`]: typeof variant !== "undefined" && isSubtle,
				[`${rootClass}--${variant}--bold`]: typeof variant !== "undefined" && isBold,
```

castastrophe (changes requested):
Added a few questions, loving this update

.changeset/big-jars-pump.md
cdransf: Oops! Those should have been custom properties — updating the changelog and verified added/removed custom properties and mods against `metadata.json`. 😵‍💫

I would _love_ to hack on generating these changelog sections programmatically at some point. 🤔

components/inlinealert/index.css
cdransf: Ahhh good point! I added a new mod for this case and the other similar ones.

I also found some unrelated color issues with bold/subtle that I'm going to dig into while I was looking at this. ✨

components/inlinealert/stories/inlinealert.test.js
```diff
@@ -22,5 +22,13 @@ export const InlineAlertGroup = Variants({
...
 			isClosable: true,
 		},
+		{
+			testHeading: "Subtle",
+			isSubtle: true,
+		},
+		{
+			testHeading: "Bold",
+			isBold: true,
+		},
```
aramos-adobe, line 32: I'd add "Bold + Closable" and "Subtle + Closable"

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -41,6 +41,26 @@ export default {
...
+			name: "Bold",
+			type: { name: "boolean" },
+			table: {
+				type: { summary: "boolean" },
+				category: "Component",
+			},
+			control: "boolean",
+			if: { arg: "isSubtle", truthy: false },
+		},
 		isClosable: {
```
aramos-adobe, line 64: I would add the Link component and control here. In the [Figma design,](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?node-id=2659-4531&m=dev) it's included.

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -77,8 +99,6 @@ Default.args = {};
 // ********* DOCS ONLY ********* //
 /**
  * The informative variant uses the informative semantic color (blue) and has an "information" icon to help those with color vision deficiency discern the message tone. This should be used when the message needs to call extra attention, as compared to the neutral variant.
- *
- * _Spectrum for Adobe Express uses a different icon. Use the SX_Info_18_S.svg icon in the Express workflow icon set._
  */
 export const Informative = Template.bind({});
```
aramos-adobe, line 103: How do you feel about adding a template here that includes "bold, subtle, and border" for each variant? I imagine each style has a specific use case.

aramos-adobe: Screenshot 2025-04-18 at 7 03 58 PM

aramos-adobe:
@cdransf In-line alert migration looks beautiful! I love the colors and the CSS is well structured. The only thing I'd probably add is the Link child component. It looks like every border variant the Link is `accent-color-default`, bold and subtle matches the content's type color tokens

components/inlinealert/stories/inlinealert.test.js
cdransf: Right on! Added these + link variants.

components/inlinealert/stories/inlinealert.stories.js
cdransf: Added this (and a link example to the docs). ✨

aramos-adobe (approved):
Amazing work on this migration. I also love the mapping trick you used in the test grid, really clever!

components/inlinealert/index.css
```diff
@@ -46,7 +47,34 @@
...
+	--spectrum-inlinealert-subtle-color-negative: var(--spectrum-negative-visual-color);
+	--spectrum-inlinealert-subtle-background-color-neutral: var(--spectrum-neutral-subtle-background-color-default);
+	--spectrum-inlinealert-subtle-color-neutral: var(--spectrum-neutral-visual-color);
+
+	/* Bold fill */
+	--spectrum-inlinealert-bold-background-color-informative: var(--spectrum-informative-background-color-default);
+	--spectrum-inlinealert-bold-background-color-positive: var(--spectrum-positive-background-color-default);
+	--spectrum-inlinealert-bold-background-color-notice: var(--spectrum-notice-background-color-default);
+	--spectrum-inlinealert-bold-background-color-negative: var(--spectrum-negative-background-color-default);
+	--spectrum-inlinealert-bold-background-color-neutral: var(--spectrum-neutral-background-color-default);
```
jawinn, line 71: Rather than creating all these additional custom properties, the values of the default properties can be changed for the variant and fill style class. For example:
```
.spectrum-InLineAlert {
  --spectrum-inlinealert-background-color: var(--spectrum-background-layer-2-color);
  --spectrum-inlinealert-border-and-icon-color: var(--spectrum-neutral-visual-color);
  --spectrum-inlinealert-icon-color: var(--spectrum-neutral-visual-color);
  --spectrum-inlinealert-content-color: var(--spectrum-body-color);
  --spectrum-inlinealert-border-width: var(--spectrum-border-width-200);
  ...
}

.spectrum-InLineAlert--positive {
  --spectrum-inlinealert-border-and-icon-color: var(--spectrum-positive-visual-color);
  ...

  &.spectrum-InLineAlert--subtle {
    --spectrum-inlinealert-background-color: var(--spectrum-positive-subtle-background-color-default);
    --spectrum-inlinealert-border-width: 0px;
    ...
  }

  &.spectrum-InLineAlert--bold {
    --spectrum-inlinealert-background-color: var(--spectrum-positive-background-color-default);
    --spectrum-inlinealert-border-width: 0px;
    --spectrum-inlinealert-border-and-icon-color: var(--spectrum-inlinealert-color-white);
    --spectrum-inlinealert-content-color: var(--spectrum-inlinealert-color-white);
    ...
  }
}

.spectrum-InLineAlert--negative {
  ...

  &.spectrum-InLineAlert--subtle { ... }
  &.spectrum-InLineAlert--bold { ... }
}
```

Using this technique, [almost all of this CSS](https://github.com/adobe/spectrum-css/blob/1d13b97e9d5013840abc7c67bc40a86ac43b26ec/components/inlinealert/index.css#L149-L314
) should be able to go away. We shouldn't need to be setting the component's own mods and shouldn't need to repeat the icon selector.

components/inlinealert/stories/inlinealert.test.js
jawinn, line 1: The truncated test needs to use a variant that has an icon, so its alignment can be tested. 

I tried it out and it looks like this has an issue with the icon's vertical alignment being middle instead of the top? I also shrunk it down while testing and the icon probably needs `flex-shrink: 0` to prevent it from scaling down:

Screenshot 2025-04-25 at 4 19 46 PM

components/inlinealert/stories/inlinealert.stories.js
jawinn, line 1: The Docs page needs some more info about the fill style options of border fill (default), subtle fill, or bold fill. I'd recommend something like:
1. Mentioning these in the main component description.
2. Labeling these within the existing stories. This could be within the text itself (or a Container; within the text is probably cleaner): Screenshot 2025-04-25 at 11 39 39 AM
3. Showing a Neutral (default) story that has all 3 fill styles, similar to the others.

I'd also suggest renaming some stories to adjust their order and grouping a bit. Maybe an order like:
Default
Variant: informative
Variant: negative
Variant: neutral (default)  _note: new, has all 3 fill styles_
Variant: notice
Variant: positive
With footer: closable
With footer: link

components/inlinealert/stories/inlinealert.stories.js
```diff
@@ -49,14 +69,28 @@ export default {
 				category: "Advanced",
 			},
 			control: "boolean",
+			if: { arg: "hasLink", truthy: false },
```
jawinn, line 72: Are the link and the closable button mutually exclusive? I see both displaying at the same time in the spec. If both can appear, we'll want to make sure both controls are allowed to be on, and both being used at the same time are included in the tests file.

Screenshot 2025-04-25 at 4 41 20 PM

jawinn (changes requested):

aramos-adobe (approved):

components/inlinealert/stories/inlinealert.stories.js
jawinn, line 1: Does the component support not having a title? I saw this on the S2 desktop file, so it looks like the header text is optional:
Screenshot 2025-04-30 at 3 01 50 PM
If that's the case, it will need to make its way into the controls, Docs page, and tests.

components/inlinealert/index.css
jawinn, line 1: I'm still seeing the alignment of the icons in the middle. I see it aligned top in the Figma dev playground (ideally this and the no title option should also be on the design specs; you might want to pass that along). The dev playground doesn't appear to be working so well with overlapping header text and body text, so I'd double-check on this. The top alignment does seem to be consistent with the alignment in previous components we've migrated:
Screenshot 2025-04-30 at 2 19 06 PM
Screenshot 2025-04-30 at 4 08 00 PM

components/inlinealert/index.css
```diff
@@ -120,42 +131,97 @@
...
+		.spectrum-InLineAlert-footer {
+			forced-color-adjust: none;
+		}
 
-	.spectrum-InLineAlert-icon {
-		color: var(--highcontrast-inlinealert-border-and-icon-color, var(--mod-inlinealert-border-and-icon-color-negative, var(--spectrum-inlinealert-border-and-icon-color-negative)));
+		&.spectrum-InLineAlert--subtle:not(.spectrum-InLineAlert--bold),
+		&.spectrum-InLineAlert--bold:not(.spectrum-InLineAlert--subtle) {
+			--highcontrast-inlinealert-background-color: ButtonFace;
+		}
```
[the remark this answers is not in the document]
cdransf: Updated to align with those values — I restored the subtle/bold borders in WHC as well. Should we override the button text color? It's not legible without setting the button component mods.

image

components/inlinealert/index.css
```diff
@@ -120,42 +131,97 @@
...
+
+	&.spectrum-InLineAlert--subtle:not(.spectrum-InLineAlert--bold) {
+		--spectrum-inlinealert-background-color: var(--spectrum-neutral-subtle-background-color-default);
+		--spectrum-inlinealert-border-and-icon-color: var(--spectrum-neutral-visual-color);
+	}
+
+	&.spectrum-InLineAlert--bold:not(.spectrum-InLineAlert--subtle) {
+		--spectrum-inlinealert-background-color: var(--spectrum-neutral-background-color-default);
+	}
+}
```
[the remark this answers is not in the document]
cdransf: Updated! I nested the treatment classes as well given that they're only used in conjunction with `.spectrum-InLineAlert`. ✨

components/inlinealert/index.css
```diff
@@ -120,42 +131,97 @@
...
 	&:empty {
 		display: none;
 	}
 }
 
 .spectrum-InLineAlert--info {
-	border-color: var(--highcontrast-inlinealert-border-and-icon-color, var(--mod-inlinealert-border-and-icon-color-info, var(--spectrum-inlinealert-border-and-icon-color-info)));
+	--spectrum-inlinealert-border-and-icon-color: var(--spectrum-informative-visual-color);
+
+	&.spectrum-InLineAlert--subtle:not(.spectrum-InLineAlert--bold) {
```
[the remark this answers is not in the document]
cdransf: They aren't and these were just to enforce that, but aren't strictly necessary. Went ahead and dropped those. ✨

components/inlinealert/index.css
```diff
@@ -63,29 +65,39 @@
...
+		--spectrum-inlinealert-border-width: 0px;
+	}
+
+	&.spectrum-InLineAlert--bold:not(.spectrum-InLineAlert--subtle, .spectrum-InLineAlert--notice) {
+		--spectrum-inlinealert-header-color: var(--spectrum-white);
+		--spectrum-inlinealert-content-color: var(--spectrum-white);
+		--spectrum-inlinealert-border-and-icon-color: var(--spectrum-white);
+
+		--mod-button-content-color-default: var(--spectrum-white);
+		--mod-button-border-color-default: var(--spectrum-white);
```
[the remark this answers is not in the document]
cdransf: Got it! I've applied static white/black consistent with the styles currently setting text/content colors:

```css
&.spectrum-InLineAlert--bold:not(.spectrum-InLineAlert--subtle, .spectrum-InLineAlert--notice) {
  --spectrum-inlinealert-header-color: var(--spectrum-white);
  --spectrum-inlinealert-content-color: var(--spectrum-white);
  --spectrum-inlinealert-border-and-icon-color: var(--spectrum-white);
}
```

components/inlinealert/index.css
```diff
@@ -13,40 +13,42 @@
...
 	--spectrum-inlinealert-border-and-icon-color: var(--spectrum-neutral-visual-color);
 	--spectrum-inlinealert-header-color: var(--spectrum-heading-color);
 	--spectrum-inlinealert-content-color: var(--spectrum-body-color);
-	--spectrum-inlinealert-border-and-icon-color-info: var(--spectrum-informative-visual-color);
-	--spectrum-inlinealert-border-and-icon-color-positive: var(--spectrum-positive-visual-color);
-	--spectrum-inlinealert-border-and-icon-color-notice: var(--spectrum-notice-visual-color);
-	--spectrum-inlinealert-border-and-icon-color-negative: var(--spectrum-negative-visual-color);
 
+	/* Link */
+	--mod-link-text-color: var(--mod-inlinealert-content-color, var(--spectrum-inlinealert-content-color));
```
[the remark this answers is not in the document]
cdransf: [I applied the same static white/black logic](https://github.com/adobe/spectrum-css/pull/3659#discussion_r2069551141) to the links (for now) and started a thread in the implementations channel asking for some guidance. Without that logic there are several cases where the link is either hard to read or not visible at all.

components/inlinealert/index.css
cdransf: [I included these questions in the implementations thread I started as well.](https://github.com/adobe/spectrum-css/pull/3659#discussion_r2069551968) For now I've top aligned the icon and added the option to remove the title (along with docs, a control and test case).

.changeset/big-jars-pump.md
```diff
@@ -0,0 +1,32 @@
+---
+`@spectrum-css/inlinealert`: major
+---
+
+#### S2 migration in-line alert
+
+This migrates the `in-line alert` component to S2. Custom properties have been remapped and added per the design specification.
```
castastrophe, line 7:
```suggestion
This migrates the `in-line alert` component to Spectrum 2. Custom properties have been updated and added per the design specification.
```

Minor wording nit

.changeset/big-jars-pump.md
```diff
@@ -0,0 +1,32 @@
...
+
+Subtle and bold treatments have been added for each badge variant.
```
castastrophe, line 9: Can we add a note here about how the customer might use these new treatments? Maybe show a markup example?

components/inlinealert/stories/template.js
```diff
@@ -57,17 +83,52 @@ export const Template = ({
...
+const toTitleCase = (string) => string.replace(/\w\S*/g, text => text.charAt(0).toUpperCase() + text.substring(1).toLowerCase());
+
+/**
+ * Set the appropriate treatment header text
+ */
+const setTreatmentHeaderText = (variant, treatment) => `${toTitleCase(variant)} variant with ${treatment !== "border" ? treatment : "outline"} fill`;
+
+export const AlertsWithStyleOptions = ({
+	...args
+}, context = {}) => Container({
```
castastrophe, line 112:
```suggestion
export const AlertsWithStyleOptions = (args, context = {}) => Container({
```

Since we're not breaking anything out, we don't need to use a spread for the args object.

components/inlinealert/stories/template.js
```diff
@@ -57,17 +83,52 @@ export const Template = ({
...
+			...args,
+			treatment: "subtle",
+			headerText: setTreatmentHeaderText(args.variant, "subtle"),
+		}, context)}
+		${Template({
+			...args,
+			treatment: "bold",
+			headerText: setTreatmentHeaderText(args.variant, "bold"),
+		}, context)}
+	`,
```
castastrophe, line 133:
```suggestion
	content: [
		Template({
			...args,
			headerText: setTreatmentHeaderText(args.variant, "border"),
		}, context),
		Template({
			...args,
			treatment: "subtle",
			headerText: setTreatmentHeaderText(args.variant, "subtle"),
		}, context),
		Template({
			...args,
			treatment: "bold",
			headerText: setTreatmentHeaderText(args.variant, "bold"),
		}, context),
	],
```

Might be simpler to use the content array technique instead of having to form the combination yourself here.

components/inlinealert/index.css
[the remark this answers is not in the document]
jawinn: This seems to be happening because `forced-color-adjust: none` is being applied to the footer element, and it's cascading to the button. Could you remove it from the footer to fix this? Could `forced-color-adjust` also be removed for the header and content classes? In a quick test, I wasn't noticing a visual difference when I disabled it.

components/inlinealert/index.css
jawinn: Thanks for those updates and starting a thread about the questions, I'm following it 👀 

I'm still seeing the icon vertically middle aligned on the PR link though.

components/inlinealert/index.css
```diff
@@ -13,40 +13,119 @@
...
 	/* Font */
-	--spectrum-inlinealert-heading-font-family: var(--spectrum-sans-font-family-stack);
+	--spectrum-inlinealert-heading-font-family: var(--spectrum-heading-sans-serif-font-family);
 	--spectrum-inlinealert-heading-font-weight: var(--spectrum-heading-sans-serif-font-weight);
 	--spectrum-inlinealert-heading-font-style: var(--spectrum-heading-sans-serif-font-style);
 	--spectrum-inlinealert-heading-font-size: var(--spectrum-heading-size-xxs);
 	--spectrum-inlinealert-heading-line-height: var(--spectrum-heading-line-height);
 
-	--spectrum-inlinealert-content-font-family: var(--spectrum-sans-font-family-stack);
+	--spectrum-inlinealert-content-font-family: var(--spectrum-body-sans-serif-font-family);
```
jawinn, line 22: I happened to notice that the heading and body fonts were not loading and were showing up as serif when testing in Assistivlabs. We should be continuing to use the font stack, otherwise Typekit fonts won't load (the font name has a dash). This token has a single value that will only show up when you have the font installed on your local machine. The fonts are a case where we can't match what is listed on the specs.

Screenshot 2025-05-01 at 11 26 33 AM

components/inlinealert/index.css
```diff
@@ -13,40 +13,119 @@
...
-	--spectrum-inlinealert-border-and-icon-color-notice: var(--spectrum-notice-visual-color);
-	--spectrum-inlinealert-border-and-icon-color-negative: var(--spectrum-negative-visual-color);
 
+	&.spectrum-InLineAlert--subtle {
+		--mod-inlinealert-border-width: 0px;
+	}
+
+	&.spectrum-InLineAlert--bold {
+		--mod-inlinealert-border-width: 0px;
+	}
```
jawinn, line 53:
```suggestion
	&.spectrum-InLineAlert--subtle {
		--spectrum-inlinealert-border-width: 0px;
	}

	&.spectrum-InLineAlert--bold {
		--spectrum-inlinealert-border-width: 0px;
	}
```

components/inlinealert/index.css
```diff
@@ -120,42 +191,30 @@
...
+		.spectrum-InLineAlert-header,
+		.spectrum-InLineAlert-content,
+		.spectrum-InLineAlert-footer {
+			forced-color-adjust: none;
+		}
+
+		&.spectrum-InLineAlert--subtle,
+		&.spectrum-InLineAlert--bold {
+			--highcontrast-inlinealert-background-color: Canvas;
+			--mod-inlinealert-border-width: var(--spectrum-inlinealert-border-width);
```
jawinn, line 217: Suggestion: I'd add `--highcontrast-inlinealert-border-width` to the stack where it's used in the `border-inline-width`, and then set the value here with `--highcontrast-inlinealert-border-width: var(--spectrum-border-width-200);`

jawinn (changes requested):
Thanks for those updates, it's looking a lot better now. The three major things remaining I'm seeing are the font stack (easy fix), the button not appearing in neutral, and the "without title" version not having an icon.

**Without title and icons**
If I use the "without header" control, the icons are no longer present. The icons are on the Figma and are necessary to relay the meaning without relying on just color:
Figma:
Screenshot 2025-05-01 at 9 55 05 AM
PR:
Screenshot 2025-05-01 at 9 53 27 AM

**Neutral button and link not visible**
It looks like this variant may be using the wrong button and link options (static color)? They are there but invisible:
Screenshot 2025-05-01 at 11 40 24 AM

components/inlinealert/index.css
cdransf: Fixed that! Managed to miss committing it. ✨

.changeset/big-jars-pump.md
cdransf: Can do! Added some notes and markup examples for each. ✨

components/inlinealert/index.css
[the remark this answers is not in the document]
cdransf: Added CSS-1198 to track this.

components/inlinealert/index.css
```diff
@@ -120,42 +193,24 @@
...
+		--highcontrast-inlinealert-content-color: CanvasText;
+		--highcontrast-inlinealert-border-and-icon-color: CanvasText;
 
-	.spectrum-InLineAlert-icon {
-		color: var(--highcontrast-inlinealert-border-and-icon-color, var(--mod-inlinealert-border-and-icon-color-negative, var(--spectrum-inlinealert-border-and-icon-color-negative)));
+		&.spectrum-InLineAlert--subtle,
+		&.spectrum-InLineAlert--bold {
+			--highcontrast-inlinealert-background-color: Canvas;
+			--highcontrast-inlinealert-border-width: var(--spectrum-inlinealert-border-width);
+		}
```
jawinn, line 214:
```suggestion
		--highcontrast-inlinealert-background-color: Canvas;

		&.spectrum-InLineAlert--subtle,
		&.spectrum-InLineAlert--bold {
			--highcontrast-inlinealert-border-width: var(--spectrum-border-width-200);
		}
```

1. The border isn't appearing for these variants right now because the value of the width variable is 0px (for these variants in not high contrast).
2. I'd move the background declaration up so the high contrast background is always applied; it doesn't appear to be limited to just these variants anymore.

jawinn (changes requested):
Nice work! One last update and then I'm good to approve.

jawinn (approved):

PR #3660, opened 2025-04-08
chore(combobox, datepicker,et all): add displayLabel arg

marissahuysentruyt:
## Description

After rebasing the search field migration branch, I noticed that the icons were placed awkwardly. I realized that the nested textfield in my search component was just missing a `displayLabel: false` argument. That arg was updated in the textfield S2 migration, to be set to `true` on default (based on guidance that "a textfield should always have a label"), but because search isn't utilizing the field label that ships with textfield, that arg should be set to `false`. Without the displayLabel arg set to false in the following components, the textfield was still rendering an additional label, and offsetting the icon & button within search.

After scanning the repo, the following components needed similar `displayValue: false` args: 
combobox, datepicker, form, pagination, stepper

### Jira/Specs
CSS-1174

### Validation steps

- [ ] Pull down the branch to run locally or visit the deploy preview.
- [ ] Navigate to each component's docs page. Ensure that nothing looks "broken" and most of the styles are acceptable.
    - [ ] combobox
    - [ ] date picker
    - [ ] form
    - [ ] pagination
    - [ ] search
    - [ ] stepper/number field
- [ ] By comparison, compare the fixed components to a recent deploy preview for `spectrum-two`:
    - [ ] combobox
    - [ ] date picker
    - [ ] form
    - [ ] pagination
    - [ ] search
    - [ ] stepper/number field

### Regression testing

Validate:

## To-do list

### DIFF
--- a/components/datepicker/stories/template.js
+++ b/components/datepicker/stories/template.js
@@ -64,6 +64,7 @@ export const DatePicker = ({
 				name: "field",
 				id: triggerId,
 				value: selectedDay ? new Date(selectedDay).toLocaleDateString(lang) : undefined,
+				displayLabel: false,
 				onclick: function () {
 					if (!isOpen) updateArgs({ isOpen: true });
 				},
--- a/components/combobox/stories/template.js
+++ b/components/combobox/stories/template.js
@@ -67,6 +67,7 @@ const Combobox = ({
 				name: "field",
 				isReadOnly,
 				value,
+				displayLabel: false,
 				onclick: function () {
 					if (!isOpen) updateArgs({ isOpen: true });
 				},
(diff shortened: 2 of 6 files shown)


components/datepicker/stories/template.js
```diff
@@ -64,6 +64,7 @@ export const DatePicker = ({
 				name: "field",
 				id: triggerId,
 				value: selectedDay ? new Date(selectedDay).toLocaleDateString(lang) : undefined,
+				displayLabel: false,
```
marissahuysentruyt, line 67: double check the Range variant's textfield

rise-erpelding (commented):
This definitely feels like the simpler solution and works nicely.

It is interesting to note though, looking through the guidelines they say "text fields should always have a label" and then in something like combobox, the field label can exist but the markup is outside `.spectrum-Textfield` (and `.spectrum-Combobox` for that matter).

rise-erpelding (approved):

aramos-adobe (approved):

PR #3664, opened 2025-04-10
fix(actionbutton): reduce mod interception

castastrophe:
## Description

This update aims to simplify `--mod-` access by ensuring local variants and states aren't hooking into those custom properties for overrides. This updates all local variants and states to override the `--spectrum-actionbutton-` properties instead and adjusts the specificity to ensure no regressions in rendered results.

Other changes of note:

- Removes the `.spectrum-ActionButton--sizeM` as unnecessary and duplicate as these styles are applied in the base class.
- Reduces the use of the `:not()` pseudo-class in theme overrides as these selectors are more complex to remap for web component projects.

### Validation steps

- [ ] Expect no visual changes to the component.
- [ ] Expect to be able to overwrite the following modifiers:

### Regression testing

Validate:

## To-do list


components/actionbutton/index.css
```diff
@@ -68,8 +78,51 @@
...
+		--mod-actionbutton-content-color-hover-selected: var(--mod-actionbutton-static-white-content-color-selected, var(--spectrum-black));
+		--mod-actionbutton-content-color-down-selected: var(--mod-actionbutton-static-white-content-color-selected, var(--spectrum-black));
+		--mod-actionbutton-content-color-focus-selected: var(--mod-actionbutton-static-white-content-color-selected, var(--spectrum-black));
+
+		--spectrum-actionbutton-content-color-disabled: var(--spectrum-disabled-static-white-content-color);
+		--spectrum-actionbutton-focus-indicator-color: var(--spectrum-static-white-focus-indicator-color);
+	}
+
+	/* Emphasized styles are only applied to the "selected" state of the button and are not applied in static color modes; must list after static variants */
+	&.spectrum-ActionButton--emphasized {
```
marissahuysentruyt, line 125: I think the simplest or most straightforward way to get the static color + emphasized background colors corrected is to add the `:not(.spectrum-ActionButton--staticBlack,.spectrum-ActionButton--staticWhite)` back to this selector. Do you have reservations about doing that `:not()`?

marissahuysentruyt: ooorrrr....we adjust the template. Does that work? That would reduce the need to continue to adjust the CSS you have right now.

`[`${rootClass}--emphasized`]: isEmphasized && !staticColor,`, that way the --emphasized class is only applied when no static color arg has been defined.

castastrophe: Oh good point, I wonder if the web component can enforce that too. 🤔

castastrophe: I added back the not selectors since SWC already accounts for it!

PR #3669, opened 2025-04-15
feat(tokens,card): update to @adobe/spectrum-tokens@13.3.0

castastrophe:
## Description

Pulling in the latest S2 release from `@adobe/spectrum-tokens`, v13.3.0. Details are outlined in the [changeset](https://github.com/adobe/spectrum-css/pull/3669/files#diff-f8f34138e149af373adcf16b6464425cefe58eba8a92778efaebf946ac7bbf59).

Updated card to remove `--mod-card-selected-background-color-rgb` in favor of `--mod-card-selected-background-color`. With this package update, `--spectrum-card-selection-background-color` now maps to a transparent global property of `--spectrum-transparent-(black|white)-600` depending on color context. This means the previous, hardcoded values are no longer relevant or usable.

### Validation steps

- [x] Expect the card's selected background color (applied to the ::before element) to be transparent-white-600 in light mode. (@rise-erpelding) (@cdransf)
- [x] Expect the card's selected background color (applied to the ::before element) to be transparent-black-600 in dark mode. (@rise-erpelding) (@cdransf)
- [x] Expect `--mod-card-background-color-selected` to override the previous 2 test results. (@rise-erpelding) (@cdransf)

### Regression testing

Validate:

## To-do list


components/card/index.css
rise-erpelding, line 1: There's a lingering blue background color on the quiet variant, I think because the background color gets applied to the preview for this variant? It seems like maybe there isn't a quiet variant in S2 anyway so this would go away on its own, but do we want to change that to `--spectrum-card-selection-background-color` for consistency?

image

rise-erpelding (approved):
This looks good! I left one small note on quiet + selected below in case that's something we want to adjust here.

components/card/index.css
castastrophe: Oh good catch. What would you think if I removed quiet from Card completely here even though it's not officially being migrated here?

cdransf (approved):
Looks good! All the validation steps check out. ✨

PR #3671, opened 2025-04-15
chore: merge latest from main

castastrophe:
## Description

Pull in the latest updates from the main branch to benefit the spectrum 2 development.

## Testing and validation

Expect no regressions.


jawinn (approved):
We've lost a couple things in the MDX docs removal for story and icon on main, that we can update in a follow-up PR:
1. Button "Docs" page is missing; I think this just needs "!autodocs" removed from the story default export.
3. Icon docs lost a few additions from its MDX in spectrum-two, “For a full list of all icons within this set...” text/links

We'll also need to reorganize and exclude new stories as necessary on the Docs page for Icon in that follow-up, as there are new stories that should be hidden and a few stories need a title change to organize them better.

Edit: as discussed, I'll add a commit to update that on this PR.

rise-erpelding (approved):
This looks great! I did check over the button and icon docs pages and they incorporate the needed info from `main` and maintain any new icons documentation that's more specific to S2.

PR #3677, opened 2025-04-17
feat(swatch+swatchgroup): S2 migration

cdransf:
## Description

`CSS-1026` + `CSS-1029`

This migrates the `swatch` and `swatchgroup` components to the latest Spectrum 2 designs. Custom properties have been remapped and added per the design specification.

An `Add Swatch` variant has been added with the required color tokens and the specified add UI icon.

The `Add Swatch` and `Mixed Value` variants may not be combined with background colors or images. To implement the `Add Swatch` you need to apply the `is-addSwatch` class to the `spectrum-Swatch` element. `Add swatch` keyboard focus may be applied by adding the `is-keyboardFocused` class to this combination of classes.

The `spectrum-Swatch-icon` is used to set the expected color to icons contained within the swatch.

The light and no border variants have been removed. Individual swatches have a border set to `--spectrum-gray-1000` at `42%` opacity, while the border opacity is set to `20%` in swatch groups.

### Removed custom properties

`--spectrum-swatch-border-color`
`--spectrum-swatch-dash-icon-color` (replaced by `--spectrum-swatch-icon-color`)

### New custom properties

`--spectrum-swatch-border-color-rgb`
`--spectrum-swatch-border-opacity`
`--spectrum-corner-radius-full`
`--spectrum-corner-radius-none`
`--spectrum-swatch-disabled-icon-border-opacity`
`--spectrum-swatch-icon-color`
`--spectrum-swatch-rectangle-width-multiplier`

### New mods

`--mod-add-button-background`
`--mod-add-button-background-down`
`--mod-add-button-background-hover`
`--mod-add-button-background-keyboard-focus`
`--mod-corner-radius-full`
`--mod-mixed-button-background`
`--mod-swatchgroup-border-color`
`--mod-swatch-border-color-rgb`
`--mod-swatch-border-opacity`

### Regression testing

Validate:

## To-do list


rise-erpelding (changes requested):
Hi! This looks like a pretty minor migration but I have a few comments that I added below that I'll summarize:

- Would love to see some updated testing options to show variants/states that we aren't seeing, this really helps us prevent regressions (or will once we have these changes in our main branch). Key-focus state was one that I mentioned to add, but it's also helpful to run through these manually too (like for instance, a hover test case usually shouldn't have a hover state, key-focus states shouldn't have an additional key focus state when you tab to the component).
- I think by default we should be seeing some rounding on the swatch, although it's really still unclear to me after looking through specs whether or not the no-rounding option should still exist in S2. Some clarification from design on that might be good here, but doesn't need to be a blocker, but in the mean time since all the default swatches in the specs have rounding, ours should too.
- Swatch group spacing for spacious maybe needs adjusting.

I have a couple of other questions after looking through the spec and the PR too:
- Down state is kind of mentioned in the spec, but no tokens are referenced. Does this mean that swatch gets a down state treatment (with the perspective-based transform)? Not a blocker for this PR but we should find out.
- Is that swatch border color (default) correct? I'm seeing it compute to rgba(19, 19, 19, 0.42) but the spec says gray-1000 which should be 0, 0, 0. If we adjust that though is there a difference between that default border and the light border?

I didn't check WHCM on this pass, but will plan to in the future!

components/swatch/index.css
```diff
@@ -289,19 +325,19 @@
...
+.spectrum-Swatch--roundingPartial {
 	&,
 	&::before,
 	&::after,
 	.spectrum-Swatch-fill,
 	.spectrum-Swatch-fill::before,
 	&.is-selected .spectrum-Swatch-fill,
 	&.is-selected .spectrum-Swatch-fill::before {
-		border-radius: 0;
+		border-radius: var(--mod-swatch-border-radius, var(--spectrum-corner-radius-75));
```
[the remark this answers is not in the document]
cdransf: There isn't (at least not a good one) I was just confused by the indication that no rounding was gone, partial was staying and what the default should be. I changed `none` as a variant and `partial` as the default. ✨

components/swatch/stories/swatch.stories.js
```diff
@@ -145,7 +165,7 @@ Disabled.parameters = {
 };
 
 /**
- * Default rounding and full rounding are usually used when a swatch is presented by itself near other components. A rounding of “none” is used in a swatch group to help minimize the Hermann grid illusion that happens at the intersections of white space in the group.
```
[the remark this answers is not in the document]
cdransf: Ah yep — just reverted that.

components/swatchgroup/index.css
```diff
@@ -13,24 +13,22 @@
...
-	--mod-swatchgroup-spacing-regular: var(--mod-swatchgroup-spacing-compact);
-	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-50);
+	--mod-swatchgroup-spacing: var(--mod-swatchgroup-spacing-compact, var(--spectrum-spacing-50));
 }
 
 /* Spacious */
 .spectrum-SwatchGroup--spacious {
-	--mod-swatchgroup-spacing-regular: var(--mod-swatchgroup-spacing-spacious);
-	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-100);
+	--mod-swatchgroup-spacing: var(--mod-swatch-group-spacing-spacious, var(--spectrum-swatch-group-spacing-spacious));
```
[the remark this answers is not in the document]
cdransf: I added a `mediumLarge` variant since the spacing for those two are the same. I _did_ drop compact as that's no longer present in the specs. I'm happy to restore it, but thought it might best be remvied.

components/swatch/stories/swatch.test.js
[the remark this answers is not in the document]
cdransf: Awesome! I just updated the tests and stories. The add section of the testing grid also has active and hoer states.

The disabled icon sizes are shown in the docs by adjusting the sizes (as is the case with the add icon) — the icon size also reacts to the size control in Storybook. ✨

components/swatch/index.css
```diff
@@ -301,7 +337,7 @@
 	.spectrum-Swatch-fill::before,
 	&.is-selected .spectrum-Swatch-fill,
 	&.is-selected .spectrum-Swatch-fill::before {
-		border-radius: 0;
+		border-radius: var(--mod-swatch-border-radius, var(--spectrum-corner-radius-none));
```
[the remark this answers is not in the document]
cdransf: Totally makes sense! Just updated that. ✨

components/swatchgroup/stories/swatchgroup.stories.js
```diff
@@ -37,14 +31,18 @@ export default {
 		imageUrl: { table: { disable: true } },
 		isMixedValue: { table: { disable: true } },
 		gradient: { table: { disable: true } },
+		isKeyboardFocused: { table: { disable: true } },
```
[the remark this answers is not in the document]
cdransf: Good call! I totally missed that. I've updated the swatchgroup story to account for that. 🥳

components/swatch/index.css
```diff
@@ -251,6 +279,14 @@
 	}
 }
 
+.spectrum-Swatch--border {
+	.spectrum-Swatch-fill {
+		&::before {
+			box-shadow: inset 0 0 0 var(--spectrum-swatch-border-thickness) var(--mod-swatchgroup-border-color, var(--spectrum-swatchgroup-border-color));
```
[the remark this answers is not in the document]
rise-erpelding: (My understanding of swatch border colors is unfolding in real time). I'm realizing now that there are a few border color tokens:

From [spectrum-tokens](https://github.com/adobe/spectrum-tokens/blob/%40adobe/spectrum-tokens%4013.3.0/packages/tokens/src/color-component.json) 13.3.0:
`swatch-border-color`: `gray-1000`
`swatch-border-opacity`: `42%`
`swatch-group-border-color`: `gray-1000`
`swatch-group-border-opacity`: `20%` - this doesn't seem to exist in spectrum-tokens

Then in custom tokens (this is light-vars) we also have these:
```css
	--spectrum-swatch-border-color: rgb(0 0 0 / 51%);
	--spectrum-swatch-border-color-light: rgb(0 0 0 / 42%);

	/**
	 * `--spectrum-swatch-group-border-color` / `--spectrum-swatch-group-border-opacity`
	 */
	--spectrum-swatchgroup-border-color: rgb(255 255 255 / 20%);
```

And then as far as border colors, I'm aware that we previously had light colored borders for certain circumstances, I'm wondering if that's why we're seeing an option for 42% opacity and one for 20% opacity? Maybe? I was thinking before that maybe we ought to remove the light border variant, but now I'm on the fence. Either way, we should clean up the custom tokens to avoid some redundancy between the custom tokens and the tokens that we should have available already.

components/swatch/stories/swatch.test.js
```diff
@@ -93,22 +101,41 @@ export const SwatchGroup = Variants({
 			testHeading: "Mixed value",
 			isMixedValue: true,
 		},
-	],
-	stateData: [
 		{
-			testHeading: "Disabled",
-			isDisabled: true,
+			testHeading: "Add",
+			isAddSwatch: true,
```
rise-erpelding, line 106: This coverage looks great!

One thing I noticed since I looked at both forced-colors and WHCM this time around is that we don't have a lot of contrast between the swatch and the icon (`.spectrum-Swatch-icon`), while I think we can make improvements to high contrast when we have a more defined spec for it, it would be nice to at least ensure that the contrast is decent for now on these icon swatches.

Here is (I think) forced colors above and WHCM below, you can see that the mixed value and add variants make it really difficult to see the icon here:
Screenshot 2025-04-28 at 8 47 26 AM
Screenshot 2025-04-28 at 8 46 51 AM

rise-erpelding (changes requested):
Hiya, leaving another round of suggestions, the biggest issues I'm currently noticing are the doubled key-focus states and the swatch border tokens I mentioned in the comments.

components/swatch/index.css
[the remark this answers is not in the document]
marissahuysentruyt: @rise-erpelding @cdransf Sorry to pop in here, but here's the PR where I left most of the discussion I had with Miruna about the light border stuff.

https://github.com/adobe/spectrum-css/pull/2954

If I'm understanding Miruna's message correctly, the 42% border opacity (`swatch-border-color`) is for single swatches in S2, and then the 20% is for swatches in a swatch group in S2. I don't get the impression that design wants a "light border" option any longer, since Miruna also says: "we decided that for S2 we are gonna be more restrictive with this and give the guidance to have a border for single swatch or swatch group." Re-reading this, I understand her to say design wants all swatches in S2 to have a border, regardless of contrast (I think that's why the light border option existed, even though it was added inconsistently all over the place in a one-off fashion).

I have not reviewed this PR yet, but my instinct is that --spectrum-swatch-border-color (which has the S1 51% opacity in the snippet above) should actually be the 42% opacity for S2. The --spectrum-swatch-border-color-light custom variable (created for the light border option) can be removed. Something more along these lines: 

```
--spectrum-swatch-border-color: rgba(0 0 0 / 42%);
--sspectrum-swatchgroup-border-color: rgba(255 255 255 / 20%)
```

It's super confusing, so I'm happy to chat if I can clear anything up. We worked a lot out in that PR! ☝️

components/swatch/stories/swatch.test.js
cdransf: image

Updated these! Let me know if you think we should revise further. ✨

tokens/custom/light-vars.css
```diff
@@ -56,8 +56,13 @@
 	--spectrum-steplist-current-marker-color-key-focus: var(--spectrum-blue-800);
 
 	--spectrum-swatch-border-color: rgb(0 0 0 / 51%);
-	--spectrum-swatch-border-color-light: rgb(0 0 0 / 20%);
+	--spectrum-swatch-border-color-light: rgb(0 0 0 / 42%);
```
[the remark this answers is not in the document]
cdransf: Just removed that!

tokens/custom/light-vars.css
```diff
@@ -56,8 +56,13 @@
...
 
-	--spectrum-treeview-item-background-color-quiet-selected: rgba(var(--spectrum-gray-900-rgb), 0.06);
-	--spectrum-treeview-item-background-color-selected: rgba(var(--spectrum-blue-900-rgb), 0.1);
+	/**
+	 * `--spectrum-swatch-group-border-color` / `--spectrum-swatch-group-border-opacity`
+	 */
+	--spectrum-swatchgroup-border-color: rgb(255 255 255 / 20%);
```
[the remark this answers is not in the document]
cdransf: With `swatch-group-border-color-token` added I've gone ahead and removed the custom var in light/dark and added `--spectrum-swatch-border-color-light: rgb(var(--spectrum-swatch-group-border-color) / 42%);` to the swatch styles (since the token includes the light and dark colors).

components/swatch/index.css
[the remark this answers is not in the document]
cdransf: ✨ I _think_ this should be sorted now — `spectrum-swatch-border-color` has been removed and `spectrum-swatchgroup-border-color` is set to `rgba(255 255 255 / 20%)` in `light-vars.css` and `dark-vars.css`. _Or_ I could be misunderstanding things. 😅

components/swatch/index.css
```diff
@@ -37,6 +36,7 @@
 	--spectrum-add-button-background-hover: var(--spectrum-gray-200);
 	--spectrum-add-button-background-down: var(--spectrum-gray-200);
 	--spectrum-add-button-background-keyboard-focus: var(--spectrum-gray-200);
+	--spectrum-swatch-border-color-light: rgb(var(--spectrum-swatch-group-border-color) / 42%);
```
rise-erpelding, line 39: image
Ah, I see the problem with using the token now. Maybe it would be better not to use the token in that case, but adding a todo comment that we should? Kinda rough, but maybe something like:

```css
/* TODO update to use tokens once available */
--spectrum-swatch-border-color-light: rgb(var(--spectrum-gray-1000-rgb), 0.2);
```
(I'm using 20% opacity and not 42% since that's what it's in the spec for swatch group border color)

This still feels weird because these are the tokens for `swatch-group-border-color` and `swatch-group-border-opacity` and I'm not absolutely certain that these are supposed to be the "light border" tokens or how we're supposed to use them. That part probably would be follow-up work, but this might get it "good enough" for now.

Not sure if maybe @castastrophe has a better suggestion though?

tokens/custom/dark-vars.css
```diff
@@ -57,9 +57,11 @@
 
 	--spectrum-steplist-current-marker-color-key-focus: var(--spectrum-blue-700);
 
-	--spectrum-swatch-border-color: rgb(255 255 255 / 51%);
-	--spectrum-swatch-border-color-light: rgb(255 255 255 / 20%);
+	/**
+	 * `--spectrum-swatch-group-border-color` / `--spectrum-swatch-group-border-opacity`
+	 */
```
rise-erpelding, line 62: Do we still need this? (and also in the light-vars)?

rise-erpelding (changes requested):
This is getting closer! WHCM looks good to me and a lot of the other issues I called out earlier have been resolved. Just a few callouts about focus outline and the light border.

1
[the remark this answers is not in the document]
cdransf: Oh fun! Yes. 🫠

tokens/custom/dark-vars.css
cdransf: Nope! Dropped both. ✨

components/swatch/stories/template.js
```diff
@@ -80,8 +80,11 @@ export const Template = ({
 			@click=${function() {
 				updateArgs({ isSelected: !isSelected });
 			}}
+			@focusin=${function() {
+				updateArgs({ isKeyboardFocused: true });
+			}}
 			@focusout=${function() {
-				updateArgs({ isSelected: false });
+				updateArgs({ isKeyboardFocused: false });
```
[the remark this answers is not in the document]
cdransf: All good! Reverted ✨

components/swatch/index.css
cdransf: Updating this with the revision you provided, but I agree that it feels odd not to use the tokens? I've run into other cases where the supplied tokens don't work with the partially opaque implementation called out in the design specs. ✨

components/swatch/index.css
```diff
@@ -214,28 +214,6 @@
...
-		inset: calc(-2 * var(--spectrum-swatch-focus-indicator-gap));
-
-		opacity: 0;
-
-		border-width: var(--spectrum-swatch-focus-indicator-thickness);
-		border-style: solid;
-		border-color: var(--spectrum-swatch-focus-indicator-color);
-		border-radius: var(--spectrum-swatch-focus-indicator-border-radius);
-
-		transition: opacity var(--mod-animation-duration-100, var(--spectrum-animation-duration-100)) ease-in-out;
```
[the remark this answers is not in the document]
cdransf: Ah, good catch! It'd definitely be nice to keep. I gave applying this to the outline a try by dropping the shorthand and splitting out outline-width, style and color. Instead of `outline: none` I'm zeroing out the width, setting the style and setting color to transparent. For the focus state I'm then setting the width and transitioning the color only. https://github.com/adobe/spectrum-css/pull/3677/commits/a044f6d4d516f38d86e65535d82adbf9d453f762

`--mod-animation-duration-100` is odd — I'd expect us to create a custom property that resolves to `--spectrum-animation-duration-100` and then add a mod based on the name of that custom property (if at all.) ✨

components/swatchgroup/stories/swatchgroup.stories.js
marissahuysentruyt, line 1: @cdransf Could we add `tags: ["migrated"]` to swatch group & swatch? I noticed some of our migrated components have that, and some don't, but I think that'd be a nice thing to do!

.changeset/proud-schools-reply.md
```diff
@@ -0,0 +1,42 @@
+---
+`@spectrum-css/swatchgroup`: major
+`@spectrum-css/swatch`: major
+---
+
+#### S2 migration swatchgroup
```
castastrophe, line 6:
```suggestion
#### S2 migration for Swatch group
```

What would you think of using the readable name in the title instead of the package?

.changeset/proud-schools-reply.md
```diff
@@ -0,0 +1,42 @@
...
+
+This migrates the `swatch` and `swatchgroup` components to S2. Custom properties have been remapped and added per the design specification.
```
castastrophe, line 8:
```suggestion
This migrates the `swatch` and `swatchgroup` components to the latest Spectrum 2 designs. Custom properties have been remapped and added per the design specification.
```

.changeset/proud-schools-reply.md
```diff
@@ -0,0 +1,42 @@
...
+
+An `Add Swatch` variant has been added with the required color tokens and the specified add UI icon.
+
+##### Removed custom properties
+
+`--spectrum-swatch-border-color`
+`--spectrum-swatch-dash-icon-color`
```
castastrophe, line 15: Just double checking, were these replaced with a property with a different name? Should we outline that here?

.changeset/proud-schools-reply.md
```diff
@@ -0,0 +1,42 @@
...
+`--mod-add-button-background-keyboard-focus`
+`--mod-corner-radius-full`
+`--mod-mixed-button-background`
+`--mod-swatchgroup-border-color`
+`--mod-swatch-border-color-rgb`
+`--mod-swatch-border-opacity`
+
+##### Removed mods
+
+`--mod-swatchgroup-spacing-compact`
```
castastrophe, line 42: Can we include a "instead, please use..." part here for those using this mod in their systems currently?

components/swatch/index.css
```diff
@@ -86,7 +75,9 @@
 	/* contain selection indicator and focus-ring */
 	position: relative;
 
-	outline: none;
+	outline-width: 0;
+	outline-style: solid;
+	outline-color: transparent;
```
castastrophe, line 80: Oh interesting! This ends up being roughly the same as none doesn't it? What's the benefit of calling these out separately here?

components/swatch/stories/swatch.stories.js
```diff
@@ -31,18 +31,30 @@ export default {
 		rounding: {
 			name: "Rounding",
 			description: "The amount of corner rounding for a swatch.",
-			defaultValue: "regular",
+			defaultValue: "partial",
```
castastrophe, line 34: Should we document any of these API adjustments in our changeset?

components/swatch/stories/swatch.test.js
```diff
@@ -93,22 +105,41 @@ export const SwatchGroup = Variants({
...
+		{
+			testHeading: "Keyboard focus",
+			isKeyboardFocused: true,
+		},
+		{
+			testHeading: "Disabled",
+			isDisabled: true,
+		},
+		{
+			testHeading: "Hovered",
```
castastrophe, line 135: Makes me feel so good that our system is able to support these edge cases!! <3

components/swatch/stories/template.js
```diff
@@ -31,11 +35,19 @@ export const Template = ({
 		case "none":
 			borderStyle = "noBorder";
 			break;
+		case "border":
+			borderStyle = "border";
+			break;
 		case "light":
 			borderStyle = "lightBorder";
 			break;
 	}
```
castastrophe, line 44: Note to self, this is a great place to align more closely with the web components API in future

components/swatch/stories/template.js
```diff
@@ -44,20 +56,24 @@ export const Template = ({
...
 				"is-selected": !isDisabled && isSelected,
 				"is-disabled": isDisabled,
-				"is-image": isMixedValue || typeof imageUrl !== "undefined",
-				"is-mixedValue": !isDisabled && isMixedValue,
+				"is-hover": isHovered,
+				"is-active": isActive,
+				"is-keyboardFocused": isKeyboardFocused,
+				"is-image": (isMixedValue || isAddSwatch) || typeof imageUrl !== "undefined",
+				"is-mixedValue": !isDisabled && !isAddSwatch && isMixedValue,
+				"is-addSwatch": !isDisabled && !isMixedValue && isAddSwatch,
```
castastrophe, line 68: Are these new classes outlined in the changeset as well? I know you mentioned the add swatch but can you include the class and where to add it (i.e., on the root class) as well as maybe an example of it in use?

components/swatch/stories/template.js
```diff
@@ -44,20 +56,24 @@ export const Template = ({
...
 				[`${rootClass}--rectangle`]: typeof shape !== "undefined" && shape !== "square",
 				"is-nothing": !isDisabled && (typeof swatchColor === "undefined" || swatchColor === "transparent"),
 				...customClasses.reduce((a, c) => ({ ...a, [c]: true }), {}),
 			})}
 			?disabled=${isDisabled}
 			id=${ifDefined(id)}
 			style=${ifDefined(styleMap({
-				"--spectrum-picked-color": isMixedValue ? "var(--spectrum-gray-50)" : swatchColor,
+				"--spectrum-picked-color": pickedColor,
```
castastrophe, line 76: 💃

components/swatch/stories/template.js
```diff
@@ -72,7 +88,7 @@ export const Template = ({
 				updateArgs({ isSelected: !isSelected });
 			}}
 		>
-			${when((typeof imageUrl !== "undefined") && !isDisabled && !isMixedValue, () => html`
+			${when((typeof imageUrl !== "undefined") && !isDisabled && !isMixedValue && !isAddSwatch, () => html`
```
castastrophe, line 91: Should we include a note in the changeset that you can't use an image with the add swatch but I think you can use a color right?

components/swatchgroup/index.css
```diff
@@ -13,24 +13,27 @@
...
 	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-75);
+}
 
+.spectrum-SwatchGroup {
 	display: inline-flex;
 	flex-flow: row wrap;
 	align-items: flex-start;
 	justify-content: flex-start;
+	gap: var(--mod-swatchgroup-spacing, var(--spectrum-swatchgroup-spacing));
+}
```
castastrophe, line 24: Just curious if there was a reason to split up the SwatchGroup root classes from each other. Was it an organizational thing?

components/swatchgroup/index.css
```diff
@@ -13,24 +13,27 @@
...
 }
 
-/* Compact */
-.spectrum-SwatchGroup--compact {
-	--mod-swatchgroup-spacing-regular: var(--mod-swatchgroup-spacing-compact);
-	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-50);
+/* Medium */
+.spectrum-SwatchGroup--sizeM {
+	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-100);
 }
```
castastrophe, line 34: What was our take away on using medium? If it's the default are we just rolling it into the root class now like SWC does or do we want to support it separately? I know we said we don't want to require it though so the root class needs this value defined too if both are present.

components/swatchgroup/stories/swatchgroup.stories.js
```diff
@@ -25,7 +19,7 @@ import { RoundingTemplate, Template } from "./template.js";
  *
  * ### Density
  *
- * Swatch groups come in 3 densities: regular (default), compact, and spacious. Compact and spacious densities retain the same swatch size as regular density, but have less or more padding between each swatch, respectively.
+ * Swatch groups come in several densities: regular (default), spacious, medium and large. Each density retains the same swatch size, but have less or more padding between each swatch, respectively.
```
castastrophe, line 22: Interesting that the densities include spacious, medium, and large. What's the distinction between spacious vs. large? Do they set different properties?

castastrophe: I think I'd like to see T-shirt sizing be distinct from densities since we already have a convention for size="l" for example.

components/swatchgroup/stories/swatchgroup.stories.js
```diff
@@ -74,18 +61,15 @@ export default {
 				category: "Component",
 				defaultValue: { summary: "noBorder" },
 			},
-			options: ["noBorder", "lightBorder"],
+			if: { arg: "isSelected", truthy: false },
+			options: ["noBorder", "border", "lightBorder"],
```
castastrophe, line 65: Oh so we can't set the border value if the swatch is selected? Can we include a note about that in the docs or changeset (if that's a new rule)?

components/swatch/index.css
cdransf: It does! It lets me transition the `outline-color` (unless I can still do that with the shorthand? I'd be happy to be wrong on that one. ✨)

components/swatchgroup/index.css
cdransf: Yep! That's purely to keep the property and style declarations separate.

.changeset/proud-schools-reply.md
cdransf: `--spectrum-swatchgroup-border-color` was removed altogether and `--spectrum-swatch-icon-color` is now used in place of the dash specific token. I'll update the changeset to clarify the latter case. ✨

components/swatchgroup/stories/swatchgroup.stories.js
cdransf: Can do! `The border should not be set when the swatch is selected as the border conflicts with the display of the selected state.` ✨

components/swatch/index.css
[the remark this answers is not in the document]
rise-erpelding: Thank you @marissahuysentruyt! What I'm gathering from looking over that #2954 and your comment here, the changes we would expect to see from the current version would be something like:
* Expect to see the border opacity for `.spectrum-Swatch` should be 42% (it is currently), but the border opacity for swatches in `.spectrum-SwatchGroup` should be 20%.
* Expect to see no light border option for either component. (removal of --lightBorder variant and associated controls, documentation, etc.)
* I'm fuzzier on this, but it looks like swatches always have a border? So can we remove the no border option altogether? It looks like the Add swatch should not have a border. If we always have a border, perhaps the border control can be completely removed.
* Note in the changeset that this border option (or border options) have been removed
* Is it also helpful to state that the border opacity is different between swatch and swatch group?

components/swatchgroup/stories/swatchgroup.stories.js
cdransf: Yep! Just added that.

components/swatch/stories/swatch.test.js
```diff
@@ -93,22 +95,40 @@ export const SwatchGroup = Variants({
...
 		},
+	],
+	stateData: [
 		{
 			testHeading: "Empty",
 			swatchColor: undefined,
 			imageUrl: undefined,
 			isMixedValue: false,
-			borderStyle: "default",
+			isAddSwatch: false,
```
rise-erpelding, line 109: Since we're treating this as a state, the empty + selected and empty + keyfocused combos from the spec are not getting tested:
image

I'm also wondering if we should have empty swatches in the mixed and add variants. Since we have `isMixedValue: false` and `isAddSwatch: false` here they end up getting canceled out. Seems like the appropriate behavior would be that mixed value or add swatches override empty swatches (which they do when I play with the Storybook controls!).
image

I'm not sure what the right answer is. I like the idea of keeping empty as a state, but maybe refining this test case with `ignore:` to take out the mixed value and add swatches. And then perhaps we could also add empty into the `testData` array but use ignore/include to eliminate some of the unneeded states? (At least for sure we don't need empty x empty.)

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,12 +8,16 @@ export const SwatchgroupGroup = Variants({
...
 		},
 		{
-			testHeading: "Spacious",
-			density: "spacious",
+			testHeading: "Small",
+			density: "sizeS",
+		},
+		{
+			testHeading: "Large",
+			density: "sizeL",
```
rise-erpelding, line 20: I think the "Extra small", "Small", and "Large" here describe the density, right? Can we clarify that in the test heading so as not to be confused with the swatch size?

But looking into it further, I'm a bit confused about the density options, it looks like from the spec the spacing options are compact (same spacing for all swatch sizes), regular (same spacing for all swatch sizes), and spacious (these have different tokens for different sizes)?

I'm not sure we've ever applied tokens depending on the size on a child component, but if I'm understanding correctly that that's what we need to do here, I wonder if this could be a job for `:has()`?

components/swatchgroup/stories/swatchgroup.stories.js
```diff
@@ -55,37 +48,13 @@ export default {
...
-			...Swatch.argTypes.borderStyle,
-			defaultValue: "noBorder",
-			description: "Apply the `spectrum-Swatch--lightBorder` class to a swatch in the swatch group when it has a color contrast ratio of less than 3:1.",
-			table: {
-				type: { summary: "string", required: true },
-				category: "Component",
-				defaultValue: { summary: "noBorder" },
-			},
-			options: ["noBorder", "lightBorder"],
-		},
```
rise-erpelding, line 78: Did we decide we were going to eliminate border options completely for swatch and swatch group? There's only one type of border for swatch, a different type of border for swatch group, and then the no-border option is completely gone?

There's a lingering `borderStyle` arg in the swatch group controls here and also in both templates, we'll have to adjust/remove accordingly.

image

components/swatchgroup/index.css
```diff
@@ -12,25 +12,31 @@
  */
 
 .spectrum-SwatchGroup {
-	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-75);
+	--spectrum-swatchgroup-spacing: var(--spectrum-spacing-100);
+
+	/* Border opacity (swatches in a swatch group have a 20% opacity border instead of 42% default border opacity) */
+	--spectrum-swatch-border-opacity: 0.2;
```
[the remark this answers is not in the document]
rise-erpelding: We'll probably want to add a comment here to keep Stylelint happy.

image

rise-erpelding (changes requested):
Ok, this is getting closer! I think swatch is looking pretty nice aside from some minor things, and I had some questions about the spacing in swatch group.

A couple of things that I noticed that aren't from the migration work here and therefore I couldn't single them out in the code:
* I started playing with the empty/nothing variant a little more. It's nice that you can see it in Storybook by entering `transparent` for the color! This is entirely optional in my opinion, but it'd be nice to add that to the `swatchColor` control description, something like "Supports standard color input or any valid input for the <code>background</code> property such as <code>linear-gradient(red, blue)</code> or <code>transparent</code>." Not sure if there are any other fun swatches you can access just from changing `swatchColor` to something unexpected that isn't already covered. Just a thought, really up to you!
* Another thing I noticed relating to the empty/nothing variant is that our slash goes from top left to bottom right whereas the spec's goes from top right to bottom left. Should we update that here? Playing with it in the browser, I think we've just got the rotation a bit off, but it looked like changing the rotation for the rectangular variant results in the slash ending before it's filled the whole swatch. That one might need an adjustment to the `block-size` as well?
image image
image image

components/swatchgroup/stories/swatchgroup.stories.js
cdransf: I _believe_ so — @marissahuysentruyt may understand it better. But my understanding is that the individual swatches have border style options but those are superseded by the swatchgroup border styles. 😵‍💫

I went ahead and removed that `borderStyle` arg in keeping with that understanding. ✨

components/swatchgroup/stories/swatchgroup.test.js
cdransf: Alrighty, I took a look at these changes in https://github.com/adobe/spectrum-css/pull/3677/commits/e2289793a83debab39f9ac33bb22d1e41fe98660. `--spectrum-spacing-75` is set as the default (medium variation which is typically our default), then we have `sizeS` and `sizeL` where `sizeL` leverages `:has` to apply the appropriate density spacing based on the child swatch sizing class.

I split out tests with more informative headers for each density + sizing permutation. ✨

components/swatch/stories/swatch.test.js
cdransf: Hmm, I added `empty` to `testData` and then removed `empty`, `add` and `mixed value` from it in the state data section. I _think_ this makes more sense as it gives us more coverage for the empty states and removes the cases where add/mixed value override empty. ✨

components/swatchgroup/index.css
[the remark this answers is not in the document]
castastrophe: Curious, why it unused?

components/swatchgroup/stories/swatchgroup.stories.js
marissahuysentruyt: How @cdransf explained it above was also my understanding. (I may have missed that arg in my haste to push up that commit the other day- sorry!)

components/swatch/stories/swatch.stories.js
```diff
@@ -85,18 +90,34 @@ export default {
...
+		isActive: false,
+		isKeyboardFocused: false,
+		rounding: "partial",
 		swatchColor: "rgb(174, 216, 230)",
-		borderStyle: "default",
 		shape: "square",
 		isMixedValue: false,
+		isAddSwatch: false,
 	},
 	parameters: {
```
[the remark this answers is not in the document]
cdransf: Added both! ✨

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,12 +8,28 @@ export const SwatchgroupGroup = Variants({
...
+		},
+		{
+			testHeading: "Large density, medium swatches (spacious)",
+			density: "sizeL",
+			size: "m"
+		},
+		{
+			testHeading: "Large density, large swatches (spacious)",
+			density: "sizeL",
+			size: "l"
```
rise-erpelding, line 32: We're only demonstrating the combination of density with various swatch t-shirt sizes for the large/spacious swatches. We're leaving out compact spacing with other swatch sizes and regular spacing with other swatch sizes (I guess technically these are covered in the Sizing section at the bottom?).

But IMO, we can leave those out. And personally I'd be for eliminating the swatch sizing for large/spacious density and only showing it where it's covered by default in the Sizing section.

Just some thoughts! This can stay if you'd prefer!

components/swatchgroup/stories/template.js
```diff
@@ -12,13 +12,11 @@ export const Template = ({
 	customClasses = [],
 	size = "m",
 	density = "regular",
-	rounding = "regular",
+	rounding = "partial",
```
rise-erpelding, line 15: One more small recommendation, I don't know that this is specifically where we ought to make the change though:

I see in the spec guide in S2 / Desktop that it says:

> A corner rounding of “none” should be used in a swatch group that consists of more than one row. This is to help minimize the Hermann grid illusion that happens at the intersections of the white space within the group.

From that I'm gathering that it seems like the default rounding we'd want for a single swatch is partial rounding, while a swatch group's swatches should have no rounding. I think you might have had some iteration of this before, but without the distinction of the single swatch having rounding and the swatch group not having rounding.

In many of our swatch groups (especially with many of our examples having more than one row) this isn't the rule that we're following, the swatch groups have partial rounding by default. If we change the rounding in the template to default to no rounding, would that make the proper adjustments on the docs page and in the tests so that the swatches in a swatch group have no rounding except in the cases that we're showing the other rounding options?

components/swatch/stories/swatch.test.js
```diff
@@ -99,14 +99,16 @@ export const SwatchGroup = Variants({
...
 			testHeading: "Empty",
 			swatchColor: undefined,
 			imageUrl: undefined,
-			isMixedValue: false,
-			isAddSwatch: false,
+		},
+	],
+	stateData: [
+		{
+			testHeading: "Empty",
```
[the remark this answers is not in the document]
cdransf: _Huh_ — I'd thought it would inherit from the `testData`. Just restored:

```
swatchColor: undefined,
imageUrl: undefined,
```

✨

components/swatchgroup/index.css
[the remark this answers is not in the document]
cdransf: @rise-erpelding I think that makes sense — the re-assignment is specific to the group, but a mod feels better to me. ✨

tokens/custom/dark-vars.css
```diff
@@ -57,9 +57,8 @@
 
 	--spectrum-steplist-current-marker-color-key-focus: var(--spectrum-blue-700);
 
-	--spectrum-swatch-border-color: rgb(255 255 255 / 51%);
-	--spectrum-swatch-border-color-light: rgb(255 255 255 / 20%);
+	--spectrum-swatchgroup-border-color: rgb(255 255 255 / 20%);
```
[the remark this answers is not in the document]
cdransf: Nope! Removed from dark and the changelog as well. ✨

components/swatch/stories/swatch.stories.js
```diff
@@ -145,7 +169,7 @@ Disabled.parameters = {
 };
 
 /**
- * Default rounding and full rounding are usually used when a swatch is presented by itself near other components. A rounding of “none” is used in a swatch group to help minimize the Hermann grid illusion that happens at the intersections of white space in the group.
```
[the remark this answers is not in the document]
cdransf: Restoring this now. ✨

components/swatchgroup/stories/swatchgroup.test.js
cdransf: Ok! So, I've dropped those large/spacious combos you mentioned, so we can model those in the sizing section, but left small/compact as compact isn't demonstrated there (and I don't think we need different size combinations for that case? 😅) ✨

components/swatchgroup/stories/template.js
cdransf: Alrighty! I reverted the `partial` naming to regular and this argument to `partial`. I checked against `spectrum-two` and then set the default control for the `swatchgroup` story to `none` to align with that original implementation. So the `swatchgroup` tests match that previous state. ✨

components/swatch/stories/template.js
cdransf: I added usage notes on the add swatch and icon classes. Mixed value was pre-existing, so I left that out of the changeset. ✨

components/swatch/stories/swatch.stories.js
cdransf: This particular change was reverted to maintain backward compatibility/consistency (a new name but same functionality, so reverting to `regular` was perfectly safe). ✨

rise-erpelding (changes requested):
This is really close! These are mostly small Storybook/Template adjustments re: borders and the default swatch size!

components/swatch/index.css
```diff
@@ -11,29 +11,51 @@
  * governing permissions and limitations under the License.
  */
 
- /* @note: default sizing for swatch is small */
+/* @note: default sizing for swatch is small */
```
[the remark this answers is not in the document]
marissahuysentruyt: Oh this is a great question, and tricky. I don't unfortunately have any other context for the default size for a swatch. Prod says medium is the default for both swatch and swatch group, which I think we got from the design guidelines Rise listed. I probably added the documentation during the docs-to-storybook migration, but didn't touch the CSS comments that say the default size is small (since we were avoiding CSS updates).

Just looking at Figma, the each swatch group size (xs, s, m, l) gets passed to the swatches themselves, so a large swatch group has large swatches. If I open up or inspect the playground swatch group (which I _assume_ would be the default), everything is a medium. Does that mean that swatch's default size is also medium?

swatchgroup

That pattern however, where the playground component was the default, doesn't really carry into the swatch playground component though. They're all sorts of sizes and variants. 😢 The token specs file also uses medium and large swatches, so it's unclear to me still. 🤷‍♀️ Personally, I'd probably update the CSS to treat medium as the default.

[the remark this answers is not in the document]
marissahuysentruyt: Maybe there's more to come on swatch group- I just noticed that the token specs for swatch group are "partially published."

components/swatch/stories/template.js
```diff
@@ -114,6 +130,13 @@ export const Template = ({
 								}[size] || "100"),
 								useRef: false,
 							}, context)] : []),
+							...(isAddSwatch ? [Icon({
+								customClasses: [`${rootClass}-icon`],
+								setName: "workflow",
+								size,
+								iconName: "Add",
```
marissahuysentruyt, line 137: Could you make a ticket to follow up on this icon? Figma is calling for the Add UI icon (not the Add workflow icon), but it is part of that group of icons that's missing from our UI package right now. Larz is working on it though!

I just don't want us to forget to eventually use the correct icon. I can share the thread with you in Slack if you want to add it in the the follow-up ticket, too.

components/swatch/stories/template.js
```diff
@@ -31,11 +35,19 @@ export const Template = ({
 		case "none":
 			borderStyle = "noBorder";
 			break;
+		case "border":
+			borderStyle = "border";
+			break;
```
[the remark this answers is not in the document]
cdransf: Dropped these references ✨

.changeset/proud-schools-reply.md
```diff
@@ -0,0 +1,43 @@
...
+
+This migrates the `swatch` and `swatchgroup` components to the latest Spectrum 2 designs. Custom properties have been remapped and added per the design specification.
+
+An `Add Swatch` variant has been added with the required color tokens and the specified add UI icon.
+
+The `Add Swatch` and `Mixed Value` variants may not be combined with background colors or images. To implement the `Add Swatch` you need to apply the `is-addSwatch` class to the `spectrum-Swatch` element. `Add swatch` keyboard focus may be applied by adding the `is-keyboardFocused` class to this combination of classes.
+
+The `spectrum-Swatch-icon` is used to set the expected color to icons contained within the swatch.
+
+The border should not be set when the swatch is selected as the border conflicts with the display of the selected state.
```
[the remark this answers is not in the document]
cdransf: Updated that as well! ✨

components/swatch/index.css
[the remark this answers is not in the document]
cdransf: I went ahead and updated the default to medium — that's how we had the control set in Storybook/templates and we default to medium for other components. ✨

components/swatch/stories/template.js
```diff
@@ -146,70 +132,37 @@ export const Template = ({
 
 /* Shows a single group of swatches with all rounding options. */
 export const RoundingGroup = (args, context) => Container({
-	withBorder: false,
```
rise-erpelding, line 126: Can we revert the removal of these `withBorder` references in `Container` components? These affect the border of the container and not the swatch so I think they can stay!

image

components/swatch/stories/template.js
```diff
@@ -146,70 +132,37 @@ export const Template = ({
...
-		}, context)}
-	`
-}, context);
-
 /* Shows a single group of swatches that are empty/nothing in various shapes and rounding. */
 export const EmptyGroup = (args, context) => Container({
-	withBorder: false,
 	content: html`
 		${Container({
-			withBorder: false,
```
rise-erpelding, line 179: We can restore these also as they're on the Container!

image

components/swatch/stories/template.js
```diff
@@ -218,31 +171,25 @@ export const EmptyGroup = (args, context) => Container({
...
 export const DisabledGroup = (args, context) => Container({
-	withBorder: false,
 	content: html`
 		${Container({
-			withBorder: false,
 			withHeading: false,
 			content: Template(args, context),
 		}, context)}
 		${Container({
-			withBorder: false,
```
rise-erpelding, line 206: Also these!

image

components/swatch/stories/template.js
```diff
@@ -218,31 +171,25 @@ export const EmptyGroup = (args, context) => Container({
...
 			withHeading: false,
 			content: Template({...args, rounding: "full", }, context),
 		}, context)}
 	`
 }, context);
 
 export const SizingGroup = (args, context) =>Container({
-	withBorder: false,
```
rise-erpelding, line 214: These too!

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,13 +8,9 @@ export const SwatchgroupGroup = Variants({
 			testHeading: "Default",
```
rise-erpelding, line 8:
```suggestion
			testHeading: "Default (regular density)",
```

components/swatchgroup/stories/swatchgroup.test.js
rise-erpelding, line 1: Completely forgot to check this on the last review!

With the updates on density you've done, could we:
* Mention the density for Default
* Remove the "Small Size" in the heading for compact? (these are medium size swatches and referring to the density as small is confusing)
* Bring back the test heading for "Spacious"?

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,13 +8,9 @@ export const SwatchgroupGroup = Variants({
...
 		},
 		{
-			testHeading: "Compact",
+			testHeading: "Small size (compact density)",
```
rise-erpelding, line 11:
```suggestion
			testHeading: "Compact density",
```

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,13 +8,9 @@ export const SwatchgroupGroup = Variants({
...
 			density: "compact",
 		},
-		{
-			testHeading: "Spacious",
-			density: "spacious",
-		},
 		{
 			testHeading: "Full rounding",
```
rise-erpelding, line 15:
```suggestion
			testHeading: "Full rounding (regular density)",
```

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -8,13 +8,9 @@ export const SwatchgroupGroup = Variants({
...
 		},
 		{
-			testHeading: "Compact",
+			testHeading: "Small size (compact density)",
 			density: "compact",
 		},
-		{
-			testHeading: "Spacious",
-			density: "spacious",
-		},
```
rise-erpelding, line 17: Can we bring this back? I've made other suggestions here that have "density" in the test heading, so something like
```
			testHeading: "Spacious density",
			density: "spacious",
```

would be great if you want to go that route!

components/swatchgroup/stories/swatchgroup.test.js
```diff
@@ -23,28 +19,5 @@ export const SwatchgroupGroup = Variants({
 			testHeading: "Regular rounding",
```
rise-erpelding, line 19:
```suggestion
			testHeading: "Regular rounding (regular density)",
```

tokens/custom/dark-vars.css
rise-erpelding, line 1: One more thing on tokens! Can we add tokens to the changeset (or make a separate changeset for it)?

rise-erpelding (changes requested):
Just another handful of small things, I think this is just about there though!

components/swatch/stories/template.js
cdransf: Absolutely! ✨

components/swatchgroup/stories/swatchgroup.test.js
cdransf: Committed your suggestions and restored the spacious test heading. ✨

tokens/custom/dark-vars.css
cdransf: Yep! Added that to the existing changeset ✨

rise-erpelding (approved):
LOOKS GREAT!!

components/swatch/stories/template.js
```diff
@@ -114,6 +116,13 @@ export const Template = ({
 								}[size] || "100"),
 								useRef: false,
 							}, context)] : []),
+							...(isAddSwatch ? [Icon({
+								customClasses: [`${rootClass}-icon`],
+								setName: "workflow",
```
aramos-adobe, line 121: @cdransf I know we don't have the UI icon for Add just yet, but once it's updated we have to change the setName later on. It looks beautiful now with the workflow icon. Maybe we should have a `TODO` on updating this later.

cdransf: Ohh good catch! We added `CSS-1206` as a follow up for when that's available. ✨

aramos-adobe (approved):

PR #3680, opened 2025-04-18
chore: harden recently updated scripts

castastrophe:
## Description

Hind-sight is 20:20! I found a couple super minor updates to the scripts I recently worked on for the style-dictionary and Windows compatibility effort. I added notes inline to the code to help clarify these updates. It should just ensure we get more predictable results.

### Validation steps

- [x] `yarn builder tokens` (@rise-erpelding )

### Regression testing

Validate:

## To-do list

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -137,6 +137,8 @@ jobs:
             styles_modified_files: ${{ needs.changed_files.outputs.styles_modified_files }}
             eslint_added_files: ${{ needs.changed_files.outputs.eslint_added_files }}
             eslint_modified_files: ${{ needs.changed_files.outputs.eslint_modified_files }}
+            mdlint_added_files: ${{ needs.changed_files.outputs.mdlint_added_files }}
+            mdlint_modified_files: ${{ needs.changed_files.outputs.mdlint_modified_files }}
         secrets: inherit
 
     # -------------------------------------------------------------
--- a/.storybook/assets/base.css
+++ b/.storybook/assets/base.css
@@ -29,14 +29,29 @@ body {
 }
 
 .spectrum {
+	/* Gradient that changes with the color theme. */
+	--spectrum-examples-gradient: linear-gradient(45deg, var(--spectrum-magenta-1500), var(--spectrum-blue-1500));
+
+	/* Gradients that do not change with the color theme, for use in static color backgrounds. */
+	--spectrum-examples-gradient-static-black: linear-gradient(45deg, rgb(255 241 246), rgb(238 245 255));
+	--spectrum-examples-gradient-static-white: linear-gradient(45deg, rgb(64 0 22), rgb(14 24 67));
+
 	color: var(--spectrum-neutral-content-color-default);
 	background-color: var(--spectrum-background-base-color);
-	-webkit-tap-highlight-color: rgba(0, 0, 0, 0%);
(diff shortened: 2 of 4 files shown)


.github/workflows/development.yml
```diff
@@ -137,6 +137,8 @@ jobs:
             styles_modified_files: ${{ needs.changed_files.outputs.styles_modified_files }}
             eslint_added_files: ${{ needs.changed_files.outputs.eslint_added_files }}
             eslint_modified_files: ${{ needs.changed_files.outputs.eslint_modified_files }}
+            mdlint_added_files: ${{ needs.changed_files.outputs.mdlint_added_files }}
```
castastrophe, line 140: These exports don't seem to be working so we're linking just all the files!

.storybook/assets/base.css
```diff
@@ -29,16 +29,31 @@ body {
...
+	--spectrum-examples-gradient: linear-gradient(45deg, var(--spectrum-magenta-1500), var(--spectrum-blue-1500));
+
+	/* Gradients that do not change with the color theme, for use in static color backgrounds. */
+	--spectrum-examples-gradient-static-black: linear-gradient(45deg, rgb(255 241 246), rgb(238 245 255));
+	--spectrum-examples-gradient-static-white: linear-gradient(45deg, rgb(64 0 22), rgb(14 24 67));
+
 	color: var(--spectrum-neutral-content-color-default);
 	background-color: var(--spectrum-background-base-color);
-	-webkit-tap-highlight-color: rgba(0, 0, 0, 0%);
+	-webkit-tap-highlight-color: rgb(0, 0, 0, 0%);
```
castastrophe, line 41: This was just an auto-fix ala VSCode stylelinting

tasks/clean-up-after-migration.sh
```diff
@@ -13,11 +13,6 @@
 # After migrating from our old build system, there are a lot of empty folders hanging out still
 # This script aims to clean up those folders
 
-# Legacy tools folder (included storybook & generator)
-# test -d "tools" && rm -rf tools
-
-test -d "tools" && rm -rf tools
```
castastrophe, line 19: Not sure why this is still here - we like and need the tools folder!!

tokens/utilities/data-json-formatter.js
```diff
@@ -23,11 +23,11 @@ export const format = ({ dictionary, platform, file, options }) => {
 			if (data.ref) {
 				data.ref = convertRef(data.ref);
 
-				if (data.ref === data.value) {
+				if (data.ref?.toString() === data.value?.toString()) {
```
castastrophe, line 26: This just ensures that the ref is converted to the same type before comparing (sometimes they're numbers).

.storybook/assets/base.css
rise-erpelding: Can you say more about this and the other change below from hsla to hsl? I would have thought it'd be more correct to use rgba and hsla, but it seems like that's not the case.

rise-erpelding (approved):
No issues here! Just left a small question. Also, thanks so much for all the helpful comments you added in the PR!

.storybook/assets/base.css
castastrophe: I just learned about this the other day too! Apparently rgba and hsla are actually just aliases for rgb and hsl! PostCSS preset env tooling suggests we use the color function directly instead of the alias.

PR #3682, opened 2025-04-22
feat(tag): migrate to S2

rise-erpelding:
## Description

STILL TODO:
Clear button color on hover - currentColor doesn't seem like it works for states
Add removable hover/focus to testing grid for default, selected, and emphasized.

This work migrates the tag component to Spectrum 2 tokens and specifications:
- The invalid variant has been removed.
- Borders remain in high contrast but are otherwise transparent.
- Thumbnail has been added as a visual option.
- Tag now has a max-inline-size.
- Read-only tags are now supported.

Additionally, Tag has an expanded testing grid to test more states and state combinations, an updated Docs page in Storybook, and updated Storybook controls to demonstrate Tag's supported content and states.

There are a few outstanding questions that I have, but I don't consider these to be blocking and will make follow-up tickets as needed to address these items:
- Down state seems to behave erratically, but this appears to happen with other components too, navigating to and from the testing grid appears to affect this.
- We don't have WHCM specs for S2, so the styles here are best-guess.
- The `neutral-selected-background-color-default` and its accompanying hover, key-focus, down tokens don't appear to exist, I used `--spectrum-neutral-background-color-default` colors, which seem to mostly match the spec colors.
-  I saw that the S1 guidelines mentioned a read-only Tag, but this doesn't appear to have been implemented. This is pretty minor to support, it won't affect implementations if they don't use it, and could be removed easily. I'll follow up on whether or not we want this and if there are any additional guidelines on what it should look like.

### Validation steps
Here are some things that I generally look at when reviewing other people's migrations, plus a few specific callouts for this work that I'd love feedback on:
1. Check against token spec:
    - [ ] Validate that all tokens used appropriately/as spec'd?
2. Check through changed files:
    - [ ] Nothing out of scope appears to have been changed
    - [ ] Double-check for typos or anything else that feels inaccurate or illogical
4. Check Windows High Contrast Mode (I noted that this looked different in AssistivLabs and forced-colors rendering in Chrome, it's not a bad idea to check both if you have the bandwidth):
    - [ ] Colors are high contrast and make sense given the way other components are often styled
6. Check the Docs page:
    - [ ] No typos, information covers the changes appropriately
    - [ ] I left documentation about the removal of the invalid tag out, but could add it, open to opinions here!
8. Check the Storybook controls:
    - [ ] They work and cover all variants, states and combinations thereof
10. Check the testing grid:
    - [ ] Is there sufficient test coverage? Are there any states that aren't covered? Any ideas for organizing this better?
    - [ ] Specifically, I didn't cover focus + hover, is there any need to for Tag?
11. View read-only state. I'll likely start a Slack conversation about this, but I think this work can be reviewed without knowing the answer:
    - [ ] Read-only behaves as expected - they appear as normal tags, but you can't interact with them

### Regression testing

Validate:

## Screenshots
Spectrum 2:
image

Spectrum 1:
image

## To-do list


components/tag/index.css
```diff
@@ -12,195 +12,190 @@
...
 .spectrum-Tag {
-	/* TODO - replace placeholder disabled avatar opacity with correct token once avatar is migrated */
-	--spectrum-avatar-opacity-disabled: 0.3;
-
 	--spectrum-tag-animation-duration: var(--spectrum-animation-duration-100);
-	--spectrum-tag-border-width: var(--spectrum-border-width-100);
+
+	/* border color - transparent unless modified or high contrast mode */
+	--spectrum-tag-border-width: var(--spectrum-border-width-100); /* borders remain in WHCM but appear transparent otherwise */
+	--spectrum-tag-border-color: var(--highcontrast-tag-border-color, var(--mod-tag-border-color, transparent));
```
rise-erpelding, line 19: Colors have var-stacks at the top level, other properties tend to be numeric or change between sizes, so I left those as-is.

components/tag/index.css
```diff
@@ -12,195 +12,190 @@
...
+	--spectrum-tag-background-color-active: var(--highcontrast-tag-background-color-active, var(--mod-tag-background-color-active, var(--spectrum-gray-200)));
+	--spectrum-tag-background-color-focus: var(--highcontrast-tag-background-color-focus, var(--mod-tag-background-color-focus, var(--spectrum-gray-200)));
+
+	--spectrum-tag-background-color-selected: var(--highcontrast-tag-background-color-selected, var(--mod-tag-background-color-selected, var(--spectrum-neutral-background-color-default)));
+	--spectrum-tag-background-color-selected-hover: var(--highcontrast-tag-background-color-selected-hover, var(--mod-tag-background-color-selected-hover, var(--spectrum-neutral-background-color-hover)));
+	--spectrum-tag-background-color-selected-active: var(--highcontrast-tag-background-color-selected-active, var(--mod-tag-background-color-selected-active, var(--spectrum-neutral-background-color-down)));
+	--spectrum-tag-background-color-selected-focus: var(--highcontrast-tag-background-color-selected-focus, var(--mod-tag-background-color-selected-focus, var(--spectrum-neutral-background-color-key-focus)));
```
rise-erpelding, line 36: This doesn't match the token spec because the token used there doesn't appear to exist.

components/tag/index.css
```diff
@@ -12,195 +12,190 @@
...
+
+	--spectrum-tag-visual-spacing-inline-start: var(--spectrum-component-edge-to-visual-100);
+	--spectrum-tag-visual-spacing-inline-end: var(--spectrum-text-to-visual-100);
 
 	--spectrum-tag-icon-spacing-block-start: var(--spectrum-component-top-to-workflow-icon-100);
 	--spectrum-tag-icon-spacing-block-end: var(--spectrum-component-top-to-workflow-icon-100);
-	--spectrum-tag-icon-spacing-inline-end: var(--spectrum-text-to-visual-100);
 
-	--spectrum-tag-avatar-spacing-block-start: var(--spectrum-tag-top-to-avatar-medium); /* 6px 9px */
+	--spectrum-tag-avatar-size: var(--spectrum-avatar-size-100);
```
rise-erpelding, line 87: Avatar and thumbnail size don't correspond to the usual 75 = small, 100 = medium, 200 = large sizing conventions.

components/tag/index.css
```diff
@@ -209,54 +204,45 @@
...
-		--spectrum-clearbutton-fill-size: fit-content;
-		--spectrum-clearbutton-fill-background-color: transparent;
-		/* @passthrough end */
-
 		box-sizing: border-box;
 
-		padding-block-start: calc(var(--mod-tag-clear-button-spacing-block, var(--spectrum-tag-clear-button-spacing-block)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
-		padding-block-end: calc(var(--mod-tag-clear-button-spacing-block, var(--spectrum-tag-clear-button-spacing-block)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
+		margin-block-start: calc(var(--mod-tag-clear-button-spacing-block, var(--spectrum-tag-clear-button-spacing-block)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
+		margin-block-end: calc(var(--mod-tag-clear-button-spacing-block, var(--spectrum-tag-clear-button-spacing-block)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
```
rise-erpelding, line 228: The padding for block spacing and margin for inline spacing that was here before was making the cross icon's focus outline distorted. It's definitely too tight as it is now. I'm kind of at a loss about what to do with this, open to ideas.

S1:
image

S2:
image

components/tag/index.css
```diff
@@ -12,195 +12,190 @@
...
+	/* @passthrough start for nested components */
+	--mod-avatar-inline-size: var(--mod-tag-avatar-size, var(--spectrum-tag-avatar-size));
+	--mod-avatar-block-size: var(--mod-tag-avatar-size, var(--spectrum-tag-avatar-size));
+	--mod-thumbnail-size: var(--mod-tag-thumbnail-size, var(--spectrum-tag-thumbnail-size));
+	--mod-clear-button-width: var(--mod-tag-clear-button-size, var(--spectrum-tag-clear-button-size));
+	--mod-clear-button-height: var(--mod-tag-clear-button-size, var(--spectrum-tag-clear-button-size));
+	--mod-clear-button-background-color: transparent;
+	--mod-clear-button-icon-color: currentColor;
+	--mod-clear-button-icon-color-hover: currentColor;
+	--mod-clear-button-icon-color-key-focus: currentColor;
```
rise-erpelding, line 105: Using a `--mod` passthrough to set the icon color instead of setting the color on `.spectrum-Tag-clearButton`, I also noticed in S1 that if the clear button is hovered or focused it was turning a black color which didn't look great for emphasized or selected states, it _should_ now just inherit the content color from the label.

S1:
image

S2:
image

components/tag/index.css
```diff
@@ -12,195 +12,190 @@
...
+	border-color: var(--spectrum-tag-border-color);
+	background-color: var(--spectrum-tag-background-color);
+	color: var(--spectrum-tag-content-color);
 
 	border-radius: var(--mod-tag-corner-radius, var(--spectrum-tag-corner-radius));
 	border-width: var(--mod-tag-border-width, var(--spectrum-tag-border-width));
 	border-style: solid;
 
-	padding-inline-start: calc(var(--mod-tag-spacing-inline-start, var(--spectrum-tag-spacing-inline-start)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
+	padding-inline-start: calc(var(--mod-tag-label-spacing-inline, var(--spectrum-tag-label-spacing-inline)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
```
rise-erpelding, line 182: Inline spacing is the same for start and end if the content is only a label (no visuals or clear button), so this seemed like a good place to update the custom property as such.

components/tag/index.css
```diff
@@ -209,54 +204,45 @@
...
 		margin-block-end: calc(var(--mod-tag-avatar-spacing-block-end, var(--spectrum-tag-avatar-spacing-block-end)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)));
-		margin-inline-end: var(--mod-tag-avatar-spacing-inline-end, var(--spectrum-tag-avatar-spacing-inline-end));
+	}
+
+	.spectrum-Tag-itemIcon,
+	.spectrum-Avatar,
+	.spectrum-Thumbnail {
+		/* set visual inline-start spacing, but take away label spacing, this will result in negative margin - this calc does not need border width */
+		margin-inline-start: calc(var(--mod-tag-visual-spacing-inline-start, var(--spectrum-tag-visual-spacing-inline-start)) - var(--mod-tag-label-spacing-inline, var(--spectrum-tag-label-spacing-inline)));
+		margin-inline-end: var(--mod-tag-visual-spacing-inline-end, var(--spectrum-tag-visual-spacing-inline-end));
```
rise-erpelding, line 220: All visuals have the same inline spacing, but avatar and icon have specific block spacing tokens, and thumbnail has none, but is centered vertically so looks fine without them.

components/tag/index.css
```diff
@@ -306,106 +298,73 @@
...
+	color: var(--spectrum-tag-content-color-disabled);
 	pointer-events: none;
 
 	.spectrum-Avatar {
 		opacity: var(--mod-avatar-opacity-disabled, var(--spectrum-avatar-opacity-disabled));
 	}
+
+	.spectrum-Thumbnail {
+		opacity: var(--mod-thumbnail-opacity-disabled, var(--spectrum-thumbnail-opacity-disabled));
+	}
```
rise-erpelding, line 367: Not in the spec but seems to work fine for now, I'll follow up and we can change it later if needed.

components/tag/stories/tag.stories.js
```diff
@@ -100,15 +103,21 @@ export default {
 		},
 		packageJson,
 		metadata,
+		downState: {
+			selectors: [".spectrum-Tag:not(:disabled)"],
+		},
 	},
+	decorators: [
+		withDownStateDimensionCapture,
```
rise-erpelding, line 111: I've noticed this sometimes stops working if you switch to the testing grid and back, but this happens with other components too. Something else I'd like to follow up on. It generally works fine if you reload.

components/tag/stories/tag.stories.js
```diff
@@ -154,31 +169,37 @@ Emphasized.tags = ["!dev"];
...
+export const ReadOnly = TagsDefaultOptions.bind({});
+ReadOnly.storyName = "Read only";
+ReadOnly.tags = ["!dev"];
+ReadOnly.args = {
+	isReadOnly: true,
+};
+ReadOnly.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
```
rise-erpelding, line 202: This is a follow-up item also, it wasn't in the design spec but has existed in the S1 design guidelines.

components/tag/stories/template.js
```diff
@@ -35,32 +40,50 @@ export const Template = ({
...
 			)}
+			${when(thumbnailUrl, () =>
+				Thumbnail({
+					imageURL: thumbnailUrl,
+				}, context)
+			)}
 			<span class="${rootClass}-itemLabel">${label}</span>
 			${when(hasClearButton, () =>
 				ClearButton({
-					size,
```
rise-erpelding, line 63: Many of these (clear button, avatar, thumbnail) have their size set by CSS, there doesn't seem to be a lot of value to passing down the size in the template, but let me know if it seems like I'm missing something.

components/tag/index.css
rise-erpelding: Also the inline spacing is a headache here as it is for many of these components. Here's a small rundown of how it works (let's assume we're working with LTR to simplify the explanation):
- The tag label spacing on the left is set with this `padding-inline-start` here
- Tag label spacing on the right is set with `margin-inline-end` on `.spectrum-Tag-itemLabel`
- If the tag has any visuals (icon, thumbnail, avatar), there's less spacing from the left edge, so the left margin of the visual is set to the difference (visual spacing minus label spacing), which results in negative margin to move the spacing further left. Then the visual has a `margin-inline-end` to set the spacing between the visual and the label.
- If the tag has a clear button on the right, the clear button has a `margin-inline-start` for left side spacing set to the difference between that label-to-clear-icon spacing and the label-spacing tokens. That gives the clear button a negative margin for the small tag, but should zero out for other sizes. On the right side, `margin-inline-end` sets the spacing we want between the clear button and the edge.
- We generally subtract border width in most cases, but if we're taking the difference (those situations I mentioned above where we set negative margin), it's not necessary.

components/tag/stories/tag.test.js
```diff
@@ -1,8 +1,9 @@
 import { Variants } from "@spectrum-css/preview/decorators";
-import { Template } from "./template.js";
+import { TagsDefaultOptions, Template } from "./template.js";
 
 export const TagGroups = Variants({
 	Template,
+	SizeTemplate: TagsDefaultOptions,
```
rise-erpelding, line 6: Realized after looking at Swatch yesterday that this is a much cleaner way to handle showing a group for sizing, so I made a few adjustments to the template, I also realized we probably don't need the selected template anymore so removed it.

components/tag/index.css
marissahuysentruyt: Thanks for this explanation- it's pretty wild to get these spacing values correct. 

I did notice an inconsistency with the label's margin-inline-end when it's next to the clear button. I _think_ maybe we don't need to subtract the border width? The medium and large spacing between the label text and clear button is off by one if I compare to Figma. Technically, the small spacing checks out correctly, but that's because the token is set to 9px instead of 8px like Figma says. 

Unfortnately, I don't really have a solution to offer. I can spend some time and see what I come up with.

marissahuysentruyt (changes requested):
Another PR, back from the dead! Thanks for the explanation on the padding/margin stuff with all of the new options- very helpful. I think overall the tokens are in a really good spot- there was only one I didn't see in the CSS (about the down state).

I found some style criss-crossing when I started to really mix up the controls, so I left some ideas on how to expand the test file so we covered more of those. 

But I had just one extra thing to add, that I couldn't add to the code...

- There's a note about the tooltip on hover to reveal the full text in the docs for text overflow. Is it helpful at all the mention that our component doesn't do that? 😆

components/tag/index.css
rise-erpelding: Leaving a todo comment for myself, since we talked about this: we'll remove the clear button focus outline/remove the clear button from the tabindex. And it would also be nice to make sure that the target for the clear button is of an accessible size.

components/tag/stories/template.js
rise-erpelding: I think this question about which approach is better was the origin of why we decided that it would be nice to have a style guide (and it was also in relation to this tag migration the first time around). I'll take a look at this too, I think my initial thought is always "but we have a token for it, so we should use it" and it helps ensure that we get it shipped with the right size, but on the other hand, extra unnecessary code isn't always the best thing. I'll take another look at this one too.

components/tag/stories/tag.test.js
```diff
@@ -1,51 +1,102 @@
...
+		// read-only
+		{
+			testHeading: "Default, read-only",
+			isReadOnly: true,
+		},
+		{
+			testHeading: "Emphasized, read-only",
+			isReadOnly: true,
+			isEmphasized: true,
+		},
```
jawinn, line 67: This one was showing emphasized colors but I would expect it to show read-only colors.

components/tag/stories/template.js
```diff
@@ -35,32 +40,50 @@ export const Template = ({
...
+				if (isReadOnly || isDisabled) return;
+				updateArgs({ isSelected: !isSelected });
+			}}
+			@focusin=${function() {
+				if (isReadOnly || isDisabled) return;
+				updateArgs({ isFocused: true });
+			}}
+			@focusout=${function() {
+				updateArgs({ isFocused: false });
+			}}
```
jawinn, line 63: I'd recommend removing this, as the focus events are not the same behavior as `:focus-visible`. So currently if you click on the tag in the Default story, the focus indicator will appear when it shouldn't.

components/tag/stories/tag.stories.js
jawinn, line 1: Docs addition suggestion: I think it's worth mentioning in the Truncated story description that tags have a maximum width set. Since that might not be expected until it was encountered.

components/tag/index.css
```diff
@@ -265,26 +251,32 @@
...
+		transform: perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down))) translateZ(var(--spectrum-component-size-difference-down));
 	}
 
-	&:focus-visible,
-	&.is-focused {
-		border-color: var(--highcontrast-tag-border-color-focus, var(--mod-tag-border-color-focus, var(--spectrum-tag-border-color-focus)));
-		background-color: var(--highcontrast-tag-background-color-focus, var(--mod-tag-background-color-focus, var(--spectrum-tag-background-color-focus)));
-		color: var(--highcontrast-tag-content-color-focus, var(--mod-tag-content-color-focus, var(--spectrum-tag-content-color-focus)));
+	&:not(.is-disabled, .is-readOnly):focus-visible,
+	&:not(.is-disabled, .is-readOnly).is-focused {
```
jawinn, line 275: Do we still need `.is-focused` for S2 if it's already handled by the `:focus-visible`? Is this a legacy thing? If needed for the implementation to apply, it should probably be documented.

components/tag/index.css
jawinn, line 36: Could "selected" tokens be used for the value of existing custom properties in `.spectrum-Tag.is-selected`, instead of creating new custom properties? For example:
```
&.is-selected {
  --spectrum-tag-background-color: var(--highcontrast-tag-background-color-selected, var(--mod-tag-background-color-selected, var(--spectrum-neutral-background-color-default)));
  --spectrum-tag-content-color: var(--highcontrast-tag-content-color-selected, var(--mod-tag-content-color-selected, var(--spectrum-gray-25)));
  --spectrum-tag-border-color: var(--highcontrast-tag-border-color-selected, var(--mod-tag-border-color-selected, transparent));
}
```

jawinn (changes requested):

components/tag/index.css
rise-erpelding: It does fit better with medium and it looked like this avatar size is how React has it, so I changed it and moved it to my list of things to ask about, but to be fair, the avatar's margin is still slightly (2px I think) out of bounds in the large tag: 
image

components/tag/index.css
rise-erpelding: This one I haven't touched yet because of my concerns about clear button, wanting to expand its target, but realizing this might mean refactoring this spacing system.

components/tag/stories/tag.stories.js
rise-erpelding: There's a ticket for this!

components/tag/stories/tag.stories.js
rise-erpelding: I removed `user-select: none;` from the tag, so now the text is selectable (but not copy-able, is there a native way to do that in CSS?). Happy to restore it if that causes unintended issues though.

components/tag/stories/template.js
rise-erpelding: Ok, it is size-mapped in the template... but we're also using the tokens. Overkill? It feels like putting in the CSS is nice, because the visuals will be appropriately sized without needing to add it in the implementation, but is there an argument against that?

components/tag/stories/tag.stories.js
```diff
@@ -119,32 +128,38 @@ WithForcedColors.parameters = {
 };
 
 // ********* DOCS ONLY ********* //
-
+/**
+ * Tags should always include a label to represent search terms, filters, or keywords. Tags also
+ * have the option to include an icon, avatar, or thumbnail in addition to the label.
```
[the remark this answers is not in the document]
rise-erpelding: Links added!

components/tag/stories/tag.stories.js
```diff
@@ -154,31 +169,37 @@ Emphasized.tags = ["!dev"];
...
-Invalid.args = {
-	isInvalid: true
-};
-
-Invalid.parameters = {
-	chromatic: { disableSnapshot: true },
-};
-
+/**
+ * Tags have the option to be removable or not. Removable tags have a small close ("x") button.
```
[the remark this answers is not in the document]
rise-erpelding: The robots wrote that one, it's been corrected!

components/tag/stories/template.js
```diff
@@ -35,32 +40,50 @@ export const Template = ({
 					typeof size !== "undefined",
 				"is-emphasized": isEmphasized,
```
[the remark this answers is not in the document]
rise-erpelding: Eeesh, good catch. It's now `.spectrum-Tag--emphasized` (or `--emphasized` on the root class).

components/tag/stories/template.js
```diff
@@ -35,32 +40,50 @@ export const Template = ({
...
 				"is-disabled": isDisabled,
-				"is-invalid": isInvalid,
 				"is-selected": isSelected,
+				"is-hover": isHovered,
+				"is-focus-visible": isFocused,
+				"is-active": isActive,
+				"is-readOnly": isReadOnly,
```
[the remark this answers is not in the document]
rise-erpelding: Let me know if this seems like it's been resolved, the :not in the CSS is being applied differently now.

components/tag/stories/tag.test.js
```diff
@@ -1,51 +1,102 @@
...
-			testHeading: "Is removable",
+			testHeading: "Emphasized",
+			isEmphasized: true,
+		},
+		{
+			testHeading: "Selected",
+			isSelected: true,
+		},
+		{
+			testHeading: "Emphasized, selected",
```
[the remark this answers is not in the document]
rise-erpelding: Ok, I didn't make it super clear in the documentation, but can if that's helpful. But I do have it now so that emphasized unselected looks the same as default unselected, and the emphasized color only applies when selected. I'm wondering now if my testing could be improved in light of that change though. 🤔

components/tag/index.css
```diff
@@ -306,106 +298,73 @@
 
 /* selected */
 .spectrum-Tag.is-selected {
```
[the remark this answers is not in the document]
rise-erpelding: I think that's what I have now.

components/tag/index.css
```diff
@@ -306,106 +298,73 @@
...
-			background-color: var(--highcontrast-tag-background-color-invalid-selected-focus, var(--mod-tag-background-color-invalid-selected-focus, var(--spectrum-tag-background-color-invalid-selected-focus)));
-		}
+		border-color: var(--spectrum-tag-border-color-selected);
+		background-color: var(--spectrum-tag-background-color-selected-focus);
+		color: var(--spectrum-tag-content-color-selected);
 	}
 }
 
 /* emphasized */
 .spectrum-Tag.is-emphasized {
```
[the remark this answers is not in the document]
rise-erpelding: I think that's been updated, let me know if it doesn't seem correct!

components/tag/stories/tag.stories.js
```diff
@@ -80,14 +81,16 @@ export default {
 		rootClass: "spectrum-Tag",
 		size: "m",
```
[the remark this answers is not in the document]
rise-erpelding: Small is the default size now! With extra comments and documentation so devs know it's deliberate.

components/tag/index.css
```diff
@@ -265,26 +251,32 @@
...
-		background-color: var(--highcontrast-tag-background-color-active, var(--mod-tag-background-color-active, var(--spectrum-tag-background-color-active)));
-		color: var(--highcontrast-tag-content-color-active, var(--mod-tag-content-color-active, var(--spectrum-tag-content-color-active)));
+	&:not(.is-disabled, .is-readOnly):active {
+		/* mod/highcontrast specified at top */
+		border-color: var(--spectrum-tag-border-color-active);
+		background-color: var(--spectrum-tag-background-color-active);
+		color: var(--spectrum-tag-content-color-active);
+
+		/* stylelint-disable-next-line spectrum-tools/no-unknown-custom-properties -- height and width are set by implementations */
+		transform: perspective(max(var(--spectrum-downstate-height), var(--spectrum-downstate-width) * var(--spectrum-component-size-width-ratio-down))) translateZ(var(--spectrum-component-size-difference-down));
```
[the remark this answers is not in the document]
rise-erpelding: My understanding based on how this is documented on our [Foundations page](https://pr-3682--spectrum-css.netlify.app/?path=/docs/foundations-down-state--docs) is that we use minimum perspective for little components, and calculated perspective (which needs the down state decorator) for anything with a width larger than 24px. It's true that it's possible for `tag-minimum-width-small` to be 21px for desktop, but I believe in that case the height for the small tag would be 24px and so, using the max formula to determine whether the height or width is larger, we'd get the same perspective value as if we'd used the `--component-size-minimum-perspective-down` token (which is 24px).

My guess is that `--component-size-minimum-perspective-down` might be included in the spec as sort of a catchall that they add to all components that receive the down state treatment?

components/tag/index.css
```diff
@@ -295,7 +287,7 @@
 			inset-block-end: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
 			inset-inline-start: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
 			inset-inline-end: calc(-1 * var(--mod-tag-focus-ring-gap, var(--spectrum-tag-focus-ring-gap)) - var(--mod-tag-border-width, var(--spectrum-tag-border-width)) - var(--mod-tag-focus-ring-thickness, var(--spectrum-tag-focus-ring-thickness)));
```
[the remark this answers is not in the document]
rise-erpelding: Done!

components/tag/stories/template.js
rise-erpelding: Ok! I'm using the keyboard focused instead of focused control now as well. I think that I'd thought it would be nice to see the control toggling when you keyboard focus, but that does have unintended side effects.

components/tag/stories/tag.stories.js
rise-erpelding: This shouldn't have had a maximum width set on the custom styles, since tag already has a max inline size set. I updated this story to show the truncated, max-width tag for all 3 sizes.

components/tag/index.css
rise-erpelding: I'll have to come back to this one, I made an attempt but started getting tangled when it came to hover/focus/active states. Is button a good component to reference to see where else it's done like this?

components/tag/index.css
```diff
@@ -12,195 +12,189 @@
...
+	min-inline-size: var(--mod-tag-min-inline-size, var(--spectrum-tag-min-inline-size));
 	position: relative;
 	display: inline-flex;
 	align-items: center;
 	box-sizing: border-box;
 	vertical-align: bottom;
-	max-inline-size: 100%;
+	max-inline-size: calc(var(--mod-tag-height, var(--spectrum-tag-height)) * var(--mod-tag-maximum-width-multiplier, var(--spectrum-tag-maximum-width-multiplier)));
 	outline: none;
-	user-select: none;
```
rise-erpelding, line 198: Is there a reason to keep this?

components/tag/stories/template.js
marissahuysentruyt: I agree! We should probably keep it in the CSS to ensure the sizes get defined with the tokens. 👍

components/tag/index.css
rise-erpelding: Done in 96247550faf1b2c87f7a9a9cb042d9d8acf539a8

components/tag/index.css
rise-erpelding: Resolving this as I think what I have now brings clear button to an accessible target size and matches the spacing in the spec.

components/tag/stories/tag.stories.js
marissahuysentruyt: Is the text still not copy-able for either of you? I can copy and paste it from a read-only tag.
Screenshot 2025-05-07 at 5 04 28 PM

> Tag label that i copied

components/tag/index.css
aramos-adobe: @rise-erpelding I think it's worth keeping it. `user-select: none` prevents selection or highlighting of text elements. If you double click it you'll see the text element is highlighted by the browser. If you add `user-select: none` the highlight disappears.

Screenshot 2025-05-07 at 5 19 15 PM

components/tag/index.css
marissahuysentruyt: @rise-erpelding I'm not certain where the discrepancy is, so don't take the fact that I'm commenting here to heart too much! This felt sort of like a similar conversation...

I'm wondering about the truncation and the clear button now. I can really smoosh that clear button to the edge of the tag with a long label. How do we avoid that?

Screenshot 2025-05-07 at 4 19 14 PM
Screenshot 2025-05-07 at 4 15 55 PM

.changeset/khaki-icons-hammer.md
```diff
@@ -0,0 +1,62 @@
+---
+"@spectrum-css/tag": major
+---
+
+Tag now uses Spectrum 2 tokens and specifications:
+
+- The invalid variant has been removed.
```
aramos-adobe, line 7: I would also add it's been removed because there are no individual invalid tags and errors are displayed as tag groups.

marissahuysentruyt (changes requested):
Thanks- a bunch of my initial feedback is complete and addressed! 

I think my biggest curiosity, that I haven't quite worked out yet, is the truncation and close button issue I mentioned. Maybe we need 
```
position: absolute;
inset-inline-end: calc(that-you-already-have); // margin-inline-end changes to inset-inline-end
```
And then maybe some max-inline-size control for the label itself, so it doesn't "go past" the clear button? Sort of like the wacky padding we have for inputs to accommodate icons within fields?
Screenshot 2025-05-07 at 5 38 54 PM
Screenshot 2025-05-07 at 5 43 17 PM

components/tag/index.css
```diff
@@ -209,54 +237,49 @@
...
+	.spectrum-Tag-itemIcon,
+	.spectrum-Avatar,
+	.spectrum-Thumbnail {
+		/* set visual inline-start spacing, but take away label spacing, this will result in negative margin - this calc does not need border width */
+		margin-inline-start: calc(var(--mod-tag-visual-spacing-inline-start, var(--spectrum-tag-visual-spacing-inline-start)) - var(--mod-tag-label-spacing-inline, var(--spectrum-tag-label-spacing-inline)));
+		margin-inline-end: var(--mod-tag-visual-spacing-inline-end, var(--spectrum-tag-visual-spacing-inline-end));
 	}
 
 	/* clear button */
 	.spectrum-Tag-clearButton {
```
aramos-adobe, line 262: @rise-erpelding, I also spotted what @marissahuysentruyt discovered when the longer the string in truncation mode the clear button magically slides over to the edge. One thing I tried to add:

```css
.spectrum-Tag-clearButton {
    inset-inline-end: var(--spectrum-tag-edge-to-clear-icon);
    position: relative;
    padding-inline-start: var(--spectrum-tag-label-to-clear-icon);
}
```

The longer the text it doesn't move over.

components/tag/stories/tag.stories.js
rise-erpelding: Yeah, usually it resolves itself and starts defining those custom props again if you reload, at least for me. I think that's mentioned in the ticket but if you're seeing otherwise, please add to it!

components/tag/stories/tag.stories.js
rise-erpelding: I don't believe that functionality is native to HTML or CSS; it'd be a JavaScript implementation. I don't think I'm mentioning that in the docs anywhere but should I? I would hope that consumers would look at the design guidelines instead of solely trusting our Storybook documentation, but I know sometimes we mention that a tooltip or other thing that's outside of the scope of our implementation should be available.

rise-erpelding:
New changes! ✨ 

- test(tag): refactor testing template
    I'm hoping I didn't lose any test cases in doing this. Selected states are now combined with unselected states. Since we have flex-shrink problems sometimes, I'm also adding visuals (icon, avatar, thumbnail) and visuals + clear button to the truncated section to test those with longer text.
- fix(tag): keep clear button from shrinking
    The clear button is a flex item! It just needed a `flex-shrink: 0;` to keep it from shrinking with longer text, I think.
- chore(tag): add migrated tag
    Adds the correct taggage.
- chore(tag): update changeset
    Explains the removal of the invalid variant a little better.

rise-erpelding:
Forgot a couple of notes in my haste to reopen this for review!

I've noticed that sometimes the truncated text doesn't go to the edge of the itemLabel (like inside of the blue in the screenshot here), making it look like there's extra space. I think this is ok and just dependent on whatever is in the label text.
Screenshot 2025-05-09 at 8 45 22 AM

This is an older screenshot of the testing grid but the emphasized + selected colors in WHCM does bug me, but it's not different from what we have for emphasized in Foundations. I'd like to clarify WHCM specs at some point and we do have a ticket for that.
Screenshot 2025-05-09 at 8 31 01 AM

components/tag/stories/tag.stories.js
```diff
@@ -192,16 +223,22 @@ Sizing.parameters = {
...
+	Template: TagGroups,
+	withHeading: false,
+	withBorder: false,
+	...args,
+}, context);
 TextOverflow.tags = ["!dev"];
 TextOverflow.args = {
 	hasIcon: true,
 	iconName: "CheckmarkCircle",
 	label: "An example of text overflow behavior. When the button text is too long for the horizontal space available, it will truncate and stay on one line.",
```
aramos-adobe, line 241: Super minor, but you mentioned "when the tag text" in the JS Doc but says "button text" here in the label prop - should be "tag text" too

.changeset/khaki-icons-hammer.md
```diff
@@ -0,0 +1,62 @@
...
+- Default t-shirt size for tag is small.
+- Clear button has accessibly sized target.
+- Emphasized tags have accent coloring only when they are selected.
+
+Changed mods:
+
+- "--mod-tag-avatar-spacing-inline-end" --> "--mod-tag-visual-spacing-inline-end",
+- "--mod-tag-icon-spacing-inline-end" --> "--mod-tag-visual-spacing-inline-end",
+- "--mod-tag-spacing-inline-start" --> "--mod-tag-label-spacing-inline",
+- "--mod-tag-label-spacing-inline-end" --> "--mod-tag-label-spacing-inline"
```
aramos-adobe, line 22: Add a comma at the end here

components/tag/stories/tag.stories.js
```diff
@@ -100,15 +103,24 @@ export default {
...
+		downState: {
+			selectors: [".spectrum-Tag:not(:disabled)"],
+		},
+	},
+	decorators: [
+		withDownStateDimensionCapture,
+	],
+	status: {
+		type: "migrated",
 	},
```
marissahuysentruyt, line 115: This `status` object should go in `parameters` (at line 109ish) in order to get picked up properly- sorry I didn't clarify in my last comment.

components/tag/index.css
marissahuysentruyt: @rise-erpelding I agree with the removal, and don't see a reason to keep `user-select: none`. If we want to be able to copy and highlight that text, like [the guidelines](https://spectrum.corp.adobe.com/page/tag/#Read-only) say for read-only, and I think you and Josh already we talking about, removing `user-select: none` should give us that ability.

components/tag/stories/template.js
```diff
@@ -85,36 +113,23 @@ export const TagsDefaultOptions = ({
 	},
 	content: html`
 		${Template(args, context)}
-		${!args.isInvalid ?
-			Template({
+		${args.isRemovable ? "" : Template({
+			...args,
+			isRemovable: true,
+			}, context)}
```
marissahuysentruyt, line 119: If I haven't said it already, this is solution to showing the removable tags in this template is great! 
swedishchefkiss

marissahuysentruyt (approved):
Wonderful work on this, Rise! I'm going to approve, but I do suggest making the change on the location of the `status` object for the migrated badge before we merge! Looks like there's a few from Aziz from Friday, but overall, the migration looks great to me, and all of my concerns from the clear button are ffffiiiixxxed! Thanks for all your hard work!

components/tag/stories/tag.stories.js
rise-erpelding: I definitely should have been checking to see if this was doing anything 😂

components/tag/index.css
rise-erpelding: I think slapping a `user-select: none;` next to the `pointer-events: none;` for disabled is a good call. Made that addition.

PR #3684, opened 2025-04-23
feat(accordion): spectrum 2 migration

jawinn:
### Accordion Spectrum 2 migration
Accordion now uses Spectrum 2 tokens and specifications. New sized tokens are used for corner rounding, the spacing around the disclosure icon, and the spacing around the content area.

A few notes on tokens: **Medium top-to-text spacing tokens** appear to be the wrong values (they appear to be the large values, not the values seen in the spec), I've followed up on this with design and assume that the correct values will be added later. **Background color tokens** also don't match the spec, I've followed up on this as well and have temporary colors that match the spec in place. Finally, so far we've noticed that the design spec appears to have a **focus indicator gap** but the token spec is 0 (no gap), so that is another follow-up item. The **`accordion-item-to-divider`** token is not used here per previous guidance from the designer.

#### New features

- Adds the optional "quiet" style, which does not show dividers between accordion items.
- Adds the "no inline padding" variant. This variant will be used within "standard panel".
- Adds CSS transition to animate the rotation of the disclosure indicator when expanding and
  collapsing the accordion item.
- Per design spec, accordion items have a min-width and default width. Default width can be
  overridden with `--mod-accordion-item-width` but should not be smaller than
  `--spectrum-accordion-item-minimum-width`.

#### Markup changes

The HTML markup has changed slightly for the accordion header. The disclosure icon has been moved
from outside the the button (`spectrum-Accordion-itemHeader`), to within the button. The extra
element with class `spectrum-Accordion-itemIconContainer`, previously wrapping the icon, has
been removed. A span with class `spectrum-Accordion-itemTitle` has been added around the heading
text.

#### Mod changes

The following `--mod` custom properties have been renamed to better reflect how they are used:

- `--mod-accordion-item-height` has been renamed to `--mod-accordion-item-minimum-height`
- `--mod-accordion-min-block-size` has been renamed to `--mod-accordion-item-min-block-size`
- `--mod-accordion-component-edge-to-text` has been renamed to `--mod-accordion-content-padding-inline`

New mods:

- `--mod-accordion-animation-duration`
- `--mod-accordion-edge-to-content-area`
- `--mod-accordion-item-minimum-width`
- `--mod-accordion-top-to-disclosure-indicator`

CSS-1018

Note: this does not include the new "direct actions" feature. That will be done in a follow-up PR/task.

### Validation steps

1. Open the [storybook](https://pr-3684--spectrum-css.netlify.app/?path=/docs/components-accordion--docs) for the accordion component:
    - [ ] Verify the boolean Quiet Style Variant has been added. When applied it should not show dividers between accordion items.
    - [ ] Verify the No Inline Padding variant has been added . When applied it should remove inline padding from the accordion.
    - [ ] Verify working CSS transition that animates the rotation of the disclosure indicator when expanding and
  collapsing the accordion item
    - [ ] Verify all related documentation has been added
    - [ ] Verify token additions align with token spec (see notes above in the description)

### Regression testing

Validate:

## To-do list


components/accordion/stories/accordion.test.js
```diff
@@ -39,8 +202,8 @@ export const AccordionGroup = Variants({
 	],
 	stateData: [
 		{
-			testHeading: "Disabled",
-			disableAll: true,
+			testHeading: "Accordion item states",
+			items: statesTestContent,
```
rise-erpelding, line 206: This is great!

components/menu/dist/metadata.json
rise-erpelding, line 1: Can we double-check the latest `spectrum-two` to determine if we still need to commit these metadata files for menu, swatch, and swatch group?

.changeset/chatty-lands-attack.md
```diff
@@ -0,0 +1,33 @@
...
+Accordion now uses Spectrum 2 tokens and specifications.
+
+New sized tokens are used for corner rounding, the spacing around the disclosure icon, and
+the spacing around the content area.
+
+#### New features
+
+- Adds the optional "quiet" style, which does not show dividers between accordion items.
+- Adds CSS transition to animate the rotation of the disclosure indicator when expanding and
+  collapsing the accordion item.
```
rise-erpelding, line 16: I think something about the no inline padding variant would also go here.

rise-erpelding (changes requested):
Nice work, I did some review on the documentation for a first pass! I will also take a look at the CSS and tokens but decided to put that in a separate review since those usually take a bit to validate and I already have quite a few comments in this review.

components/accordion/index.css
```diff
@@ -46,31 +50,34 @@
...
-
-	/* Focus indicator */
-	--spectrum-accordion-focus-indicator-color: var(--spectrum-focus-indicator-color);
-
-	--spectrum-accordion-min-block-size: max(
-		var(--spectrum-accordion-item-height),
-		calc(var(--mod-accordion-item-header-top-to-text-space, var(--spectrum-accordion-item-header-top-to-text-space)) + var(--mod-accordion-item-header-bottom-to-text-space, var(--spectrum-accordion-item-header-bottom-to-text-space)) + (var(--mod-accordion-item-header-font-size, var(--spectrum-accordion-item-header-font-size)) * var(--mod-accordion-item-header-line-height, var(--spectrum-accordion-item-header-line-height))))
+	/* Calculated minimum block-size that accounts for density spacing changes and/or increases in font size and line-height.
+	   This gives us the block-size of the header with a single line of text. */
+	--spectrum-accordion-item-min-block-size: max(
```
marissahuysentruyt, line 69: holy smokes this is such an intense custom prop! Just for my own sanity, `--spectrum-accordion-item-min-block-size` is the larger value between: 

the defined accordion-item-minimum-height (whether it's a mod or not)

AND

the top padding + the bottom padding + whatever the value is when you multiply the font size by the line height

frog-wow-scroll

components/accordion/index.css
```diff
@@ -218,124 +262,78 @@
...
-	&:focus-visible {
+	&:focus-visible &:not(.is-disabled) {
 		border-radius: var(--mod-accordion-corner-radius, var(--spectrum-accordion-corner-radius));
-
-		outline: var(--mod-accordion-focus-indicator-thickness, var(--spectrum-accordion-focus-indicator-thickness)) solid var(--mod-accordion-focus-indicator-color, var(--spectrum-accordion-focus-indicator-color));
 		background-color: var(--mod-accordion-background-color-key-focus, var(--spectrum-accordion-background-color-key-focus));
 		color: var(--mod-accordion-item-header-color-key-focus, var(--spectrum-accordion-item-header-color-key-focus));
-		outline-offset: calc(var(--mod-accordion-focus-indicator-gap, var(--spectrum-accordion-focus-indicator-gap)) * -1);
+		outline: var(--mod-accordion-focus-indicator-thickness, var(--spectrum-accordion-focus-indicator-thickness)) solid var(--mod-accordion-focus-indicator-color, var(--spectrum-accordion-focus-indicator-color));
+		outline-offset: var(--mod-accordion-focus-indicator-gap, var(--spectrum-accordion-focus-indicator-gap));
```
marissahuysentruyt, line 307: `--spectrum-accordion-focus-indicator-gap` is the actual token, so should we be creating a similar custom property to use in our styles, instead of directly using the token? I thought that was our preference, unless that's changed.

components/accordion/index.css
```diff
@@ -218,124 +260,78 @@
...
 		color: var(--mod-accordion-item-header-color-hover, var(--spectrum-accordion-item-header-color-hover));
 		background-color: var(--mod-accordion-background-color-hover, var(--spectrum-accordion-background-color-hover));
-
-		+ .spectrum-Accordion-itemIconContainer {
-			color: var(--mod-accordion-item-header-color-hover, var(--spectrum-accordion-item-header-color-hover));
-		}
 	}
 
-	&:focus-visible {
+	&:focus-visible &:not(.is-disabled) {
```
marissahuysentruyt, line 302: I'm curious if we need the space in this selector. With the space between `&:focus-visible` and `&:not(.is-disabled)`, I don't actually see any of the accordion interactive states (hover, focus, active, active+hovered, etc). 

If I remove the space, I see the styles. If I remove the space and the next ampersand, I see the styles. I _don't_ however, know which of those options are the right fix.

marissahuysentruyt: Actually, looking at the token specs more, it almost looks like the focus ring applies to the `.spectrum-Accordion-item`, as opposed to `.spectrum-Accordion-itemHeader` (which is the button element and probably what is _actually_ getting the focus). Josh and/or SaaSha may have already asked if it's ok that our border-block-end peeks through behind the focus ring.

Figma (the divider gets cut off by the focus ring)
Screenshot 2025-06-11 at 5 33 11 PM

Ours (the divider shows behind the focus ring)
Screenshot 2025-06-11 at 5 32 22 PM

.changeset/chatty-lands-attack.md
```diff
@@ -0,0 +1,34 @@
...
+text.
+
+#### Mod changes
+
+The following `--mod` custom properties have been renamed to better reflect how they are used:
+
+- `--mod-accordion-item-height` has been renamed to `--mod-accordion-item-minimum-height`
+- `--mod-accordion-item-width` has been renamed to `--mod-accordion-item-minimum-width`
+- `--mod-accordion-min-block-size` has been renamed to `--mod-accordion-item-min-block-size`
+- `--mod-accordion-component-edge-to-text` has been renamed to `--mod-accordion-content-padding-inline`
```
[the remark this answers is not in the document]
rise-erpelding: I updated the changeset in 423ee339d. It does not include `--mod-accordion-item-padding-block` because I've removed that custom property and its mod after reading through the Slack thread Josh left behind where the designer confirmed that it would not be necessary to use that token.

components/accordion/index.css
rise-erpelding: Yep, I'm not touching this one if I don't have to.

components/accordion/stories/accordion.stories.js
```diff
@@ -200,8 +186,56 @@ Spacious.args = {
 Spacious.parameters = {
 	chromatic: { disableSnapshot: true },
 };
-Spacious.storyName = "Density - Spacious";
+Spacious.storyName = "Density: Spacious";
+
+/**
+ * Individual accordion items (of class `.spectrum-Accordion-item`) can be styled as disabled by applying the `is-disabled` class.
+ * This example markup also applies the `disabled` attribute on the heading button.
```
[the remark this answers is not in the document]
rise-erpelding: Our implementation closes them when you disable all items, but what SWC has doesn't. There isn't any guidance either way in the design guidelines, so I'd feel better leaving this out for now.

components/accordion/stories/accordion.stories.js
```diff
@@ -68,101 +82,19 @@ export default {
...
 		packageJson,
 		metadata,
+		status: {
+			type: "migrated",
+		},
+		tags: ["migrated"],
 	},
 };
 
-/* Content sourced from: https://www.adobe.com/products/catalog.html#:~:text=Frequently%20asked%20questions. */
```
rise-erpelding, line 74: The removal of this chunk means we're using the content from the accordion.test.js file. There are some smaller differences in the content; this version looks like it's a bit longer in the "worth it?" section, and adds a "how much" section that's missing from the tests, plus this version does _not_ include notes about whether or not typography was used. Overall though, I still think it's worthwhile to reuse the same example rather than use repeated, overlapping content in both.

Open to suggestions on the test content, there are a couple of sections that don't note how the content is styled, but if we'd like to add that, I can!

marissahuysentruyt (changes requested):
Thanks for getting this polished up! Overall, I think things are in a really good place. Most of my initial comments were taken care of, so I resolved those- thank you!

I will admit, that I have not tracked down those threads that Josh and SaaSha and Kami had, or dug through the backlog. I think partly related to that is my question about the switch & action button options in the "Embedded components" section in Figma. Are we waiting to implement those separately? Do we have follow up tickets already created?

components/accordion/index.css
marissahuysentruyt: I left another comment about it but I'm not convinced it's _just_ the gap that would need some attention. 

This screenshot has the 2px gap on that bottom focused item, and that makes the overall width of that item too large. 

Screenshot 2025-06-13 at 3 02 26 PM

.changeset/chatty-lands-attack.md
```diff
@@ -0,0 +1,44 @@
...
+- Adds no inline padding variant (`.spectrum-Accordion--noInlinePadding`) in which the item
+  header does not have padding.
+- Per design spec, accordion items have a min-width and default width. Default width can be
+  overridden with `--mod-accordion-item-width` but should not be smaller than
+  `--spectrum-accordion-item-minimum-width`.
+
+#### Markup changes
+
+The HTML markup has changed slightly for the accordion header. The disclosure icon has been moved
+from outside the the button (`spectrum-Accordion-itemHeader`), to within the button. The extra
```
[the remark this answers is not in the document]
rise-erpelding: Updated in 5954349442e06f7c6eb1ab26c37cfa0d94c9d1ea!

components/accordion/index.css
```diff
@@ -218,124 +261,79 @@
...
 		border-radius: var(--mod-accordion-corner-radius, var(--spectrum-accordion-corner-radius));
-
-		outline: var(--mod-accordion-focus-indicator-thickness, var(--spectrum-accordion-focus-indicator-thickness)) solid var(--mod-accordion-focus-indicator-color, var(--spectrum-accordion-focus-indicator-color));
-		background-color: var(--mod-accordion-background-color-key-focus, var(--spectrum-accordion-background-color-key-focus));
-		color: var(--mod-accordion-item-header-color-key-focus, var(--spectrum-accordion-item-header-color-key-focus));
-		outline-offset: calc(var(--mod-accordion-focus-indicator-gap, var(--spectrum-accordion-focus-indicator-gap)) * -1);
+		background-color: var(--spectrum-accordion-background-color-key-focus);
+		color: var(--spectrum-accordion-item-header-color-key-focus);
+		outline: var(--mod-accordion-item-focus-indicator-thickness, var(--spectrum-accordion-item-focus-indicator-thickness)) solid var(--spectrum-accordion-item-focus-indicator-color);
+		outline-offset: var(--mod-accordion-item-focus-indicator-gap, var(--spectrum-accordion-item-focus-indicator-gap), 0);
```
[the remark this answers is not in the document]
rise-erpelding: Ope, I see the mistake now.

I did ask about this and Kami said that we do want the regular 2px focus gap as shown in the design. I asked about whether I should be using the standard spectrum-focus-indicator-gap token instead and haven’t gotten a reply as of now, but I’m going to assume that’s the one to use since it’s the one we have everywhere else.

I changed out the token, how does that look? I did assume the focus indicator was put on `.spectrum-Accordion-itemHeader` and not the item because that’s the button, but maybe I’m missing something?

components/accordion/stories/accordion.test.js
```diff
@@ -1,46 +1,208 @@
...
+			content: "Yes, Adobe offers free products like Acrobat Reader, Aero, Fill & Sign, Photoshop Express, and Adobe Scan. You can also use Creative Cloud Express, Fresco, and Lightroom Mobile for free, with the option of making in-app purchases.",
+		},
+	],
+	[
+		"Are Adobe products worth it?",
+		{
+			content: Typography({
+				semantics: "body",
+				content: [
+					"This is an example of text content within the content area that uses the spectrum-Body typography class on the paragraphs. Adobe makes some of the most widely used software applications in the world, many of which are industry standard.",
```
marissahuysentruyt, line 23:
```suggestion
"This is an example of text content within the content area that uses the `.spectrum-Body` typography class on the paragraphs. Adobe makes some of the most widely used software applications in the world, many of which are industry standard."
```

not technically blocking, but I think you & I just talked about this!

components/accordion/stories/accordion.stories.js
```diff
@@ -180,33 +112,109 @@ WithForcedColors.parameters = {
...
 Compact.args = {
-	items: content,
+	items: accordionContent,
 	density: "compact",
 };
 Compact.parameters = {
 	chromatic: { disableSnapshot: true },
 };
-Compact.storyName = "Density - Compact";
+Compact.storyName = "Density: Compact";
```
marissahuysentruyt, line 145: If we prefer sentence-case, would this be "Density: compact" even with the colon? Does the "sentence-case" rule  change in this case? I don't actually know hahah 🤷‍♀️

I don't think it's technically blocking anyways, just something I noticed. If you decided to change this one, we should do the same for the spacious story name.

marissahuysentruyt (approved):
Great work on this!

PR #3745, opened 2025-05-15
chore(storybook): add migrated components list

marissahuysentruyt:
## Description

**DISCLAIMER: MOST CODE GENERATED BY CURSOR**

- adds a MigratedComponentsList() block to ComponentDetails.jsx. That block gets used in the S2 Migration guide MDX file
- adds a migrated-component-scanner.js task to scan the components directory and generate a list of migrated components based on their status (status: type: "migrated")
- adds a migrated-components.json file to the .storybook/data directory
- updates the project.json script to include a new `report` target that runs `migrated-component-scanner.js` so that it builds and captures the migration status
- captures the component title (so there is proper capitalization) and generates the url fragment so that the list of components links back to their corresponding docs pages

### Jira/Specs

CSS-1259

### Validation steps

- [x] Pull down the branch to run locally or visit the deploy preview. [@rise-erpelding]
- [x] `yarn nx run .storybook:report` should run the `migrated-component-status.js` task. You should get some feedback like this in the terminal:  [@rise-erpelding]

Screenshot 2025-07-25 at 12 14 18 PM

Don't be alarmed if you see changes in the `migrated-components.json` file- new components merged so that's a good sign that this script is working!
- [x] The branch should run as normal. [@rise-erpelding]
- [x] Verify that each of the links on the migration guide page correctly navigates to the docs page of the component. [@rise-erpelding]

### Regression testing

Validate:

## Screenshots

Screenshot 2025-07-25 at 12 14 18 PM

https://github.com/user-attachments/assets/446a5d8b-3aa8-4138-bb35-3924b1d9d5d1

## To-do list


package.json
```diff
@@ -15,7 +15,7 @@
 	"scripts": {
 		"build": "yarn builder tag:component,ui-icons",
 		"build:docs": "yarn build:preview --output-dir ../dist/",
-		"build:preview": "yarn report && nx build storybook",
+		"build:preview": "yarn report && yarn component:data && nx build storybook",
```
castastrophe, line 18: I wonder if there's a way to include this task as a storybook report. In the storybook project.json, if we use this script as the report, would it get run automatically with the others? What's the downside of doing that, maybe it's run more than it needs to be?

castastrophe: I'm picturing you could run it like `nx report storybook`

.storybook/data/migrated-components.json
marissahuysentruyt, line 1: Is this file something we need and/or want to keep committed? At first, this was basically just verify that I was grabbing the right data, but I'm not sold that it needs to be committed to the repo. Thoughts? 🤷‍♀️

.storybook/project.json
```diff
@@ -119,6 +119,15 @@
...
 			}
+		},
+		"report": {
+			"executor": "nx:run-commands",
+			"options": {
+				"commands": [
+					"node tasks/migrated-component-scanner.js --output=.storybook/data/migrated-components.json"
+				],
+				"cwd": "{workspaceRoot}"
+			}
```
marissahuysentruyt, line 130: @castastrophe I added a `report` target to the storybook/project.json. Is that sort of [what you were going for](https://github.com/adobe/spectrum-css/pull/3745/files#r2159082940)? 

Right now, we can trigger this manually with `yarn nx run .storybook:report`

According to Cursor, yes it would run probably more than it needs to. 

> you can add it as a dependency for other targets (like "build" or "start"), so it runs automatically before those.
>
> Example, to run it before build:
> ```
> "build": {
>  ...
>  "dependsOn": ["report", ...]
> }
> ```
>
> Downsides / Considerations
> Performance: If you add "report" as a dependency for "build" or "start", it will run every time you build or start Storybook, even if the component list hasn't changed. This could be unnecessary if the migration status doesn't change often.
> Redundancy: If the script is slow or does heavy file I/O, it could slow down your workflow.
>
> Best Practice: Only add "report" as a dependency if you want the report to always be up-to-date for every build. Otherwise, run it manually or as part of CI.

.storybook/data/migrated-components.json
marissahuysentruyt, line 1: Is this file something we need and/or want to keep committed? At first, this was basically just verify that I was grabbing the right data, but I'm not sold that it needs to be committed to the repo. Thoughts? 🤷‍♀️ Maybe we add it to the .gitignore or something? Maybe there's some options to update it occasionally, sort of like renovate PRs? Whenever we deploy the site, we have to make sure this is up-to-date or something? I don't know...

.storybook/blocks/utilities.js
```diff
@@ -1,7 +1,48 @@
...
+export function getComponentsByStatus(options = {}) {
+	// Check if we're in a Node.js environment
+	if (!isNode || !componentUtils) {
+		console.warn('getComponentsByStatus can only be used in a Node.js environment');
+		return [];
+	}
+
+	const { statusType } = options;
+
+	return componentUtils.getComponentsByStatus(statusType);
```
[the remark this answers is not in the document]
marissahuysentruyt: @rise-erpelding Ok, so from Cursor, here's why there's 2: 

> They serve different purposes and environments.
tasks/migrated-component-scanner.js: The real Node/FS implementation (and CLI) that scans the filesystem for story files and filters by status. This is Node-only.
.storybook/blocks/utilities.js: A safe, Storybook-facing wrapper that only calls into the Node implementation when it’s actually running in Node. In the browser it returns an empty array and warns, preventing bundlers from pulling in fs/path and avoiding runtime errors.

What do you think about me renaming it to `getComponentByStatusWrapper`? I could add some JSDoc comments to explain it a little more as well.

.storybook/data/migrated-components.json
```diff
@@ -0,0 +1,317 @@
...
+    {
+      "name": "taggroup",
+      "title": "Tag group",
+      "url": "tag-group"
+    },
+    {
+      "name": "textfield",
+      "title": "Text area",
+      "url": "text-area"
+    },
```
rise-erpelding, line 299: Hmmm, I was wondering why textfield didn't show up on the list for me, and it is here... just pretending to be text area instead?

tasks/migrated-component-scanner.js
```diff
@@ -0,0 +1,199 @@
...
+				title: dir,
+				url: dir,
+			};
+		}
+
+		const storyFiles = fs.readdirSync(storiesDir).filter(file => file.endsWith(".stories.js"));
+		let title = null;
+		for (const file of storyFiles) {
+			title = extractTitleFromStoryFile(path.join(storiesDir, file));
+			if (title) break;
```
rise-erpelding, line 147: I think the logic around here needs updating if we want to be able to pick up both text field and text area. When I log `storyFiles` for text field I get `[ 'textarea.stories.js', 'textfield.stories.js' ]`, but then we have it set up to only extract one title and in this weird edge case we have two.

I think that's the only place where we have that situation though. I wonder if there are other places where this is potentially causing an issue? If so we could consider modifying the way textfield and textarea are packaged?

My vote based on what I know currently would be trying to update the logic here rather than doing that.

.storybook/data/migrated-components.json
```diff
@@ -0,0 +1,317 @@
...
+    {
+      "name": "statuslight",
+      "title": "Status light",
+      "url": "status-light"
+    },
+    {
+      "name": "stepper",
+      "title": "Number field",
+      "url": "number-field"
```
rise-erpelding, line 258: Does this one show up out of alphabetical order for you in the list?

image

Not a deal breaker at all, but worth mentioning.

.storybook/project.json
rise-erpelding: This feels separate-ticket worthy to figure out how we keep the list up-to-date without having to manually do it. In the CI would be good, or if we can run it when we run `yarn start` locally, then we remember to commit the changes to that file like we do to metadata.json, that would work too.

rise-erpelding (commented):
Just a few more little things! I called out the console.log issue when running the script in the terminal, but I think the main thing that may need addressing is ensuring that we have both text field and text area on the list, or at least minimally, we want text field showing up on the list instead of text area.

Also we'll want to have a follow-up ticket to make sure that we continue to update the list as new components are added!

.storybook/data/migrated-components.json
marissahuysentruyt: It's probably because only that component's `title` in Storybook is Number field- it's package and component name that we ship are still Stepper.

tasks/migrated-component-scanner.js
marissahuysentruyt: Great call, thank you so much. Yes, I think it's the only place right now. We separated meter from progress bar, and those would have been in the same boat. 

Anyways. the script has been updated to account for multiple story files in a directory. 👍 

Screenshot 2025-08-15 at 5 18 06 PM

.storybook/data/migrated-components.json
marissahuysentruyt: lemme know what you think of the sorting functions! 

Screenshot 2025-08-15 at 5 18 00 PM

rise-erpelding (approved):

PR #3799, opened 2025-05-20
feat(table): s2 table migration layout part-1

marissahuysentruyt:

### Regression testing

Validate:

## To-do list


components/table/stories/table.stories.js
```diff
@@ -394,17 +524,44 @@ Scrollable.parameters = {
...
+		},
+	],
+};
+Visuals.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Visuals.storyName = "With visuals";
+
+
+// TODO: The design team doesn't have support for collapsible rows in the table component, so they are removed from the docs page for now.
```
marissahuysentruyt, line 564: I have a couple `TODO` notes throughout the stories file here, reflecting on the conversation I had with Stefan. 

Design doesn't have collapsible row designs, or dropzone designs done (and they're unsure if they will). React only has beta drag & drop table components. 

So I didn't really want to get rid of our stories or tests for those table variants/options. What are our thoughts on just removing these 5 or 6 stories regarding collapsible rows and dropzones from the autodocs?

components/table/stories/template.js
```diff
@@ -80,19 +113,23 @@ export const TableRowItem = ({
...
+		tabindex="0"
 	>
 		${when(showCheckbox && !isSectionHeader, () => html`
 			<${cellTag}
 				role="gridcell"
 				class=${classMap({
 					[`${rootClass}-cell`]: true,
 					[`${rootClass}-checkboxCell`]: true,
+					[`${rootClass}-cell--alignEnd`]: getTextAlignment(0) === "end",
+					...cellCustomClasses?.[0]?.reduce((a, c) => ({ ...a, [c]: true }), {}),
```
marissahuysentruyt, line 125: I wanted to show the individual cell focus states, so I had to pass custom classes to the cells.

components/table/stories/template.js
```diff
@@ -101,59 +138,73 @@ export const TableRowItem = ({
...
 				role=${ifDefined(showCheckbox ? "gridcell" : useDivs ? "cell" : undefined)}
 				class=${classMap({
 					[`${rootClass}-cell`]: true,
-					[`${rootClass}-cell--thumbnail`]: useThumbnail,
+					[`${rootClass}-cell--visual`]: useVisuals,
 					[`${rootClass}-cell--divider`]: useColumnDividers,
+					[`${rootClass}-cell--alignEnd`]: getTextAlignment(1) === "end",
+					...cellCustomClasses?.[showCheckbox ? 2 : 1]?.reduce((a, c) => ({ ...a, [c]: true }), {}),
 				})}
+				tabindex="0"
```
marissahuysentruyt, line 196: The rows should be keyboard focusable, as well as the cells (according to the designs), so I added the tabindex in quite a few places.

Any accessibility concerns with how everything is laid out right now? I don't know the flow for tables. Should each focused row move to the next focusable row, or would a user focus on a row, then move through the cells?

components/table/stories/template.js
```diff
@@ -204,65 +285,75 @@ export const Template = ({
 		id=${ifDefined(id)}
 		role=${ifDefined(useCheckboxCell ? "grid" : useDivs ? "table" : undefined)}
 		aria-multiselectable=${ifDefined(useCheckboxCell ? "true" : undefined)}
-		style="max-width: 800px;"
```
marissahuysentruyt, line 207: Should I keep this? Is it necessary? I'm fine with adding this max-width back on, just let me know!

components/table/stories/template.js
```diff
@@ -204,65 +285,75 @@ export const Template = ({
...
 						class="spectrum-Table-headCell spectrum-Table-checkboxCell"
 						role=${ifDefined(useDivs ? "columnheader" : undefined)}
 					>
-						${Checkbox({
-							size,
-							isEmphasized: isEmphasized,
-							isChecked: false,
-							isIndeterminate: true,
-							customClasses: [`${rootClass}-checkbox`],
+						${when(selectionMode === "multiple", () => html`
```
marissahuysentruyt, line 302: Checkboxes should only appear in the `thead` if users can select more than one row (and this puts the indeterminate checkbox in that header row). Otherwise, checkboxes for single select rows just appear in their rows.

components/table/stories/template.js
```diff
@@ -204,65 +285,75 @@ export const Template = ({
...
+					</${thTag}>`,
 				)}
+				<${thTag}
+					class="${rootClass}-headCell is-sortable"
+					role=${ifDefined(useDivs ? "columnheader" : undefined)}
+					aria-sort="none"
+					tabindex="0"
+				>
+					${Icon({
+						iconName: "Sort",
```
marissahuysentruyt, line 320: I added each of the new sort icons to each header cell.

components/table/stories/table.stories.js
```diff
@@ -47,16 +46,17 @@ export default {
 			},
 			control: "boolean",
 		},
-		showThumbnails: {
-			name: "Show thumbnails in content",
-			description:
-				"Uses the thumbnail component at the start of the first column's cells.",
-			type: { name: "boolean" },
+		selectionMode: {
```
marissahuysentruyt, line 49: I got "selection mode" from the menu component, as well as React.

components/table/index.css
```diff
@@ -12,511 +12,348 @@
  */
 
 .spectrum-Table {
```
marissahuysentruyt, line 14: A lot of the CSS has just been reorganized to more or less match what we are doing in other components. There was a lot of consolidating of the custom properties up at the top that needed to happen.

components/table/index.css
```diff
@@ -12,511 +12,348 @@
...
+/********* TABLE BODY *********/
 .spectrum-Table-body {
 	position: relative;
 	border: none;
-	border-radius: var(--mod-table-border-radius, var(--spectrum-table-border-radius));
+	border-end-start-radius: calc(var(--mod-table-border-radius, var(--spectrum-table-border-radius)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
+	border-end-end-radius: calc(var(--mod-table-border-radius, var(--spectrum-table-border-radius)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
 	display: table-row-group;
 
 	&.is-drop-target {
```
marissahuysentruyt, line 359: I'm curious if anybody has ideas on the body drop zone. Right now, I get little cell borders that interrupt the body outline.

Screenshot 2025-05-22 at 2 15 21 PM

components/table/index.css
```diff
@@ -551,10 +390,41 @@
...
+		z-index: 1; /* Ensure cell focus indicator appears above row focus indicator */
+
+		&::after {
+			content: "";
+			position: absolute;
+			inset: calc(-1 * var(--mod-table-border-width, var(--spectrum-table-border-width)));
+			border-radius: calc(var(--mod-table-border-radius, var(--spectrum-table-border-radius)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
+			border: var(--mod-table-focus-indicator-thickness, var(--spectrum-table-focus-indicator-thickness)) solid var(--highcontrast-table-cell-focus-indicator-color, var(--highcontrast-table-focus-indicator-color, var(--mod-table-focus-indicator-color, var(--spectrum-table-focus-indicator-color))));
+			pointer-events: none;
+		}
```
marissahuysentruyt, line 417: The cells have a different focus indicator border radius, than they do when non-focused. I'd love some feedback on this `::after` approach to achieve that.

components/table/index.css
```diff
@@ -608,48 +482,55 @@
...
-	}
+	&:focus {
+		outline: none;
 
-	&:active {
-		--highcontrast-table-row-text-color: var(--highcontrast-table-row-text-color-hover);
-		--highcontrast-table-icon-color: var(--highcontrast-table-row-text-color-hover);
-		--spectrum-table-cell-background-color: var(--highcontrast-table-row-background-color-hover, var(--mod-table-row-active-color, var(--spectrum-table-row-active-color)));
-	}
+		/* First cell of focused rows need fancy new focus indicator borders & no border-radius, and the thicker row focus indicator */
```
marissahuysentruyt, line 491: This was pretty complex, so again- feedback is welcome! 

The table row itself can be focused.
It's got a 4px blue side focus indicator, as well as (basically) a border that goes around the entire row. 
Some rows don't have border radii, and some do. But- those border radii are created with the individual table cells in that row.

🥵🥵🥵

components/table/index.css
```diff
@@ -873,20 +728,14 @@
...
 	flex-direction: row;
 	flex-wrap: nowrap;
 	justify-content: flex-start;
-	align-items: flex-start;
-
-	.spectrum-Table-collapseContent {
-		padding-block-start: calc(var(--mod-table-row-top-to-text, var(--spectrum-table-row-top-to-text)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
-		padding-block-end: var(--mod-table-row-bottom-to-text, var(--spectrum-table-row-bottom-to-text));
-	}
+	align-items: center;
```
marissahuysentruyt, line 738: Calling all feedback again! 

The [guidance](https://spectrum.adobe.com/page/table/#Respect-vertical-alignment) says: "All content should be centered vertically within the row to offer the right visual balance." That's why I added the align-items: center, ripped out all of the custom padding, and centered everything everywhere else. That does however, prove difficult in collapsible rows for a number of reasons:

- the button component (with the disclosure icon) is just tall, so the padding around the cells beefs the height up. 
- it's really weird to have a cell with wrapped text where everything is aligned to the center

Screenshot 2025-05-22 at 3 06 35 PM

However....collapsible rows aren't quite designed out so we don't have any true guidance on how to handle any of that spacing. I removed them from the docs page (pending others' agreement/approval). Should I reimplement some of this padding and alignment handling since the collapsible table variants are still in the test grid? How would we like to handle this?

marissahuysentruyt:
@marissahuysentruyt TODO: write changeset

components/table/index.css
```diff
@@ -608,48 +482,55 @@
...
-		--highcontrast-table-row-text-color: var(--highcontrast-table-row-text-color-hover);
-		--highcontrast-table-icon-color: var(--highcontrast-table-row-text-color-hover);
-		--spectrum-table-cell-background-color: var(--highcontrast-table-row-background-color-hover, var(--mod-table-row-active-color, var(--spectrum-table-row-active-color)));
-	}
+		/* First cell of focused rows need fancy new focus indicator borders & no border-radius, and the thicker row focus indicator */
+		.spectrum-Table-cell:first-child {
+			position: relative;
+			border-inline-start: 0;
+			border-block: var(--spectrum-table-row-focus-indicator-outline-width) solid var(--highcontrast-table-focus-indicator-color, var(--mod-table-focus-indicator-color, var(--spectrum-table-focus-indicator-color)));
+
```
marissahuysentruyt, line 496: TODO: Because the ::before element technically removes the table's inline border, the first cell needs the entire cell padding to avoid horizontal layout shift when a row is focused. */

`padding-inline-start: var(--spectrum-table-cell-inline-spacing);`

components/table/index.css
marissahuysentruyt: I sent a message to Stefan to clarify the "borders around a focused row." That _might_ just be for an emphasized table? I need some clarity on that.

components/table/stories/table.stories.js
```diff
@@ -154,29 +160,38 @@ const ExampleRowItems = [
...
-export const Sizing = (args, context) => Sizes({
-	Template,
-	withHeading: false,
-	withBorder: false,
-	...args,
-}, context);
-Sizing.args = {};
-Sizing.tags = ["!dev"];
-Sizing.parameters = {
+export const EmptyState = Template.bind({});
```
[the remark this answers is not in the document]
5t3ph: I would not render any of the table semantics. We _could_ talk to design about having a "skeleton" loading state to populate a few rows which would be wrapped with `aria-busy="true"` until the data loaded, as described by [Sara](https://www.sarasoueidan.com/blog/accessible-notifications-with-aria-live-regions-part-1/#aria-busy%3A-please-wait-until-the-changes-are-complete). That would also have a benefit of holding space and possibly reducing a page jump when content came in. It might be appropriate to couple with a live region as well, which would be more app-dependent.

components/table/index.css
5t3ph: If you apply it as an `::after` for `tbody`, that will allow it to appear over the row borders. You'll also need to set the border-radius to match the end, and use `pointer-events: none`. There's possibly other side effects that I'm not seeing checking it via devtools.

components/table/index.css
5t3ph: While text that's truly that long is probably an outlier, I could see an argument for start-aligning the icon with the text specifically as an accessibility benefit for users of zoom or screen magnification who see less of the screen at one time. So, having the collapse icon so separated from the labeling text might create a barrier in noticing the visual cue of the icon. (same issue possible for selectable rows)

So - maybe we can bring up this point with them and hold on further changes as far as this PR is concerned?

components/table/stories/template.js
5t3ph: I wonder if we should remove these as far as the default display, and only have them within a Story related to sorting? Then we can also include any relevant docs for using sorting.

components/table/index.css
5t3ph: Looks like you added a different TODO comment re: the shift?

components/table/stories/template.js
marissahuysentruyt: Sure I can do that!

components/table/index.css
marissahuysentruyt: https://github.com/adobe/spectrum-css/pull/3799/commits/c7a3a1ece739856b8e86bcb0e1719b55415a8966

components/table/index.css
```diff
@@ -608,48 +493,58 @@
...
+			/* Because the ::before element technically removes the table's inline border, the first cell needs the entire cell padding to avoid horizontal layout shift when a row is focused. */
+			padding-inline-start: var(--spectrum-table-cell-inline-spacing);
```
marissahuysentruyt, line 509: Since it already exists, do you think I should add the mod here?

components/table/stories/template.js
5t3ph: Ok, I think that tabindex will go away when you update this to be a button instead (per Slack discussion), since that will provide the tab stop.

Scrolling via keyboard works as I expect 👍

5t3ph:
Possibly need a background color for the table header for "Quiet" styling to prevent this clash on scrollable tables:
image

5t3ph:
For WHCM - I think there should be a difference between "Focused selected" vs. "Focused unselected" just like there is in regular mode:

image

marissahuysentruyt:
@rise-erpelding Thanks for reviewing this! To address your comments: 

- About vertically centering content- that is a great question. I added vertical-align: middle wherever I could. The token specs list out block spacing, AND leaves a note about vertically centering the icons in the header cells. As you mentioned, [the guidance](https://spectrum.adobe.com/page/table/#Respect-vertical-alignment) also says: "All content should be centered vertically within the row..." I sent a message to Stefan to ask, but when I actually inspect the cell components in Figma, they also are not centered. This definitely seems to warrant a ticket to readdress- maybe the guidelines are going to change. 🤷‍♀️ 

Screenshot 2025-06-03 at 11 10 36 AM

- For the empty table- I removed all of those controls for that story. The only one that I passed through is density. 
- About the section header quiet- great catch! Removed ✅ 
- For emphasized control- I'm not sure on this one. "Emphasized" is sort of the default for a table. A non-emphasized table wasn't part of the first iteration of the S2 table (Stefan was concerned about having too much complexity in all of the focus/selected/hover etc states). I believe in the screenshot you sent, that row is only focused. If you add the is-selected class also or instead, and then toggle the emphasized control, there is a change. The row and cell states story responds to the isEmphasized toggle pretty well too. Does that alter your preference? 

https://github.com/user-attachments/assets/2a5ff344-fe4c-44ee-935b-256fffcbd312

https://github.com/user-attachments/assets/d506f462-2acd-4d47-a78f-13d305eed09c


I think the emphasized styles apply to the selected backgrounds and the checkboxes, so maybe I can update the description to make that clearer, instead of removing the argTypes again? (videos of the emphasized toggle)

- About the text alignment- You're definitely right. Jon & I spent time trying to figure out why during a 1:1 yesterday too. 😭 It was worse before I refactored the buttons into the markup, but I think the issue is coming from the icon, and maybe has to do with the icon's margin? I haven't quite figured it out. 

Screenshot 2025-06-03 at 2 08 31 PM

components/table/index.css
marissahuysentruyt: According to Stefan, the guidance is out of date. He also mentioned that the "uneven" padding is temporary until the new font is incorporated.

components/table/index.css
```diff
@@ -712,44 +583,36 @@
 		padding-block-end: calc(var(--mod-table-section-header-block-end-spacing, var(--spectrum-table-section-header-block-end-spacing)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
```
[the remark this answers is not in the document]
marissahuysentruyt: Oh yeah, great catch. I think you're right. I think most of the table dividers are created by using border-block-start, which is why we _should_ be subtracting the border-width for any padding-block-starts...

https://github.com/adobe/spectrum-css/pull/3799/commits/7b94b04f956188164ad5edb763a17b7f8f7b6133

I think it should be corrected now. Thanks- let me know if you see anything else out of place!

components/table/stories/table.stories.js
```diff
@@ -242,22 +426,58 @@ Quiet.parameters = {
...
+Sortable.tags = ["!dev"];
+
+/**
+ * Tables can have additional actions in the header cells that could trigger a menu. The content of the header cell is an [button](/docs/components-button--docs). Oftentimes, the menu is used to show users the available sorting options.
+ */
+export const WithMenuButton = Template.bind({});
+WithMenuButton.args = {
+	...Default.args,
+	hasMenu: true,
+};
```
rise-erpelding, line 455: I was expecting this story to have a different icon (the chevron/disclosure); should it? Should it have an icon at the start and the end?

image

components/table/stories/template.js
```diff
@@ -204,65 +283,76 @@ export const Template = ({
...
+						() => when(hasMenu,
+							() => Button({
+								size: "m",
+								iconName: "SortUp",
+								iconSet: "workflow",
+								label: "Column title",
+								trailingIcon: "Chevron100",
+								trailingIconSet: "ui",
+								customClasses: [`${rootClass}-tableButton`],
+							}, context),
```
rise-erpelding, line 339: I have some questions about this too!

* I don't see `trailingIcon` (or `trailingIconSet`) defined as an arg anywhere in the button template, and removing it isn't doing much that I can tell, do we need it?
* `!isSortable && hasMenu` shows the sort up icon, and `isSortable && hasMenu` shows the sort icon, is that expected?
image
image

components/table/index.css
```diff
@@ -12,529 +12,440 @@
...
-		--mod-table-row-bottom-to-text: var(--mod-table-row-bottom-to-text--spacious, var(--spectrum-table-row-bottom-to-text-extra-large-spacious));
-
-		/* Row Selection */
-		--mod-table-row-checkbox-block-spacing: var(--mod-table-row-checkbox-block-spacing--spacious, var(--spectrum-table-row-checkbox-to-top-extra-large-spacious));
-
-		/* Collapsible and Thumbnails */
-		--mod-table-thumbnail-block-spacing: var(--mod-table-thumbnail-block-spacing-spacious, var(--spectrum-table-thumbnail-to-top-minimum-extra-large-spacious));
-		--mod-table-thumbnail-size: var(--mod-table-thumbnail-size-spacious, var(--spectrum-thumbnail-size-800));
-	}
+.spectrum-Table-scroller {
```
rise-erpelding, line 254: Leaving this here because I don't have another great place to put it and I can't find where @5t3ph wrote her comment about this earlier:

Is this issue with quiet & scrollable addressed in Part 2?

image

rise-erpelding (commented):
Hi! Mostly questions and grammar nits for this round. This PR is in good shape though and I know between Part 2 and the follow-up tickets this will need that it'll get there. Not approving because I feel like I still have a lot of questions, but am unsure whether it needs changes so didn't want to hit that option either. I'll leave it here in "Comment" purgatory but am happy to take another look if you think it needs some changes, or approve if you think the things I brought up here are non-blocking!

One more question re: the emphasized styling control, I _do_ see it working sometimes... but are there stories in which I shouldn't expect it to work? For example if I have this visuals story and use the inspector to trigger focus/focus-visible on a row, should the the background color change from gray to blue? This also happens to me on the default story. Is it just not meant to behave like that?
Screenshot 2025-06-04 at 9 30 22 PM

components/table/index.css
marissahuysentruyt: Yes- in part 2 the background color of the header columns gets updated so this doesn't happen.

components/table/stories/table.stories.js
marissahuysentruyt: oh how weird! The storybook link isn't rendering that icon? I get it locally, but hadn't checked the preview link. Let me look into this.
Screenshot 2025-06-05 at 9 45 45 AM

components/table/index.css
```diff
@@ -709,48 +646,35 @@
 		block-size: var(--mod-table-section-header-min-height, var(--spectrum-table-section-header-min-height));
 
 		padding-block-start: calc(var(--mod-table-section-header-block-start-spacing, var(--spectrum-table-section-header-block-start-spacing)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
-		padding-block-end: calc(var(--mod-table-section-header-block-end-spacing, var(--spectrum-table-section-header-block-end-spacing)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
+		padding-block-end: var(--mod-table-section-header-block-end-spacing, var(--spectrum-table-section-header-block-end-spacing));
```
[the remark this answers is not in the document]
marissahuysentruyt: I'm for doing it in a different ticket, sure! For now, we cannot vertically center, but once the new font is available, the spacing tokens _should_ be the same at least.

.changeset/petite-toys-greet.md
```diff
@@ -0,0 +1,87 @@
...
+
+### S2 table migration
+
+Compared to the S1 table, this component has updated corner rounding, updated color tokens, some updated spacing, and an overall refreshed look.
+
+#### Net-new features
+
+- The S2 table supports an empty state, rendering an illustrated message component.
+- As data is loading to the table, this component will render a progress circle during the loading state.
+- There are 2 selection modes: single-select and multi-select. Multi-select tables (`selectionMode: "multiple"`) render an indeterminate checkbox in the `thead`/`div` equivalent. Single-select tables (`selectionMode: "single"`) do not render the indeterminate checkbox in the header row.
```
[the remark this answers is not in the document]
marissahuysentruyt: I'm going to keep the selection mode in, but I did add a note that the storybook control is removed from the table.

I would guess no on the `aria-multiselectable`. I've updated the component to check for `selectionMode` instead of `useCheckboxCell`.

I added a link to checkbox in regards to the indeterminate checkbox for mutli-select 👍

components/table/stories/table.stories.js
marissahuysentruyt: I made some updates to the source path I was using with `mask-image`. It looks like it builds correctly now in [the preview](https://pr-3799--spectrum-css.netlify.app/?path=/docs/components-table--docs)! 🥳

rise-erpelding (approved):
Good to go! On to Part 2!

components/table/stories/table.stories.js
```diff
@@ -242,22 +425,58 @@ Quiet.parameters = {
...
+	sortIcon: "Sort",
+};
+Sortable.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+Sortable.storyName = "Sortable columns";
+Sortable.tags = ["!dev"];
+
+/**
+ * Tables can have additional actions in the header cells that could trigger a menu. The content of the header cell is an [button](/docs/components-button--docs). Oftentimes, the menu is used to show users the available sorting options.
```
5t3ph, line 448: Implementation looks good! Just a couple suggestions:
1. Can this be re-ordered so it is the story right after sortable columns?
2. I think we should demonstrate in the markup - and mention - that `aria-sort` should still be used when sorting is active
3. It looks like an extra class of `spectrum-Table-tableButton` is needed on the button, which I would elevate into both descriptions.
4. Also note the extra class of `spectrum-Table-menuButton` in this description
5. Additional slight wording update as suggested (feel free to edit further)

```suggestion
 * Tables can have additional actions in the header cells that could trigger a menu. Oftentimes, the menu is used to show users the available sorting options.  The content of the header cell is an [button](/docs/components-button--docs), which should contain the sorting icon when applicable as well as the header label.
```

marissahuysentruyt: > Can this be re-ordered so it is the story right after sortable columns?

✅ YES! 

> I think we should demonstrate in the markup - and mention - that aria-sort should still be used when sorting is active

:white Let me know if this still needs tweaking or improvement. Is this what you were thinking?
"The `aria-sort` attribute should be set on the `.spectrum-Table-headCell .is-sortable` element for the column that is currently sorted, and when sorting is active."
"The `aria-sort` attribute should be set on the `.spectrum-Table-hasMenuButton` element for the column that is currently sorted, and when sorting is active."

> It looks like an extra class of spectrum-Table-tableButton is needed on the button, which I would elevate into both descriptions.

I did adjust this slightly. The classes have been updated from `.spectrum-Table-menuButton` to `.spectrum-Table-hasMenuButton`, as well as `.spectrum-Table-tableButton` to `.spectrum-Table-sortButton`. As I was working through the suggestions, they were tripping me up, so hopefully that clears up the CSS and the documentation in one go.

I mention `.spectrum-Table-sortButton` in both descriptions, and believe that class is being added to the button appropriately in the markup. Let me know if you're not seeing that.

> Also note the extra class of spectrum-Table-menuButton in this description

✅ mentioned `.spectrum-Table-hasMenuButton`

> Additional slight wording update as suggested (feel free to edit further)
✅

5t3ph: LGTM!

5t3ph (approved):
whooohoo, excellent work!! 🙌

PR #3818, opened 2025-05-21
feat(table): migrates remaining table styles to s2 part2

marissahuysentruyt:

### Regression testing

Validate:

## To-do list


components/table/index.css
```diff
@@ -702,10 +745,6 @@
      --mod-table-current-header-height should be dynamically updated with JS to match the table header height. */
 	scroll-padding-top: var(--mod-table-current-header-height, calc((var(--mod-table-header-line-height, var(--spectrum-table-row-line-height)) * var(--mod-table-header-font-size, var(--spectrum-table-row-font-size))) + var(--mod-table-header-row-top-to-text, var(--spectrum-table-header-row-top-to-text)) + var(--mod-table-header-row-bottom-to-text, var(--spectrum-table-header-row-bottom-to-text)) + var(--mod-table-border-width, var(--spectrum-table-border-width))));
 
-	&.spectrum-Table--quiet {
-		border-block-start: none;
-	}
```
marissahuysentruyt, line 707: Quiet tables still have a border-block-start for the header row.

components/table/index.css
```diff
@@ -547,6 +591,19 @@
...
+		.spectrum-Table-cell {
+			/* Remove bottom border by default for all selected rows to conditionally add it back. */
+			border-block-end: none;
+		}
+	}
+
+	/* Adding the bottom border only to the last selected row in a sequence achieves 1px border between adjacent selected rows */
+	&.is-selected:not(:has(+ .is-selected)) .spectrum-Table-cell {
+		border-block-end: var(--mod-table-border-width, var(--spectrum-table-border-width)) solid var(--highcontrast-table-focus-indicator-color, var(--mod-table-focus-indicator-color, var(--spectrum-table-focus-indicator-color)));
+	}
```
marissahuysentruyt, line 605: I got some help from Cursor with this, so please review and let me know if there's red flags. I'm particularly concerned with the `:not(:has())` situation. In order to test this, you'll have to add the `is-selected` class to an adjacent row.

Basically, this is the code to produce 1px of blue border, between adjacent selected rows. I noted in the PR description that during my talks with Stefan, he said design's preference is to have only 1px of border between adjacent rows. He did say though that he understands it's quite complicated, and had already discussed the possibility of needing 2px in between adjacent selected rows. Although the 2px isn't ideal, if we decide that's probably the best route long-term, it's acceptable. 

What are our thoughts on this approach to achieving the single pixel border between adjacent selected rows? I do believe there would be 1px of content shift, so does that mean we should toss this idea out? I couldn't achieve only 1px between ALL adjacent rows, when some are selected and some are not, so maybe this approach is overkill, since there's 2px borders in between other rows anyways?

.changeset/petite-toys-greet.md
```diff
@@ -0,0 +1,87 @@
...
+Compared to the S1 table, this component has updated corner rounding, updated color tokens, some updated spacing, and an overall refreshed look.
+
+#### Net-new features
+
+- The S2 table supports an empty state, rendering an illustrated message component.
+- As data is loading to the table, this component will render a progress circle during the loading state.
+- There are 2 selection modes: single-select and multi-select. Multi-select tables (`selectionMode: "multiple"`) render an indeterminate checkbox in the `thead`/`div` equivalent. Single-select tables (`selectionMode: "single"`) do not render the indeterminate checkbox in the header row.
+- For tables with sortable column, there are three new S2 icons used: `Sort` to indicate "general" sorting, `SortUp` for ascending sort direction, `SortDown` for descending sort direction.
+- Tables support thumbnail, avatar, and icon components as content within a cell.
+- Focus indicators for entire rows have been updated for rows to include a side focus indicator.
```
marissahuysentruyt, line 16: add a note about the sparkline variant

components/table/stories/table.stories.js
```diff
@@ -24,11 +22,15 @@ export default {
 			control: "select",
 		},
 		isQuiet,
-		isEmphasized,
+		isEmphasized: {
+			...isEmphasized,
+			description: "Emphasized styling on the table affects colors of selected rows and any checkboxes."
+		},
```
marissahuysentruyt, line 28: TODO: is this control necessary for all stories? Is it ok if we disable the control per story? Now that colors are in place, revisit some of the questions Rise posed in #3799

components/table/index.css
```diff
@@ -608,48 +623,57 @@
...
+		}
 	}
 
-	&:active {
-		--highcontrast-table-row-text-color: var(--highcontrast-table-row-text-color-hover);
-		--highcontrast-table-icon-color: var(--highcontrast-table-row-text-color-hover);
-		--spectrum-table-cell-background-color: var(--highcontrast-table-row-background-color-hover, var(--mod-table-row-active-color, var(--spectrum-table-row-active-color)));
+	/* Adding the bottom border only to the last selected row in a sequence achieves 1px border between adjacent selected rows */
+	&.is-selected:not(:has(+ .is-selected)) .spectrum-Table-cell {
+		border-block-end: var(--mod-table-border-width, var(--spectrum-table-border-width)) solid var(--highcontrast-table-focus-indicator-color, var(--mod-table-focus-indicator-color, var(--spectrum-table-focus-indicator-color)));
```
saashapina, line 636: In the spec it also looks like there should be 2px border width on the active down state. Is that missing?

components/table/index.css
saashapina: hmm, yea this is always tricky. I am seeing the 1px shift when adding and removing selected state. I also noticed in the mockup it looks like the row is retaining its top borders. Not sure if that was an oversight or if its intentional. But if it is we could probably try using something like `box-shadow: inset 0 -1px 0 var(--highcontrast-table-focus-indicator-color, var(--mod-table-focus-indicator-color, var(--spectrum-table-focus-indicator-color)));` here instead of the border. Using insets my fix the shifting issue. In the end it might come down to whether or not design is okay with a content shift or the 2px border.
Screenshot 2025-06-06 at 9 40 01 AM

saashapina (commented):
This is looking good so far Marissa. I just had a couple initial questions and I took a look at the selected issue you were seeing.

components/table/index.css
marissahuysentruyt: No- Stefan clarified with me that the 2px was incorrect in Figma. I asked about the same thing, and he said the only time anything gets the 2px border width is when an individual cell is focused.

components/table/index.css
```diff
@@ -12,533 +12,513 @@
...
+	--spectrum-table-drop-zone-outline-color: var(--spectrum-accent-visual-color);
 
-		--highcontrast-table-selected-row-background-color-focus: Highlight;
-		--highcontrast-table-selected-row-text-color-focus: HighlightText;
-		--highcontrast-table-row-background-color-hover: Highlight;
-		--highcontrast-table-row-text-color-hover: HighlightText;
+	/* Row States */
+	/* @todo Refactor or remove these properties once the RGB stripped value is available. */
+	--spectrum-table-row-background-color-hover: rgb(var(--spectrum-gray-900-rgb), var(--spectrum-table-row-hover-opacity));
+	--spectrum-table-row-background-color-active: rgb(var(--spectrum-gray-900-rgb), var(--spectrum-table-row-down-opacity)); /* active/down state background color */
```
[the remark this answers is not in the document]
marissahuysentruyt: Personally, I think it's ok to have the single comment for both of these properties. There's a chunk of similar custom props farther up in the file that have the same comment.

components/table/index.css
```diff
@@ -220,26 +254,22 @@
 
 .spectrum-Table-headCell {
```
aramos-adobe, line 255: In regards to my test comment on sortable and head cell with menu button and it's just a suggestion:

```css
    --highcontrast-button-background-color-default: Canvas;
    --highcontrast-button-content-color-default: CanvasText;
    --highcontrast-button-background-color-hover: Canvas;
    --highcontrast-button-content-color-hover: Highlight;
    border: none;
```
Keeps focus on the table rows that are selected and that may have checkboxes.

aramos-adobe: Screenshot 2025-06-09 at 11 35 58 AM
Screenshot 2025-06-09 at 11 35 52 AM

components/table/index.css
```diff
@@ -340,6 +390,28 @@
...
+/* Quiet table header rows do not have inline borders */
+.spectrum-Table--quiet .spectrum-Table-headRow:first-child {
+	.spectrum-Table-headCell:first-child {
+		border-inline-start: 0;
+	}
+
+	.spectrum-Table-headCell:last-child {
+		border-inline-end: 0;
+	}
+}
```
aramos-adobe, line 413: Duplicate selectors here

marissahuysentruyt: Thanks for catching that! It must have come over during a rebase 🥴

components/table/stories/table.test.js
```diff
@@ -492,6 +492,31 @@ export const TableGroup = Variants({
 					isSelected: true,
 				}
 			],
+		},
+		{
+			testHeading: "Sparkline charts",
```
marissahuysentruyt, line 497: @rise-erpelding I'm plopping this comment here to sort of get it in a closer spot. 

You mentioned potentially adding hover states. I feel like we've been trying to get rid of storybook only classes in our selectors, in favor of just using the `:focus-visible` or `:hover`. If that's the case, I'm not sure I quite know how to add hover tests to the testing grid without args and `is-hovered` classes. I absolutely don't mind doing it - I think it's a great idea - but is there a better way than just adding an `isHover` arg to the template so that it adds an `.is-hovered` class?

components/table/stories/template.js
```diff
@@ -322,6 +330,7 @@ export const Template = ({
 					${when(hasMenu || isSortable, () => html`
 						${when(isSortable, () => Button({
 								size: "m",
+								variant: "secondary",
```
marissahuysentruyt, line 333: @rise-erpelding I don't mind changing this to the secondary button, but I feel like I still needed all of the passthroughs anyways. I think I can remove the `--mod-button-height`, alongside a `min-block-size` to help out the funky alignment we were seeing, but that felt like that was all 😞

components/table/index.css
```diff
@@ -124,18 +133,24 @@
...
 	--mod-button-background-color-focus: transparent;
 	--mod-button-background-color-down: transparent;
-	--mod-button-content-color-default: var(--spectrum-table-header-text-color);
-	--mod-button-content-color-hover: var(--spectrum-table-header-text-color);
-	--mod-button-content-color-focus: var(--spectrum-table-header-text-color);
-	--mod-button-content-color-down: var(--spectrum-table-header-text-color);
+	--mod-button-content-color-default: var(--mod-table-header-text-color, var(--spectrum-table-header-text-color));
+	--mod-button-content-color-hover: var(--mod-table-header-text-color, var(--spectrum-table-header-text-color));
+	--mod-button-content-color-focus: var(--mod-table-header-text-color, var(--spectrum-table-header-text-color));
+	--mod-button-content-color-down: var(--mod-table-header-text-color, var(--spectrum-table-header-text-color));
```
marissahuysentruyt, line 139: @rise-erpelding So far, I only updated these to have the table mods included the the button passthrough definitions. Do you think I should do that for any of the other mods?

components/table/index.css
```diff
@@ -131,11 +140,17 @@
 	--mod-button-edge-to-text: var(--spectrum-table-cell-inline-spacing);
 	--mod-button-padding-label-to-icon: var(--spectrum-table-visual-to-text);
 	--mod-button-height: var(--spectrum-table-min-header-row-height);
-	--mod-button-focus-ring-border-radius: var(--spectrum-table-focused-cell-border-radius);
+	--mod-button-focus-ring-border-radius: calc(var(--mod-table-border-radius, var(--spectrum-table-focused-cell-border-radius)) - var(--mod-table-border-width, var(--spectrum-table-border-width)));
```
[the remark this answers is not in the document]
marissahuysentruyt: ah yeah, that's a really good point. I like this idea!

components/table/index.css
```diff
@@ -825,10 +913,10 @@
 	border: 0;
 	color: var(--spectrum-table-icon-color);
 
-	transform: var(--spectrum-logical-rotation,);
+	transform: var(--spectrum-logical-rotation);
 
 	&[aria-expanded="true"] {
-		transform: var(--spectrum-logical-rotation,) rotate(90deg);
```
marissahuysentruyt, line 831: @rise-erpelding fixed!

components/table/index.css
```diff
@@ -299,7 +349,6 @@
 		background-color: var(--mod-table-header-background-color, var(--spectrum-table-header-background-color));
 		border-block-start: var(--mod-table-border-width, var(--spectrum-table-border-width)) solid var(--highcontrast-table-border-color, var(--mod-table-border-color, var(--spectrum-table-border-color)));
 
-		transition: color var(--highcontrast-table-transition-duration, var(--mod-table-transition-duration, var(--spectrum-table-transition-duration))) ease-in-out;
```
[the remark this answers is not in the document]
marissahuysentruyt: Also, about this `--highcontrast-table-transition-duration`. I believe this is probably set to 0 so that there's not any animation for it. For instance, when a collapsed row expands, in WHCM it would just pop right open, instead of slide/transform. I can't replicate that behavior though, so it's sort of a guess.

components/table/index.css
```diff
@@ -170,36 +185,51 @@
...
 	--spectrum-table-border-radius: var(--mod-table-border-radius-quiet, 0px);
 	--spectrum-table-outer-border-inline-width: var(--mod-table-outer-border-inline-width-quiet, 0px);
-	--spectrum-table-header-background-color: var(--mod-table-header-background-color-quiet, var(--spectrum-transparent-white-100));
-	--spectrum-table-row-background-color: var(--mod-table-row-background-color-quiet, var(--spectrum-transparent-white-100));
-	--spectrum-table-border-color: transparent;
 }
 
-.spectrum-Table-row {
-	&:hover,
+.spectrum-Table-scroller {
```
marissahuysentruyt, line 196: @rise-erpelding I'm going to plop this comment here too, in response to your scrollable question and the borders. Currently, yes it's intentional that the selected row doesn't have inline borders. 

So in the non-scrolling tables, it's the cells themselves that end up creating the entire, outermost border of the table. It's easier to target the inline borders of a selected row. But to get a scrolling table, we add an extra wrapper element around the table and that scroller wrapper is what's actually creating the outermost border of the table instead of the cells. I'm sure I could target the same cells and add inline borders for selected rows in a scrolling table, but that would add extra borders inside of the scroll wrapper borders. 

Does that make sense at all? We can pair on this if I didn't explain it super well!

marissahuysentruyt:
@rise-erpelding 

> Are drop targets a variant, similar to collapsible, that aren't explicitly in S2? The borders there are a little funny, but not worth fixing if they're not in the spec. What is the plan for those variants that aren't in S2 but we haven't removed? Did design have an opinion on that?

Drop targets are a variant that aren't explicitly in S2, and they are not in the spec. I can make a card to revisit the borders (I think I saw the body dropzone is the funkiest, maybe?). When I was talking with Stefan about the dropzones, he referenced React's drag and drop functionality, and that it is currently "beta." Stefan suggested that we could treat our dropzones the same. Design isn't certain if they will work on dropzone/drag and drop stuff specifically, so they don't have a time frame, but it is something they've explored and want to come back to in the future.

components/table/index.css
marissahuysentruyt: Thanks for these! I like most of the suggestions, but the only one I'm uncertain of is removing the border for that button. If we style the rest of the button content just like the other header cells, AND remove the border, there's nothing visually in WHCM that can notify a user that it's actually a button and they can do something with it.

I did try to compromise and adjusted the button border radius to better match the cells. What do you think about things now? Should I go back to design and double check this?

components/table/index.css
```diff
@@ -33,8 +35,10 @@
...
 
-	/* Row Selection */
-	--spectrum-table-row-active-color: rgb(var(--spectrum-gray-900-rgb), var(--spectrum-table-row-down-opacity)); /* active/down state background color */
+	/* Row States */
+	/* @todo Refactor or remove these properties once the RGB stripped value is available. */
+	--spectrum-table-row-background-color-hover: rgb(var(--spectrum-gray-900-rgb), var(--spectrum-table-row-hover-opacity));
+	--spectrum-table-row-background-color-active: rgb(var(--spectrum-gray-900-rgb), var(--spectrum-table-row-down-opacity)); /* active/down state background color */
 
 	/* Border */
 	--spectrum-table-border-color: var(--spectrum-gray-300);
```
marissahuysentruyt, line 44: @rise-erpelding Just wanted to let you know I made that card for refactoring the table `border-block-start`s to `border-block-end`s that match Figma & React. 

CSS-1244
Our original discussion: https://github.com/adobe/spectrum-css/pull/3799#discussion_r2127796546

components/table/stories/table.test.js
rise-erpelding: Yeah, I think so! Now that I'm looking more closely, I _think_ you can also remove the `.is-focused` class from the css as well unless there's something I'm missing about needing it?

The way I understand it, the PostCSS [config](https://github.com/adobe/spectrum-css/blob/main/postcss.config.js#L81) will prefix the pseudo-classes listed there ("focus-visible", "focus-within", "hover", "active", "disabled") with `.is-` and transform it so that in the `dist/index.css` anywhere where you've used `:focus-visible` for example should also have an `.is-focus-visible` class applied as well.

Then, if you change the template to apply the `.is-focus-visible` class instead of `.is-focused` for the `isFocused` arg, those focus styles would be linked to the control. You could do something with `:hover` also since that `.is-hover` class is already being applied?

rise-erpelding (changes requested):
I think this just needs a few little changes that I found places for within the comments! I'm going to double-check that I addressed all your questions as well.

tokens/custom/dark-vars.css
```diff
@@ -57,6 +57,8 @@
 
 	--spectrum-steplist-current-marker-color-key-focus: var(--spectrum-blue-700);
 
+	--spectrum-table-selected-row-background-color-rgb: var(--spectrum-blue-800-rgb);
```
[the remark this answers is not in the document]
marissahuysentruyt: Ah, yes! thanks!

components/table/index.css
marissahuysentruyt: Good catch! it was picking up the text-hover colors still in WHCM.

aramos-adobe (approved):

rise-erpelding (approved):
Crushed it!

PR #3861, opened 2025-05-28
fix(link): fixes keyboard focus shift by changing padding to outline-offset

cdransf:
## Description

The previous implementation with padding would cause the component to shift when focused using the keyboard. Replacing `padding` with `outline-offset` resolves this issue.

### Validation steps

1. Open the deployment link for this PR.
2. Navigate to the `link` component docs page.
3. Focus the link component using your keyboard.
4. Verify that the focused link state no longer causes a layout shift.

## To-do list

### DIFF
--- a/.changeset/lovely-geckos-hammer.md
+++ b/.changeset/lovely-geckos-hammer.md
@@ -1,0 +1,5 @@
+---
+"@spectrum-css/link": minor
+---
+
+Fixes keyboard focus shift by changing padding to outline-offset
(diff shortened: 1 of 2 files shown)


aramos-adobe (approved):
Tested it on other components:

- Accordion
- Inline Alert
- Contextual Help
- Well

No more layout shifts ⭐️ great use of `outline-offset`

.changeset/lovely-geckos-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@spectrum-css/link": minor
```
marissahuysentruyt, line 2: Personally, I'd probably just call this a patch, since we categorized it as bug. I'll leave it up to you though.

marissahuysentruyt (approved):
Thanks for tracking this down!

PR #3871, opened 2025-05-29
chore(swatch): updates the add swatch icon to use the ui rather than workflow icon

cdransf:
## Description

When the `swatch` component was originally migrated we only had the `workflow` add icon. This updates the component to use the `ui` icon now that the latter is available.

## To-do list

### DIFF
--- a/components/swatch/stories/template.js
+++ b/components/swatch/stories/template.js
@@ -118,9 +118,14 @@ export const Template = ({
 							}, context)] : []),
 							...(isAddSwatch ? [Icon({
 								customClasses: [`${rootClass}-icon`],
-								setName: "workflow",
+								setName: "ui",
 								size,
-								iconName: "Add",
+								iconName: "Add" + ({
+									xs: "75",
+									s: "75",
+									m: "100",
+									l: "200",
+								}[size] || "100"),
 								useRef: false,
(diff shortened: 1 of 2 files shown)


components/swatch/stories/template.js
```diff
@@ -118,7 +118,7 @@ export const Template = ({
 							}, context)] : []),
 							...(isAddSwatch ? [Icon({
 								customClasses: [`${rootClass}-icon`],
-								setName: "workflow",
+								setName: "ui",
```
rise-erpelding, line 121: It looks like from the spec we'll want the sizing to follow the same sizing pattern as with Dash for mixed value, 75 for xs and s, 100 for m, and 200 for l, without specifying it looks like it pulls the default sizing (50, 75, 100, and 200), does it need to be 75 for xs instead of 50?

If so it looks like that'd be just a simple fix in the `iconName` for Add just like what Dash has!

rise-erpelding (changes requested):
Correct me if I'm wrong! I think this just needs a tiny sizing adjustment for XS!

components/swatch/stories/template.js
cdransf: Ahhh good catch! Just updated that. ✨

PR #3950, opened 2025-06-11
test(alertbanner): updates alertbanner test file to restore text and action button label control functionality

cdransf:
## Description

`CSS-1245`

This removes an unnecessary test template that was overriding the Storybook controls for the alert text, action button label and semantic variants. Removing this template restores the functionality of all said controls and the previously displayed, fixed variants can be viewed using the controls or testing preview.

## To-do list

### DIFF
--- a/components/alertbanner/stories/alertbanner.stories.js
+++ b/components/alertbanner/stories/alertbanner.stories.js
@@ -61,7 +61,7 @@ export default {
 		isOpen: false,
 		variant: "neutral",
 		actionButtonText: "Action",
-		text: "Alert banner message. Use a short phrase to describe what's happening.",
+		text: "Your trial has expired",
 		showCloseButton: true,
 	},
 	parameters: {
--- a/components/alertbanner/stories/alertbanner.test.js
+++ b/components/alertbanner/stories/alertbanner.test.js
@@ -2,7 +2,7 @@ import { Variants } from "@spectrum-css/preview/decorators";
 import { Template } from "./template.js";
 
 export const AlertBannerGroup = Variants({
-	Template: Template,
+	Template,
 	TestTemplate: Template,
 	stateDirection: "column",
 	wrapperStyles: {


components/alertbanner/stories/alertbanner.test.js
```diff
@@ -1,38 +1,8 @@
...
-			text: "Alert banner message. Use a short phrase to describe what's happening.",
-			actionButtonText: "Action",
-		}, context),
-	],
-}, context);
-
 export const AlertBannerGroup = Variants({
-	Template: AlertBannerVariantsTemplate,
+	Template: Template,
 	TestTemplate: Template,
```
marissahuysentruyt, line 6: Because the `Template` that we're using on the story's page is now the same as the `TestTemplate`, I'm pretty sure we can delete the `TestTemplate` line. 

```suggestion
Template,
```

components/alertbanner/stories/alertbanner.stories.js
```diff
@@ -61,7 +61,7 @@ export default {
 		isOpen: false,
 		variant: "neutral",
 		actionButtonText: "Action",
-		text: "Your trial has expired",
+		text: "Alert banner message. Use a short phrase to describe what's happening.",
```
marissahuysentruyt, line 64: Just a question- do we have VRT baselines for the `spectrum-two` branch? If we're changing text here, it'd cause a diff for this snapshot, but if we don't really have baselines yet, does it matter? 

I guess my point it that we don't really need to change this text. I sort of like your change, and I vote that if it doesn't cause confusion in any sort of VRT run (again, not sure if we technically have baselines or not 🤔), let's keep this new message text.

marissahuysentruyt (changes requested):
Looking good! I left small suggestions to polish it off- we can remove the changeset since we didn't change CSS, and a small deletion in the test file! Nice catch on this!

components/alertbanner/stories/alertbanner.test.js
cdransf: Good catch! All done. ✨

castastrophe (approved):

marissahuysentruyt (approved):
yes

PR #3965, opened 2025-06-17
fix(popover): restores drop shadow for nested popovers

cdransf:
## Description

`CSS-1066`

Restores drop shadow for nested popovers.

Per design and the coach mark component in our Figma library, nested popovers should have drop shadows enabled.

## Screenshots

image
image

## To-do list


castastrophe (approved):

marissahuysentruyt (commented):
Another note for posterity:

The box shadow (pink) renders smaller than the filter: drop-shadow() (green). _However_, the box shadow looks to match the Figma drop shadow better- this is where I'd need to get with a designer to ask which they like better. The values are identical for X, Y, blur, and color, but: 
- the box-shadow and filter: drop-shadow() use different blur algorithms. 🤦‍♀️  
- Apparently, there's also an extra compositing layer with filter: drop-shadow(), which can affect how the blur spreads. 🤦‍♀️ 
- Maybe there's some bounding container fudging going on which may make the blur extend further. I wonder if the filter is considering the tip as "part" of bounds, and is pretty much adding 8 extra pixels to the bounding container when it shouldn't really, giving the illusion of a larger, softer drop-shadow.

I had everything in dark mode to try to see the shadows better.

Screenshot 2025-06-20 at 3 26 56 PM

marissahuysentruyt:
Sorry one more comment. I may have a fix, but I'd definitely need some design eyes I think (which we'll get eventually since this is S2). I flipped everything to dark mode and added some wild colors just to experiment and get a better understanding. 

So this is with my idea to use box-shadow at line 72 (instead of filter). I left some comments that we can commit or just use here as notes. 

We'd define a new custom prop after that filter prop: 
```
/* filter is for popovers with a tip. we divide the blur by 2 to match the box-shadow rendering better. */
--spectrum-popover-filter: drop-shadow(var(--mod-popover-drop-shadow-x, var(--spectrum-popover-drop-shadow-x)) var(--mod-popover-drop-shadow-y, var(--spectrum-popover-drop-shadow-y)) calc(var(--mod-popover-drop-shadow-blur, var(--spectrum-popover-drop-shadow-blur)) / 2) var(--mod-popover-drop-shadow-color, var(--spectrum-popover-drop-shadow-color)));

/* box-shadow is for popovers without a tip. */
--spectrum-popover-box-shadow: var(--mod-popover-drop-shadow-x, var(--spectrum-popover-drop-shadow-x)) var(--mod-popover-drop-shadow-y, var(--spectrum-popover-drop-shadow-y)) var(--mod-popover-drop-shadow-blur, var(--spectrum-popover-drop-shadow-blur)) var(--mod-popover-drop-shadow-color, var(--spectrum-popover-drop-shadow-color));
```

at line 72ish, replace filter with box-shadow: 
```
...
background-color: var(--mod-popover-background-color, var(--spectrum-popover-background-color));
box-shadow: var(--mod-popover-box-shadow, var(--spectrum-popover-box-shadow));

/* has tip triangle */
&.spectrum-Popover--withTip {
...
```

Then keep the filter and box-shadow: none styles you have in the `--withTip` selector block.

That should give us this: 

Screenshot 2025-06-20 at 3 26 39 PM

marissahuysentruyt (approved):
This is looking good to me! I did leave a suggestion for the changeset, since we ended up changing it a little more dramatically that we had initially. 

I'm also curious about the stylelint PR comment that `--spectrum-popover-border-width` is undefined. It was already there during [the popover migration](https://github.com/adobe/spectrum-css/pull/3365/files), so I'm trying to figure out where that variable went missing...that's a different PR I guess! Thanks for your hard work on this!

PR #3968, opened 2025-06-18
feat(meter): s2 migration, full corner radius, help text

aramos-adobe:
### S2 Meter Migration

This migration updates the meter component and the progress bar component. Both needed the track to have a full corner radius to make the rounded look. The progress bar's HTML template now includes the help text component but it is only used in the meter.

#### New updates

- Includes `--sizeM` and `--sizeXL`
- Includes help text component for more context
- Shares progress bar's field label typography
- Added help text visual test

Since the progress bar and meter are now using the same tokens for track thickness, font size, and spacing, the following mods have been redacted:

- `--mod-progressbar-thickness`
- `--mod-progressbar-font-size`
- `--mod-progressbar-spacing-top-to-text`

#### New mods

- `--mod-meter-help-text-to-progress-bar`

### Regression testing

Validate:

## Screenshots

Screenshot 2025-06-18 at 12 31 09 PM

## To-do list


components/progressbar/index.css
```diff
@@ -126,6 +126,8 @@
 		block-size: var(--mod-progressbar-thickness, var(--spectrum-progressbar-thickness));
 		transition: width 1s;
 		background: var(--highcontrast-progressbar-fill-color, var(--mod-progressbar-fill-color, var(--spectrum-progressbar-fill-color)));
+		border-start-end-radius: var(--spectrum-progressbar-corner-radius);
+		border-end-end-radius: var(--spectrum-progressbar-corner-radius);
```
rise-erpelding, line 130: I wonder if we missed this in the progress bar migration or if it was added after we migrated?

Nice catch adding this! 💥

components/meter/stories/meter.stories.js
rise-erpelding, line 1: I think some cleanup on the docs page might be good. A few things I noted reading through the docs page:

* I notice that we're saying "The meter component is a visual _representations_..." which should really be "The meter component is a visual representation..."
* I think we previously only offered small or large, right? So it says that large is the default around L68, this area also seems like a good opportunity for some grammar cleanup ("By default, meters have a blue fill and are medium-sized.")
* Also the Sizing story still says "Meters come in two sizes: large and small" so that section would need updating too.
* I'd also like to see static colors stories here, I know we've done them for other components (such as button) so that seems doable here!

.changeset/sour-donuts-go.md
```diff
@@ -0,0 +1,25 @@
...
+- Includes `--sizeM` and `--sizeXL`
+- Includes help text component for more context
+- Shares progress bar's field label typography
+- Added help text visual test
+
+#### Removed mods
+
+- `--mod-meter-thickness`
+- `--mod-meter-font-size`
+- `--mod-meter-spacing-top-to-text`
```
rise-erpelding, line 21: I think the removed mods here are actually the progressbar ones: `--mod-progressbar-thickness`, etc. They're actually not removed because they can still be used, we're just not passing anything through them anymore. Any thoughts on a good way to handle/document that? 🤔

.changeset/sour-donuts-go.md
```diff
@@ -0,0 +1,25 @@
...
+
+#### New mods
+
+- `--mod-meter-field-label-progress-bar`
```
rise-erpelding, line 25: I suggested a change to this custom property, if you end up changing its name, don't forget to update the changeset too!

components/meter/stories/meter.test.js
rise-erpelding, line 1: image

Welp, after we had that chat about the progress bar side label variant, I found it after all. And I see that it surfaces in the meter tests here too.

It's not in the spec for meter, and it looks a little wonky here (the % label is above the track slightly). I'm wondering if it's worth the trouble to try to troubleshoot and fix? It looks like it _might_ also be a problem in progress bar, but only if the label wraps to more than 1 line.

aramos-adobe: Removed side label story because it's not a meter config, but fixed it!

components/progressbar/index.css
```diff
@@ -155,6 +158,7 @@
 		margin-inline-end: var(--mod-progressbar-spacing-label-to-text, var(--spectrum-progressbar-spacing-label-to-text));
 		margin-block-end: 0;
 		margin-block-start: 0;
+		align-self: center;
```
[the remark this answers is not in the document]
aramos-adobe: Ah yes good catch!

components/meter/stories/meter.test.js
```diff
@@ -28,6 +24,10 @@ export const MeterGroup = Variants({
 			testHeading: "Without value label",
 			showValueLabel: false,
 		},
+		{
+			testHeading: "Help text",
+			helpText: "Help text message to add more context",
+		},
```
marissahuysentruyt, line 30: Do you think it's worth it to add the help text to the sizing test template, sort of like Rise did with tag group? The reason I'm asking is because the help text changes its t-shirt size when the meter does, so do we want to have a VRT baseline for that behavior?

Screenshot 2025-06-25 at 4 58 48 PM

I know you reviewed tag group, but just in case: https://github.com/adobe/spectrum-css/pull/3966/files#diff-e4784e05af5c8f7ab43b11c1d8211a05747916857afebf6f74225141329178a6

components/progressbar/index.css
marissahuysentruyt, line 1: Do you think we should explicitly set any of the passthroughs for field label? I know the font sizes are already correct, but I think we got some requests to be more explicit about the cascade, so would that mean creating custom properties (something like `--spectrum-progressbar-percentage-font-size: var(--spectrum-font-size-100);`, and then using mods for field label (--mod-field-label-font-size: var(--spectrum-progresbar-percentage-font-size);`)? 

It feels like we did that in other components, just to be _really_ sure that we're declaring the nested component styles properly. And we're already "redeclaring" the line-height property, even though it's also the same as field label's line-height. Just want to make sure we're doing all the components the same when it comes to embedded components!

components/meter/stories/meter.stories.js
```diff
@@ -33,12 +33,26 @@ export default {
 			options: ["info", "notice", "positive", "negative"],
 			control: "select",
 		},
+		helpText: {
```
marissahuysentruyt, line 36: Did we want to add a help text story for meter? I did see an example in the token specs, but help text doesn't show up on the desktop designs. It's also not in the Figma properties, so maybe we don't need it (maybe that part of the token specs is unnecessary and could be removed?)

Screenshot 2025-06-25 at 5 18 26 PM

I feel like if we have a control and styles for it, we should probably have a story too.

components/progressbar/index.css
```diff
@@ -40,9 +40,12 @@
 	--spectrum-progressbar-text-color: var(--spectrum-neutral-subdued-content-color-default);
 	--spectrum-progressbar-track-color: var(--spectrum-gray-300);
```
marissahuysentruyt, line 41: Oh interesting- we're "technically" using the wrong token here! The specs for both progress bar and meter request `--spectrum-track-color`, which resolves to gray-300 anyways, but I wonder if we didn't have that token when we originally migrated progress bar. 

I'll make a follow up ticket to fix this, unless you want to just correct it now! Let me know 😊

marissahuysentruyt (changes requested):
Overall, this is looking pretty good! I think I found some missing tokens, specifically the meter thickness ones for each t-shirt size. And then I'd love your perspective on the border radius- is it supposed to be on both ends of the fill? 

I did have a question about the animation for meter's (and progress bar's) animation. It looks like the CSS animation has that jitter that we solved over in SWC (https://github.com/adobe/spectrum-web-components/pull/5535). Do you think this migration is the right place to fix that, too? Or do you think that's a separate bug fix PR? If we don't pull the fix into our repo for S2, would that mean when SWC takes in any new S2 CSS, we'd break the animation again? 

Looks like there's some missing tokens: the meter-thickness-* t-shirt sizing tokens. Looking through the tokens/dist, the values for them are all slightly different than the current progress-bar-thicknes-* t-shirt sizing.  Maybe we need some mods to pass to meter again to use those? There also may be some context I don't have!

components/progressbar/index.css
```diff
@@ -161,6 +166,7 @@
 		margin-inline-start: var(--mod-spacing-progressbar-label-to-text, var(--spectrum-progressbar-spacing-label-to-text));
 		margin-block-end: 0;
 		margin-block-start: 0;
+		align-self: center;
```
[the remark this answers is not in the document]
aramos-adobe: If the side label on the left wraps to another line the percentage label is fixated at the top. That's why I added the `align-self: center`

components/meter/stories/meter.stories.js
```diff
@@ -90,6 +108,26 @@ FillColors.parameters = {
...
+StaticWhite.args = {
+	staticColor: "white",
+	value: 50,
+};
+StaticWhite.tags = ["!dev"];
+StaticWhite.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
+export const StaticBlack = MeterGroup.bind({});
```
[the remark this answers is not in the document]
aramos-adobe: Changing them to use `Template.bind({})`

components/progressbar/index.css
```diff
@@ -181,6 +187,23 @@
 	}
 }
 
+/* Static Black */
```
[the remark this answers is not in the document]
aramos-adobe: Good idea!

components/progressbar/index.css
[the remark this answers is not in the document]
aramos-adobe: Uploading Screen Recording 2025-06-25 at 6.10.54 PM.mov…

components/progressbar/index.css
aramos-adobe: We probably didn't. But I think we can correct it here in this migration

aramos-adobe:
@marissahuysentruyt The reason why I didn't use the token `mod-progressbar-thickness` is because in the design the thickness between progressbar and meter is identical. Unless you see something different in the `token/dist`. But it appears to be the same. Since the tokens for meter-thickness does exist in the token library I'll use the mods accordingly. Maybe something might in the future

components/progressbar/index.css
aramos-adobe: I moved the `--spectrum-progressbar-font-size` token to the label and percentage selector. It was in the rootClass selector and it wasn't causing any effect there.

components/meter/stories/meter.stories.js
```diff
@@ -90,6 +107,39 @@ FillColors.parameters = {
...
+export const HelpText = Template.bind({});
+HelpText.args = {
+	helpText: "This is a help text",
+	value: 50,
+};
+HelpText.tags = ["!dev"];
+HelpText.parameters = {
+	chromatic: { disableSnapshot: true },
+};
+
```
marissahuysentruyt, line 142:
```suggestion
HelpText.storyName = "Help text";
```

Let's use sentence-case for our story display names. (thanks for adding this story, by the way!)

components/meter/index.css
```diff
@@ -15,15 +15,19 @@
 .spectrum-Meter {
 	--mod-progressbar-max-size: var(--mod-meter-max-width, var(--spectrum-meter-maximum-width));
 	--mod-progressbar-min-size: var(--mod-meter-min-width, var(--spectrum-meter-minimum-width));
+	--spectrum-meter-help-text-to-progress-bar: var(--spectrum-spacing-75);
+	--mod-progressbar-thickness: var(--spectrum-meter-thickness-medium);
```
marissahuysentruyt, line 19: @aramos-adobe thanks for adding the meter tokens in! You're right that they're the same as progress bar. When I made that comment yesterday, I was looking through the large/mobile values 🤦‍♀️ , which are, in fact, different. But either way, I think it's a good practice for us to be explicit with the tokens like this (even if they're the same for the time being). Thanks!

marissahuysentruyt (approved):
Lovely! I had a few super small suggestions for some docs, but tokens are in great shape! I'm going to approve anyways. 
excited-seinfeld

PR #4020, opened 2025-07-09
feat(accordion): add direct actions

rise-erpelding:

### Regression testing

Validate:

## To-do list


components/accordion/index.css
```diff
@@ -72,6 +74,12 @@
 		var(--mod-accordion-item-header-top-to-text-space, var(--spectrum-accordion-item-header-top-to-text-space)) + var(--mod-accordion-item-header-bottom-to-text-space, var(--spectrum-accordion-item-header-bottom-to-text-space)) + (var(--mod-accordion-item-header-font-size, var(--spectrum-accordion-item-header-font-size)) * var(--mod-accordion-item-header-line-height, var(--spectrum-accordion-item-header-line-height)))
 	);
 
+	/* Calculated vertical spacing for action button and switch to center them within the accordion item */
+	--spectrum-accordion-item-direct-actions-vertical-spacing: calc(
+		(var(--mod-accordion-item-min-block-size, var(--spectrum-accordion-item-min-block-size)) -
+		var(--mod-accordion-item-direct-actions-height, var(--spectrum-accordion-item-direct-actions-height))) / 2
+	);
+
```
rise-erpelding, line 82: Did I miss a better way to do vertical spacing for this situation? I was finding that this got pretty complex, pretty fast:

* An `align-items: center` on the `.spectrum-Accordion-itemDirectActions` isn't ideal when the text wraps because the designs show that we want direct actions to align with the first line of text, instead of all of them.
* Accordion items have a min-height of `component-height-75`/`component-height-100`/`component-height-200`/`component-height-300`, same with the direct actions, but we don't usually see that min-height in accordion (except for compact) because of the calculations for spacing and font-size * line-height making it larger

components/accordion/index.css
```diff
@@ -315,6 +328,20 @@
...
 }
 
+.spectrum-Accordion-itemDirectActions {
+	margin-inline-end: var(--mod-accordion-edge-to-content-area, var(--spectrum-accordion-edge-to-content-area));
+	display: inline-flex;
+	gap: var(--mod-accordion-item-direct-actions-spacing, var(--spectrum-accordion-item-direct-actions-spacing));
+
+	& .spectrum-ActionButton,
+	& .spectrum-Switch {
+		margin-block-start: var(--mod-accordion-item-direct-actions-vertical-spacing, var(--spectrum-accordion-item-direct-actions-vertical-spacing));
```
marissahuysentruyt, line 338: Although I'm not sure I have a way to avoid this calc (I think it looks the best, honestly), I wonder if we can slap the block-size on that itemDirectActions container instead, to keep it from growing.

If we put: 
```
margin-block-start: var(--mod-accordion-item-direct-actions-vertical-spacing, var(--spectrum-accordion-item-direct-actions-vertical-spacing));
block-size: var(--spectrum-accordion-item-direct-actions-height);
```

on `.spectrum-Accordion-itemDirectActions`, I think we can remove this nested block with the action button and switch.

marissahuysentruyt (commented):
Thanks for getting this into accordion!

Separately, I want to look into the align-items: flex-start in the switch component. To me, in this specific case, the action button icon and the switch don't look center-aligned to each other, but maybe that's ok. Inspecting Figma, there's 6px of space around the action button icon, but 8px of space around the switch, so it feels like there's an accommodation of that flex-start. I'm not sure this is the PR to investigate though.

components/accordion/index.css
```diff
@@ -268,15 +280,16 @@
 
 /* Focusable button that expands/collapses the accordion item. */
 .spectrum-Accordion-itemHeader {
+	overflow-wrap: anywhere;
```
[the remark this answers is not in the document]
rise-erpelding: If there are other methods out there, I'd love to hear them, but I found [this](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#break-word) from MDN:

> `word-break: break-word` Has the same effect as `overflow-wrap: anywhere` combined with `word-break: normal`

We can't use `word-break: break-word` because it's been deprecated (see #4044), but in this case substituting with `overflow-wrap: break-word` (which is what I see in #4044) doesn't work.

Reading what MDN says about `word-break: normal`, I believe `word-break: normal` is the default so it didn't need to be set explicitly... I set it here on the off-chance it gets set somewhere else unintentionally.

marissahuysentruyt (changes requested):
I think I only have one real suggestion in adding the disabled arg! Nice work on this- for something that seems to straightforward, that calc to align the action button/switch is so complicated.

A couple of other things I noticed this time around:

- I noticed that the `accordion-top-to-text-compact-medium` token is 8px, so eventually design will have to update that to 5px like the spec shows in Figma. I'm assuming 5px is correct, because 8px makes the text sit way too low to my eyes. The matching bottom-to-text is 5px as well. Are they aware that it's an incorrect value, or is that intentional?

Screenshot 2025-07-23 at 1 01 29 PM

Screenshot 2025-07-23 at 1 03 48 PM

- I saw your question about the space between the itemHeader text and the direct actions container being spacing-100. I also think it's odd that the measurement is to the action button icon. I'm not sure if we could do some sort of clamp(), but that really only gets the space closer to the desired 8px, it doesn't actually address the problem. I had something like: 

```
.spectrum-Accordion-itemHeading:has(.spectrum-Accordion-itemDirectActions) .spectrum-Accordion-itemHeader {
    padding-inline-end: clamp(5px, calc(var(--mod-accordion-edge-to-content-area, var(--spectrum-accordion-edge-to-content-area)) - var(--spectrum-spacing-100)), var(--spectrum-spacing-100));
}
```
...and I picked 5px as the min, since I saw the action button padding was 3px. But I don't know how we account for the changes in the action button padding when the t-shirt sizing changes. I wonder if at the very least we can account for the spacing-100 in the variant without inline padding. Right now, it looks like all of the padding around itemHeader also goes away, instead of just the outermost padding. Am I understanding that story right? There should technically be _some_ sort of spacing there right?

https://github.com/user-attachments/assets/aa18c1e0-9174-46a4-8889-ea28f7050b51

components/accordion/index.css
rise-erpelding: Yeah it 100% bugs me too that the action button doesn't fit within the accordion item and that the `itemDirectActions` container is bigger than the accordion item, but I think this is only for XL compact (I guess technically for large compact we also end up with a negative margin but I think the browser might handle that as 0 since it's fractional?).

I'd love to get more design feedback on this one at some point. Do we think now is the better time to do that? One reason I can think of to hold back is that I'd thought the implementation of Adobe Clean Spectrum was going to help resolve some of our line height problems, which give us component height problems in certain components. Josh made that calc that we use here in accordion to set `--spectrum-accordion-item-min-block-size` to either the `--spectrum-component-height-xx` tokens or the sum of the top and bottom spacing and font size times line height. My understanding was that if the font's rolled out as proposed, the tokens would set font-size, line-height, and vertical-padding in a way that will always add up to the corresponding component height.

I'm wondering if we get that all resolved, if that will change anything related to the way the compact accordion item is spaced. It might not. Thoughts?

I'm also wondering if once design has more concrete guidance in place if they might decide to recommend that direct actions and XL accordions shouldn't coexist because it doesn't leave enough space for the heading text?

components/accordion/stories/template.js
```diff
@@ -66,6 +73,28 @@ export const AccordionItem = ({
 					}, context)}
 					<span class="${rootClass}Title">${heading}</span>
 				</button>
+				${when(
+					hasActionButton || hasSwitch,
+					() =>
+						html`
+							<div class="${rootClass}DirectActions">
+								${when(hasActionButton, () =>
+									ActionButton({
```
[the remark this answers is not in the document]
rise-erpelding: Whoops! Good idea.

rise-erpelding:
> I noticed that the accordion-top-to-text-compact-medium token is 8px, so eventually design will have to update that to 5px like the spec shows in Figma. I'm assuming 5px is correct, because 8px makes the text sit way too low to my eyes. The matching bottom-to-text is 5px as well. Are they aware that it's an incorrect value, or is that intentional?

Yep, I noticed it too, when I'd originally asked I didn't notice that it was done only for the regular density and not the compact/spacious. It looks like it was addressed in [13.12.0](https://github.com/adobe/spectrum-tokens/releases/tag/%40adobe%2Fspectrum-tokens%4013.12.0), so that hopefully will _just work_ once we're taking in that version of tokens. We're a bit behind at the moment but hopefully we can address this in #4019?

Still looking into some of the other issues!

components/accordion/index.css
```diff
@@ -32,6 +32,7 @@
 	--spectrum-accordion-item-content-area-bottom-to-content: var(--spectrum-accordion-content-area-bottom-to-content);
 	--spectrum-accordion-divider-thickness: var(--spectrum-divider-thickness-small);
 	--spectrum-accordion-item-direct-actions-height: var(--spectrum-component-height-100);
+	--spectrum-accordion-item-header-to-direct-actions-space: var(--spectrum-spacing-100); /* same for all sizes */
```
rise-erpelding, line 35: After examining Figma further, the spacing used between the end of the header text and the start of the direct actions is always 8px/spacing-100 regardless of size. This will also be maintained even if there's no inline padding on the ends of the accordion items.

image

components/accordion/index.css
```diff
@@ -336,12 +333,10 @@
 }
 
 .spectrum-Accordion-itemDirectActions {
+	margin-block: var(--mod-accordion-item-direct-actions-vertical-spacing, var(--spectrum-accordion-item-direct-actions-vertical-spacing));
 	margin-inline-end: var(--mod-accordion-edge-to-content-area, var(--spectrum-accordion-edge-to-content-area));
 	display: inline-flex;
 	gap: var(--mod-accordion-item-direct-actions-spacing, var(--spectrum-accordion-item-direct-actions-spacing));
-
-	/* margin needs to be set on top and bottom to keep compact XL items vertically centered and prevent them from growing vertically */
```
rise-erpelding: I still think top and bottom margin being set is a good idea although it is probably less necessary now. I removed the comment since it no longer applies.

components/accordion/index.css
```diff
@@ -13,7 +13,7 @@
 
 .spectrum-Accordion {
 	/* Layout and spacing */
-	--spectrum-accordion-item-minimum-height: var(--spectrum-component-height-100);
+	--spectrum-accordion-item-minimum-height: var(--spectrum-component-height-200);
```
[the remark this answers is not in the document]
cdransf: If they're consistent with the values in Figma I'm all for updating them here. ✨

cdransf (approved):
Looks great! ✨

tokens/custom/large-vars.css
```diff
@@ -138,4 +138,7 @@
 	--spectrum-field-edge-to-icon-100: 9px;
 	--spectrum-field-edge-to-icon-200: 11px;
 	--spectrum-field-edge-to-icon-300: 14px;
+
+	--spectrum-accordion-top-to-text-compact-medium: 5px;
+	--spectrum-accordion-top-to-text-spacious-medium: 15px;
```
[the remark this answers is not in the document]
marissahuysentruyt: Yeah, you're right- it does look uneven still. 🤦‍♀️  I hope the new font helps out with all the uneven paddings!

PR #4043, opened 2025-07-16
feat(clearbutton): s2 migration

cdransf:
## Description

`CSS-1169`

## To-do list


components/clearbutton/stories/clearbutton.stories.js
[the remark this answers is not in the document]
cdransf: Added those! ✨

cdransf:
> Figma always gives the same 3 down state tokens for every component where it's used, but you probably don't need all of them. 

We _do_ have the color tokens in place for each state, but that predates the migration. ✨

> ... since the clear button is larger than 24px but the icons are all smaller, it _seems_ like it might be ok to use the `component-size-minimum-perspective-down` token?

That makes the most sense to me — I've gone ahead and run with `component-size-minimum-perspective-down` for the latest update.

> We'll probably want to capture this in the tests, as well as the hover state

Added tests for hover and down. ✨

> I'm wondering how this will look in the removable Tag, which uses the clear button. The down state for Tag applies the down state to the whole Tag, but I'm guessing that if we were only targeting the clear button within Tag, that ideally we'd want just the clear button to have the down state and not the whole tag? That might be another ticket to review and confirm with design later on.

Hmmm, that's a good question. Looking at the removable tag now, the down state applied to the whole tag doesn't feel quite right. I'd be inclined to apply it to the child clear button like you suggest. I'll open up a ticket.

rise-erpelding (approved):

components/clearbutton/stories/clearbutton.stories.js
```diff
@@ -1,13 +1,13 @@
 import { Sizes } from "@spectrum-css/preview/decorators";
 import { disableDefaultModes } from "@spectrum-css/preview/modes";
-import { isDisabled, isQuiet, size, staticColor } from "@spectrum-css/preview/types";
+import { isDisabled, size } from "@spectrum-css/preview/types";
```
marissahuysentruyt, line 3: If we use the `isHovered` and `isActive` types, we could have those controls in the table for Storybook 🤔 

Screenshot 2025-07-23 at 4 38 50 PM

There'd be a couple of things left to do, like add those args to the default list so the boolean is set in the control table, change around a few things in the template and test files, but I think it'd be nice to get those controls in.

marissahuysentruyt (changes requested):
Nice work! There's some very small things I think we should add, like the migrated status & tag, and then the active and hover controls. I think the tokens are in a good but, I had a couple of extra things I wanted to bring up! 

Do we know if the corner radius needs to be anything specific for clear button? Should it be fully rounded into a circle, or follow the infield button border radius specs? I very well may have missed it if you already asked this question somewhere! I noticed there were no rounded corners at all when I was playing with the mod for the background color, that's all.

I noticed that we get the browser default `focus-visible` outline on the clear button. But looking at [the infield button CSS](https://github.com/adobe/spectrum-css/blob/bc7c72455cc1573d83ae76cce463404c76ddc72c/components/infieldbutton/index.css#L132) (which this is sort of a variant of), the outline is removed completely. Do you think we should also remove the outline for keyboard focus in clear button?

Screenshot 2025-07-23 at 3 08 51 PM

.changeset/neat-schools-take.md
```diff
@@ -0,0 +1,11 @@
...
+"@spectrum-css/clearbutton": major
+---
+
+#### S2 migration for clear button
+
+This migrates the clear button component to the latest Spectrum 2 designs. Custom properties have been remapped and added per the design specification.
+
+- The quiet variant has been remove as all clear buttons have a transparent background.
+- The over-background and static color variants have been removed.
+- Test and stories for deprecated variants have been removed.
```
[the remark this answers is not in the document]
cdransf: Added! ✨ `- The down state has been added, corresponding controls and tests have been added.`

components/clearbutton/stories/clearbutton.stories.js
```diff
@@ -85,17 +78,14 @@ Disabled.parameters = {
...
 
 /**
- * The `.spectrum-ClearButton--quiet` class will use a transparent background (including when the Express theme is active).
+ * The clear button component supports the following sizes:
+ * - `s`: 75px
+ * - `m`: 100px
+ * - `l`: 200px
+ * - `xl`: 300px
+ *
+* The default size is medium.
```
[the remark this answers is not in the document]
cdransf: I like that! (It looks better too.) Dropped that change in. ✨

components/clearbutton/index.css
```diff
@@ -80,7 +59,6 @@
 .spectrum-ClearButton {
 	block-size: var(--mod-clear-button-height, var(--spectrum-clear-button-height));
 	inline-size: var(--mod-clear-button-width, var(--spectrum-clear-button-width));
-	border-radius: 100%;
 
 	background-color: var(--mod-clear-button-background-color, transparent);
```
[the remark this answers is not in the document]
cdransf: I set it to `background-color: transparent;` like the infield button so the two are consistent. ✨

.changeset/neat-schools-take.md
```diff
@@ -0,0 +1,12 @@
+---
+"@spectrum-css/clearbutton": major
+---
+
+#### S2 migration for clear button
+
+This migrates the clear button component to the latest Spectrum 2 designs. Custom properties have been remapped and added per the design specification.
+
+- The quiet variant has been remove as all clear buttons have a transparent background.
```
castastrophe, line 9:
```suggestion
- The `quiet` variant has been **removed** as all clear buttons have a transparent background now, making this variant irrelevant.
```

Minor formatting and wording knit.

.changeset/neat-schools-take.md
```diff
@@ -0,0 +1,12 @@
...
+- The over-background and static color variants have been removed.
```
castastrophe, line 10:
```suggestion
- The `over-background` and `static color` variants have been removed.
```

What does this translate to in the code? I like this general note but might be good to translate this into, _X_ class was removed for example.

.changeset/neat-schools-take.md
```diff
@@ -0,0 +1,12 @@
...
+- Test and stories for deprecated variants have been removed.
```
castastrophe, line 11: I like this note but it doesn't really impact customers so maybe we remove this point?

.changeset/neat-schools-take.md
```diff
@@ -0,0 +1,12 @@
...
+- The down state has been added, corresponding controls and tests have been added.
```
castastrophe, line 12: What does that translate to for customers? Can you clarify in the docs here what the introduction of the down state looks like for customers adopting this change?

components/clearbutton/index.css
```diff
@@ -80,9 +59,8 @@
 .spectrum-ClearButton {
 	block-size: var(--mod-clear-button-height, var(--spectrum-clear-button-height));
 	inline-size: var(--mod-clear-button-width, var(--spectrum-clear-button-width));
-	border-radius: 100%;
 
-	background-color: var(--mod-clear-button-background-color, transparent);
+	background-color: transparent;
```
castastrophe, line 63: Nice, let's make sure this gets documented clearly in the changelog that this mod no longer works here.

PR #4115, opened 2025-08-06
feat(search): collapsed search 

aramos-adobe:
## Description

The search component allows for a minimized state where the search field is collapsed to an action button. When clicked the the search field is visible.

### Validation steps

  Include steps for the PR reviewer that explain how they should test your PR. Example test outline:
  1. Open the storybook for the search component:
  - [x] Check `true` is collapsed state (@cdransf)
  - [x] Click the action button to show the search field (@cdransf)

### Regression testing

Validate:

### States
Screenshot 2025-08-15 at 11 11 18 AM

### Keyboard Navigation
- `Tab` key to access the search button and press `Enter` to expand the search field

## To-do list


.changeset/long-carrots-sleep.md
```diff
@@ -0,0 +1,15 @@
...
+"@spectrum-css/search": major
+---
+
+## S2 Minimized search field
+
+The search component allows for a minimized state where the search field is collapsed to a button.
+
+### Anatomy
+
+The collapsed state consists of a single action button that only has a hover and disabled state. This state is triggered by the `is-collapsed` class on the search component. When the search field is in this state, the textfield is hidden and the search button is displayed. The button can be hovered and focused, and will expand the search field when clicked.
```
marissahuysentruyt, line 11: since we added keyboard focus back in, can we update this first sentence just a touch so that it doesn't say that it "only" has hover and disabled states? You mention focus later in the paragraph, even if you decide not to change it. 👍

components/search/index.css
```diff
@@ -149,6 +149,20 @@
 	.spectrum-HelpText {
 		margin-block-start: var(--mod-search-to-help-text, var(--spectrum-search-to-help-text));
 	}
+
+	/* Animation for collapsible search expansion */
+	&.is-collapsed {
+		transition: inline-size 0.3s cubic-bezier(0.4, 0, 0.2, 1);
```
marissahuysentruyt, line 155: how would you feel about maybe using of any of the `--spectrum-animation-duration-*` tokens? Maybe we can use that here? Or was the 0.3 design directed? And just because I'm nosy, where did you find these timing curves?! 

Should the cubic-beziers and even the transition-duration be set as custom props? Maybe it's not necessary, but we do use the same curve & durations twice 🤷‍♀️

marissahuysentruyt (commented):
Nice work on this. I feel like I still mainly have questions, so I left a few for you 😊 

I might consider just updating the changeset heading to "Collapsed" instead of "Minimized." Just personal preference though.

Have you reached out to design to ask about some of the focus and keyboard focus behaviors? I found the collapsed usage section in Figma, and the "focus state" of this button is just...open? That feels odd to me for some reason. Like if I was tabbing along, does it make sense for me to tab to this action button, and then it just be a completely different component? Anyways, I think this is more along the lines that I would expect, where I'm tabbing along, and I have to trigger the rest of the search field to open. I may be wrong though. Is that worth bringing up with design?

components/search/index.css
aramos-adobe: I'm just experimenting with animation curves here. I don't know what the intended animation should be

marissahuysentruyt: oh nice! I've been learning about each of the values of the `cubic-bezier` function, so I was just curious!

PR #4139, opened 2025-08-18
fix(accordion): clean up spacing issues

rise-erpelding:
## Description
Resolves the previous issue with accordion where, when the action button was added to an item's direct actions area, the action button was taller than the accordion item's height. (This was only an issue with compact XL.)
image

The updates made to accordion adjust the spacing tokens so that the action button always fits within the accordion item, although its focus outline may be outside of the accordion item (just as the accordion item's focus outline is outside of the item).
image

After confirming with design, the tokens used for minimum height have been updated to be more spacious, which better matches the designs. Note that these tokens aren't always used; there is a calculation that sets the min height to the maximum between the token or the sum of the space needed (top & bottom padding + font-size * line-height) to ensure that the accordion item always has enough space even with adjustments to padding, font-size, or line-height.

Since accordion items have more space, some adjustments were also made to center the accordion's itemTitle vertically.

Also addresses a small issue with the hover/key focus state where the itemTitle extends to the edge of the direct actions, this was more noticeable if there was a switch direct action, but not an action button:
Before:
image
image
After:
image
image

CSS-1260

### Validation steps

Use the PR preview or check out the branch locally:
- [x] Check every combination of t-shirt size/density to confirm that minimum height looks as expected and items are centered/aligned vertically. (@cdransf)
- [x] Check hover state and key focus with switch direct action only to confirm that this looks as expected (@cdransf)
- [x] Check hover state and key focus with both direct actions (switch and action button) to confirm that this looks as expected (@cdransf)
- [x] Check hover state and key focus with no direct actions to confirm that this looks as expected (@cdransf)
- [x] If the line-height is adjusted (for instance if we switch to Japanese), the accordion item will make more space to accommodate that line height. Does this look ok? (It's occurred to me that this does throw the alignment of the disclosure icon off a little bit, are we ok with that for now or do we want to look more closely for a solution that will align the chevron and the text for this use case?) (@cdransf)
- [x] Are there any other items that might be helpful to address re: accordion while we're here? Do tests/documentation look good, for instance? (@cdransf)

### Regression testing

Validate:

- [x] Any VRT changes have been accepted (by reviewer and/or PR author), <strike>or there are no changes</strike>.

## To-do list

- [ ] <strike>I have tested these changes in Windows High Contrast mode.</strike>
- [ ] <strike>If my change impacts **other components**, I have tested to make sure they don't break.</strike>

### DIFF
--- a/components/accordion/stories/accordion.test.js
+++ b/components/accordion/stories/accordion.test.js
@@ -16,39 +16,20 @@ export const testsContent = new Map([
 	[
 		"Architecture",
 		{
-			content: Typography({
-				semantics: "body",
-				content: [
-					"This is an example of text content within the content area that uses the spectrum-Body typography class on the paragraphs. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eleifend est mollis ligula lobortis, tempus ultricies sapien lacinia. Nulla ut turpis velit.",
-					Link({
-						url: "https://www.adobe.com/creativecloud/plans.html",
-						text: "Learn more about Adobe Creative Cloud plans and pricing.",
-					}),
-				],
-			}),
+			content: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eleifend est mollis ligula lobortis, tempus ultricies sapien lacinia. Nulla ut turpis velit. Sed finibus dapibus diam et sollicitudin. Phasellus in ipsum nec ante elementum congue eget in leo. Morbi eleifend justo non rutrum venenatis. Fusce cursus et lectus eu facilisis. Ut laoreet felis in magna dignissim feugiat.",
 			isOpen: true,
 		},
 	],
 	[
 		"Nature",
 		{
-			content: Typography({
-				semantics: "body",
-				content: [
-					"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eleifend est mollis ligula lobortis, tempus ultricies sapien lacinia. Nulla ut turpis velit. Sed finibus dapibus diam et sollicitudin. Phasellus in ipsum nec ante elementum congue eget in leo. Morbi eleifend justo non rutrum venenatis. Fusce cursus et lectus eu facilisis. Ut laoreet felis in magna dignissim feugiat.",
-					Link({
-						url: "https://www.adobe.com/acrobat.html",
-						text: "Learn more about Acrobat.",
-					}),
-				],
-			}),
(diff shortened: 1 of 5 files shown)


components/accordion/stories/accordion.test.js
```diff
@@ -16,39 +16,20 @@ export const testsContent = new Map([
...
-				semantics: "body",
-				content: [
-					"This is an example of text content within the content area that uses the spectrum-Body typography class on the paragraphs. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eleifend est mollis ligula lobortis, tempus ultricies sapien lacinia. Nulla ut turpis velit.",
-					Link({
-						url: "https://www.adobe.com/creativecloud/plans.html",
-						text: "Learn more about Adobe Creative Cloud plans and pricing.",
-					}),
-				],
-			}),
+			content: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec eleifend est mollis ligula lobortis, tempus ultricies sapien lacinia. Nulla ut turpis velit. Sed finibus dapibus diam et sollicitudin. Phasellus in ipsum nec ante elementum congue eget in leo. Morbi eleifend justo non rutrum venenatis. Fusce cursus et lectus eu facilisis. Ut laoreet felis in magna dignissim feugiat.",
```
rise-erpelding, line 19: When showing the Storybook preview to Kami, she noticed the discrepancies between the item content formatting that we had before, where some of the items use plain text styled according to the design spec, but some demonstrate use of the Typography component, which has different line spacing and a larger font size (even if we use the same t-shirt size). I see how that's confusing and inconsistent, so I've taken this out of the default story.

Since it is useful within the custom width story that has links within its content, the typography classes are still used there.

components/accordion/stories/accordion.test.js
```diff
@@ -80,9 +61,10 @@ export const longerContent = new Map([
 		"Are Adobe products worth it?",
 		{
 			content: Typography({
+				size: "s",
```
rise-erpelding, line 64: The typography font size doesn't line up with the spec'd font size for accordion item content, so we're using a smaller typography size here.

.changeset/pre.json
```diff
@@ -142,6 +142,7 @@
     "itchy-shrimps-help",
     "itchy-waves-work",
     "khaki-icons-hammer",
+    "kind-cycles-check",
```
cdransf, line 145: Should this be included as part of the changes for this PR? ✨

